PR #78, opened 2015-04-02
Fixes enqueue eror that occurs when multiple copies of CMB2 are used

LeachCreative:
Added composer dependency support to handle the loading of CMB2, which fixes the problem I reported on your website about enqueuing errors with multiple copies of CMB2.


DevinWalker:
@LeachCreative thanks for the PR. The issue is that Give will soon be going up on WordPress.org - CMB2 needs to be packaged within the plugin for it to work. Composer won't work for WP plugins on the repo - correct?

DevinWalker:
FYI - here's the bug report that was submitted: https://github.com/WordImpress/Give/issues/77

LeachCreative:
I'm not too familiar with the requirements for plugins on Wordpress.org, however this plugin doesn't appear to be having any issues with using composer, so I think it should be possible.

https://wordpress.org/plugins/composer/

LeachCreative:
Hmm, upon reading it again, I'm not sure that I read what you said properly last time.  When you say it needs to be  packaged with the plugin you mean that a copy of CMB2 needs to be within the Give directory?  If thats the case then what I committed should work, as CMB2 is in the vendor directory, and is being pulled in through Composer's autoload.

DevinWalker:
Composer's autoload won't work with WordPress plugins. I'll take a hard look into the conflicts you mentioned, but for now we're going to package CMB2 directly within Give. :]

PR #99, opened 2015-04-09
Some readme.txt fixes

chriscct7:


DevinWalker:
Hey hey... I just reviewed your PR and you're proof reading helps a lot. I think my only issue with the edits is with the word "Add-on" being changed to "addon" - we've thought about this for awhile and want to keep it "Add-on" because it's a noun describing our products. I'm no grammar wiz but we've standardized this on the site and other places so preferably want to keep it the same. Other than that, really enjoy the edits.

DevinWalker:
@chriscct7 I made some changes to the two readmes like you recommended but I would still love to get a PR from you with "Add-ons" kept the same. I believe I have also addressed the "gate" issue haha

Issue #131, opened 2015-05-14
Composer package


stopspazzing:
Could you please add to https://getcomposer.org/ ? Would making wordpress theme devs happy for those who use composer to manage packages.

DevinWalker:
To be honest I haven't done much with composer but hear a lot about it. This will give me a good excuse to get hands on with it :+1:

stopspazzing:
Never created one myself, however, should be pretty easy:
```xml
{
  "name": "roots/soil",
  "type": "wordpress-plugin",
  "license": "MIT",
  "description": "A WordPress plugin which contains a collection of modules to apply theme-agnostic front-end modifications",
  "homepage": "https://roots.io/plugins/soil/",
  "authors": [
    {
      "name": "Ben Word",
      "email": "ben@benword.com",
      "homepage": "https://github.com/retlehs"
    },
    {
      "name": "Scott Walkinshaw",
      "email": "scott.walkinshaw@gmail.com",
      "homepage": "https://github.com/swalkinshaw"
    }
  ],
  "keywords": [
    "wordpress"
  ],
  "support": {
    "issues": "https://github.com/roots/soil/issues",
    "forum": "https://discourse.roots.io/"
  },
  "require": {
    "php": ">=5.4.0",
    "composer/installers": "~1.0"
  }
}```

Example. Should take a minute, literally to do.

**Comment by michaelbeil** [comment]
@stopspazzing, `composer.json` was added. @DevinWalker can you add to Packagist?

Also, Give is on WPackagist: http://wpackagist.org/search?q=give&type=plugin&search=.

**Comment by DevinWalker** [comment]
@michaelbeil I'm adding it now

**Comment by stopspazzing** [comment]
Working great! 
```
composer require wordimpress/give @dev
```

PR #135, opened 2015-05-20
Composerify

michaelbeil:
#131

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,27 @@
+{
+  "name": "wordimpress/give",
+  "type": "wordpress-plugin",
+  "license": "GPL",
+  "description": "Give - The best way to accept donations on WordPress",
+  "homepage": "http://givewp.com",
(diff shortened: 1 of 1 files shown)


DevinWalker:
Thanks @michaelbeil - question: since we are using other packages like CMB2 - should that also be included as one of our "dependencies"?

michaelbeil:
Well, it looks you are not using CMB2 as a submodule, but are simply pulling in any changes directly, right?

DevinWalker:
No it's not a submodule - since we have a lot of custom fields we created to work for Give we prefer to update it incrementally.

michaelbeil:
You should be good to go then.

If you want to experiment, check out this tut form the guys over at Roots: https://roots.io/using-composer-with-wordpress.

DevinWalker:
Cool thanks!

michaelbeil:
Sure thing.

PR #137, opened 2015-05-21
Added Donation Limiting for Logged-In Users

d4mation-:
I was working on donations for a political campaign website and needed to limit the total donation (Past and Current) to a certain value. I guess they couldn't accept donations from any singular party over $500.

This adds that functionality per-form. It works best when Guest Donations are turned off as those cannot be as easily tracked. I didn't want to filter by IP because that could blacklist entire office buildings for example.

I apologize for the indentation in some of these files. It looked fine in Brackets, but the DIffs show otherwise.


DevinWalker:
This looks pretty good @d4mation- I have some concerns about allowing Guest donations in conjunction with a cap. It seems like it should force the login option if a cap is set and define that clearly to the user. What do you think? Also, it may be useful to have a global option and a per-form option similar to how we do with default checkout (form options override globals) to provide further control. 

I'm creating a new branch of Donation Limiting - would you mind submitting your PR to that new branch?

DevinWalker:
@d4mation-  Would you mind submitting your PR to https://github.com/WordImpress/Give/tree/donation_limits

d4mation-:
I agree. I initially didn't code it that way because I figured I'd see if anyone could think of a use case for it. Perhaps it could be toggled for hide/show with Enabling Guest Donations and then put a check for if Guest Donations are disabled in the AJAX Error Handler? I have tested it with Registering within the form and it doesn't cause any problems, as long as Guest Donations are disabled things would work just fine.

Sure, I'll go ahead and submit it to that branch.

DevinWalker:
Thanks for moving the PR over

PR #138, opened 2015-05-21
Added Donation Limiting for Logged-In Users.

d4mation-:
Moved the Pull Request from Master -> donation_limits.

Text from PR on Master:

I was working on donations for a political campaign website and needed to limit the total donation (Past and Current) to a certain value. I guess they couldn't accept donations from any singular party over $500.

This adds that functionality per-form. It works best when Guest Donations are turned off as those cannot be as easily tracked. I didn't want to filter by IP because that could blacklist entire office buildings for example.

I apologize for the indentation in some of these files. It looked fine in Brackets, but the DIffs show otherwise.


DevinWalker:
Thanks @d4mation- , we'll continue refining in this branch. We want to add this feature to core though, just need to ensure it's done right before merging into master.

PR #160, opened 2015-06-29
fix PHP notice for Undefined index: _give_price. closes #159.

### LINKED ISSUE #159: PHP notice
When viewing a form on the front end with `WP_DEBUG` enabled I am seeing

>Notice: Undefined index: _give_price in /wp-content/plugins/give/includes/forms/template.php on line 323

helgatheviking:


DevinWalker:
Thanks @helgatheviking for taking the time to correct this :+1:

PR #198, opened 2015-08-04
fix the [give_form] "show_title" attribute

pryley:
[give_form id="276" show_title=false] now hides the title


pryley:
Also fixes "Disable CSS" option not working

PR #211, opened 2015-08-21
Extensibility improvements

cwackerman:
### Purpose
- - -
The purpose of this pull request is primarily to address some pain points I recently encountered while working with the Give plugin on a client project. They wanted to update the text in places, and I found that difficult to do in some cases.

### Solution
- - -
- Added a Wordpress filter to allow developers to modify the 'Select Payment Method' legend text.
- Modified existing Wordpress filter controlling submit button text so that it passes `$form_id` to hooks. This allows developers to access form options when generating submit button text. In my case this was necessary to output default donation amount, e.g. "Donate $5.00 Now".

### DIFF
--- a/includes/forms/template.php
+++ b/includes/forms/template.php
@@ -1064,7 +1064,7 @@ function give_checkout_button_purchase( $form_id ) {
 		<span class="give-loading-animation"></span>
 	</div>
 	<?php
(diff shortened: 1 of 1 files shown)


DevinWalker:
Looks great thanks for this PR :+1:

Issue #204, opened 2015-08-12
trigger JS events after any ajax requests that change the form DOM


pryley:
I have a custom script that styles the radio/checkbox/select fields and converts labels to floating labels (similar to this: http://clubdesign.github.io/floatlabels.js/).

![ccard](https://cloud.githubusercontent.com/assets/134939/9217105/0e1c2150-40f6-11e5-997e-1151970eaffb.png)

I need to be able to trigger a refresh anytime the DOM is changed via ajax. All this requires is something like the following inserted at the end of any successful ajax request: 

`jQuery( document ).trigger( 'give_gateway_loaded', [response] );`

Example: `function give_load_gateway() // give-ajax.js:197`

DevinWalker:
@pryley would you be interested in submitting a pull request for this?

pryley:
@DevinWalker sure.

DevinWalker:
Cool! :+1:

pryley:
@DevinWalker Hey were you referring to the event trigger or the float labels script? I understood the latter from the issue tags so I converted an internal script to a plugin and integrated it into Give #231 .

If not, no worries.

PR #231, opened 2015-09-03
Feature/floatlabels #204

pryley:
#204 : Uses my float-labels.js plugin: http://geminilabs.github.io/float-labels.js/


mathetos:
Paul that link seems to be down. Do you have an active link to see a demo of the Float labels?

DevinWalker:
Thanks for the PR. Question: Are the floating labels enabled by default or is there an option to enable them?

pryley:
I added an option to enable them:

screenshot 2015-09-04 00 26 45

I will put up a live example soon, for now here is a screenshot with the twentyfifteen theme enabled:

screenshot 2015-09-04 00 28 35

pryley:
@mathetos The link should work now. There is no live example at the moment, just instructions on usage.

DevinWalker:
Hey @pryley how come here https://github.com/WordImpress/Give/pull/231/files#diff-143c1ae2e57a1cf5852d709be4b5aa85L531 you change the input to type `tel` for CC fields?

pryley:
@DevinWalker `tel` input types show the numeric keyboard on mobile devices.

Reference: http://mobileinputtypes.com/

DevinWalker:
Cool! I figured. So, I just took a look at this and it's a really great PR. Did you get a chance to run a few test transactions through the new CC fields to verify they pass to the gateways in a valid format still? I had an issue with the recent credit card mask upsetting PayPal with extra spaces in the CC fields. I just want to ensure we don't disrupt the way the fields are currently passed.

pryley:
I took a quick look at the output of the `give_get_purchase_cc_info()` function which builds the CC array and everything looked fine, from looking at the code that seemed to be the area to verify.

pryley:
A few things to consider:

1. Changing the input types to `tel` is great for accessibility, but the Twenty Fifteen Wordpress theme (and maybe the others also) does not include css styles for that type, so when the theme is enabled and the `float labels` option is disabled, the form looks like this:

screenshot 2015-09-04 01 39 01

This is Wordpress' fault as their theme styles are incomplete.

2. With `Float Labels` enabled, the select styling will likely not match all themes (borders, background, etc.). It may be necessary to have a slightly more opinionated view on the default Give form styles.

3. The `Float Labels` Select angle-down icon is css generated. You may want to use a icon font instead (i.e. add an `angle-down` icon to the icomoon font).

DevinWalker:
That's ok with the style concern. I'm going to touch up the styles anyways for 1.3.

PR #239, opened 2015-09-07
Swedish translation

mepmepmep:
Adding swedish translation files .po and .mo


DevinWalker:
Hmm, it looks like you may have submitted this PR to the wrong branch. Can you submit it to the master branch?

PR #241, opened 2015-09-07
Create give-sv_SE.mo

mepmepmep:
Add translation file (.mo) for Swedish


mepmepmep:
I opened the file in Notepad and copied the source and then pasted it into a new file here. I have no idea if thats the recommended method? Hope it works.

DevinWalker:
Good enough! Typically you'd use a code editor to do that but notepad works. In fact, notepad is like the oldest code editor. haha

Thanks!

mepmepmep:
Ok.. great. :)

Issue #249, opened 2015-09-15
`give_require_billing_address` filter is always set to `false`


pryley:
The `give_require_billing_address` filter (`/includes/process-purchase.php:275`) is `true` by default, but will always be set to `false` because of `give_manual_no_cc_validation()`.

Here is a alternate way to write the function that allows other functions to modify the filtered value:
```
/**
 * Manual Gateway does not need CC billing address validation, so remove it.
 *
 * @since 1.0
 *
 * @param bool $bool
 *
 * @return void
 */
function give_manual_no_cc_validation( $bool ) {
	if ( isset( $_POST['give-gateway'] ) && $_POST['give-gateway'] == 'manual' ) {
		$bool = false;
	}

	return $bool;
}

add_filter( 'give_require_billing_address', 'give_manual_no_cc_validation' );
```

Since the default filtered value is `true`, you will need to add a similar function to any gateways that do not require billing address validation (i.e. PayPal Standard, Offline, etc.).

---

There is also an issue with the Braintree gateway plugin where form validation is not performed via ajax, making it impossible to use the single Form Payment Field `Reveal Upon Click / Modal Window Upon Click` options successfully.

I mention this here because I am sending you an email with a preliminary fix for the Braintree plugin, but it is not complete without being able to also validate the billing address fields.

DevinWalker:
Hmm, I'm going to dig into this one. Seems alarming off the bat.

DevinWalker:
@pryley how about we simply remove `give_manual_no_cc_validation()` it seems entirely unnecessary?

DevinWalker:
I'm trying to wrap my head around this... 

I have removed `give_manual_no_cc_validation()` but now currently PayPal and Manual (Test) are broken because of the true value returned here: https://github.com/WordImpress/Give/blob/master/includes/process-purchase.php#L281

Perhaps we can write a function similar to `give_is_cc_verify_enabled()` that checks whether or not the current gateway has CC fields enabled then return true/false dynamically depending on the chosen gateway.

DevinWalker:
@pryley we're getting close here. The only issue now that I can see is with `give_is_field_required` here https://github.com/WordImpress/Give/blob/master/includes/forms/functions.php#L206 when there's no `$_POST` available the fields required labels don't display properly.

pryley:
[...]

PR #251, opened 2015-09-15
Re: #249

pryley:
Please refer to the comment on #249 for the reason of this pull request.

You'll probably need to test this with all of your gateway plugins to ensure that any which don't require billing address validation call the `give_require_billing_address` hook to disable it.

### DIFF
--- a/includes/gateways/manual.php
+++ b/includes/gateways/manual.php
@@ -18,16 +18,24 @@
 add_action( 'give_manual_cc_form', '__return_false' );
 
 /**
- * Manual Gateway does not need a CC form validation, so remove it.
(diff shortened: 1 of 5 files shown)


DevinWalker:
Please see my comment over on https://github.com/WordImpress/Give/issues/249

pryley:
How about something like this?

pryley:
The function simply checks that 'billing_country' is in $_POST and if so, returns true to require validation.

I added a hook so gateways can modify the return value (they can check for their gateway name in $_POST).

I'm not sure why Scrutinizer cancelled. It should be fine.

DevinWalker:
Checking it out now

PR #262, opened 2015-09-16
Fix for Issue/249

### LINKED ISSUE #249: `give_require_billing_address` filter is always set to `false`

DevinWalker:
@mathetos @pryley check out this PR to fix #249 

I've tested and it's working well with the gateways I've tested (braintree, offline, paypal standard, manual/test)


mathetos:
Just tested with Stripe (test keys, but Give in Live Mode) and it failed. Here's the error:

`Warning: Missing argument 1 for give_purchase_form_required_fields(), called in ...\wp-content\plugins\give\includes\process-purchase.php on line 598 and defined in ...\wp-content\plugins\give\includes\process-purchase.php on line 274`

DevinWalker:
@mathetos can you check now?

mathetos:
Looks good now. Used a test credit card number and when Give was LIVE it declined the card, and when I put it in TEST the donation went through perfectly.

DevinWalker:
@mathetos thanks for checking stripe. The issue we're looking for is whether gateways that do NOT require billing fields (address/zip/country) that they do not incorrectly respond back with validation errors saying these fields are required (when they aren't even output). 

We need to confirm:

- [ ] PayPal Standard works
- [ ] Offline Works
- [x ] Stripe Works
- [ ] Manual / Test works
- [ ] PayPal Pro works
- [x] Braintree works
- [ ] Authorize.net works

mathetos:
Updated list: 
- [x] PayPal Standard works*
- [x] Offline Works*
- [x] Stripe Works
- [x] Manual / Test works
- [ ] PayPal Pro works
- [x] Braintree works
- [ ] Authorize.net works
- [ ] Dwolla works*

DevinWalker:
Confirmed Dwolla & PPP

DevinWalker:
Seems pretty solid. I'd like to hear from @pryley

DevinWalker:
Confirmed Authorize just now too

pryley:
It looks good to me.

You could also make $form_id optional in `give_field_is_required()` as it seems to be only the billing address fields that use it (?). That would mean getting the $payment_mode in `give_require_billing_address()` and then a check for only if form_id !== false...or something like that.

DevinWalker:
Thinking future-wise I think it'll be useful to pass the `$form_id` var for form-specific validation.

Issue #165, opened 2015-07-08
Shortcode Attributes


mathetos:
We should add a couple basic attributes in the shortcode and shortcode builder. Like:

- [x] goals="true/false"
- [x] form-title="true/false"
- [x] form-content="true/false"

DevinWalker:
Moving to 1.2 since we're going to be doing the goal shortcode there as well.

mathetos:
Updated list of attributes here:

- [ ] show_goals="true/false"
- [ ] show_title="true/false"
- [ ] show_content="true/false"
- [ ] display_style="onpage/reveal/modal"

DevinWalker:
Closing due to many PRs regarding this issue - somehow this issue wasn't closed in the merging of those PRs you see above.

PR #268, opened 2015-09-20
Add Shortcode Attributes to [give_form]

pryley:
Implements: #165 

- [x] show_goals="true/false"
- [x] show_title="true/false"
- [x] show_content="true/false"
- [x] display_style="onpage/reveal/modal"

### DIFF
--- a/includes/forms/template.php
+++ b/includes/forms/template.php
@@ -34,7 +34,6 @@ function give_get_donation_form( $args = array() ) {
 		$post_id = $args['id'];
 	}
 
-
 	$defaults = apply_filters( 'give_form_args_defaults', array(
 		'form_id' => $post_id
 	) );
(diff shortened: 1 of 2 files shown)


pryley:
`show_goal` and `show_content` attributes only take effect if the corresponding form options are enabled.

DevinWalker:
@pryley this looks really great. We *really* appreciate your PRs.

DevinWalker:
Hey @pryley I see we're passing the `$args` now through a few functions - Question: would it be better to swap out the `$form_id` for the `$args` param throughout? Then we can add a method for retrieving the ID based on the `$args`? 

Just thinking aloud here... would love to hear your thoughts.

pryley:
It sounds good, only caveat I can think of is that if you remove the $form_id from the do_action() hooks and only provide the form $args, someone hooking into them would likely have to do their own ID extraction logic.

PR #282, opened 2015-09-30
Hotfix/fix floatlabel on events

pryley:


DevinWalker:
Sweet

DevinWalker:
Checking it out

pryley:
The only problem I see now is if a user has a form showing twice on the same page (i.e. using the widget). When this happens, the form and form element ID's are duplicated which messes with the form events.

DevinWalker:
Yeah I am seeing issues too. I have it enabled globally and it doesn't seem to be taking effect like it should. Seeing strange behaviour

pryley:
That should do it.

DevinWalker:
Yup, just confirmed it's working great.

Issue #294, opened 2015-10-01
Form Contents does not show in Single Form


farookibrahim:
![image](https://cloud.githubusercontent.com/assets/10401366/10219247/ed45ed3a-685a-11e5-8c68-8451e1672452.png)

![image](https://cloud.githubusercontent.com/assets/10401366/10219261/059d9054-685b-11e5-8daa-5a1d48d95520.png)

pryley:
I got this.
It was introduced with the shortcode attributes feature because I neglected to test the single-forms (they were disabled in options), only the widgets and pages with a form shortcode.

PR #296, opened 2015-10-01
Fix: #294

pryley:
The shortcode feature did not take into account the single-form post_type view.


DevinWalker:
Hey @pryley thanks

PR #297, opened 2015-10-01
Show the shortcode button/dropdown conditionally

pryley:
Use the `give_shortcode_button_condition` hook to hide the Give shortcut button on Give Form editors, etc.


DevinWalker:
Looks good thanks - I didn't notice it was displaying.

PR #298, opened 2015-10-01
The previous pull-request was a little unfinished/borked (sorry) This should do it.

pryley:
Ok, the shortcode "show_content" attribute now accepts three values:
- none
- above
- below


DevinWalker:
Yeah I just noticed. Ok merging.

Issue #326, opened 2015-10-14
id="give-amount" output multiple times on the page


mathetos:
## A11Y Report Notes
http://b8d.3e8.myftpupload.com/
**DONATION FORM WIDGET**

* Error /  priority 65%
* Duplicate ID found
* (line: 138, 3)

`<input id="give-amount" class="give-amount-hidden" type="hidden" name="give-amount" value="5.00" required="" style="">`

**Web Content Accessibility Guidelines (WCAG) 2.0, Level A: 4.1.1 Parsing**

An ID attribute value has been used more than once. In cases where the ID is used to reference a UI control this is likely to cause issues for users of assistive technologies

## Possible Solutions
* Dynamically add the Form ID to the end
* Determine whether input actually requires an ID at all, perhaps not

mathetos:
@TheGeniusWM What's your take on this? Do input even need an ID at all? Are there accessibility issues with inputs having ID's? basically, maybe we can just remove the ID safely to avoid this problem and not adversely affect the styles or functionality at all. Let me know your thoughts.

TheGeniusWM:
Let me dig into this one before you work on it. I will comment with a definitive answer shortly.

TheGeniusWM:
Ok so it is recommended that all form inputs have proper labels, and that labels that exist on the same page need to be unique.
For someone using a screenreader, inputs with the same ID can get combined and inputs could potentially be missed,
A hidden label would be a valid fix. As long as the assistive technology can read it in the code we are good.

mathetos:
This should be resolved with PR #358

Issue #330, opened 2015-10-14
id="give-amount" is output twice on the page, once as an input, once as a span


mathetos:
## A11Y Report Notes
http://b8d.3e8.myftpupload.com/donations/basic-donation-form/
**Basic Donation Form**
* Error /  priority 65%
* Duplicate ID found
* (line: 127, 3)

`<input id="give-amount" class="give-amount-hidden" type="hidden" name="give-amount" value="5.00" required="" style="">`

**Web Content Accessibility Guidelines (WCAG) 2.0, Level A: 4.1.1 Parsing**
An ID attribute value has been used more than once. In cases where the ID is used to reference a UI control this is likely to cause issues for users of assistive technologies.

## Possible Solutions
* Change the input ID to `give-amount-input`
* Change the span ID to `give-amount-text`
* Determine whether either is necessary.

TheGeniusWM:
your possible solutions are valid fixes, go with it,

mathetos:
The span ID was necessary for styling of the fixed amount field. So I changed it to `give-amount-text` and added the styling to that new ID as well.

Issue #328, opened 2015-10-14
<input class="give-text-input" name="give-amount"> doesn't have a label


mathetos:
## A11Y Report Notes

http://b8d.3e8.myftpupload.com/donations/basic-form-with-custom-amount-enabled/
**Basic Form with Custom Amount Enabled**
* Error /  priority 89%
* Form element has no label
* (line: 130, 5)

`<input class="give-text-input" id="give-amount" name="give-amount" type="tel" placeholder="" value="20.00" required="" autocomplete="off">`

Provide a label for all controls. This allows all users to understand the type of information expected in the field.

## Possible Solutions
* Can we add a hidden label?
* Are there other methods for labelling the input without having the actual text spit out above the field that also suits the WCAG 2.0 Requirements?

Need Jordan's input on this one.

mathetos:
@TheGeniusWM We REALLY don't want a visible label on this input. Would this input be accessible if we added a hidden label?

TheGeniusWM:
A hidden label for any input is a valid fix, as long as each label is unique. As long as the assisstive technology device can recognize it, the label does not actually need to be visible.
It is recommended that every form input have a unique label.

PR #358, opened 2015-10-22
Added hidden label to input.give-text-input

mathetos:
This resolves accessibility issues for both #328 and #330


mathetos:
I believe it also resolved #326

PR #380, opened 2015-10-30
Refactored includes/forms/template.php to have unique IDs

### LINKED ISSUE #379: Make ALL the ids be unique to avoid multiple identical ids when multiple forms are on the page
Currently because many of the IDs are generic we'll always produce a not-fully-accessible page if there are even just two Give forms on the page at the same time. To avoid that we just need to add the form_id to each and every ID, or remove unnecessary IDs.

mathetos:
Fixes #379 

Also cleaned up markup to be more semantic. Required changes to JS files as well, so careful review would be necessary.


DevinWalker:
It's looking really good! I think the one thing that needs to be standardized is the underscore vs hyphen usage within the selectors. I think we should stick with the hypens as I a rarely see underscores used for classes/ids. Sound good? :)

mathetos:
@DevinWalker The PR is refreshed to make all IDs and Classes consistently use dashes instead of underscores. The "show terms" reveal link was also not working correctly previously, that's corrected in the refreshed PR.

DevinWalker:
@mathetos I also see you update the name attribute with hyphens which can cause issues with the way donations are processed. Let's discuss this PR together tomorrow.

includes/forms/template.php
```diff
@@ -455,20 +455,20 @@ function give_user_info_fields( $form_id ) {
...
 				<?php _e( 'First Name', 'give' ); ?>
 				<?php if ( give_field_is_required( 'give_first', $form_id ) ) { ?>
 					<span class="give-required-indicator">*</span>
 				<?php } ?>
 				<span class="give-tooltip give-icon give-icon-question" data-tooltip="<?php _e( 'We will use this to personalize your account experience.', 'give' ); ?>"></span>
 			</label>
-			<input class="give-input required" type="text" name="give_first" placeholder="<?php _e( 'First name', 'give' ); ?>" id="give-first" value="<?php echo is_user_logged_in() ? $user_data->first_name : ''; ?>"<?php if ( give_field_is_required( 'give_first', $form_id ) ) { echo ' required '; } ?>/>
-		</p>
+			<input class="give-input required" type="text" name="give-first" placeholder="<?php _e( 'First name', 'give' ); ?>" id="give-first-<?php echo $form_id; ?>" value="<?php echo is_user_logged_in() ? $user_data->first_name : ''; ?>"<?php if ( give_field_is_required( 'give-first', $form_id ) ) { echo ' required '; } ?>/>
+		</div>
```
DevinWalker: It's OK for name attributes to use underscores

includes/forms/template.php
```diff
@@ -490,9 +490,9 @@ function give_user_info_fields( $form_id ) {
 				<span class="give-tooltip give-icon give-icon-question" data-tooltip="<?php _e( 'We will send the purchase receipt to this address.', 'give' ); ?>"></span>
 			</label>
 
-			<input class="give-input required" type="email" name="give_email" placeholder="<?php _e( 'Email address', 'give' ); ?>" id="give-email" value="<?php echo is_user_logged_in() ? $user_data->user_email : ''; ?>"<?php if ( give_field_is_required( 'give_email', $form_id ) ) { echo ' required '; } ?>/>
+			<input class="give-input required" type="email" name="give-email" placeholder="<?php _e( 'Email address', 'give' ); ?>" id="give-email-<?php echo $form_id; ?>" value="<?php echo is_user_logged_in() ? $user_data->user_email : ''; ?>"<?php if ( give_field_is_required( 'give_email', $form_id ) ) { echo ' required '; } ?>/>
```
DevinWalker: Same here

includes/forms/template.php
```diff
@@ -853,42 +853,42 @@ function give_get_login_fields( $form_id ) {
...
 
-		<p id="give-user-pass-wrap" class="give_login_password form-row form-row-last">
-			<label class="give-label" for="give-password"><?php _e( 'Password', 'give' ); ?></label>
+		<div id="give-user-pass-wrap" class="give-login-password form-row form-row-last">
+			<label class="give-label" for="give-user-pass-confirm-form-id-<?php echo $form_id; ?>"><?php _e( 'Password', 'give' ); ?></label>
 			<input class="<?php if ( give_no_guest_checkout( $form_id ) ) {
 				echo 'required ';
-			} ?>give-input" type="password" name="give_user_pass" id="give_user_pass" placeholder="<?php _e( 'Your password', 'give' ); ?>" />
+			} ?>give-input" type="password" name="give-user-pass" id="give-user-pass-form-id-<?php echo $form_id; ?>" placeholder="<?php _e( 'Your password', 'give' ); ?>" />
 			<input type="hidden" name="give-purchase-var" value="needs-to-login" />
```
DevinWalker: Name attribute change here as well.

DevinWalker:
Closing this since it's out of date and incompatible now.

PR #406, opened 2015-11-15
#405 : SiteOrigin Widgets Bundle plugin compat

### LINKED ISSUE #405: Shortcode builder dialogs do not work when the "SiteOrigin Widgets Bundle" plugin is active
Will submit a PR.

pryley:
This changes the load sequence of the "mce_external_plugins" filter so it is triggered correctly when the SiteOrigin Widgets Bundle plugin is active.

Fixes #405


DevinWalker:
Thanks @pryley I'll take a look and merge it in

Issue #407, opened 2015-11-17
Warning: "Sorry, it appears the viewing window for this donation receipt has expired or you do not have the permission to view this donation receipt."


kurund:
Here are the steps to replicate the give plugin donation receipt warning
- Go to wordpress user list. 
- Create user with email id.
- Create donation form and make donation with the above email id without login.
- You will get warning message as the email id is already present in the user list.

Warning: "Sorry, it appears the viewing window for this donation receipt has expired or you do not have the permission to view this donation receipt."

DevinWalker:
Thanks for reporting this @kurund we'll take a look asap

andrejcremoznik:
This only happens with offline donations. I don't think there's anything wrong with it as you can't really show a recipt for an offline donation. I think the error message should be changed to a notification saying something along the lines of "Thank you for your donation. After we confirm the transaction we will email you the receipt."

ibndawood:
I know exactly what is causing the issue. The above error will be reproduced in the following case:
* **When a guest makes a donation using the email of a registered user.**

**The Reason**

* The code here : https://github.com/WordImpress/Give/blob/master/includes/shortcodes.php#L269
* Precisely this part which checks if a guest can view the receipt or not:  ``( ( $donor_id == 0 || $donor_id == '-1' ) && ! is_user_logged_in() && give_get_purchase_session() )``
* In our case the guest has a registered email address so `donor_id` will be equal to the `user_id` and hence ``( $donor_id == 0 || $donor_id == '-1' )`` fails right away causing the warning to be triggered.

I believe the condition ```( $donor_id == 0 || $donor_id == '-1' )``` does not serve its purpose of making sure that it is a guest and it should be removed.

cc: @DevinWalker

mathetos:
:+1:  @ibndawood

PR #416, opened 2015-11-22
Removed donor_id check for guest users before displaying donation receipt

ibndawood:
This is with reference to #407


DevinWalker:
That's I'm looking to clean this up a bit more than this fix. I'll show you the PR once that's complete

PR #420, opened 2015-11-24
Fixed textdomain and wrong use of escape functions

valeriosouza:


DevinWalker:
@valeriosouza thanks!

PR #459, opened 2016-01-11
Correct for/id relationship in credit card fields

joedolson:
$form_id wasn't being added to the far attribute for card number, name, and cvc fields, causing the label/form field relationship to be invalid. See also #379, which added the form IDs but appears to have missed a few.


DevinWalker:
Hey, thanks! :+1:

joedolson:
No problem. Glad to help.

Issue #17, opened 2015-03-13
Frontend List of Donors with Gravatars & Opt-in/out


mathetos:
Currently all Gravatar settings are global via the Settings panel. They should instead be available in each form for specificity.

DevinWalker:
I've disabled the Gravatar functionality until this can be hashed out a bit more. I agree it should be form specific.

DevinWalker:
Items needed for this:

- [ ] Donor opt-in/out of public display
- [ ] Columns attribute to display donors in 2/3/4/6 columns
- [ ] Form ID filter; if none specify pull globally

More consideration: Sidebar widget

DevinWalker:
Related to #83

Benunc:
Still regularly getting this request from users. is it still on track for this release?

DevinWalker:
We're creating a free Add-on for this

DevinWalker:
Follow along on the repo for this here: https://github.com/WordImpress/Give-Per-Form-Confirmation-Messages 

This will be a free plugin when released.

mathetos:
@DevinWalker I think you linked to the wrong Repo for this issue. Isn't this one going to be called "Display Donors"? Or something to that effect?

DevinWalker:
Yup, I fixed the link above.

DevinWalker:
Reopening this as I'd really like to see a "Donor Wall" in core.

Benunc:
Another request for this today. all requests will be tagged "anonymous donation" in help scout.

PR #460, opened 2016-01-12
Rename constructor method to __contruct

lots0logs:
Use of constructor methods with same name as class is deprecated.


lots0logs:
@DevinWalker *bump* :innocent:

DevinWalker:
We don't really have this working into the core plugin yet... pending #17 but this makes sense so it's getting merged :)

lots0logs:
Cool. My debug log thanks you :wink:

Issue #390, opened 2015-11-09
Ability to set a minimum donation amount 


DevinWalker:
**Problem:** Can I set the minimum price for a donation? For example, I want the min donation at $5 and they also can donate custom amount but not below $5.

Reported via Help Scout request. I think it's an interesting request that we should consider adding to the form functionality. @mathetos what do you think?

mathetos:
I agree that it would be useful. But there should be an alert of some sort to alert the donor that a minimum is in place. Perhaps just a small description under the amount field. 

We get this request often enough I think it would be valuable.

pryley:
This is an important feature. Sometime a donation form will be used to test multiple credit card numbers, each time attempting to process a tiny donation (i.e. $3-$5). The problem with PayPal Pro (and maybe other gateways?) is when these transactions are appealed, flagged as fraudulent and then reversed, Paypal adds a $20 reversal processing fee. Having a minimum donation amount discourages this type of behaviour.

mathetos:
That's a really good point @pryley particularly since we have heard reports from a few users saying they are getting those kinds of automated attacks on their forms. Combining this with a honeypot for security should be a priority for a near-future version I think.

pryley:
A couple articles about this:

https://philanthropy.com/article/Fraud-Alert-Criminals-Test/233197
http://www.bankinfosecurity.com/stolen-cards-tested-on-charity-sites-a-7608/op-1

DevinWalker:
This has been merged via https://github.com/WordImpress/Give/pull/498

PR #489, opened 2016-02-23
Issue/390

pryley:
#390 

image

A couple notes:

1. There is no detection to check if the set amount is less than the custom amount minimum on Edit/Quick Edit form. I decided it was better to leave this in the user's court and only inform the user via the "Set Amount" label on the Edit Form page to make sure the set amount was greater than the custom amount minimum.

2. I'm not convinced that setting the `$minimum_price` variable in the `Give_Donate_Form` object is the correct thing to do. But it's related at times to the form's `$price` so maybe it's fine.


DevinWalker:
Thanks I'll check it out!

DevinWalker:
I wonder why scrutinizer screwed up

pryley:
I merged your recent changes and adjusted `give-checkout-global` to work with it.

DevinWalker:
Updated PR via: https://github.com/WordImpress/Give/pull/489

PR #491, opened 2016-02-23
Detect card-number class instead of ID

### LINKED ISSUE #490: Card type image not showing when floating-labels is enabled
image

pryley:
fixes #490

The only files changed are `.gitignore` and `give.js`, the rest is due to gulp.

`.gitignore` was changed to allow switching branches without removing the  node_modules and bower dirs.


DevinWalker:
See https://github.com/WordImpress/Give/pull/491

PR #515, opened 2016-03-10
Show "cancelled" transactions on Transaction page

### LINKED ISSUE #514: Transactions with a "cancelled" status are not shown in the wp-admin Transactions table
The `cancelled` status was introduced I think in preparation for the recurring donations addon.

**What it looks like:**

image

**What it _should_ look like:**

image

pryley:
Fixes #514 

@DevinWalker please review.


DevinWalker:
Just tested and it works well - thanks for your effort here.

PR #517, opened 2016-03-10
Adding unit tests for the Give_Donate_Form class

cklosowski:
Just adding the beginnings of some unit tests for the `Give_Donation_Form` class. fixed `is_single_price_mode` to return a default value like the `has_variable_prices` does.


DevinWalker:
Sweet! Thanks for your contribution @cklosowski

cklosowski:
@DevinWalker Anytime! Let me know what else we can add here if you want more in it.

PR #519, opened 2016-03-11
Check $_POST for enabled gateways before $give_options

### LINKED ISSUE #512: You have to ( save + reload ) after enabling a gateway in order to select it as the default
**Problem:** When enabling a gateway, you have to reload the page after saving the settings in order to select it in the `Default Gateway` select dropdown. Same goes for when you disable a gateway, once you save the settings the gateway is still in the `Default Gateway` select dropdown until you refresh the page.

image

pryley:
Fixes #512


DevinWalker:
Thanks dude! Nice and elegant. Me likey.

Issue #529, opened 2016-03-16
$0 set donation won't 'stick' in dashboard


DevinWalker:
As reported here: https://wordpress.org/support/topic/0-set-donation-wont-stick-in-dashboard?replies=1

PR #540, opened 2016-03-23
Allow for $0 set donations to be saved properly #529

DevinWalker:
Also fixed issue with new Goal Format field not being properly hidden/shown in admin #


DevinWalker:
#529

PR #552, opened 2016-03-30
Issue/521

### LINKED ISSUE #521: Session cookies are created on every page load preventing various caches
Reported by several users including here: https://wordpress.org/support/topic/session-cookie-set-on-every-page-load-causes-cache-miss?replies=1

```
It looks like Give sets two PHP session cookies on every page load as long as the plugin is merely installed and active. This prevents Varnish from working in some Managed WordPress hosting and visitors never get cached pages as a result. With any kind of load from concurrent users, overall performance degrades badly. Do you have a workaround for situations like this?
``` 

Related to: Session cookies are created on every page load: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/4284

DevinWalker:
Fixes #521


DevinWalker:
@ibndawood it would be great to get your eyes on as well. See #521 for how to test if you can

DevinWalker:
Testing is going pretty well with both PHP Sessions on and off... nearing ready for merge to master

PR #553, opened 2016-03-30
Issue/496

### LINKED ISSUE #496: Refresh Sessions or Login Functionality when expired sessions for receipt links & r
**Scenario:** A donor receives the confirmation email with the receipt link and clicks it after the expired time.

**Problem:** The donor sees the error message: 

```
Sorry, it appears the viewing window for this donation receipt has expired or you do not have the permission to view this donation receipt.
```

This isn't very user friendly at all and is confusing and frustrating for donors. We need to give the donor the option to either log in or receive an updated link via the email used when they gave. Once they login, they should be properly redirected back to the receipt page.

DevinWalker:
Fixes #496


DevinWalker:
@ibndawood it would be great to get your eyes on this see #521 for how to test if you can

ibndawood:
@DevinWalker , I will take a look into this a little later today and get back to you

PR #625, opened 2016-05-24
Fix bug on current_url retrival

Rahe:
Fix bug for https://github.com/WordImpress/Give/issues/622

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -198,7 +198,7 @@ function give_get_current_page_url() {
 	if ( is_front_page() ) {
 		$current_url = home_url( '/' );
 	} else {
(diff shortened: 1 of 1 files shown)


DevinWalker:
Thanks @Rahe 

It looks like this is breaking some unit tests though... https://travis-ci.org/WordImpress/Give/builds/132518886

If you have a chance can you look into that? Otherwise, I'll have to check it out soon.

Rahe:
Wrong variable name :( I Will change the PR

Rahe:
Hello,

I have fixed the variable name ;)

DevinWalker:
Thanks for fixing this... I'm wondering do we need the `untrailingslashit` ? It's breaking some unit tests...

Rahe:
Hello, this is like WordPress is doing it 
https://github.com/WordPress/WordPress/blob/master/wp-includes/nav-menu-template.php#L590

So the tests shouldn't fail :/

PR #643, opened 2016-06-08
Add a hook

Benunc:
Adds a hook between donation amount and amount choices, for users who want to place something there (several requests to put the recurring checkbox there).

Not sure the logistics/best practice of adding it to the $output but it worked in my tests.


DevinWalker:
Hmm, let's discuss this tomorrow. I think there may be a better route to go.

Benunc:
sounds good.

PR #661, opened 2016-06-28
Corrects instances of "products" in comments

Benunc:


DevinWalker:
Sweet! Can you also remove ` * @hooked wc_print_notices - 10`

Benunc:
I just removed that, as well

DevinWalker:
Thanks

PR #700, opened 2016-07-07
Adds "forgot password" link to login form.

Benunc:
Proof-of-concept based on support ticket user feedback. https://secure.helpscout.net/conversation/224241808/2731/?folderId=672199

### DIFF
--- a/includes/forms/template.php
+++ b/includes/forms/template.php
@@ -955,6 +955,12 @@ function give_get_login_fields( $form_id ) {
 			<input type="hidden" name="give-purchase-var" value="needs-to-login"/>
 		</div>
 
+		<div id="give-forgot-password-wrap-<?php echo $form_id; ?>" class="give_login_forgot_password">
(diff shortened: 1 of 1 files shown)


ramiy:
You should change the "**Forgot Password**" text with "**Reset Password**".

We are trying to do the same in WordPress core (https://core.trac.wordpress.org/ticket/35736).

DevinWalker:
Hey @Benunc I just checked this out and the link should open in a new window so that the donor is not taken away from the donation page. Optimally we would load the reset password form via AJAX. I'll open up an issue for this...

DevinWalker:
@Benunc I have created the issue for AJAX loading the password reset form itself here: https://github.com/WordImpress/Give/issues/723 

I figure we can add this base functionality in now until we add the AJAX loading of the reset form later. I will also update the text to read "Reset Password" as @ramiy suggested.

PR #707, opened 2016-07-11
i18n Improvments

ramiy:
Avoid using HTML tags in translation strings.


ramiy:
The second change is replacing this translation string:
"**%1$sView the receipt in your browser %2$s**"

With that:
"**View the receipt in your browser**"

DevinWalker:
Thanks this is great!

PR #710, opened 2016-07-11
More i18n improvments

ramiy:
Like the previous pull request, this one has many more needed i18n fixes.


DevinWalker:
Awesome, thanks!

Issue #451, opened 2015-12-23
Fix translatable strings according to GlotPress


DevinWalker:
Now that we are added on GlotPress I'm seeing issues with the strings that are being translated. We should be providing descriptions using `_x()` https://codex.wordpress.org/Function_Reference/_x more often and reduing the amount of HTML within the strings found here: https://translate.wordpress.org/projects/wp-plugins/give

DevinWalker:
Closing this due to it being a vague issue and all the recent PRs by @ramiy https://github.com/WordImpress/Give/pull/710 and https://github.com/WordImpress/Give/pull/707

PR #732, opened 2016-07-14
Few more i18n improvments

ramiy:
This time I'm fixing the _x(). Translations with context. Update context text in some places, remove context in other places, and few extra translation string updates.

### DIFF
--- a/includes/admin/class-give-settings.php
+++ b/includes/admin/class-give-settings.php
@@ -592,7 +592,7 @@ class Give_Plugin_Settings {
 						array(
 							'id'      => 'donation_subject',
 							'name'    => __( 'Donation Email Subject', 'give' ),
-							'desc'    => __( 'Enter the subject line for the donation receipt email', 'give' ),
+							'desc'    => __( 'Enter the subject line for the donation receipt email.', 'give' ),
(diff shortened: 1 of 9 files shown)


ramiy:
The last commit for similar strings based on:

DevinWalker:
Thanks @ramiy - Related to #451

DevinWalker:
How come you're removing usage of `_x()` string explanations? Do you think they add more confusion for translators?

ramiy:
one more:

ramiy:
As for the use of _x() function:

In WordPress core, we use this function when a string is used in many places, but we want one of those strings to have a different meaning. When we add the context function, this string will appear as a separate string in translate.wordpress.org .

ramiy:
one more:

ramiy:
Another similar string merge:



@DevinWalker Here you can see the string with the unneeded context.

DevinWalker:
Oh yeah, I see that. Thanks for the example

ramiy:
few more:





Merge similar strings and remove context functions.

DevinWalker:
This is great, thanks a lot for the effort here.

PR #737, opened 2016-07-15
Permission Error Messages

ramiy:
This pull request is unifying different permission error messages:

Source:
perm

Note: permission error messages fired by current_user_can() when the user has no capability to do something.


DevinWalker:
👍

PR #740, opened 2016-07-16
Text Changes: periods

ramiy:


DevinWalker:
Nice, thanks! All hail the great `.`

PR #742, opened 2016-07-18
adds form id to terms link class

### LINKED ISSUE #741: Multiple modal popups with "show terms" on one page lead to JS issues.
Customer reported here: https://secure.helpscout.net/conversation/227726404/2868/?folderId=672194
Steps to replicate:

1. Create at least two forms with payment fields set to "Modal Window Upon Click", and Terms and conditions set to "Yes."
1. Insert those two forms into one page using a shortcode.
1. Click the "Donate Now" button, and the "show terms" link.

Terms and conditions will open/close for the number of times you've included shortcodes for forms with modal windows with terms available.

Upon further investigation, it appears the issue shows up regardless of modal window settings. so in step one, you could replicate with any form with Terms and conditions set to "yes"

Benunc:
fixes js conflict with multiple forms on a page. 

fixes #741

### DIFF
--- a/includes/forms/template.php
+++ b/includes/forms/template.php
@@ -1063,8 +1063,8 @@ function give_terms_agreement( $form_id ) {
 				?>
 			</div>
 			<div id="give_show_terms">
-				<a href="#" class="give_terms_links-<?php echo $form_id;?>"><?php _e( 'Show Terms', 'give' ); ?></a>
-				<a href="#" class="give_terms_links-<?php echo $form_id;?>" style="display:none;"><?php _e( 'Hide Terms', 'give' ); ?></a>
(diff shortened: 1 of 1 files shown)


DevinWalker:
@Benunc rather can you leave the class the same and add an `id` with the same to the link? That way anyone who has css/js attached to that class name won't break.

Benunc:
ok, @DevinWalker made those changes and tested. all appears functional on my end.

DevinWalker:
Ah, crud - you can't have two IDs on the same page for accessibility reasons and valid HTML markup. I didn't notice there were two links attached to the single js function. So, how about this - remove the ID and add an addition class with the form ID attached to it. Sound good?

Benunc:
How's that, @DevinWalker ?

DevinWalker:
Nice! 👏

PR #745, opened 2016-07-19
i18n Improvments

ramiy:
More i18n Improvments:

1. Use `__()` in two more strings, making them translatable to other languages.
2. use sprintf() to show the `%s` placeholder. Not in all languages the tag-list is in the end. This why we should use placeholders. Translation editor will decide where to place the placeholder.


DevinWalker:
Nice!

mathetos:
So great! Thanks @ramiy !

PR #755, opened 2016-07-19
fixes unrendered html in email subjects

### LINKED ISSUE #754: Unrendered html in multi-level donation confirmation emails.
Steps to replicate:

1. set a `{donation}` email tag in an email subject (it is in the multi-form offline donation template by default)
1. process a transaction on a multi-level form with a title for the selected level.

The email sent will have unrendered html in the subject.

Benunc:
fixes #754

### DIFF
--- a/includes/emails/class-give-email-tags.php
+++ b/includes/emails/class-give-email-tags.php
@@ -525,7 +525,7 @@ function give_email_tag_receipt_id( $payment_id ) {
  */
 function give_email_tag_donation( $payment_id ) {
 	$payment    = new Give_Payment( $payment_id );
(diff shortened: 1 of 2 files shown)


DevinWalker:
Hey @Benunc good catch here. Thanks. Can we use `strip_tags()` rather than two if statements?

`if($is_email){  strip_tags($form_title); }`

Benunc:
good call. We could probably just strip the tags in class-give-email-tags.php and leave the other file alone.

Will investigate

Benunc:
Gonna submit a much simpler one.

PR #762, opened 2016-07-19
strips the tags from the donation form

### LINKED ISSUE #754: Unrendered html in multi-level donation confirmation emails.

Benunc:
much simpler. fixes #754


DevinWalker:
Noice!

Issue #781, opened 2016-07-21
esc_html doesn't expect a textdomain


DevinWalker:
The WP core function `esc_html` only expects one parameter yet we are passing two with the text domain here: https://github.com/WordImpress/Give/blob/release/1.6/includes/scripts.php#L254%23L271

I believe https://codex.wordpress.org/Function_Reference/esc_html_2 should be used in place.

Please check uses of `esc_html` due to merge of #776

ramiy:
Good catch!

DevinWalker:
Resolved with PR merge #811

PR #782, opened 2016-07-21
i18n: use escape functions with text-domains

ramiy:
Fixing #781


DevinWalker:
Can you submit this to the `release/1.6 branch for me?

PR #786, opened 2016-07-22
Makes country names translatable

Benunc:
Per customer request: https://secure.helpscout.net/conversation/229632064/0/


DevinWalker:
Looks good!

Issue #736, opened 2016-07-15
RTL Audit & Resolution


ramiy:
See the attached screenshot:

![give add ons page rtl issues](https://cloud.githubusercontent.com/assets/576623/16879155/fe743946-4ab7-11e6-85e1-082cd6f0715c.png)

ramiy:
The plugin needs and RTL audit. Many pages are missing RTL view.

ramiy:
![rtl](https://cloud.githubusercontent.com/assets/576623/16890257/9e572792-4af5-11e6-87ed-d43c78f574c2.png)

ramiy:
Front end issues:

![front1](https://cloud.githubusercontent.com/assets/576623/16893815/3c564940-4b4c-11e6-8a34-b1f0807e2972.png)

![front2](https://cloud.githubusercontent.com/assets/576623/16893818/43f7025c-4b4c-11e6-9b02-51ca5d40496d.png)

ramiy:
![addons-rtl](https://cloud.githubusercontent.com/assets/576623/17086534/6362218c-51fd-11e6-9922-72bed5807bbc.png)

Fixed RTL addons page

DevinWalker:
There are a few lingering issues here after merging in #796 

Still an issue:

![3c564940-4b4c-11e6-8a34-b1f0807e2972](https://cloud.githubusercontent.com/assets/1571635/17112813/7a4dddfa-525c-11e6-833e-21f1535e68a5.png)

ramiy:
@DevinWalker yes I know.

I fixed the admin issues first, it's easier. Now I'm moving to the FrontEnd.

DevinWalker:
Thanks for the update @ramiy it's looking good - I can almost read RTL now 👀

ramiy:
@DevinWalker,

For now the RTL is working. Not 100% but I guess it fixes over 95% of the cases.

Currently, **rtl.scss** is updated manually. In the future we will need to do that automatically using a gulp taks.

DevinWalker:
Hey @ramiy I just reviewed #812 and notice a few items:


1. Should the border and icon be moved to the right for all notices?
![rtl1](https://cloud.githubusercontent.com/assets/1571635/17154902/cd3d7f76-5337-11e6-9d7a-6c98a2290653.png)

2. This border appears to be missing:
[...]

PR #788, opened 2016-07-24
RTL

ramiy:
#736


DevinWalker:
Thanks @ramiy I have checked it out and it's looking good. I made an RTL stylesheet for the frontend called `rtl.sass` and also updated a few styles for the tooltip content and improved the padding and margins for the gateway radios.

DevinWalker:
The new PR with your commits attached: #796

Issue #471, opened 2016-02-03
Incorporate proper sanitization for i18n strings


DevinWalker:
@Shelob9 has pointed out some potential security concerns with using `__(` rather than `esc_html_` for internationalization strings. We should follow suit. Here's a good article: https://www.chriswiegman.com/2015/12/sanitization-is-for-translations-too/

DevinWalker:
This isn't really a bug.. removed label

ramiy:
Yes, some hackers target po/mo files, injecting JS code.

When you properly escape your translation strings, you prevent the script from loading.

Another method to harden your plugin.

ramiy:
More info at:

https://codex.wordpress.org/Validating_Sanitizing_and_Escaping_User_Data#Escaping:_Securing_Output

DevinWalker:
@ramiy I'm seeing some HTML being displayed now after merging: https://github.com/WordImpress/Give/pull/776

Can you double-check for me that the strings using `esc_html` do not contain any html themselves?

![screen shot 2016-07-20 at 11 14 28 am](https://cloud.githubusercontent.com/assets/1571635/16997684/5c1bdcaa-4e6b-11e6-9e52-a2b6903dbdd9.png)

PR #800, opened 2016-07-25
i18n sucurity: escaping translation strings #471

ramiy:


DevinWalker:
@ramiy can you resolve the conflict for this PR for me?

PR #807, opened 2016-07-26
adds filter for offline donation labels on front and backend.

Benunc:
We've had multiple requests to change this text from customers.


DevinWalker:
Hey there's a snippet for that, it's already being passed through a filter called `give_payment_gateways`:

https://github.com/WordImpress/Give-Snippet-Library/blob/master/gateway-customizations/customize-payment-gateway-labels.php

PR #812, opened 2016-07-26
Give frontend RTL

ramiy:
See #736


DevinWalker:
I added a few items I noticed here: #736

DevinWalker:
Looking good! Thanks @ramiy 🎉

Issue #820, opened 2016-07-28
Accessibility Headings


ramiy:
See: https://make.wordpress.org/plugins/2015/08/03/4-3-change-to-plugin-dashboard-pages/
and: https://core.trac.wordpress.org/ticket/31650

We need to update give.

ramiy:
https://github.com/WordImpress/Give/pull/819/commits/ab453d70389bdb45f1decb3cb5ddb895e6e1f825
https://github.com/WordImpress/Give/pull/819/commits/feb0cc86ff565aaddc87c276e8e9167d6a759d29

ramiy:
Also, at the end of the year, we will need to apply this change: https://core.trac.wordpress.org/ticket/37257

ramiy:
@DevinWalker, you can tag this as `a11y` and `UI/UX`. Add 1.6 milestone. And close the issue. PR https://github.com/WordImpress/Give/pull/819 was merged.

DevinWalker:
Yep, thanks for the reminder @ramiy

PR #829, opened 2016-08-01
a11y: use WordPress default screen-reader class

ramiy:
You added `give-hide-heading` to hide `<H1>` titles in the admin (#820). We should use the default WordPress class for screen readers.

For more info, read this: https://make.wordpress.org/accessibility/2015/02/09/hiding-text-for-screen-readers-with-wordpress-core/

### DIFF
--- a/assets/scss/admin/settings.scss
+++ b/assets/scss/admin/settings.scss
@@ -57,11 +57,6 @@ Settings
 
 }
 
-// Section Titles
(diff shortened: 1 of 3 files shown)


DevinWalker:
I did not know about this - cool! 👍

ramiy:
That is why I'm here. :-)

Issue #274, opened 2015-09-23
Email Previews & Add-on Specific Email Tags Don't Display


DevinWalker:
**Scenario:** A user purchases the Give PDF add-on which adds the {pdf_receipt} email tag and then attempts to preview the email. 

**Problem:** The email preview only outputs the tag and therefore the user thinks the plugin doesn't work unless they actually test out the functionality themselves by running a test donation - hardly idea. 

**Spawned from:** https://givewp.com/support/ticket/pdf-receipt-shortcode-not-generating-link-in-emails/#post-5290

PR #834, opened 2016-08-01
Multiple enhancements to the donation receipt preview functionality #274

DevinWalker:
Donation receipt email previews now allow you to view actual payment data and switch between transactions. Add-on specific email tags will render when actual data is previewed. - https://github.com/WordImpress/Give/issues/274


DevinWalker:
Dang - wrong branch :(

PR #858, opened 2016-08-10
a11y in give shortcode

ramiy:
Remove unneeded title attribute.

### DIFF
--- a/templates/shortcode-login.php
+++ b/templates/shortcode-login.php
@@ -38,7 +38,7 @@ if ( ! is_user_logged_in() ) {
 			</div>
 
 			<div class="give-lost-password give-login">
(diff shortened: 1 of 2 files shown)


DevinWalker:
Thanks! Can you leave the "title" attribute on the `anchor` tag? It's cool to remove it from the `inputs`

ramiy:
We can remove from `inputs` because the screen reader has the context in the `label`.

As for the `a` tag, when title is equal to the content inside the `a` tag, the screen reader reads the same text twice. In our case, it can be removed.

DevinWalker:
I gotcha, I was thinking for typical users when they hover the anchor and see the browsers tooltip that it's pretty helpful.

Issue #866, opened 2016-08-10
Make transaction columns "Donation Form" and "Status" sortable


DevinWalker:
These two columns need to be sortable:

![screenshot at aug 10 16-46-43](https://cloud.githubusercontent.com/assets/1571635/17574966/3c7922c6-5f1a-11e6-9163-a6d700bd88da.png)

ramiy:
You have **Status Links** above the table... :-)

ramiy:
I've played around with the UI, and it looks like it's a very simple fix - only 2 lines of code to add the sorting functionality.

![sorting](https://cloud.githubusercontent.com/assets/576623/17648542/2a58a9e4-6222-11e6-8326-c64585fdc056.png)

DevinWalker:
@ravinderk is saying this isn't working for him, so we're going to QA it together before closing again.

PR #886, opened 2016-08-15
Make transaction columns "Donation Form" and "Status" sortable

ramiy:
Resolving issue #866 .

See screenshots in the issue page.


DevinWalker:
This appears to allow sorting, but the sorting itself for the `Donation Form` and `Status` columns appear incorrect.

Issue #887, opened 2016-08-15
Transaction page improvements


ramiy:
Single transaction page in the admin need accessibility love.

- [x] **Update Payment** - use `label` tags.

- [x] **Payment Meta** - use `strong` tags.

- [x] **Donation Information** - replace the table with other a column based layout like the billing address metabox #767 .

- [x] **Donor Details** - rethink the layout.

----

Note: this screen is not responsive.

ramiy:
Current:
![accessibility-donations-screen](https://cloud.githubusercontent.com/assets/576623/17662578/180d66bc-62ef-11e6-89d7-5d4471291fa4.png)

New:
![accessibility-donations-screen-new](https://cloud.githubusercontent.com/assets/576623/17662587/27e939bc-62ef-11e6-9448-749bea307c0c.png)

ramiy:
Rethink the **Donor Details Metabox**.

Old:
![donation-details-old](https://cloud.githubusercontent.com/assets/576623/17717941/c479b984-6419-11e6-8031-35639eb0699a.png)

New:
![donation-details-new](https://cloud.githubusercontent.com/assets/576623/17717947/cac117ba-6419-11e6-9c6f-e86069d36a6c.png)

Changes:
* The 1st column is displaying the current donor as a simple text, with a link to the donor page.
* The 2nd column allows you to change to other donor from the existing donors list.
* The 3rd column allows you to create a new donor.

What do you think?

mathetos:
@ramiy I like this idea, but I might take it one step further and just show the Donor name as a link -- like you suggested, then next to it another smaller link that says "Edit Donor" which would then show those additional fields. No need to clutter that metabox with additional information unless it's needed.

ramiy:
Additional donor information is a good idea, make the metabox more useful. As for the "Edit Donor", the 3rd column looks empty when this is the only link there.

How about this:
![donation-details-new2](https://cloud.githubusercontent.com/assets/576623/17725106/ce62cf1c-6450-11e6-9a37-43f80a33a3af.png)

PR #888, opened 2016-08-15
a11y: transaction page

ramiy:
Resolving #887 .

Before:
accessibility-donations-screen

After:
accessibility-donations-screen-new


DevinWalker:
Looks good, thanks

PR #894, opened 2016-08-15
Docs: document action hooks

ramiy:
I think I finished documenting all the action hooks.


DevinWalker:
So great

ramiy:
My mistake, found much more hooks...

Issue #897, opened 2016-08-16
Dynamic Hooks


ramiy:
I think we should gradually adopt wordpress core standards.

Dynamic hooks like:

`do_action( 'give_hook_name_' . $dynamic_field );`

Should be replaced with:

`do_action( "give_hook_name_{$dynamic_field}" );`

One more step to **Give Developer Resources**, just like **[WordPress Code Reference]( https://developer.wordpress.org/?s=%7B&post_type%5B%5D=wp-parser-hook)**.

DevinWalker:
Sounds good. I'm fine with that.

ramiy:
@DevinWalker,

You closed the issue before I had the chance to commit the fix.

DevinWalker:
It's not in here? https://github.com/WordImpress/Give/pull/899/files

DevinWalker:
More PRs, reopening

ramiy:
All actions and filters updated.

PR #899, opened 2016-08-16
Document action hooks

ramiy:
More docs, and implementing #897 .

I need to mention that the `init` uses `give_action` and the `admin_init` uses `give-action`. Mistake?


DevinWalker:
🎉

PR #912, opened 2016-08-17
Terminology: Transaction > Donation

ramiy:
issue #896


DevinWalker:
This is looking niceeee 🎉

mathetos:
@DevinWalker Has this been tested against any Add-ons? For example, PDF Receipts has a {transaction_id} email tag:
https://givewp.com/documentation/add-ons/pdf-receipts/

ramiy:
@mathetos email tags are set using a separate class. This PR updated only the translation strings and some of the phpDocs. We haven't updated class/function names, yet.

Issue #915, opened 2016-08-19
Id columns


ramiy:
The "logs" screen has a beautifully design id column:

![id-column](https://cloud.githubusercontent.com/assets/576623/17800181/9035c09c-65e9-11e6-9366-a4f593a5e77f.png)

I think we need to add id columns to other screens.

- [x] Donors
- [ ] Donations / Transactions
- [ ] Forms

ramiy:
**DONORS**

Before:

![donors-id](https://cloud.githubusercontent.com/assets/576623/17800816/ea4e4d2e-65ee-11e6-8c5b-899671795e31.png)

After:

![donors-id2](https://cloud.githubusercontent.com/assets/576623/17800826/f9171f34-65ee-11e6-9c8a-c1d295727f32.png)

ramiy:
**Donations / Transactions**

Before:

![donation-old](https://cloud.githubusercontent.com/assets/576623/17808896/73071278-661b-11e6-854b-9fc29459350a.png)

After:

![donation-new](https://cloud.githubusercontent.com/assets/576623/17808899/79ee8f76-661b-11e6-89e3-feea46c23976.png)

Before committing, I need your feedback on this screen.

ramiy:
Or maybe moving more links to the "Actions Row" ?

![donation-new2](https://cloud.githubusercontent.com/assets/576623/17809369/bbcfda6e-661e-11e6-82aa-01f8779fe227.png)

What do you think?

DevinWalker:
Hey @ramiy we actually just redid the transaction / donations screen here https://github.com/WordImpress/Give/issues/814 

The `id` isn't very important to the users. While I do think it looks nice on the logs screen, I think we're OK on the others.

PR #916, opened 2016-08-19
ID columns

ramiy:
Issues #915


DevinWalker:
See my note https://github.com/WordImpress/Give/issues/915

PR #971, opened 2016-08-29
Fix typo: "login" is a noun or adjective. "log in" is the verb.

Benunc:
The irony is not lost on me that the commit text has a typo in it.


DevinWalker:
Makes sense to me!

PR #974, opened 2016-08-29
Make exception message translatable

ramiy:


ramiy:
@DevinWalker The CMB2 strings are also updated but not in the Give plugin files, but directly in the CMB2 source code. See this PR: https://github.com/WebDevStudios/CMB2/pull/724

PR #975, opened 2016-08-29
Issue/962

### LINKED ISSUE #962: Change some settings to properly reflect Enable/Disable Correctly
Throughout the plugin we often mixed the purposes of settings between Enable/Disable, or Show/Hide in ways that can cause confusion.

Here's one example:
enable-disable-settings-confusion

Three different settings, checking one enables, while checking another disables. 

I propose something like the following instead:
enable-disable-settings-suggestion

This would apply to every setting that uses language like True/False or On/Off, Show/Hide, Enable/Disable. Consistency in our settings is important to the user experience.

mathetos:
Resolves #962 

But creates a new backward compatibility issue which I'll create a new issue on now.

### DIFF
--- a/includes/admin/class-give-settings.php
+++ b/includes/admin/class-give-settings.php
@@ -752,7 +752,7 @@ class Give_Plugin_Settings {
 						array(
 							'id'      => 'recaptcha_secret',
 							'name'    => esc_html__( 'reCAPTCHA Secret Key', 'give' ),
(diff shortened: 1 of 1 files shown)


DevinWalker:
I agree this needs to be standardized, but before we do I want to merge in @ravinderk's work on the settings before this PR. Closing until that's merged.

Issue #943, opened 2016-08-23
New Email tag for just the donation title without level


Benunc:
Here's the customer ticket that spawned this: https://secure.helpscout.net/conversation/241888682/3344/?folderId=672199

Option 1: create a different tag that "opts out" of the donation level provided by `give_get_payment_form_title()` 

Option 2: filter the entire `give_email_tag_donation` function's output to give developers control over it.

DevinWalker:
How about a snippet that uses the `give_email_tags` filter?

DevinWalker:
For record, here's what the issue says:

> In Settings>Emails, for the donation receipt emails, I entered:
> 
> I saw that you made a donation of {price} to {donation}.
> 
> With credit card donations, the donor gets this with an added hyphen, which isn't much of a problem:
> 
> I saw that you made a donation of $1 to PASA - .
> 
> But with PayPal donations, if the donor clicks the button to choose their own amount, the thank-you letter says
> 
> I saw that you made a donation of $1 to PASA - Type in your own amount.
> 
> Is there a way to get rid of the "- Type in your own amount"?

ramiy:
Source:
https://github.com/WordImpress/Give/blob/master/includes/emails/class-give-email-tags.php#L519-L534

ramiy:
### Issue
The email-tag retrieves the form title using `give_get_payment_form_title()` function. This functions accepts 3 parameters:
* (array) $payment
* (bool) $level_title - default false
* (string) $separator - default empty

In this email-tag, [we force the separator `'-'` but we hide the level](https://github.com/WordImpress/Give/blob/master/includes/emails/class-give-email-tags.php#L530). Strange behavior. We need to remove the `'-'`.

### Solution
Remove the `'-'` from the email-tag.

ramiy:
@DevinWalker @Benunc need feedback and testing.

ramiy:
Or, we can check if the form has variable prices and only then output the level.

Replacing the current code:

```
function give_email_tag_donation( $payment_id ) {
	$payment    = new Give_Payment( $payment_id );
	$form_title = strip_tags( give_get_payment_form_title( $payment->meta, false, '-' ) );
	return ! empty( $form_title ) ? $form_title : esc_html__( 'There was an error retrieving this donation title.', 'give' );
}
```

With:

```
function give_email_tag_donation( $payment_id ) {
	$payment      = new Give_Payment( $payment_id );
	$payment_meta = $payment->meta;
[...]

PR #979, opened 2016-08-30
Issue/943 - Remove the title seperator from the {donation} email template tag

ramiy:
Resolve issue #943


DevinWalker:
This looks good thanks - can we roll this into 1.6.1, though?

PR #988, opened 2016-08-30
Issue/943 - Fix the title level text in {donation} email template tag

ramiy:
Resolve issue #943 (master branch)


Benunc:
This bug appears to be back in 1.6.3. I replicated it, and a customer did as well.

PR #1004, opened 2016-09-05
Issue/1003 - Remove "Form Labels" functions from translation strings

### LINKED ISSUE #1003: Remove "Form Labels" functions from translation strings
Give has those two functions: `give_get_forms_label_singular()` and `give_get_forms_label_plural()`.

**From a developer point of view** it's easier to update the labels in one place and to see all the labels in all the strings updated accordingly.

BUT, **from a translator point of view** it can be a nightmare. We use `%s` placeholder in translation strings for the form labels. But it's a bad practice. In many cases it's makes it hard to understand what the plugin is trying to say and how to translate the string.

Two examples I encountered when translating to Hebrew:
* When using placeholders, in some languages it's hard to translate the string without the gender of the placeholder.
* In other translation strings, you use singular placeholders (for English), but in other languages we should use plural placeholders.

I'm not saying we need to deprecate those functions, I'm just saying we need to remove them from translation strings.

ramiy:
Resolve #1003

### DIFF
--- a/includes/admin/customers/customers.php
+++ b/includes/admin/customers/customers.php
@@ -505,7 +505,7 @@ function give_customers_view( $customer ) {
 		<table class="wp-list-table widefat striped donations">
 			<thead>
 			<tr>
-				<th scope="col"><?php echo give_get_forms_label_singular(); ?></th>
+				<th scope="col"><?php esc_html_e( 'Form', 'give' ); ?></th>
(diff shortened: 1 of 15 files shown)


DevinWalker:
This looks good, thanks!

mathetos:
Love this, less functions and letting i18n do the heavy lifting with singular/plural instead. Nice @ramiy !

PR #1014, opened 2016-09-09
Issue/1013 - Use `$current_user = wp_get_current_user()` instead of `global $current_user`

### LINKED ISSUE #1013: STOP USING `global $current_user`
One less global to improve our scrutinizer score at #675

See also: https://generatewp.com/quick-tip-stop-using-get_currentuserinfo/

ramiy:
Resolve #1013


DevinWalker:
Crushing those globals! Nice @ramiy

PR #1022, opened 2016-09-12
Docs: globals

ramiy:


DevinWalker:
Can't for get about those docs! :)

PR #1023, opened 2016-09-12
Issue/1013 - Use `$current_user = wp_get_current_user()` instead of `global $current_user`

ramiy:
Seems like I missed this global in #1013.


ramiy:
@DevinWalker

PR #1027, opened 2016-09-13
Issue/1024 - stop using `$give_options` global

### LINKED ISSUE #1024: STOP USING `global $give_options`
Replace `global $give_options;` with `$give_options = give_get_settings();`.

We call this global 26 times in 13 different files. We need to stop using this global!

This will also improve our scrutinizer score at #675.

ramiy:
Resolve #1024

### DIFF
--- a/give.php
+++ b/give.php
@@ -298,7 +298,6 @@ if ( ! class_exists( 'Give' ) ) :
 		 * @return void
 		 */
 		private function includes() {
-			global $give_options;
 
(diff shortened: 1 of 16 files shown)


DevinWalker:
@ramiy what about all our Add-ons that use `$give_options`? We need to be sure it's still accessible until we update all of them and then officially deprecate it's usage.

ramiy:
@DevinWalker It's a very good point!

I will revert some of the changes for backwards-compatibility.

DevinWalker:
@ramiy looking good now, this is nice

PR #1029, opened 2016-09-13
Issue/1028 - Start using `get_admin_page_title()` for sreen titles

### LINKED ISSUE #1028: Use get_admin_page_title() for admin page titles
We need to replace the H1 titles in different Give admin screens to use `get_admin_page_title()` function.

This way we will be able to use the `$page_title` defined in `add_{menu|submenu}_page()` function, and reduce further more the amount of translation strings.

ramiy:
Resolve #1028


DevinWalker:
Nice! Good call.

PR #1067, opened 2016-10-02
Translate deprecated hook notice

ramiy:
## Description
The notice for deprecated hook is not translatable. This patch adds i18n function to the translation string.

## Motivation and Context
The notice in `_give_deprecated_function()` is translatable, but the stings passed by `give_deprecated_action_mapping()` and `give_deprecated_filter_mapping()` are not translatable. The result is wired.


DevinWalker:
Good catch. Thanks!

PR #1106, opened 2016-10-10
Fix admin_url() function usage

ramiy:
## Description

In some place the `admin_url()` function was used not as it should be, for example the `$path` parameter was outside of the function or has a leading `/` character. Or in one case we used only the path, without the `admin_url()` function.

Any way, this PR fixes those issues.


DevinWalker:
Nice catches 👍

PR #1135, opened 2016-10-19
Issue/1131

### LINKED ISSUE #1131: Updating donor profile results in email error
## Issue Overview

The following error seems to prevent any type of update to donor info from the donor profile screen.

Attempting to edit donor information using the `Edit Donor` button results in a server-side error upon submission: `Error: Please enter a valid email address.`

I thought that possibly the email field farther down the page was triggering the error, but entering a valid email in that field still results in the same error. This error occurs when attempting to edit registered and unregistered donors.

## Expected Behavior

User should be able to update donor info from donor profile screen.

## Current Behavior

Clicking `Update Donor` results in an error that references a valid email address. User is left confused because an email field is not part of the form they are updating.

## Possible Solution

Investigate server-side validation to remove the condition causing the error.

1. Make a donation with address fields.
2. View donor profile.
3. Click `Edit Donor` button at top of profile.
4. Change address fields.
5. Click `Update Donor`
6. Error appears: `Error: Please enter a valid email address.`

The same error occurs for unregistered users that only have a `User ID` field to edit.

## Screenshots

Error when attempting to update regisered user's address.

image

Error when attempting to update unregisered user's name.

image

## Context

Discovered while investigating address fields for #1112.

## Your Environment

* Give version used: 1.7
* Browser Name and version: Chrome 53.0.2785.143 (64-bit)
* Operating System and version (desktop or mobile): OS X Yosemite 10.10.5

kevinwhoffman:
## Description

Attempting to update donor info was causing an email validation error described in #1131.

As of 1.7 donor email updates are handled in their own functions outside of `give_edit_customer()`. For that reason, references to the email field are removed from `give_edit_customer()`.

Donor info can now be updated without error.

## How Has This Been Tested?

- Updated donor name without error.
- Updated donor address without error.
- Connected donor to user ID without error.
- Confirmed email validation and editing still work in the Donor Emails section added in 1.7.

## Types of changes

Fixes #1131

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/admin/customers/customer-actions.php
+++ b/includes/admin/customers/customer-actions.php
@@ -50,15 +50,10 @@ function give_edit_customer( $args ) {
 
 	$defaults = array(
 		'name'    => '',
-		'email'   => '',
 		'user_id' => 0
 	);
 
 	$customer_info = wp_parse_args( $customer_info, $defaults );
-
-	if ( ! is_email( $customer_info['email'] ) ) {
-		give_set_error( 'give-invalid-email', esc_html__( 'Please enter a valid email address.', 'give' ) );
-	}
 
 	if ( (int) $customer_info['user_id'] != (int) $customer->user_id ) {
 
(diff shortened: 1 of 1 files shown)


DevinWalker:
I just tested this and it's working great. Thanks for catching that bug before it ever had a chance to get out in the wild.

PR #1140, opened 2016-10-21
Issue/509

### LINKED ISSUE #509: Determine an appropriate [donation_history] "thank you" text by the transaction status.
The `[donation_history]` shortcode always shows "Thank you for your donation!" regardless of whether the donation status is failed/cancelled/abandoned/etc.

I suggest you change this text depending on the transaction status.

image

kevinwhoffman:
## Description

Improves receipt-viewing experience discussed in #509 with the following changes:

- Receipts now display on Donation History page instead of Donation Confirmation page, allowing for a more predictable UI when clicking into an individual receipt.
- Whether user sees full donation history or individual receipt depends on presence of `payment_key` query arg in URL.
- Link back to all donations is provided below receipt.
- `[give_receipt]` shortcode has two new attributes:
    - `payment_status` determines whether status row within receipt is displayed. Default `false`.
    - `status_notice` determines whether notice above receipt is displayed. Default `true`.
- `give_receipt_status_notice` filter allows users to filter the notice.

## How Has This Been Tested?

- Created a donation with each possible status.
- Confirmed each status displays the relevant notice.
- Confirmed Donation Status row is hidden by default.
- Confirmed shortcode atts can override default behavior.
- Tested against non-English locales.

509-give-receipt-notice-low

## Types of changes

Resolves #509

## Checklist:
- [x] My code is tested.

## Documentation Updates

@mathetos We will need to update https://givewp.com/documentation/core/shortcodes/give_receipt/ with the following changes:

- [ ] Clarify when receipt appears on Donation Confirmation page vs. Donation History page.
- [ ] Document new `[give_receipt]` shortcode attributes.
- [ ] Document new `give_receipt_status_notice` filter with examples.

### New Filter: `give_receipt_status_notice`

```
/**
 * Filters payment status notice for receipts.
 *
 * By default, a success, warning, or error notice appears on the receipt
 * with payment status. This filter allows the HTML markup
 * and messaging for that notice to be customized.
 *
 * @since 1.7.0
 *
 * @param string $notice HTML markup for the default notice.
 * @param int    $id     Post ID where the notice is displayed.
 * @param string $status Payment status.
 * @param array  $meta   Array of meta data related to the payment.
 */
echo apply_filters( 'give_receipt_status_notice', give_output_error( $notice_message, false, $notice_type ), $id, $status, $meta );
```

### Basic Example

Customize an error notice based on donation status.

```
function prefix_custom_failed_notice( $notice, $id, $status, $meta ) {
	if ( 'failed' === $status ) {
		return give_output_error( __( 'Whoops! Something went wrong.' ), false, 'failed' );
	}

	return $notice;
}
add_filter( 'give_receipt_status_notice', 'prefix_custom_failed_notice', 10, 4 );
```

### Advanced Example

Personalize a success notice with the donor's first name. This is possible because we are passing `$meta` to the hooked function.

```
function prefix_personalized_success_notice( $notice, $id, $status, $meta ) {
	if ( 'publish' === $status ) {
		$first_name = $meta['user_info']['first_name'];
		$notice_message = sprintf( __( '%s, your generous donation is most appreciated!', 'text_domain' ), $first_name );

		return give_output_error( $notice_message, false, 'success' );
	}

	return $notice;
}
add_filter( 'give_receipt_status_notice', 'prefix_personalized_success_notice', 10, 4 );
```

### DIFF
--- a/includes/forms/functions.php
+++ b/includes/forms/functions.php
@@ -263,6 +263,21 @@ function give_listen_for_failed_payments() {
 
 add_action( 'template_redirect', 'give_listen_for_failed_payments' );
 
+/**
+ * Retrieve the Donation History page URI
+ *
+ * @access      public
+ * @since       1.7
+ *
+ * @return      string
+ */
+function give_get_history_page_uri() {
+	$give_options = give_get_settings();
+
+	$history_page = isset( $give_options['history_page'] ) ? get_permalink( absint( $give_options['history_page'] ) ) : get_bloginfo( 'url' );
+
+	return apply_filters( 'give_get_history_page_uri', $history_page );
+}
 
 /**
  * Check if a field is required
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -238,7 +238,7 @@ function give_receipt_shortcode( $atts ) {
 		'payment_method' => true,
 		'payment_id'     => true,
 		'payment_status' => false,
-		'notice'         => true,
+		'status_notice'  => true,
 	), $atts, 'give_receipt' );
 
 	//set $session var
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -25,6 +25,11 @@ if ( ! defined( 'ABSPATH' ) ) {
  */
 function give_donation_history() {
 
+	// If payment_key query arg exists, return receipt instead of donation history.
+	if ( isset( $_GET['payment_key'] ) ) {
+		return give_receipt_shortcode( array() );
+	}
+
 	$email_access = give_get_option( 'email_access' );
 
 	//Is user logged in? Does a session exist? Does an email-access token exist?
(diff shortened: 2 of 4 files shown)


DevinWalker:
Love the examples! Thanks!

mathetos:
@kevinwhoffman Looks really awesome. Only one nit-picky item, I think the filter should be called `give_receipt_status_notice` ... don't have plans for more receipt notices right now, but don't want to have the need and not have good filter name options.

kevinwhoffman:
@mathetos I'm good with that, but I think we should be consistent and do the same for the shortcode attribute, which is currently named `notice`.

In summary, the new additions would be:
- New Filter - `give_receipt_status_notice`
- New shortcode att - `payment_status` (determines whether status is displayed within receipt)
    - I named it like this to match the other existing atts which you can toggle on/off.
- New shortcode att - `status_notice` (determines whether notice is displayed above receipt)

mathetos:
@kevinwhoffman Ya, makes sense. Why are these checks failing?

kevinwhoffman:
@mathetos I pushed those changes and edited my first comment in this PR so the documentation and examples will reflect the correct name usage.

PR #1159, opened 2016-10-27
Add select box to choose logo size on email tab in settings

philwp:
## Description

When adding a logo image in the 'emails' settings tab you cannot select which size image you want to display.  I added a select box which uses the native WP image sizes to control the image size displayed in the email.  
   
## How Has This Been Tested?

I used multiple images testing all sizes in the email preview view.  I also tested using a URL from an image hosted on another site.  

## Types of changes

- New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My code is tested.

### DIFF
--- a/includes/admin/class-give-settings.php
+++ b/includes/admin/class-give-settings.php
@@ -543,6 +543,19 @@ class Give_Plugin_Settings {
 							'type' => 'file'
 						),
 						array(
+						    'id'      => 'email_logo_size',
+						    'name'    => esc_html( 'Select Logo Size', 'give' ),
+						    'desc'    => esc_html( 'Choose which size to display logo', 'give' ),
+						    'type'    => 'select',
+						    'options' => array(
+						        'full' 		=> __( 'Full', 'give' ),
+						        'large'   	=> __( 'Large', 'give' ),
+						        'medium'    => __( 'Medium', 'give' ),
+						        'thumbnail' => __( 'Thumbnail', 'give' )
+						    ),
+						    'default' => 'full',
+						),
+						array(
 							'id'      => 'from_name',
 							'name'    => esc_html__( 'From Name', 'give' ),
 							'desc'    => esc_html__( 'The name donation receipts are said to come from. This is most likely the name of your organization or website.', 'give' ),
--- a/templates/emails/header-default.php
+++ b/templates/emails/header-default.php
@@ -58,7 +58,9 @@ $header_content_h1 = "
 	font-weight: 500;
 	line-height: 1.2;
 ";
-$header_img = give_get_option( 'email_logo', '' );
+$header_img_id = give_get_option( 'email_logo_id', '' );
+$logo_size = give_get_option( 'email_logo_size', '' );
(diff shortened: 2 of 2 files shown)


templates/emails/header-default.php
```diff
@@ -58,7 +58,9 @@
 	font-weight: 500;
 	line-height: 1.2;
 ";
-$header_img = give_get_option( 'email_logo', '' );
+$header_img_id = give_get_option( 'email_logo_id', '' );
+$logo_size = give_get_option( 'email_logo_size', '' );
+$header_img = $header_img_id == '' ? give_get_option( 'email_logo' ) : wp_get_attachment_image_src( $header_img_id, $logo_size )[0];
```
ravinderk: Please validate `wp_get_attachment_image_src` before getting first element or array because this function can be return false.

for ref: https://developer.wordpress.org/reference/functions/wp_get_attachment_image_src/#return

includes/admin/class-give-settings.php
```diff
@@ -543,6 +543,19 @@ public function give_settings( $active_tab ) {
 							'type' => 'file'
 						),
 						array(
+						    'id'      => 'email_logo_size',
```
ravinderk: @philwp nowadays emails are responsive, so I did not think to give the admin the option to select logo size has more benefits than giving admin an option for email logo, so admin has full freedom to select upload their own size.

philwp: @ravinderk I ran into an issue where the image that one of my clients wanted to use was too large to look good in an email.   The image was over 500px wide and I needed to manually enter the path to a smaller version of it.

ravinderk: @philwp It seems the most specific problem that general. It is easy to crop and upload  an image and then assign as email logo.

@DevinWalker What do you think?

DevinWalker:
@ravinderk I personally have had the issue of uploading an image that is too large and felt frustrated by that. Personally, I like this field. @mathetos @kevinwhoffman what do you think?

DevinWalker:
@philwp can you switch this over to the `release/1.6.4` branch?

kevinwhoffman:
I just tested and it is a nice addition especially for users who may already have a high-resolution logo uploaded and lack the graphics editing skills to create a specially sized logo just for emails.

With that said, we should also have a max-width style applied to that image in the event that an oversized image is selected.

DevinWalker:
@kevinwhoffman I agree on the max-width style for the image. I just set a huge image and tested and it looked pretty terrible.

ravinderk:
@DevinWalker I already added max-width into `release/1.8` with new setting api but I think i is too late.
I also experimented with adding images sizes  model. Let me know what do you think.

screen shot 2016-10-28 at 1 03 46 am

DevinWalker:
Ah, I see @ravinderk do in `release/1.8` the modal has a size select field. Is that correct?

ravinderk:
@DevinWalker I developed model now but max width already exists. Let me know if you good with this model , I will create pr then

DevinWalker:
Thanks @philwp we've merged #1168 thanks to you. Good idea. 👍

PR #1165, opened 2016-10-28
Issue/1163

### LINKED ISSUE #1163: If theme doesn't register image sizes our we get a PHP Warning on the Give Settings page
## Issue Overview
 Install a theme that doesn't register any image sizes at all. I'm testing with a customer's theme called "Enigma-Pro". When going to the Give Settings page, it shows the following PHP Warnings:

intermediate_image_sizes_warning

`Warning: Invalid argument supplied for foreach() in ...\wp-content\plugins\give\includes\admin\class-give-settings.php on line 1139`

## Expected Behavior
Shouldn't have a warning

## Possible Solution
The problem is that we're looking for intermediate images sizes which aren't there. We're using the Core function `get_intermediate_image_sizes`. We need to check whether there are some before executing that foreach.

PR coming soon

mathetos:
## Description
Resolves #1163

## How Has This Been Tested?
Locally:
  - [x] reproduced the problem.
  - [x] added conditional to prevent the problem
  - [x] verified that when the conditional is swapped the problem reappears

## Types of changes

## Checklist:

  - [x] My code is tested.

### DIFF
--- a/includes/admin/class-give-settings.php
+++ b/includes/admin/class-give-settings.php
@@ -1139,7 +1139,7 @@ function give_get_featured_image_sizes() {
 
 	// check whether intermediate image sizes exist first
 	if ( empty( $get_sizes ) ) {
-		return false;
+		$get_sizes = array( 'thumbnail', 'medium', 'medium_large', 'large' );
 	}
 
 	foreach ( $get_sizes as $_size ) {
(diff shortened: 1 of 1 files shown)


DevinWalker:
I reviewed this and when `give_get_featured_image_sizes` returns false the select field displays blank:



However, it should return with the default WordPress image sizes of `thumbnail`, `medium`, etc.

mathetos:
@DevinWalker 
https://github.com/WordImpress/Give/pull/1165/commits/69da98b3c4d05eefa9e2a74562f8f77f3457a664 Ensures that the default image sizes are returned in the case that get_intermediate_image_sizes() ever returns false.

To test it, I verified that the image sizes were shown in the "Featured Image SIze" dropdown as intended, and that when that setting was changed it had the expected result on the form featured image as well.

Issue #744, opened 2016-07-18
Include custom form fields values in CSV donor exports?


daveis:
Could this feature request be considered in a future update? I greatly need the CSV exports to display all of the data I have captured with my form, not just the core customer/transaction details.

Asparagirl:
+1

mathetos:
@daveis Just a heads up that we slated this for version 1.7 which should be released in early November.

daveis:
@mathetos That is fantastic news! Thanks for the heads up and the great work you do

daveis:
@mathetos - looks like this issue got moved into the 1.9 milestone? So does this mean it's more likely to roll out in 2017 now?

drzraf:
a first part in PR #1287

drzraf:
You **can** currently export custom fields:
`give_pre_setup_payment` put your custom forms/POST in a custom object attribute (but *not* userinfo)
Ex:
```php
/*
  In any case we can hook give_setup_payment but, user_info are reset by built-in filter
  Thus ony give_pre_setup_payment is usable to temporarily stores and transmit the data in a custom object attribute
*/
function PA_give_pre_setup_payment( $give_payment, $payment_id) {
  $give_payment->custom_fields  = $give_payment->payment_meta['user_info']['custom_fields'];
}
add_filter('give_pre_setup_payment', 'PA_give_pre_setup_payment', 10, 2);
```

In `give_setup_payment` store this meta inside WordPress (here use `update_user_meta()` to store inside `wp_usermeta` or `Give_Customer::update_meta` to store inside `wp_give_customers`). Personnally, I choosed the former.

```php
function PA_give_setup_payment( $give_payment, $payment_id) {
  if (! isset($give_payment->custom_fields)) return;
  $customer = new Give_Customer( $give_payment->customer_id );
  // $customer->update_meta() stores inside wp_give_customers
  // var_dump($customer->update_meta('give_custom_fields', $give_payment->custom_fields));
  // update_user_meta() stores inside wp_usermeta
  update_user_meta($customer->user_id, 'give_custom_fields', $give_payment->custom_fields);
}
add_filter('give_setup_payment', 'PA_give_setup_payment', 10, 2);
```

With the above meta are attached to the donor rather than the transaction (as it shown in the old tutorial).
Then add these fields to the donor CSV:

```php
function pa_custom_fields_export_custom_data($data) {
  foreach($data as $i => $exported_user) {
    // load the previously stored meta
[...]

PR #1287, opened 2016-12-01
donors export: show custom fields to be added inside the CSV export

drzraf:
## Description
show custom fields to be added inside the CSV export (see issue #744 )

## How Has This Been Tested?

Go inside wp-admin/edit.php?post_type=give_forms&page=give-reports&tab=export
Visual testing: HTML difference only if hook is implemented

## Types of changes

screenshot-2016-11-30_22-30-51-crop

HTML changes in the backoffice for hooks implementers
(no change if `give_export_csv_cols_customers` isn't implemented)

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/admin/reporting/reports.php
+++ b/includes/admin/reporting/reports.php
@@ -420,8 +420,19 @@ function give_reports_tab_export() {
 											<label for="give-export-donation-sum"><input type="checkbox" checked name="give_export_option[donation_sum]" id="give-export-donation-sum"><?php esc_html_e( 'Total Donated', 'give' ); ?>
 											</label>
 										</li>
+										<?php
+										$custom_cols = [];
+										$custom_cols = apply_filters( 'give_export_csv_cols_customers', $custom_cols );
+										foreach($custom_cols as $col_id => $col_name):
+										?>
+										  <li>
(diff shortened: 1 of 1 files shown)


drzraf:
Forgot to answer to the question: Why are the checkbox forcefully checked?
Because Give's hook mechanism does not pass the result of this form POST to the underlying hooks.
`give_export_csv_cols_customers` is passed `$custom_cols` only.
No way to choose whether these custom fields are requests  inside the resulting file or not.

Let's say that, for now, implementers of `give_export_csv_cols_customers` also correctly implement `give_export_get_data_customers` (otherwise ...eck... troubles).

DevinWalker:
Hey @drzraf this is a good starting base for what we're moving to do. We have open issues to address the `payment_meta` https://github.com/WordImpress/Give/issues/1132 and better custom payment meta fields in the export process as well as the donation edit screen. 

As well, we want to make your life easier as a developer. Seeing your comment:
> You can currently export custom fields (but it's hard to find, understand and implement):
give_pre_setup_payment put your custom forms/POST in a custom object attribute (but do not userinfo since it's reseted by normal fields handling)

Your function there is a good workaround! Good job. Looking at this PR I see a number of quick fixes that need to be done like you can't use PHP short tags in WP plugins `<?` and arrays need to be created like `$custom_cols = array()`. I'm going to close this PR not because of that though, it's because we really need to refactor _how_ custom payment meta is handled and exported. Your `foreach` loop will definitely be in there 😉

drzraf:
About the form:
* These are not "short open tags" but (short) "template tags" which AFAICT are not prohibited inside WP plugin.
* When you write:
> it's because we really need to refactor how custom **payment** meta is handled and exported. 

this applies to **donor** meta too, isn't?

PR #1350, opened 2016-12-21
Update to use `cmb2_init` action and `new_cmb2_box` function.

jtsternberg:
The `cmb2_meta_boxes` filter is mostly for back-compatibility and it is recommended to use the action and the API to add boxes/fields. I have updated to use the `cmb2_init` and to loop through the metaboxes and register them with `new_cmb2_box`. This is untested, but should work exactly the same, but be more resilient as it does not depend on [others using the filter correctly](https://wordpress.org/support/topic/fatal-error-upon-activation-24/).

### DIFF
--- a/includes/admin/forms/metabox.php
+++ b/includes/admin/forms/metabox.php
@@ -14,16 +14,12 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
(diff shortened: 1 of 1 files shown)


DevinWalker:
Looks good, thanks @jtsternberg

PR #1376, opened 2017-01-05
Fixes #619

### LINKED ISSUE #619: Hint.css for our Tooltips instead of QTips2
While QTip is a good library, it's heavily JS based. When tiny JS errors that we can't control happen on the form page the QTips often just break.

[Hint.css](https://github.com/chinchang/hint.css) is pure CSS based. Of course there's potential that themes might mess with it in one way or another, but their repo is very active and they put a lot of attention on making their styles very strict in to avoid conflicts like that. Plus it would make styling them very easy.

@DevinWalker would love your input.

mathetos:
Fixes #619 and submitted specifically to release/1.9

Does the following:

* Removed qtips from being enqueued anywhere
* Removes qtips from bower, and any pre-processing tools
* Removes qtips from tests
* Removes qtips markup throughout the plugin
* Adds hint.css to be enqueued via Admin and frontend
* Adds hint.css markup throughout the plugin

No strings were changed at all.

There's room for improving this implementation of hint.css to give users more flexibility, but this PR is just about replacing qtips with hint.css out of the box. Additional features can come later.

I tested each individual tooltip in the Admin and front-end. If we merge this, these changes will have to be done in Add-ons as well.


DevinWalker:
Thanks @mathetos I'll check this out in depth. Looks good from a code standpoint. To address the breaking unit tests can you fix the broken file check:

mathetos:
@DevinWalker Just fixed. Thought I had done that previously. Weird. Thanks!

DevinWalker:
Fixed merge conflict and additional updates. Commits were preserved and new PR here: https://github.com/WordImpress/Give/pull/1376

mathetos:
You rock @DevinWalker !

PR #1433, opened 2017-01-31
Issue/1430

### LINKED ISSUE #1430: Shortcode parameters broken in 1.8
## Issue Overview

The following shortcode that works as expected in 1.7.2 has issues in 1.8.

`[give_form id="8" show_title="false" show_goal="false" show_content="none" display_style="modal"]`

- `show_title` is NOT reflected (title shows no matter what).
- `show_goal` is NOT reflected (goal shows no matter what).
- `show_content` IS reflected.
- `display_style` IS reflected.

## Expected Behavior

Title and Goal should not be displayed when relevant parameters are set to `false`.

## Current Behavior

Title and Goal are displayed even when parameters are set to `false`.

## Possible Solution

Check shortcode function for changes since 1.7.2.

1. Create form with Goal and Content.
2. Create post. Use shortcode builder to set parameters as shown above.
3. View post to confirm Title and Goal are still showing.

kevinwhoffman:
Fixes #1430 by evaluating `give_form` shortcode arguments with a boolean filter so that `"true"` and `"false"` are correctly evaluated as `true` and `false`.

`show_title` and `show_goal` are now reflected in the form output.

### DIFF
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -79,8 +79,8 @@ function give_form_shortcode( $atts ) {
 	), $atts, 'give_form' );
 
 	// Convert string to bool.
-	$atts['show_title'] = (bool) $atts['show_title'];
-	$atts['show_goal']  = (bool) $atts['show_goal'];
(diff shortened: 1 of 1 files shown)


DevinWalker:
Good catch!

PR #1552, opened 2017-02-25
Issue/1551

kevinwhoffman:
This change allows the token to be set in the database, which then allows the user to view donation history via email access.

## Testing
The token validation appears to only work on the first view. When revisiting the donation history page, it tells me:

> It looks like you haven't made any donations.

I believe after the initial visit, cookie-based authentication should take over. That part needs testing.

### DIFF
--- a/includes/class-give-email-access.php
+++ b/includes/class-give-email-access.php
@@ -246,12 +246,10 @@ class Give_Email_Access {
 		// Must have a token.
 		if ( ! empty( $token ) ) {
 
-			// Token & verify key must be valid.
-			if (
-				! $this->is_valid_token( $token )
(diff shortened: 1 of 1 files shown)


kevinwhoffman:
Donation History should now work for return visits as long as token remains valid. See commit message for full details.

As far as I can tell, this was an old bug, not caused by any recent changes. It most likely went unnoticed because most users only use the email access link once, and do not return to the Donation History page multiple times.

kevinwhoffman:
Update: The donation history breaking after the first visit actually was a recent bug introduced in the "code improvements" commit. It had to do with changing `get_var()` to `get_row()`. I reverted it to use `get_var()` and it's working as expected.

PR #1574, opened 2017-03-12
PL translation added

code2prog:
I also added the same translation on https://translate.wordpress.org/projects/wp-plugins/give/stable/pl/default. Not sure when it's going to be approved there.


DevinWalker:
Thanks @code2prog! We actually prefer to have WP.org handle the language files for downloading etc. so bundling within the plugin won't be necessary.

Issue #844, opened 2016-08-08
Give Workers need to have expanded user roles.


Benunc:
"Give Worker" users should be equivalent to contributors, with the addition of being able to create give forms.

Currently, a Give Worker can create a form, but not create (or even submit for review) a post or page with a shortcode of that form on it.

Give Workers should be able to edit but not publish posts and pages.

One potential use case would be making a fundraising site where individuals can create pages or posts to be approved by an admin.

DevinWalker:
Related to https://github.com/WordImpress/Give/issues/662

PR #1697, opened 2017-05-17
Expanded Give Worker User Role - Issue/844

mehul0810:
## Description
This PR is for #844. Added Capabilities to Give Worker user role for editing posts & pages and submitting it for review to admin.

## How Has This Been Tested?
Tested this feature after creating a user with Give Worker role and then creating and editing the posts and pages.

## Types of changes
New feature (non-breaking change which adds functionality

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/admin/upgrades/upgrade-functions.php
+++ b/includes/admin/upgrades/upgrade-functions.php
@@ -49,6 +49,11 @@ function give_do_automatic_upgrades() {
 		case version_compare( $give_version, '1.8.7', '<' ) :
 			give_v187_upgrades();
 			$did_upgrade = true;
+
+        case version_compare( $give_version, '1.8.8', '<' ) :
+			give_v188_upgrades();
+			$did_upgrade = true;
+
 	}
 
 	if ( $did_upgrade ) {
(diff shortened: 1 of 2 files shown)


includes/admin/upgrades/upgrade-functions.php
```diff
@@ -892,3 +897,31 @@ function give_v187_upgrades(){
...
 }
+
+/**
+ * Update Capabilities for Give_Worker User Role.
+ *
+ * This upgrade routine will update access rights for Give_Worker User Role.
+ *
+ * @since      1.8.8
+ */
+function give_v188_upgrades() {
```
ravinderk: @mehul0810 We do not need an upgrade for capabilities.

screen shot 2017-05-17 at 10 35 04 am

WordPress will automatically find capabilities attached to `give_worker` role.
Can you test this without an upgrade?

mehul0810: @ravinderk I've checked the code again and tried without the upgrade script for capabilities. But, it is not affecting to existing users without an upgrade. If I run an upgrade, it is immediately adding capabilities to existing users with `give_worker` role

ravinderk: @mehul0810 You are right. It will not work without an upgrade. Because roles and capabilities store in database which needs to update 👍

Issue #1050, opened 2016-09-22
Dynamically hide/show and unrequire/require state fields based on country.


Benunc:
## Issue Overview

We currently populate the state/province dropdown with separate state lists, depending on the country. Would be nice to (especially in places like the UK and Germany where there is no "state") either populate that field with "n/a" or make it go away all together. 

And it definitely needs to not be a required field if we don't have a case for it in the list at https://github.com/WordImpress/Give/blob/master/includes/country-functions.php#L60

DevinWalker:
Good call @Benunc 

Please review how WooCommerce handles this per country. It's how I'd like our address fieldset to work when changing countries. Depending on the country:

1. If the country has states, display the state field - either as a text field or dropdown select if a list is available (like the US)
2. If the country does not have states, remove the field, it's not necessary

**TODO:** Test for this. Q: Does Woo have unit tests for this?

raftaar1191:
Working on the about issue

raftaar1191:
To Make the same that is being offered by WooCommerces we need to refactor the code.

1. **Recommendation on how we can achieve**
Need to make a list wherein country states are been added. If in the list country doest not find any states then no drop-down will not be shown and if it finds the states then there will be drop-down option show which will display the list of state that country has.

2.**our current functionality**
I have gone throughout our functionality we need to create some functions and need to depreciate the old one like the example: give_get_brazil_states_list which provided the list of states in Brazil and make all that thing into one array. Where in the key will be the country code and the value will be in array format containing the States pf the respected country.

3. **Time**
Around 10 hours.

raftaar1191:
Notes where to check for dropdown
1> /wp-admin/edit.php?post_type=give_forms&page=give-donors&view=overview&id=3
2> http://localhost/givewp/wp-admin/edit.php?post_type=give_forms&page=give-payment-history&view=view-payment-details&id=1764

3> http://SITENAME/donations/first-form-all-fields-support-non-login-user-as-well/

DevinWalker:
@raftaar1191 that sounds good. Let's proceed with the refactor.

raftaar1191:
@ravinderk  
The above functionality is being added on the following PR please review and merge the PR
https://github.com/WordImpress/Give/pull/1858

DevinWalker:
@raftaar1191 I noticed an issue with this functionality which is why I'm reopening it.

## Reopened Issue Description

[...]

PR #1858, opened 2017-07-12
Issues 1050

raftaar1191:
Description
This PR is for #1050

How Has This Been Tested?
Tested for all the display option that can be set from the form display tab one by one.
Have also tested the JS that it should not get a break with the current functionality.

## Types of changes

## Checklist:
- [x] My code is tested.


DevinWalker:
@raftaar1191 is this ready for review?

raftaar1191:
Yes

includes/ajax-functions.php
```diff
@@ -192,35 +192,39 @@ function give_ajax_get_form_title() {
...
-
-	echo $response;
-
-	give_die();
+	$response = array(
+		'success' => true,
+		'states_found' => $states_found,
+		'data' => $data,
+	);
+	wp_send_json( $response );
```
ravinderk: @raftaar1191 any reason not to use `wp_send_json_success` function

ravinderk: @raftaar1191 I agree as we discuss on call. As per current solution, state field will not print as text field as it was happening before. We can add a filter here to show/hide text field on basis of state if admin wants by default we can hide text field for `Germany`, `UAE` etc.

raftaar1191: wp_send_json_success internalty used the same function.
Here there is no possibility of wp_send_json_success or wp_send_json_error so I used wp_send_json in whcih parameter I can definne if it successfully or not

https://core.trac.wordpress.org/browser/tags/4.8/src/wp-includes/functions.php#L0

DevinWalker:
@raftaar1191 I'm seeing an issue with the functionality with the state field being removed completely. See here when I make a change from the US to the United Kingdom:



The field is removed incorrectly. The functionality should work like WooCommerce does:

raftaar1191:
@DevinWalker 

We can add a filter here to show/hide text field on basis of state if admin wants by default we can hide text field for Germany etc.

DevinWalker:
@raftaar1191 definitely. Though, by default it should work like Woo does and show the text field for state and depending on the country have it required or not.

raftaar1191:
@DevinWalker @ravinderk 

Done, now is working as in WooCommerces

DevinWalker:
@raftaar1191 I'm still noticing differences. For instance, for Argentina WooCommerce loads the states whereas ours does not. Please ensure the states are loaded properly for all countries.

raftaar1191:
@DevinWalker 

A States can be easily added to the country by using the `give_states_list` filter 
<a href="https://github.com/WordImpress/Give/pull/1858/files#diff-060a5c321b244e3d4a8903b351323c75R372" >FIlter link </a>

Was not know that need to add new states for the rest country as well.

Adding that now.

DevinWalker:
This is working well now. Thanks @raftaar1191 !

Issue #1952, opened 2017-08-03
Multi-level dropdown needs max-width


mathetos:
## Issue Overview
Multi-level dropdown needs max-width. This is problematic when the levels are very wide, and the form is presented in modal or button mode.

## Expected Behavior
The select dropdown should never go outside the modal area.

## Current Behavior
Currently it goes outside the modal area:
![image](https://user-images.githubusercontent.com/930724/28946888-bf541d2a-7861-11e7-90ec-ef472fa93de8.png)

## Possible Solution
Add `max-width: 100%` to `form[id*=give-form] select.give-select-level`

DevinWalker:
Woah! Yeah definitely need to fix that asap.

DevinWalker:
@mehul0810 can you also take care of the max-width issue with the chosen2 dropdown that appears throughout the plugin? It needs a max-width as well. See if I make a really long title:

![2017-08-07_20-19-26](https://user-images.githubusercontent.com/1571635/29054983-7ded456e-7bae-11e7-9efc-31ec3d4955c3.png)

![2017-08-07_20-19-37](https://user-images.githubusercontent.com/1571635/29054984-7f47b6ba-7bae-11e7-8aef-dd714ec8b34a.png)

PR #1965, opened 2017-08-08
Issue/1952

mehul0810:
## Description
This PR is for #1952 

## How Has This Been Tested?
I'm unable to reproduce this case in Chrome as well Safari of MAC using the `Display Mode` Button.
However, I've added `max-width:100%` as suggested.

## Types of changes

## Checklist:
- [x] My code is tested.


mehul0810:
@DevinWalker I've updated the PR with changes for select field with chosen js. I've found issue of width on `Donations` page only.
image

PR #1976, opened 2017-08-10
Upgrade Give to float-labels v3.0.3

pryley:
## Description

This pull request upgrades the Float Labels plugin to v3.0.3.

Reasons:
- Simplifies Give's javascript as previous hacks (i.e. adjusting the DOM after float-labels has been triggered or after an ajax call) are no longer required.
- Earlier versions of float-labels.js are no longer actively supported
- No dependencies

## How Has This Been Tested?

I have tested the upgrade on all official WordPress themes (twentyten - twentyseventeen), as well as on Divi theme v2.5.3 (as that is the only version I have access to). All themes display the Float Labels fields correctly and ajax (i.e. switching gateways, changing country/state) works as expected. 

## Types of changes

Each commit on the pull request has been titled appropriately and changes should be easy to follow.

## Possible breaking changes

IE compatibilty has changed from IE9+ to IE10+. Specifically, the class helper methods (i.e. `addClass`, `removeClass`, `hasClass`) have been removed in favor of using [classList](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) directly.


DevinWalker:
Works great! The floating labels seem smoother and more polished now too. Thanks!

PR #1980, opened 2017-08-10
remove [type=“url”] from float-labels exclude option

pryley:
You will need to run gulp after merging this fix


pryley:
Also please see: https://github.com/WordImpress/Give-Form-Field-Manager/issues/160#issuecomment-321579459

DevinWalker:
Thanks @pryley

PR #2003, opened 2017-08-20
Issues 1996

raftaar1191:
## Description
PR for issues #1996 

## How Has This Been Tested?
Manually Tested.
  
## Types of changes

## Checklist:
- [ ] My code is tested.


includes/payments/actions.php
```diff
@@ -253,3 +253,29 @@ function give_refresh_thismonth_stat_transients( $payment_ID ) {
...
+ * Add meta in payment that store page id and page url.
+ *
+ * Will add/update when user add click on the checkout page.
+ * The status of the donation doest not matter as it get change when user had made the payment successfully.
+ *
+ * @since 1.8.13
+ *
+ * @param {integer} $payment_id Payment id for which the meta value should be updated.
+ */
+function give_payment_save_page_data( $payment_id ) {
```
DevinWalker: I would rather this go directly within `give_insert_payment()` rather than creating a separate function

raftaar1191: @DevinWalker 
Why I did this because when I am adding Donations from importer or from Give-Manual-Donations then I don't need this functionality to run

So I can use remove_action function that will help me out to remove this action, so it won't run at the time of Importing or adding from Give-Manual-Donations  add-on.

raftaar1191: @DevinWalker 
Please let me know should I add it to the main function where the importing is being added or should I keep here only

DevinWalker: Just use `remove_action` within the importer.

raftaar1191: Okay

Issue #1938, opened 2017-08-01
Provide an option to set the PayPal Standard locale


DevinWalker:
## Issue Overview
We sometimes get the request to set the PayPal Standard locale (language). For the most part, PayPal does a decent job of setting it for the donor automatically. However sometimes admins want to have it always set to one location. I suggest we add an dropdown with the default option set to "Let PayPal Decide the Locale" and then a list of available locales after that to set. 

This is how PayPal determines the user's locale automatically. Which I think is good and can remain the default behavior:

> Localization factors
> 
> To determine the locale for the checkout page, PayPal uses the locale code and these other factors:
> 
> 1. The shipping address country provided by the merchant in the request.
> 2. The country taken from the locale code parameter passed by the merchant.
> 3. The country of the currently logged-in user.
> 4. The merchant's country.

### Additional Links of Interest

- PayPal Locales: https://developer.paypal.com/docs/classic/api/locale_codes/
- Current Snippet: https://github.com/WordImpress/Give-Snippet-Library/blob/master/gateway-customizations/paypal-custom-checkout-locale.php
- The `ln` variable explained: https://developer.paypal.com/docs/classic/paypal-payments-standard/integration-guide/Appx_websitestandard_htmlvariables/

PR #2004, opened 2017-08-21
Issue/1938

mehul0810:
## Description
This PR is for #1938

## How Has This Been Tested?
I've tested changing to locale to various different countries and the locale based on the country reflected on PayPal checkout page.

image

## Types of changes

## Checklist:
- [x] My code is tested.


DevinWalker:
@mehul0810 I've decided not to include this feature in Give core because it would be too confusing to support properly and we have a snippet that works fine.

Issue #1999, opened 2017-08-18
Generating API keys does not resolve 404 on API endpoint


Benunc:
## Issue Overview
Before you create an API key, API endpoints will 404. This is expected behavior.

Once you generate API key, the endpoints should not 404.

## Expected Behavior
As soon as the keys are created, the endpoints should be accessible.

## Current Behavior
Currently you have to refresh permalinks to make the endpoints accessible.

## Possible Solution
from @kevinwhoffman:

> Flush the permalinks after the first API key is generated.

## Steps to Reproduce (for bugs)
1. Fresh install.
2. Confirm 404 on API endpoint. (e.g. yoursite.dev/give-api/stats or similar)
3. Create new API key at Donations > Tools > API
4. Confirm it’s still 404 on the endpoint.
5. Refresh permalinks at Settings > Permalinks.
6. Confirm endpoint is working.

## Todos
- [ ] Tests

PR #2005, opened 2017-08-21
Issue/1999

mehul0810:
## Description
This PR is for #1999 

## How Has This Been Tested?
I've reproduced the issue and made changes. Then, created a new instance, installed Give and then proceed with generating Give API which worked for the first time.

## Types of changes

## Checklist:
- [x] My code is tested.


includes/api/class-give-api.php
```diff
@@ -1735,6 +1735,10 @@ public function output( $status_code = 200 ) {
 	function user_key_field( $user ) {
 
 		if ( ( give_get_option( 'api_allow_user_keys', false ) || current_user_can( 'manage_give_settings' ) ) && current_user_can( 'edit_user', $user->ID ) ) {
+
+			// Flush Rewrite Rules to avoid 404 error using Give API for first time.
+			flush_rewrite_rules();
```
DevinWalker: @mehul0810 this isn't the proper place to hook in because it would call `flush_rewrite_rules()` way too many times and that's resource exhaustive and not best practice. I suggest only hooking in a _new_ API is created. Most likely within the `generate_api_key()` method.

mehul0810: @DevinWalker I've implemented the required change and simplified the usage of `generate_api_key` fn to achieve the functionality from single fn. Also, tested the case from `User Profile` as well as `Tools > API` and checked Generate API, Regenerate API and Revoke API.

ravinderk:
@mehul0810 It is not working for me when generating user access token on API screen.
screen shot 2017-08-22 at 3 29 28 pm

includes/api/class-give-api.php
```diff
@@ -1849,40 +1850,51 @@ public function process_api_key( $args ) {
...
 
 		$user = get_userdata( $user_id );
 
+		// Bail Out, if user object doesn't exists.
 		if ( ! $user ) {
 			return false;
 		}
 
+		// Flush Rewrite Rules for Give API to work for the first time.
+		flush_rewrite_rules();
```
ravinderk: @mehul0810 @DevinWalker use `flush_rewrite_rules` here is not good. Can we use this on plugin activate and deactivation?
For ref: https://codex.wordpress.org/Function_Reference/flush_rewrite_rules

includes/api/class-give-api.php
```diff
@@ -150,8 +150,8 @@ public function __construct() {
 		add_filter( 'query_vars', array( $this, 'query_vars' ) );
 		add_action( 'show_user_profile', array( $this, 'user_key_field' ) );
 		add_action( 'edit_user_profile', array( $this, 'user_key_field' ) );
-		add_action( 'personal_options_update', array( $this, 'update_key' ) );
-		add_action( 'edit_user_profile_update', array( $this, 'update_key' ) );
+		add_action( 'personal_options_update', array( $this, 'generate_api_key' ) );
```
ravinderk: @mehul0810 Any specific reason to rename function

mehul0810: @ravinderk I've not renamed the function `update_key`. I've just removed it as `generate_api_key` and `update_key` both are using similar code and functioning.

ravinderk: @mehul Got it

DevinWalker:
Nice, thanks @mehul0810

Issue #1975, opened 2017-08-09
"Email address ... already active for another user" even when it isnt. 


Benunc:
## Issue Overview
One time donations without login create donations, and donors. Then if the same email is used later to attempt to create a user, the error `The email address provided is already active for another user.` displays. That's not entirely true, and makes it impossible for a donor who once donated without login to later create a login. 

Additionally, that error would make a donor think that they have a login for the site, when in reality they don't. 

## Expected Behavior
If there's no WordPress user with the email address, Give should create that user. 

## Current Behavior
An error message that is neither correct nor helpful displays:

<img width="588" alt="screen shot 2017-08-09 at 5 25 10 pm" src="https://user-images.githubusercontent.com/9629102/29144649-ba930f60-7d27-11e7-918e-4820a318eafd.png">

## Steps to Reproduce (for bugs)

1. Create a form that accepts guest donations (form A) and doesn't display login or registration fields. 
2. Create a form that doesn't accept guest donations (form B). Make sure that form B displays registration fields (at least).
2. Log out of WordPress.
2. In one browser (or an incognito browser window) submit a donation to form A using the email john@doe.com
3. In another browser, or after closing the incognito browser window and reopening a new one, make a donation to form B using the same email, and attempt to create a new user with that email.
4. You'll see the above error, despite the fact that if you log back in and check the WP user list, there is no such user.

## Todos
- [ ] Tests

DevinWalker:
For reference here's the doc @mehul0810 created to outline the issue problems and solutions: https://docs.google.com/document/d/1AoXhN0ckRZ5gF_5oDbKmSu00NtH0elsHzpC-jFnQClk/

PR #2011, opened 2017-08-22
Issue/1975

mehul0810:
## Description
This PR is for #1975 

## How Has This Been Tested?
This issue is generated after the fix of this issue https://github.com/WordImpress/Give/issues/1722
So, I've tested this PR considering both the issues and the flowchart which was created to overcome both the issues at the same time.

**Here are the scenario's checked:**
#### Scenario 1
1. Create a Donation Form with Guest Checkout enabled and no registration/login and say it `Form 1`.
2. Complete the donation using Form 1 with a new email which is not a donor nor a user. (For Example, john@doe.com)
3. Now, Create a new Donation for with Guest Checkout disabled and registration enabled and say it `Form 2`.
4. Use the same primary email (i.e. john@doe.com) to complete the donation using Form 2. Then, you should be able to complete the donation as a part of fix for #1975 

#### Scenario 2
1. Add a new email (i.e. additional.john@doe.com) as an `additional email` to donor with email (i.e. john@doe.com) via admin.
2. Now, complete a donation using Form 2 with email (i.e. additional.john@doe.com). Then, you should not be able to complete the donation as a part of fix for #1722 

@DevinWalker @ravinderk Hope these scenario's helps you recreate the issue and check the fix. Let me know if you have any queries.

## Types of changes

## Checklist:
- [x] My code is tested.


DevinWalker:
This is working well. Thanks for the unit tests also!

Issue #1997, opened 2017-08-18
Suggest move banner, makes UI confusing


gingerling:
## Issue Overview

On the settings page a banner displays between the two navigation areas (presumably not all the time, but it was there for me).

This makes the top navigation less visible than the should be, so, for example, I came to the settings page (having not used it in a while) and it wondered why there were only 3 settings tabs (General | Currency | Access Control ).

I later realised there were more tabs at the top (General, Payment Gateways, Display Options, Emails, Advanced) but only because I was looking for specifically the Payment settings as I had configured this before on a previous site. Would be confusing to a new user.
 
![screenshot from 2017-08-18 12-53-46](https://user-images.githubusercontent.com/2853040/29458123-7efa33e4-8415-11e7-93d3-27213089aa4b.png)

## Possible Solution

Suggest moving the banner to above the top tabs.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Todos
- [ ] Tests
- [ ] Documentation

mathetos:
Thanks @gingerling -- totally agree. We'll look into it.

gingerling:
Great

DevinWalker:
Very nice UI/UX improvement here. Thanks @mehul0810 for the implementation and @gingerling for the suggestion!

gingerling:
thanks :)

PR #2016, opened 2017-08-25
Issue/1997

mehul0810:
## Description
This PR is for #1997 

## How Has This Been Tested?
This issue is due to improper settings html formatting. To make all the notices above the tabs, we should place an `<h2>` above the tabs and inside `<div class="wrap">` to work properly.

@DevinWalker @ravinderk I've kept `<h2>` empty right now to solve the issue. If you would like to add some heading text then we can add that in `<h2>` as well. Let me know your thoughts.

#### Before
image

#### After
image

## Types of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/admin/class-i18n-module.php
+++ b/includes/admin/class-i18n-module.php
@@ -94,7 +94,7 @@ class Give_i18n_Banner {
 			return;
 		}
 
-		if ( ! $this->hide_promo() ) {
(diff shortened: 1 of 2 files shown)


DevinWalker:
@mehul0810 I tested this out and it's not how I'd like it yet. Can you make it so that all notices appear _above_ the tabs and don't flash anywhere above/below like the following:

mehul0810:
@DevinWalker I've updated the PR. Please review and let me know your thoughts.

PR #2027, opened 2017-08-30
Validate

Umangvaghela:
If we use **[wp give cache]**   that time comes warning error.When we pass action=delete that time work complete.
To add action is required at this command so add --action=<cache_action> instead [--action=<cache_action>].

Another point I have added  **wp give report** to list all give the report.


DevinWalker:
Thanks @Umangvaghela

PR #2029, opened 2017-08-31
change

Umangvaghela:
add changes


collizo4sky:
@Umangvaghela you apparently have problem with codebase without strict type checking. Glad to see I am not alone.

I often do `true === $status` 😆

PR #2041, opened 2017-09-04
Issues 2036

### LINKED ISSUE #2036: Add setting to select default state

raftaar1191:
## Description
Admin can select default state/province (if any) while selecting/changing the country.
This PR will resolve #2036

## How Has This Been Tested?
Manually tested
Please view the video for the step 
http://screencast-o-matic.com/watch/cbQh6nI1m6 ( updated on 09.05.2017 )

## Types of changes

Video Link
http://screencast-o-matic.com/watch/cbQh6nI1m6 ( updated on 09.05.2017 )

## Checklist:
- [x] PHPCS error.
- [x] PPH Compatibility.
- [ ] My code is tested.


ravinderk:
@raftaar1191 It is not working when you reset country.

okokkp4l3a

raftaar1191:
@ravinderk @DevinWalker 

Fixed all the issues and also add the video that covers all the test case where the default states are being used

DevinWalker:
Thanks @raftaar1191 the video was good. 

I have a few questions:

1. One thing I didn't see though was if you tested the `[give_profile_editor]` shortcode on the frontend. Actually, when I tested with it I noticed a bug with the state field not being updated entirely. Can you fix that too while also ensuring it works accordingly? 



2. What will happen when users update with the new state field? For backwards compatibility it should just be blank. I don't think we want to presume to set a state for them.

raftaar1191:
@DevinWalker 

Fixed all the issues

assets/js/admin/admin-scripts.js
```diff
@@ -467,11 +467,43 @@ jQuery.noConflict();
 	var Give_Settings = {
 
 		init: function () {
+			this.setting_change_Country();
```
DevinWalker: How come the "C" is capped here? Can you just keep them all lowercase?

raftaar1191: Fixed

DevinWalker:
@raftaar1191 please take care of what we reviewed with the ability to save an "empty" state field like so:

Issue #2081, opened 2017-09-15
Using reCAPTCHA on email access fails with more than one shortcode on the page


Benunc:
## Issue Overview

If you enable the reCAPTCHA fields in Donations > Settings > General > Access Control and then put both a [donation_history] and [give_subscriptions] on the same page, reCAPTCHA will fail every time.

## Expected Behavior

Users should be able to put donation history and subscriptions on the same page.

## Current Behavior

It won't let you log in via email access with reCAPTCHA enabled.

## Steps to Reproduce (for bugs)

1. Enable email access, and configure the reCAPTCHA fields
2. Create a new page with both [donation_history] and [give_subscriptions] on it.
3. In an incognito browser, visit that page.
4. Attempt to complete one of the reCAPTCHA fields and submit the form. (it will not work).

## Todos
- [ ] Tests
- [ ] Documentation

jaydeeprami:
@DevinWalker and @Benunc, I can able to reproduce this issue and working on it.

Thanks,

PR #2093, opened 2017-09-19
Fixed: reCAPTCHA on email access fails with more than one shortcode on the page #2081

jaydeeprami:
## Description

Fix reCAPTCHA on email access fails with more than one shortcode on the page #2081

## How Has This Been Tested?
Manually tested.

## Video URL:
http://www.screencast.com/t/HRI1NDDV7SI

## Types of changes
- Removed 'Google reCAPTCHA' script from the 'email-login-form.php' file because it's load two times when we have two shortcodes on the same page. so I have added before the '</head>'.
- Add script for unified both google reCaptcha form. so Google reCaptcha will work for both.

I found 2 issues while working on it which I have mentioned in my Video.
1) Recurring donation used same email access token so if I will be varified Donation History Google reCaptcha then it will auto work for `[give_subscriptions]` shortcode. 
2) Currently, Email access form load from 'email-login-form.php' for the both shortcode. so It will duplicate and conflict when two shortcodes on the same page.

## Checklist:
- [X] My code is tested.


DevinWalker:
@jaydeeprami I could barely hear you in that video - please check the audio next time.

>  I found 2 issues while working on it which I have mentioned in my Video.
> 
> 1. Recurring donation used same email access token so if I will be varified Donation History Google reCaptcha then it will auto work for [give_subscriptions] shortcode.
> 2. Currently, Email access form load from 'email-login-form.php' for the both shortcode. so It will duplicate and conflict when two shortcodes on the same page.

1. That's intended 
2. The email access shortcode should only ever be output _one time_ on the page. Even if there are two shortcodes:



@jaydeeprami I'm going to close this in favor of my PR here: https://github.com/WordImpress/Give/pull/2094

Issue #1721, opened 2017-05-22
If user is disconnected from associated donor, an error should show when user tries to login


mehul0810:
## Issue Overview
As discussed with @ravinderk today, we should restrict user to login with meaningful error message, if that user is disconnected with associated donor profile. Even, we can send an email to admin that user is trying to login which is disconnected from donor. Do you want to reconnect user again to donor? or need to create new donor profile for the user?

Related to #1720 

## Expected Behavior
User should not be able to access features of Give plugin when user is disconnected from donor as donor is the main entity representing donations for Give plugin.

## Current Behavior
User can login to his account even if user is disconnected from the donor profile.

## Possible Solutions
Need to investigate the issue in more detail and based on which we can come up with best possible solution. I think restricting user with after login action should work.

## Todos
- [ ] Tests
- [ ] Documentation

mehul0810:
@DevinWalker Today, I had a discussion with @ravinderk regarding this issue and had few questions related to this issue.

### Scenario
This issue will affect those users who are donors but user and donor are disconnected by admin from donor detail page.

![image](https://user-images.githubusercontent.com/1852711/30537553-b9a6d148-9c87-11e7-90fc-ebbf8b60a0d0.png)

### Questions
1. Do we need to restrict user with WP Admin Login as well as Donation Form Login ?
2. We will be sending an email to admin, if user will try to login and restricted to login due to this reason - We have discussed to have hardcoded email right now and later on in `release 2.0/2.1` we will provide flexibility to modify emails. Let me know your thoughts.
3. What should be the functionality of the `Reconnect User` button in the email and what page should open on clicking it?
4. Secondary CTA in the email will be `Create New Donor` and on clicking to it will redirect to `Create Donor` page so that admin can link the user to the newly created donor. Also, we don't have a separate page for creating new donor in admin. So, how do we think to manage this? Let me know your thoughts.

**My thought on point (3):** We can open the donor detail page when admin clicks the `Reconnect User` button and will be connecting the respective donor which was disconnected earlier and an admin notice will be displayed `We have successfully connected the user <email> to its respective donor.

DevinWalker:
## Responses
1. No don't restrict the user from WP admin login
2. Hard coded is fine. I doubt people will want to edit it as it will be seldom sent.
3. It should go to the donor's profile page in wp-admin like you said - that sounds good.
4. Don't worry about creating a new donor for now.

PR #2103, opened 2017-09-21
Issue/1721

mehul0810:
## Description
This PR is for #1721 

## How Has This Been Tested?
I have tested this with different scenario's:
1. Connecting and Disconnecting User from Donor Profile.
2. Disconnecting, Changing and Connecting User from Donor Profile.
3. Disconnecting User from Donor Profile and then try logging in. Then, user shouldnot be able to login and an email is sent to admin for reconnecting user.
4. Clicking on `Reconnect User` button in email should reconnect donor with user and display a success notice.

image

## Types of changes

## Checklist:
- [x] My code is tested.


includes/process-donation.php
```diff
@@ -715,6 +715,23 @@ function give_donation_form_validate_user_login() {
...
+
+					$disconnected_donor_id = get_user_meta( $user_data->ID, '_give_disconnected_donor_id', true );
+
+					// Send Email.
+					give_admin_email_user_donor_disconnection( $user_data->ID, $disconnected_donor_id );
+
+					// Donor and User are disconnected.
+					give_set_error(
+						'donor_user_disconnected',
+						__( 'Your Donor and User profile are no longer connected. Please contact site administrator.', 'give' )
```
DevinWalker: Change the text to:
> Your Donor and User profile are no longer connected. Please contact the site administrator.

DevinWalker:
@mehul0810 let's discuss this when you come online. I have a few questions:

1. When a user and donor profile are disconnected this PR is preventing the user from logging into WordPress admin through the give login shortcode. However, if they simply access `/wp-admin` or `/wp-login.php` they will be able to login. Right? 
2. It's my thought that would shouldn't prevent them from logging in but rather prevent them from accessing the donations without proving their identity via email confirmation. So if I attempt to check my donations profile when logged in to a disconnected account I'll see something like "Your Donor and User profile are no longer connected. Please enter your email below to confirm your identity." The user then would receive an email at the _donor_ email address and then once clicked be able to access it. Does that make sense?

mehul0810:
@DevinWalker Please check my comments

1. When a user and donor profile are disconnected this PR is preventing the user from logging into WordPress admin through the give login shortcode. However, if they simply access /wp-admin or /wp-login.php they will be able to login. Right?
-- **Yes. If user and donor profile are disconnected. Then, user can login via `wp-admin`. But, not via login in donation form.**

2. It's my thought that would shouldn't prevent them from logging in but rather prevent them from accessing the donations without proving their identity via email confirmation. So if I attempt to check my donations profile when logged in to a disconnected account I'll see something like "Your Donor and User profile are no longer connected. Please enter your email below to confirm your identity." The user then would receive an email at the donor email address and then once clicked be able to access it. Does that make sense?
-- **Yes. we can allow login user and restrict the Give Core area for donors like accessing donations, Give profile editor and other relating shortcodes.**

Need more info on why we want to send email to donor email address? I think donor and user email address will be same in most of the cases. We should send email to admin and admin should have those rights to allow donor access to user. Let me know your thoughts

ravinderk:
@mehul0810  
While saving name with profile editor, I am getting error notice.
Let me know if you also able to produce it.

oyagxc0amr

mehul0810:
@DevinWalker @ravinderk I have made changes to this PR.
1. Login via WP Admin and Shortcode no longer restrict user from logging in.
2. Processing Donation is restricted with message if user is logged in and disconnected with donor and tries to donate.
3. It will show a message and hide `give_profile_editor` form, if donor and user profile are disconnected.
4. For `donation_history` shortcode, there is a message already displayed when user and donor profile are disconnected. So, i have not added message to show as per the PR as it is not required in this case.

Please let me know your thoughts on the above points.

mehul0810:
@DevinWalker I had a discussion @ravinderk yesterday and we decided to keep a question for discussion. 

What will happen if user and donor with email `hello@give.dev` is doing a guest checkout and donor-user profiles are disconnected?

Let me know your thoughts on this.

DevinWalker:
@mehul0810 my thought is that the donation goes through. We don't want to prevent a donation from happening.

mehul0810:
@DevinWalker I have removed the restriction while processing donation. So, now, donation process will go through irrespective of user being logged in or non logged in, when donor and user profile are disconnected.

Issue #2045, opened 2017-09-05
fix(admin-menu): display Donation forms in Edit Menus tab by default


DevinWalker:
## Issue Overview
I thought we resolved this with https://github.com/WordImpress/Give/issues/1765 but I'm not seeing it on a fresh install.

## Expected Behavior
Donation forms is added by default on plugin install:
![2017-09-05_11-53-40](https://user-images.githubusercontent.com/1571635/30077569-1d4a8e3a-9231-11e7-8402-22d358d25eae.png)

## Current Behavior
It's not.

kevinwhoffman:
This works by filtering the user option in order to add `add-post-type-give_forms` into the `$initial_meta_boxes` array. This prevents Donation Forms from being hidden by default.

```php
function give_filter_initial_nav_menu_meta_boxes( $result, $option, $user ) {
	global $wp_meta_boxes;

	// Only run if user option is empty (which means first-time visit to Menus page).
	if ( ! empty( $result ) || ! is_array( $wp_meta_boxes ) ) {
		return $result;
	}

	// Include give_forms in the initial meta boxes array.
	$initial_meta_boxes = array( 'add-post-type-page', 'add-post-type-post', 'add-custom-links', 'add-category', 'add-post-type-give_forms' );
	$hidden_meta_boxes = array();

	foreach ( array_keys($wp_meta_boxes['nav-menus']) as $context ) {
		foreach ( array_keys($wp_meta_boxes['nav-menus'][$context]) as $priority ) {
			foreach ( $wp_meta_boxes['nav-menus'][$context][$priority] as $box ) {
				if ( in_array( $box['id'], $initial_meta_boxes ) ) {
					unset( $box['id'] );
				} else {
					$hidden_meta_boxes[] = $box['id'];
				}
			}
		}
	}

	$user = wp_get_current_user();
	update_user_option( $user->ID, 'metaboxhidden_nav-menus', $hidden_meta_boxes, true );

	// Return hidden meta boxes which should no longer include `add-post-type-give_forms`.
	return $hidden_meta_boxes;
}
add_filter( 'get_user_option_metaboxhidden_nav-menus', 'give_filter_initial_nav_menu_meta_boxes', 10, 3 );
```

I agree with @mehul0810 that a better approach would be to propose a core filter to modify `$initial_meta_boxes`. Without that filter, I've had to duplicate much of the `wp_initial_nav_menu_meta_boxes()` function.

mehul0810:
[...]

PR #2122, opened 2017-09-26
Issue/2045

mehul0810:
## Description
This PR is for #2045 

## How Has This Been Tested?
I have tested this by replicating the scenario of fresh install and also make sure that the function doesn't get called on every visit of `Appearance > Menus`.

Also, tested this issue by switching `Single Form Views` to `enabled` as well as `disabled`

## Types of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -854,10 +854,17 @@ function give_donation_metabox_menu() {
 	$screen = get_current_screen();
 
 	// Proceed, if current screen is navigation menus.
-	if( 'nav-menus' === $screen->id && give_is_setting_enabled( give_get_option( 'forms_singular' ) ) && ! get_user_option( 'give_is_donation_forms_menu_updated' ) ) {
+	if(
+		'nav-menus' === $screen->id &&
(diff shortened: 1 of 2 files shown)


includes/misc-functions.php
```diff
@@ -840,31 +840,45 @@ function modify_nav_menu_meta_box_object( $post_type ) {
...
+ *
+ * @todo Remove this, when WordPress Core ticket is resolved (https://core.trac.wordpress.org/ticket/16828).
+ *
+ * @return bool
  */
-function give_nav_donation_metabox_enabled() {
+function give_donation_metabox_menu() {
 
-	// Return false, if it fails to retrieve hidden meta box list and is not admin.
-	if ( ( ! $hidden_meta_boxes = get_user_option( 'metaboxhidden_nav-menus' ) ) || ! is_admin() ) {
```
kevinwhoffman: Make `! is_admin()` the first condition so if that fails, it prevents the unnecessary database call to check user option.

mehul0810: @kevinwhoffman Implemented required changes

DevinWalker: 👍

PR #2186, opened 2017-10-12
Issue/1790

mehul0810:
## Description
This PR is for #1790 

## How Has This Been Tested?
I have tested this using the `View it in browser` link with following scenario:
- Opening it in same window where donation is performed.
- Opening it in incognito window
- Opening after the access token has expired to make sure it still works after verification and regenerating access token

## Types of changes

## Checklist:
- [x] My code is tested.


DevinWalker:
This is working well. Merging.

PR #2205, opened 2017-10-13
Map postal code to zip during importing

BhargavBhandari90:
## How Has This Been Tested?
I changed the header from `Zip` to `Postal Code` into sample csv

## Screenshot

give-map-zip-to-postal-code

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/admin/tools/import/class-give-import-donations.php
+++ b/includes/admin/tools/import/class-give-import-donations.php
@@ -460,6 +460,12 @@ if ( ! class_exists( 'Give_Import_Donations' ) ) {
 		 */
 		public function selected( $option_value, $value ) {
 			$selected = '';
+
+			// Postal Code also needs to map with zip.
+			if ( 'Postal Code' === $value ) {
+				$value = 'Zip';
+			}
+
 			if ( stristr( $value, $option_value ) ) {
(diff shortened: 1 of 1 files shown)


includes/admin/tools/import/class-give-import-donations.php
```diff
@@ -460,6 +460,12 @@ public function render_dropdown() {
 		 */
 		public function selected( $option_value, $value ) {
 			$selected = '';
+
+			// Postal Code also needs to map with zip.
+			if ( 'Postal Code' === $value ) {
```
DevinWalker: Can we sanitize this so that it can either be lowercase or uppercase?

mathetos:
But will it automap if the uploaded CSV says "Zip" instead of "Postal Code"? That's the heart of the issue, those words are synonymous and both should map automatically.

DevinWalker:
Yeah any combination of "Zip", "zip", "Postal Code", "postal code" should map to the single field. Does that make sense?

mathetos:
@DevinWalker Yes, exactly, that's what I'm hoping for.

DevinWalker:
Thanks!

Issue #2581, opened 2017-12-26
[Give 2.0] WordPress Default Admin Email Goes Blank


mehul0810:
## Issue Overview
I've noticed that on performing new donation with create an account option selected, admin receives a WordPress default email which is empty.

![image](https://user-images.githubusercontent.com/1852711/34356013-dc41c516-ea5f-11e7-8a29-93844080f797.png)

## Expected Behavior
New User Registration email should have some content.

## Current Behavior
Currently, there is no content in New User Registration email which sends out to admin.

## Steps to Reproduce (for bugs)
1. Create a Donation Form with Registration Enabled.
2. Perform a donation by checking `Create an Account` checkbox.
3. Check email with subject `New User Registration`
4. You will notice the email is empty

## Todos
- [ ] Tests
- [ ] Documentation

## WordPress Environment
<details>

### WordPress Environment ###

Home URL: http://give20.local
Site URL: http://give20.local
WP Version: 4.9.1
WP Multisite: –
WP Memory Limit: 256 MB
WP Debug Mode: ✔
WP Cron: ✔
Language: en_US
Permalink Structure: /%postname%/
Show on Front: posts
Table Prefix Length: 3
Table Prefix Status: Acceptable
Admin AJAX: Inaccessible
Registered Post Statuses: publish, future, draft, pending, private, trash, auto-draft, inherit, refunded, failed, revoked, cancelled, abandoned, processing, preapproval, give_subscription

### Server Environment ###

Hosting Provider: DBH: localhost, SRV: give20.local
TLS Connection: 
TLS Connection: 
Server Info: nginx/1.10.1
PHP Version: ❌ 5.3.29 - We recommend a minimum PHP version of 5.6. See: PHP Requirements in Give
PHP Post Max Size: 1,000 MB
PHP Time Limit: 900
PHP Max Input Vars: 3000
PHP Max Upload Size: 1,000 MB
cURL Version: ❌ 7.38.0, OpenSSL/1.0.2h - We recommend a minimum cURL version of 7.40.
SUHOSIN Installed: –
MySQL Version: ❌ 5.5.55 - We recommend a minimum MySQL version of 5.6. See: WordPress Requirements
Default Timezone is UTC: ✔
fsockopen/cURL: ✔
SoapClient: ✔
DOMDocument: ✔
gzip: ✔
GD Graphics Library: ✔
Multibyte String: ✔
Remote Post: ✔
Remote Get: ✔

### Give Configuration ###

[...]

PR #2590, opened 2017-12-27
#2581  [Give 2.0] WordPress Default Admin Email Goes Blank

akashsonic:
**Fix issue:** Email is missing default content in 2.0+


DevinWalker:
Good catch. Thanks!

PR #2673, opened 2018-01-18
Issue/1472 Fix: Add give_totals shortcode

### LINKED ISSUE #1472: feat(shortcode): add `[give_totals]` shortcode
## Issue Overview

User wants to be able to show the progress of a form dynamically in text only, or sometimes as a progress bar. This is particularly helpful when you have several forms that are each part of a larger project within the organization. 

For example a homeless shelter might have 4 different forms, but they are all part of their "Spring Fundraising Campaign". If they wanted to show how much they've raised for that campaign, a shortcode like this would be really useful.

@Benunc has already started in this direction with this snippet:
https://github.com/WordImpress/Give-Snippet-Library/blob/46846070b7cec12247a5f641506e7e8df1189355/custom-shortcodes/total-earnings-shortcode.php

But there are several important elements that I think it would need in order to be good in Give Core.

1. **All Forms** the `multi_id` attribute should be able to take either a comma-separated list of unique form ids (like it does currently) or if set to `all` or `0` it would show the progress of ALL forms on the site.
2. **Categories and Tags** you should be able to display the total of all forms in a category or a tag
3. **Display Style** -- there should be a new attribute called `display_style` which takes either `text` or `progress_bar` are arguments. "Text" would only output the total as plain-text. "Progress" would show our Give Goal progress bar and total but of course be the total of all forms along with the goal of all those forms together.

jaydeeprami:
## Description

This PR will fix #1472 

@DevinWalker, As I have discussed with you on Slack that we can't extend Shortcode generator API to support Multiselect so Admin can only select one option but they can add multiple ids and cats, tags into the shortcode.

## How Has This Been Tested?

- Manually tested

## Video link: (jpeg or gifs if applicable):
https://screencast-o-matic.com/watch/cFVFlRoFXs

## Types of changes
 - New feature (non-breaking change which adds functionality)

## Checklist:
- [X] My code is tested.


raftaar1191:
@DevinWalker @jaydeeprami 

Will love to have chosen functionality for Donation Form when selecting from Donation From Dropdown

jaydeeprami:
Thanks @mehul0810, Updated since tag to 2.1.

Issue #2553, opened 2017-12-20
Change Password Update Message in give_profile_editor Shortcode 


raftaar1191:
## Issue Overview
In Give give_profile_editor Shortcode page when I change the Password of mine it shows me a message like this: 
![image](https://user-images.githubusercontent.com/22215595/34205660-375344cc-e5a9-11e7-99f3-fa306eda4b5b.png) this does not mean that I have updated the PassWord it means that my profile ( Name, Display name and other details are updated but not the password )

![image](https://user-images.githubusercontent.com/22215595/34205630-287caf7e-e5a9-11e7-9708-ce346be0c19b.png)

Also, I don't know why this message is showing me here when after the PassWord Change I am not getting login Off
![image](https://user-images.githubusercontent.com/22215595/34205541-cfd1d3fe-e5a8-11e7-8f0a-e6e0b08b2256.png)

## Related PR and Issues
https://github.com/WordImpress/Give/issues/2414

## Expected Behavior
Display Message Should Be proper 

## Current Behavior
Display Message is not getting display properly 

## Todos
- [x] Tests
- [x] Documentation

## WordPress Environment
<details>
```
### WordPress Environment ###

Home URL: http://recurring.local
Site URL: http://recurring.local
WP Version: 4.9.1
WP Multisite: –
WP Memory Limit: 256 MB
WP Debug Mode: ✔
WP Cron: ✔
Language: en_US
Permalink Structure: /%postname%/
Show on Front: posts
Table Prefix Length: 3
Table Prefix Status: Acceptable
Admin AJAX: Inaccessible
Registered Post Statuses: publish, future, draft, pending, private, trash, auto-draft, inherit, refunded, failed, revoked, cancelled, abandoned, processing, preapproval, give_subscription

### Server Environment ###

Hosting Provider: DBH: localhost, SRV: recurring.local
TLS Connection: Connection uses TLS 1.2
TLS Connection: Probably Okay
Server Info: Apache/2.4.10 (Debian)
PHP Version: 7.0.3
PHP Post Max Size: 1,000 MB
PHP Time Limit: 900
PHP Max Input Vars: 3000
PHP Max Upload Size: 1,000 MB
cURL Version: ❌ 7.38.0, OpenSSL/1.0.1t - We recommend a minimum cURL version of 7.40.
SUHOSIN Installed: –
MySQL Version: ❌ 5.5.55 - We recommend a minimum MySQL version of 5.6. See: WordPress Requirements
[...]

PR #2759, opened 2018-02-02
issue-2553 Fix success message on profile edit and logout on password…

Sidsector9:
Related #2553 

## Description

- Change of _first_name, last_name, display_name or email_  will output a success message as:
`Your profile has been updated.`
- Changing only the password will output a success message as:
`Your password has been updated`
- Change of both of the above will display a success message as:
`Your profile and password has been updated.`
- Change of password will logout and redirect to the current page with a success message corresponding 
to the type of change.

## How Has This Been Tested?
This has only been tested manually by first updating profile, then only updating password, and then updating everything all at once.

## Types of changes
Previously, on successfully updating the profile, the page redirected to:
`http://give.local/profile-editor-shortcode/?updated=true`

Now the page redirects with an additional query string parameter:
`http://give.local/profile-editor-shortcode/?updated=true&update_code={number}`

Where {number} takes values:
`0`: No change
`1`: Profile change only
`2`: Profile and password change
`3`: Password change only

## Checklist:
- [ ] My code is tested.

### DIFF
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -442,11 +442,11 @@ function give_process_profile_editor_updates( $data ) {
 		give_set_error( 'email_not_valid', __( 'The email you entered is not valid. Please use another', 'give' ) );
 
 	} elseif ( $email != $old_user_data->user_email ) {
-		// Make sure the new email doesn't belong to another user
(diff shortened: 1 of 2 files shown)


DevinWalker:
First PR! 👍

PR #2868, opened 2018-03-05
fix(bug): Only show message about options if options are available #2864

### LINKED ISSUE #2864: fix(bug): Only show message about options if options are available
## Issue Overview
Currently the Export Give Settings shows a message about checked options not being exported. This should only appear if such options are available. 

image

Sidsector9:
Closes #2864 

If the exclude from export list is empty, then the message and empty container will be hidden.

## How Has This Been Tested?
Manually tested

## Checklist:
- [ ] My code is tested.


ravinderk:
@Sidsector9 This functionality works fine. 

image
Can you convert span tag to button because tab selection is not working on this element?

PR #2893, opened 2018-03-12
build(composer): streamline local development setup #2798

### LINKED ISSUE #2798: build(composer): streamline local development setup
Instead of having to run `npm install` separately, we can hook it to composer's `post-update-cmd` event. This can simplify the setting up of the development environment for contributors.

This will also help setting up git's pre-commit hooks that is currently under development since npm will install eslinter and the pre-commit hooks will lint JS files before committing. 

What will be required from contributors is that they already have the following installed:

- Git
- Composer
- NPM
- PHPCS

@kevinwhoffman @ravinderk let me know your thoughts on this.

Sidsector9:
Closes #2798 

## Description
Now, "npm install" will execute automatically after composer's install and update commands finish their execution.

## How Has This Been Tested?
Manually tested

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
   },
   "require-dev": {
     "squizlabs/php_codesniffer": "*",
-    "wp-coding-standards/wpcs": "^0.13.1",
+    "wp-coding-standards/wpcs": "*",
     "wimg/php-compatibility": "*"
   },
   "keywords": [
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
   "scripts": {
     "post-install-cmd": [
       "\"vendor/bin/phpcs\" --config-set installed_paths vendor/wp-coding-standards/wpcs",
-      "npm install"
+      "npm start"
     ],
     "post-update-cmd": [
       "\"vendor/bin/phpcs\" --config-set installed_paths vendor/wp-coding-standards/wpcs"
--- a/composer.json
+++ b/composer.json
@@ -26,8 +26,7 @@
       "npm install"
     ],
     "post-update-cmd": [
-      "\"vendor/bin/phpcs\" --config-set installed_paths vendor/wp-coding-standards/wpcs",
-      "npm install"
+      "\"vendor/bin/phpcs\" --config-set installed_paths vendor/wp-coding-standards/wpcs"
     ]
   }
 }
--- a/composer.json
+++ b/composer.json
@@ -22,10 +22,12 @@
   },
   "scripts": {
     "post-install-cmd": [
-      "\"vendor/bin/phpcs\" --config-set installed_paths vendor/wp-coding-standards/wpcs"
+      "\"vendor/bin/phpcs\" --config-set installed_paths vendor/wp-coding-standards/wpcs",
+      "npm install"
     ],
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -22,10 +22,12 @@
...
   "scripts": {
     "post-install-cmd": [
-      "\"vendor/bin/phpcs\" --config-set installed_paths vendor/wp-coding-standards/wpcs"
+      "\"vendor/bin/phpcs\" --config-set installed_paths vendor/wp-coding-standards/wpcs",
+      "npm install"
     ],
     "post-update-cmd": [
-      "\"vendor/bin/phpcs\" --config-set installed_paths vendor/wp-coding-standards/wpcs"
+      "\"vendor/bin/phpcs\" --config-set installed_paths vendor/wp-coding-standards/wpcs",
+      "npm install"
```
kevinwhoffman: Let's remove the second `npm install`. I don't think we want to trigger `npm install` every time a developer runs `composer update`. The goal should be to streamline the initial setup of Give in a local environment, so it makes sense to hook into `composer install` but not `composer update`.

composer.json
```diff
@@ -22,10 +22,12 @@
   },
   "scripts": {
     "post-install-cmd": [
-      "\"vendor/bin/phpcs\" --config-set installed_paths vendor/wp-coding-standards/wpcs"
+      "\"vendor/bin/phpcs\" --config-set installed_paths vendor/wp-coding-standards/wpcs",
+      "npm install"
```
kevinwhoffman: `npm install` is good here 👍

composer.json
Sidsector9: @kevinwhoffman The first one `post-install-cmd` will work only with a lock file present, which won't be the case in the initial setup.

`post-update-cmd` is for both update or install _without_ the lock file present, which will be the scenario after cloning the repo.

kevinwhoffman: @Sidsector9 A `composer.lock` file should be committed to the Git repo. I created an issue for it in https://github.com/WordImpress/Give/issues/2896. Please make sure `composer.lock` gets committed and then continue with my recommendation above to remove the second mention of `npm install`.

Sidsector9: @kevinwhoffman , I'll do that :+1:

kevinwhoffman:
@Sidsector9 Your `npm install` script was firing correctly but it did not take care of the initial build necessary to put Give in a usable state because SCSS and JS assets were not processed. So I added a new `start` script that looks likes this:

`"start": "npm install && npm run watch"`

The end result is that running `composer install` will take care of the entire build process all the way through to launching browsersync. I'm going to hold off on merging until @DevinWalker takes a look but I think we've got what we're looking for. Thanks for getting this up and running.

Sidsector9:
@kevinwhoffman Sounds good! My idea was just to install all the dependencies in a single step. Although, I should have aimed to setup the project to its usable state. 

One suggestion, can we use `npm run dev` instead of `npm run watch`, if the aim is to put the environment in a usable state?

kevinwhoffman:
@Sidsector9 `npm run watch` does the same thing as `npm run dev` but it has the added bonus of launching browsersync so that your local site appears in the browser right away.

PR #2920, opened 2018-03-14
fix(form): display actual goal progress even after goal is met #2638

### LINKED ISSUE #2638: fix(form): display actual goal progress even after goal is met
## Issue Overview
When the donation goal format is set to percentage. It is set to show the correct percentage upto 100%. Once the donation is fully funded it keep showing 100% instead of showing the correct figure which should be greater than 100%.

But when the goal format is set to amount it always shows the correct figure no matter the donation is fully funded or not.

## Expected Behavior
It should show the correct figure in percentage like it shows for the donation goal when set for amount format.

## Current Behavior
It shows correct figure until 100% and then stuck at 100% only.

## Possible Solution
There is a conditional in the code which check if the percentage is greater equal to 100% ,if yes it sets the variable to 100% else to the actual value. We can remove that conditional and let the actual figure in percentage display.

1. Check this in a donation form which has exceeded the goal amount.
2. First set it to Amount and then check the donation goal in the front end. You would see the correct figure.
3. Now set the goal format to percentage and then check the front end again, you would see 100% and not the correct figure.

## Related Helpscout ticket
https://secure.helpscout.net/conversation/495071541/12794/?folderId=1457790

Sidsector9:
Closes #2638 

## Description
Donation forms which were over-funded showed 100% as the maximum progress value. Now it shows the actual value.

If the user wants to limit the progress value for 100%, then they have to use the following filter:
```php
function limit_donation_exceed_to_100( $progress, $form_id, $form ) {
	$goal_format         = give_get_meta( $form_id, '_give_goal_format', true );
	$income              = $form->get_earnings();
	$goal                = $form->goal;
	$donations_goal      = give_get_meta( $form_id, '_give_number_of_donation_goal', true );
	$donations_completed = give_get_form_sales_stats( $form_id );

	return 'donation' === $goal_format ? (
		$donations_completed >= $donations_goal ? 100 : $progress
	) : (
		$income >= $goal ? 100 : $progress
	);
};

add_filter( 'give_goal_amount_funded_percentage_output', 'limit_donation_exceed_to_100', 10, 3 );
```

## How Has This Been Tested?
This has been tested with all 3 goal formats - **Amount, Percentages** and **Number of Donations**
with donations lesser than the goal and more than the goal in all combinations.

## Checklist:
- [x]  My code is tested.

### DIFF
--- a/templates/shortcode-goal.php
+++ b/templates/shortcode-goal.php
@@ -62,9 +62,9 @@ $form_currency = apply_filters( 'give_goal_form_currency', give_get_currency( $f
 
 // Set progress to 100 percentage if income > goal or completed donations > donation goal count.
 if ( 'donation' === $goal_format ) {
-	$progress = $donations_completed >= $donations_goal ? 100 : $progress;
+	$progress_bar_value = $donations_completed >= $donations_goal ? 100 : $progress;
 } else {
-	$progress = $income >= $goal ? 100 : $progress;
+	$progress_bar_value = $income >= $goal ? 100 : $progress;
 }
 ?>
 <div class="give-goal-progress">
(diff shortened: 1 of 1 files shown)


DevinWalker:
@Sidsector9 thanks for the snippet too. I've added it here: https://github.com/WordImpress/Give-Snippet-Library/blob/master/form-customizations/limit-goal-to-100-percent.php

PR #3022, opened 2018-04-11
feat(form): allow forms to be queried based on close status #2250

### LINKED ISSUE #2250: feat(form): allow forms to be queried based on close status

### LINKED ISSUE #3027: feat(form): display goal stats on form detail page
## User Story
As an admin, I want to see goal statistics on form detail page so that it will be easy to check the goal stats when I directly come to form detail page from the frontend.

**Here are the possible areas from where an admin can visit form edit screen directly:**
1. frontend (admin bar)
2. donation listing page
3. donation detail page
4. forms tab on report page
5. log tab on tools page

## Current Behavior
I currently can't see the goal stats on form detail page.

## Expected Behavior
I expect to see the goal stats on form detail page

## Tasks
- [ ] Create and display the metabox for goal statistics
- [ ] Testing

mehul0810:
## Description
This PR resolves #2250 and resolves #3027

## How Has This Been Tested?
I've tested this PR with following scenarios:
- updated form meta to `closed` when form goal is achieved.
- updated form meta to `open` when form goal is updated with which goal is not achieved.
- tested form grid shortcode with `open` and `closed` status and without status

## Types of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/assets/src/css/admin/goals.scss
+++ b/assets/src/css/admin/goals.scss
@@ -4,7 +4,7 @@
   background: #EEE;
   border-radius: 25px;
   overflow: hidden;
-  margin: 5px 0 0;
+  margin: 5px 0 2px;
 
   > span {
 	display: block;
--- a/assets/src/css/admin/goals.scss
+++ b/assets/src/css/admin/goals.scss
@@ -1,32 +1,31 @@
-td.column-goal {
-  .give-progress-bar {
-	height: 8px;
+.give-admin-progress-bar {
+  height: 8px;
+  position: relative;
+  background: #EEE;
+  border-radius: 25px;
+  overflow: hidden;
+  margin: 5px 0 0;
+
+  > span {
+	display: block;
+	height: 100%;
+	border-top-right-radius: 8px;
+	border-bottom-right-radius: 8px;
+	border-top-left-radius: 20px;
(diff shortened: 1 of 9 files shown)


mehul0810:
@ravinderk I've implemented the requested changes.

I've tested this PR with all the possible cases.
- Add New Form screen with no goal meta box.
- Goal not achieved form edit screen. 
- Goal achieved form edit screen.
- checked form closed status when upgrade routine not processed by admin.
- checked form closed status after running upgrade routine.
- checked adding whitespaces to the value of a status attribute in form grid shortcode

**Goal is achieved**
image

**Goal in progress**
image

**Note:** When Goal is not enabled for a form or it is a new form then the Goal meta box won't be visible.

mehul0810:
@DevinWalker @ravinderk 
I've implemented changes as discussed over the call.

**Re-position the text and progress bar**
image

**Lower the priority of Goal meta box**
image

PR #3177, opened 2018-05-04
fix(admin-email): fix post date function argument 

raftaar1191:
## Description
PR to fix get_the_date argument that was getting passed 

Function documentation https://codex.wordpress.org/Function_Reference/get_the_date 

## How Has This Been Tested?
Tested by manual making donation 

## Types of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/emails/class-give-email-tags.php
+++ b/includes/emails/class-give-email-tags.php
@@ -714,9 +714,9 @@ function give_email_tag_billing_address( $tag_args ) {
  *
  * Date of donation
  *
- * @param array $tag_args
+ * @param array $tag_args Arguments which helps to decode email template tags.
  *
- * @return string date
+ * @return string $date Post Date.
  */
 function give_email_tag_date( $tag_args ) {
 	$date = '';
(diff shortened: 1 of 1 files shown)


includes/emails/class-give-email-tags.php
```diff
@@ -726,7 +726,7 @@ function give_email_tag_date( $tag_args ) {
 
 	switch ( true ) {
 		case give_check_variable( $tag_args, 'isset', 0, 'payment_id' ):
-			$date = date_i18n( give_date_format(), strtotime( get_the_date( $tag_args['payment_id'] ) ) );
+			$date = date_i18n( give_date_format(), strtotime( get_the_date( '', $tag_args['payment_id'] ) ) );
```
DevinWalker: Rather than passing empty here couldn't you pass `U` and get the unix timestamp and remove the `strtotime()` call? That would be more effecient. See: https://codex.wordpress.org/Formatting_Date_and_Time

raftaar1191: Thanks @DevinWalker what was really nice 

Updated the PR

PR #3183, opened 2018-05-04
fix(donation): …

### LINKED ISSUE #3181: fix(donation): focusing out of custom amount field without adding an amount leaves 0.00 as amount
## User Story

As a `donor`, I want `the custom amount feature` to assist me `with giving more than 0.00` so that I'm not confused `when I donate`. In 2.1.1 if I focus out out while on the custom amount field without entering a number it stays at `0.00`.

DevinWalker:
## Description

Don't allow 0 as a valid donation amount and set correct min max amount if donor attempts unallowed amount.

Resolves #3181

## How Has This Been Tested?
- Manually tested with custom amount using various separator configurations and decimal values. 
- Ensured if a level is less than the custom amount min that validation still works as expected.

## Types of changes
- UX/UI

## Checklist:
- [x] My code is tested.

### DIFF
--- a/assets/src/js/frontend/give-donations.js
+++ b/assets/src/js/frontend/give-donations.js
@@ -721,7 +721,7 @@ Give.form = {
 		 */
 		__sendBackToForm: function() {
 
-			var form_id = Give.fn.getParameterByName( 'form-id' ),
+			let form_id = Give.fn.getParameterByName( 'form-id' ),
 				payment_mode = Give.fn.getParameterByName( 'payment-mode' );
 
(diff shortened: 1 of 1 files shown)


raftaar1191:
Slack Call Summary

Participants: @devinwalker, @raftaar1191
Topic: Testing PR  
Result: As me and @DevinWalker tested this PR over the slack call we were able to generate multiple issues like it's not working for recurring and fee recovery add-on etc.

So we come to the conclusion that if donor enters no amount then with JS we are going to trigger the default value that was being selected by admin from donation form page 
Example:
image

PR #3204, opened 2018-05-10
fix(currency-switcher): amount should be formatted based on the currency I switched into.

### LINKED ISSUE #3203: fix(currency-switcher): amount should be formatted based on the currency I switched into.
## User Story
As a `Developer`, I want `to use the form's currency format setting to format the various amount inside the verify min/max amount and when displaying the amount in the price level element` so that `I can see the correct donation amount in the currency I have switched into`.

NOTE: Related to currency-switcher.

## Current Behavior
1. When verifying the min/max amount it is sanitizing the amount based on the Give's base currency. 
2. If the currency is auto-switched to another preferred currency then the amount is appearing in Give's base currency format.

## Expected Behavior 
1. min/max amount should be sanitized from the switched currency.
2. If the currency is auto-switched to another currency then, the variable amount should appear in that switched currency.

## Related
- https://github.com/WordImpress/Give-Currency-Switcher/issues/137

## Tasks
- [x] Pass the currency when displaying the variable amount.
- [x] Add new filter to the verify min/max amount to re-verify the amount through any add-on.

emgk:
## Description
This PR Fix #3203 

## How Has This Been Tested?
- By creating donation using Give-Currency-Switcher add-on.

## Types of changes
 Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/forms/template.php
+++ b/includes/forms/template.php
@@ -508,8 +508,8 @@ function give_output_levels( $form_id ) {
 			$output .= '<ul id="give-donation-level-button-wrap" class="give-donation-levels-wrap give-list-inline">';
 
 			foreach ( $prices as $price ) {
-				$level_text       = apply_filters( 'give_form_level_text', ! empty( $price['_give_text'] ) ? $price['_give_text'] : give_currency_filter( give_format_amount( $price['_give_amount'], array( 'sanitize' => false ) ) ), $form_id, $price );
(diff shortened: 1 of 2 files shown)


emgk:
@ravinderk I've made change for code improvement, Please check and let me know if anything is missing.

PR #3279, opened 2018-05-29
fix(donation): dismiss button should work when rendering Give notice using the JS.

### LINKED ISSUE #3278: fix(donation): dismiss button should work when rendering Give notice using the JS.
## User Story
As a `Developer`, I want the Give notice's dismiss event callback should be bound using the `on()` fn, so that I dismiss button in Give notices rendered using JS can work properly.

https://github.com/WordImpress/Give/blob/f647b7f2c6bc4941795ff815d059ae074a3f000a/assets/src/js/frontend/give-misc.js#L96-L101

## Current Behavior
In above JS snippet click() binding of `.give-notice-close` is "direct" binding which will only attach the handler to elements that already exist when the document loaded. It won't get bound to elements created using the JS.

## Expected Behavior
I expect that If I render the dissmissable notice using the JS then it should be dismissed when I click on the dismiss button.

## Possible Solution
bind the event callback using `on()` fn.

## Related
https://github.com/WordImpress/Give-Currency-Switcher/issues/128

## Tasks
- [x] Fix 
- [x] Test

emgk:
## Description
This PR fix #3278 

## How Has This Been Tested?
- Tested it by creating dimiss notices and removing them.
- Tested it with https://github.com/WordImpress/Give-Currency-Switcher/issues/128 

## Types of changes
 Breaking change (fix or feature that would cause existing functionality to not work as expected) 

## Checklist:
- [x] My code is tested.

### DIFF
--- a/assets/src/js/frontend/give-misc.js
+++ b/assets/src/js/frontend/give-misc.js
@@ -93,7 +93,7 @@ jQuery( function( $ ) {
 	}
 
 	// Button to close notices on front-end.
-	jQuery(document).on( 'click', '.give-notice-close', function() {
+	$(document).on( 'click', '.give-notice-close', function() {
 		$(this).hide();
 		const notice_container = $(this).closest( '.give_notices' );
 		notice_container.slideUp();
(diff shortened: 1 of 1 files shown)


Sidsector9:
Using `body` instead of `document` for event delegation is working fine. PR is good.

PR #3295, opened 2018-05-31
fix(form-grid): resolve error where Payment Method shows with only one gateway enabled. #3293

### LINKED ISSUE #3293: fix(form-grid): resolve error where Payment Method shows with only one gateway enabled.
## User Story

As a `give admin`, I want `to hide the select payment method fieldset when there is only one gateway enabled` so that `donors don't have to see it`.

The Form Grid Shortcode is forcing the Select Payment Method fieldset on forms, even when only one is active. 

## Current Behavior

I currently can't hide that section on form grids.

## Possible Solution
There's a `display:block!important` that is targeting that fieldset:
```
.give-donation-grid-item-form #give-payment-mode-select,.give-donation-grid-item-form #give_purchase_form_wrap {
    display: block!important
}
```
That needs to be removed in the case of only one gateway active.

## Steps to Reproduce
1. Create a form grid using the shortcode, to display in modal
2. Only enable one payment gateway
3. Check the form, and the payment gateway options still show. 

## Related

- https://secure.helpscout.net/conversation/591303787/19411/
- 
- 

## Tasks

- [ ] resolve the fieldset showing for only one gateway

Sidsector9:
Closes #3293 

## Description
This PR will only display the payment select fields only if there are more than 1 gateways enabled.

## How Has This Been Tested?
I have tested this on the Form Grid page and individual Form page and tried all Display options.

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/forms/template.php
+++ b/includes/forms/template.php
@@ -1402,89 +1402,89 @@ function give_payment_mode_select( $form_id, $args ) {
 	do_action( 'give_payment_mode_top', $form_id );
 	?>
 
-	<fieldset id="give-payment-mode-select" <?php if ( count( $gateways ) <= 1 ) {
-		echo 'style="display: none;"';
-	} ?>>
-		<?php
-		/**
-		 * Fires while selecting payment gateways, before the wrap div.
-		 *
-		 * @since 1.7
-		 *
-		 * @param int $form_id The form ID.
-		 */
-		do_action( 'give_payment_mode_before_gateways_wrap' );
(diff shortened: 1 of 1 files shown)


ravinderk:
@Sidsector9 Can we remove the only CSS as @Benunc suggested in the issue?

image

ravinderk:
@Sidsector9 fix this issue by removing class using filters as we discussed and remove extra css.

PR #3298, opened 2018-05-31
fix(form-grid): resolve error where Payment Method shows with only one gateway enabled. #3293

### LINKED ISSUE #3293: fix(form-grid): resolve error where Payment Method shows with only one gateway enabled.

Sidsector9:
Closes #3293 

## Description
This PR will only display the payment select fields only if there are more than 1 gateways enabled.

## How Has This Been Tested?
I have tested this on the Form Grid page and individual Form page and tried all Display options.

## Checklist:
- [x] My code is tested.

### DIFF
--- a/assets/src/css/frontend/modal.scss
+++ b/assets/src/css/frontend/modal.scss
@@ -127,7 +127,6 @@
 	height: 85vh;
 	overflow-y: auto;
 
-	#give-payment-mode-select,
 	#give_purchase_form_wrap {
 		display: block !important;
 	}
--- a/assets/src/css/frontend/modal.scss
+++ b/assets/src/css/frontend/modal.scss
@@ -135,6 +135,10 @@
 	.give-btn-modal {
 		display: none !important;
 	}
+
+	.give-form-title {
+		display: none;
+	}
 }
 
 .modal-fade-slide .give-modal--slide {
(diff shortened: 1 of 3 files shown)


assets/src/css/frontend/modal.scss
```diff
@@ -127,7 +127,6 @@
 	height: 85vh;
 	overflow-y: auto;
 
-	#give-payment-mode-select,
 	#give_purchase_form_wrap {
```
ravinderk: @Sidsector9 Why you did not remove this?

ravinderk:
@Sidsector9 

I can see that  if button mode enabled for form then you will see form title in the model but not for others. can you hide form title in the model for symmetry

image

raftaar1191:
Working fine as expected

PR #3343, opened 2018-06-11
fix(form): Be more specific with .notice-dismiss class #3337

### LINKED ISSUE #3337: fix(form): Be more specific with .notice-dismiss class
## User Story

As a Give user, I expect that the Give styles will be strong enough to avoid problems like this:

image

## Current Behavior
Currently, [this style](https://github.com/WordImpress/Give/blob/a5f8b96eeef37ce20d7efb997f79d37d7708bf29/assets/src/css/frontend/alerts.scss#L8) is very generic, it's just `.notice-dismiss` so it can easily get over-written by other themes or plugins like in the example above.

## Expected Behavior
I expect that Give can avoid this by being more specific in the notices style. 

## Possible Solution
Here's a couple options:

1. img[class*="give"].notice-dismiss {}
2. p.give_notice img.notice-dismiss {}
3. div.give_notices img.notice-dismiss {}

Any of those would help prevent the problem shown above. 

## Related
- HS Ticket: https://secure.helpscout.net/conversation/591988690/19460?folderId=672201

Sidsector9:
Closes #3337 

## Description
Refactored SCSS for alerts.

## How Has This Been Tested?
By testing all types of front-end notices.

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/class-notices.php
+++ b/includes/class-notices.php
@@ -634,7 +634,7 @@ class Give_Notices {
 		// Note: we will remove give_errors class in future.
 		$error = sprintf(
 			'<div class="give_notices give_errors" id="give_error_%1$s">
-						<p class="give_error give_notice give_%1$s" data-dismissible="%2$s" data-dismiss-interval="%3$d" data-dismiss-type="%4$s">
+						<p class="give_notice give_%1$s" data-dismissible="%2$s" data-dismiss-interval="%3$d" data-dismiss-type="%4$s">
 							%5$s
 						</p>
 						%6$s
--- a/assets/src/css/frontend/alerts.scss
+++ b/assets/src/css/frontend/alerts.scss
@@ -2,93 +2,102 @@
 Notices, Warnings & Errors
 -----------------------------------*/
 .give_notices {
-  position: relative;
+	position: relative;
 }
 
-.notice-dismiss {
-  position: absolute;
-  top: 50%;
-  right: 0;
-  transform: translate(0, -50%);
-  border: none;
-  margin: 0;
-  padding: 5px;
-  background: 0 0;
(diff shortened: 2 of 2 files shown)


includes/class-notices.php
```diff
@@ -634,7 +634,7 @@ public static function print_frontend_notice( $message, $echo = true, $notice_ty
 		// Note: we will remove give_errors class in future.
 		$error = sprintf(
 			'<div class="give_notices give_errors" id="give_error_%1$s">
-						<p class="give_error give_notice give_%1$s" data-dismissible="%2$s" data-dismiss-interval="%3$d" data-dismiss-type="%4$s">
```
Sidsector9: @kevinwhoffman The red line in the notice was caused by the hardcoded `give_error` class.

On a success notice, it displayed 3 classes `give_error give_notice give_success`, I think this shouldn't be the case. It should be either `give_error`, `give_success` or `give_warning`. Let me know your views on this, if there can be a possibility of backward compatibility issue.

kevinwhoffman: @Sidsector9 I think the `give_error` class needs to stay for now because it also provides the padding and other styles necessary to format the notice in case `give_success` or `give_warning` is not used along with it. I agree it's not ideal. There should be a base class and then modifier classes just for the colors, but we don't want to create rippling effects throughout the add-ons by changing that.

When I tested this notice with Currency Switcher activated, I found that there were styles printed in the `<head>` of the page that were actually causing the red/green mix because they were taking precedent over the stylesheet.

kevinwhoffman: Disregard. The red/green was caused by CSS that was pasted in the Customizer's Additional CSS field.

PR #3368, opened 2018-06-14
fix(form): Be more specific with .notice-dismiss class #3337

### LINKED ISSUE #3337: fix(form): Be more specific with .notice-dismiss class

Sidsector9:
Closes #3337 

## Description
This is a quick fix for issue #3337 

## How Has This Been Tested?
Manually tested by reloading the page.

noticeclose

## Checklist:
- [x] My code is tested.

### DIFF
--- a/assets/src/css/frontend/alerts.scss
+++ b/assets/src/css/frontend/alerts.scss
@@ -2,93 +2,102 @@
 Notices, Warnings & Errors
 -----------------------------------*/
 .give_notices {
-  position: relative;
+	position: relative;
 }
 
-.notice-dismiss {
-  position: absolute;
-  top: 50%;
-  right: 0;
-  transform: translate(0, -50%);
-  border: none;
-  margin: 0;
-  padding: 5px;
-  background: 0 0;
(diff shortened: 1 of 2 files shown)


raftaar1191:
Working great for me

PR #3508, opened 2018-07-19
fix(i18n): format dates to WordPress `date_format` #3502

### LINKED ISSUE #3502: fix: format dates to WordPress `date_format`
## User Story

As a `Give Admin`, I want the `start/end date filter format` in the Donations admin page to match my locale so that `I have a consistent user experience`.

My website has UK English (en_GB) as the site language so I expect all dates to be formatted as `dd/mm/yyyy`.
## Current Behavior

Currently the date filter boxes have a placeholder of `mm/dd/yyyy` and when a date is selected from the dropdown calendar it is returned in the same US date format.

## Expected Behavior

I expect to see `dd/mm/yyyy` as the placeholder and have the date returned from the calendar dropdown formatted to `dd/mm/yyyy`.

## Acceptance Criteria

- [x] The date picker supports all date format (US or non-US) based on WordPress date settings.
- [x] Supporting WP date format does not affect other functionality like donation exports.

## Related
https://secure.helpscout.net/conversation/671557137/26233
https://secure.helpscout.net/conversation/658188284/25003?folderId=672197

Sidsector9:
Close #3502 

## Description
Added support for locale `en_GB`. By default the format will be `mm/dd/yy`. As more such requests come, we can conditionally support more requests.

## How Has This Been Tested?
By switching from locale `en_US` to `en_GB` and back.

## Checklist:
- [x] My code is tested.

### DIFF
--- a/assets/src/js/admin/admin-scripts.js
+++ b/assets/src/js/admin/admin-scripts.js
@@ -105,11 +105,21 @@ var give_setting_edit = false;
 	 * @since: 1.0
 	 */
 	var enable_admin_datepicker = function () {
+		let datepicker = $('.give_datepicker'),
+		    dateFormat = '';
+
+		switch( give_vars.wp_site_language ) {
+			case 'en_GB':
+				dateFormat = 'dd/mm/yy';
+				break;
+
+			default:
+				dateFormat = 'mm/dd/yy';
+		}
 
 		// Date picker.
-		if ($('.give_datepicker').length > 0) {
-			var dateFormat = 'mm/dd/yy';
-			$('.give_datepicker').datepicker({
+		if ( datepicker.length > 0) {
+			datepicker.datepicker({
 				dateFormat: dateFormat
 			});
 		}
--- a/assets/src/js/admin/admin-scripts.js
+++ b/assets/src/js/admin/admin-scripts.js
@@ -108,19 +108,10 @@ var give_setting_edit = false;
 		let datepicker = $('.give_datepicker'),
 		    dateFormat = '';
 
-		switch( give_vars.wp_site_language ) {
-			case 'en_GB':
-				dateFormat = 'dd/mm/yy';
-				break;
-
-			default:
-				dateFormat = 'mm/dd/yy';
-		}
-
 		// Date picker.
 		if ( datepicker.length > 0) {
 			datepicker.datepicker({
-				dateFormat: dateFormat
+				dateFormat: give_vars.date_format
 			});
 		}
 	};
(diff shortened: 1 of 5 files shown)


kevinwhoffman:
@Sidsector9 I may have led you down the wrong path by mentioning the locale in the issue title. The original issue is really about Give using the same date format as the rest of the site, and the most reliable way to do that is to use the `date_format` from the site settings.

So I think we should be using `get_option('date_format')` and then make that format available to JavaScript via `wp_localize_script()`.

Sidsector9:
@kevinwhoffman That was exactly how I was going to go with this, but the problem is that the datcepicker-ui accepts a limited number of formats, see [here](https://api.jqueryui.com/datepicker/#utility-formatDate). PHP's date format for example `F j, Y` will not be acceptable by the datepicker-ui.

CC: @ravinderk

kevinwhoffman:
I see. I still think using the `date_format` rather than the locale is a better approach as it will allow for a more flexible solution that doesn't require us to add locales one by one. Here is a solution I adapted from [this SO answer](https://wordpress.stackexchange.com/a/179500).

## Proposed Solution

```php
/**
 * Converts a PHP date format for use in JavaScript.
 * 
 * @since 2.2.0
 *
 * @param string $php_format The PHP date format.
 * @return string The JS date format. 
 */
function give_convert_php_date_format_to_js( $php_format ) {
    switch( $php_format ) {
        case 'F j, Y':
            $js_format = j'MM dd, yy';
            break;
        case 'Y-m-d':
            $js_format = 'yy-mm-dd';
            break;
        case 'm/d/Y':
            $js_format = 'mm/dd/yy';
            break;
        case 'd/m/Y':
            $js_format = 'dd/mm/yy';
            break;
    }

    /**
     * Filters the date format for use in JavaScript.
     * 
     * @since 2.2.0
     * 
     * @param string $js_format The JS date format.
     * @param string $php_format The PHP date format.
     */
    $js_format = apply_filters( 'give_js_date_format', $js_format, $php_format );

    return $js_format;
}
```

This would convert the preset date formats available in WordPress settings, which would cover the majority of use cases. And if someone is using a custom date format that is not matched, then a filter is available for special use cases.

ravinderk:
@Sidsector9 Check code for date fields and update them all. 

image

Sidsector9:
@ravinderk I updated all the date fields and also took care of escaping in certain places. Kindly review this.

ravinderk:
@Sidsector9 did you check my this comment: https://github.com/WordImpress/Give/pull/3508#issuecomment-406251212

Do not just replace date picker on donation related screen. Check other plugin pages too.

PR #3542, opened 2018-07-31
fix(shortcode): provide grid fallback styles for `[give_form_grid]` #3504

### LINKED ISSUE #3504: fix(shortcode): provide grid fallback styles for [give_form_grid] and [give_donor_grid]
## User Story
As an IE user, I want fallback styles provided for the form grid so that the grid appears similarly to how ti does in modern browsers.

## Current Behavior
User feedback:
>It works in all Browsers except internet explorer. In IE it only displays a single column and the icons are massive making it not readable

## Expected Behavior
I expect the shortcode to render as a multi-column grid that respects my shortcode settings.

## Possible Solution

We can use `@suppports (display: grid)` feature query to target browsers that support Grid, and provide float styles as fallbacks for IE.

## Steps to Reproduce
1. Embed a `[give_form_grid]` shortcode.
2. View in IE 11.
3. Confirm single-column layout.

## Related
https://wordpress.org/support/topic/grid-on-ie/

Sidsector9:
Closes #3504 

## Description
In this PR, `flex` has been used as a fallback for Grid to support form grid on IE 11 browser.
Although IE does not fully support flex model layout, all the flex CSS properties I have used in this PR are supported by IE 11. 

screenshot from 2018-07-31 21-42-32
screenshot from 2018-07-31 21-41-42

## Checklist:
- [x] My code is tested.

### DIFF
--- a/assets/src/css/frontend/give-frontend.scss
+++ b/assets/src/css/frontend/give-frontend.scss
@@ -36,3 +36,13 @@
 #give-receipt{
 	@import '../../../../blocks/components/container-placeholder-animation/style';
 }
+
+.give-grid-ie-utility {
+	margin: 0 -12px;
+}
+
+@supports (display: grid) {
+	.give-grid-ie-utility {
+		margin: 0;
+	}
+}
--- a/assets/src/css/frontend/_grid.scss
+++ b/assets/src/css/frontend/_grid.scss
@@ -16,6 +16,63 @@
 	}
 }
 
+.give-wrap {
+	&:before {
+		display: block;
+		content: '';
+		clear: both;
+	}
+}
+
+// IE 11 browser support.
+@media ( min-width: 30rem ) {
+	.give-grid {
+		display: flex;
+		flex-flow: row wrap;
+		margin-bottom: 1.5rem;
+		max-width: 100%;
+
+		&__item {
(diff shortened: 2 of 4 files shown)


DevinWalker:
@Sidsector9 can you ensure this works with the new `[give_donor_grid]` shortcode as well? See as you're using the `.give-grid` class, it should but I just want to make sure you check.

assets/src/css/frontend/give-frontend.scss
```diff
@@ -36,3 +36,13 @@
 #give-receipt{
 	@import '../../../../blocks/components/container-placeholder-animation/style';
 }
+
+.give-grid-ie-utility {
```
Sidsector9: This part of CSS has been purposely added to this file because for some reason adding this in `_grid.scss` seems to have no effect. Debugging costed too much time therefore for now I'm switching to another task, this can be debugged later.

Sidsector9:
I have used flex instead of float since floated items' height depend on the content within the items. This caused all the items with different contents to have variable heights which made the overall look ugly.

Below is the video demonstration of Form Grid and Donor Wall on IE 11:
https://www.useloom.com/share/72a8e8729bfa432d9ceb8905d925585d

PR #3582, opened 2018-08-07
fix(gutenberg): error notices shouldn't display on Gutenberg #3575

### LINKED ISSUE #3575: fix(gutenberg): error notices shouldn't display on Gutenberg
## User Story

As an `admin` I don't want `Give error notices` displaying on my `Gutenberg editor`.

## Current Behavior

Error notices display and are not dismissible. 

## Expected Behavior

Error notices do not appear and disturb my editing.

Sidsector9:
Closes #3575 

## Description
This PR disables all global notices on the Gutenberg editor page.

## How Has This Been Tested?
Tested with Gutenberg enabled and disabled and on new/edit page/post screens.

## Checklist:
- [x] My code is tested.


includes/misc-functions.php
```diff
@@ -2288,3 +2288,58 @@ function give_get_safe_asset_url( $url ) {
...
+ * Returns if the editor page is a Gutenberg editor.
+ *
+ * @since 2.2.3
+ *
+ * @return boolean
+ */
+function give_is_gutenberg_editor() {
+
+	// Check if Gutenberg is active.
+	$is_gutenberg_active = is_plugin_active( 'gutenberg/gutenberg.php' );
```
DevinWalker: @Sidsector9 what happens when Gutenberg is merged into WP Core? Will this return properly since it will no longer be a plugin? Please take into account when it is merged into core. Is there a codex option that's necessary?

DevinWalker: @Sidsector9 can you simply check the version of WP here for 5.0 plus so we don't have to go back and refactor?

Sidsector9: @DevinWalker I have added check for WPv5 and also for active Gutenberg plugin if WP version is < 5.

includes/misc-functions.php
```diff
@@ -2288,3 +2288,68 @@ function give_get_safe_asset_url( $url ) {
...
+}
+
+/**
+ * Returns if the editor page is a Gutenberg editor.
+ *
+ * @since 2.2.3
+ *
+ * @return boolean
+ */
+function give_is_gutenberg_editor() {
```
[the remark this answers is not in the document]
Sidsector9: @ravinderk Any suggestions?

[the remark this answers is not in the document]
ravinderk: @Sidsector9 check if Gutenberg defines any custom constant or hook.

includes/class-notices.php
```diff
@@ -209,6 +209,14 @@ public function render_admin_notices() {
 			return;
 		}
 
+		// Do not render notice on Gutenberg editor page.
```
ravinderk: @Sidsector9 @DevinWalker For now we can use the Gutenberg core function to detect editor page.

PR #4016, opened 2019-02-21
Feature/give update donor information post create filter

andrewminion-luminfire:
## Description
Adds a filter to modify the `$donor` object after retrieving/creating it.

Use case: we’re building a plugin to add matching donations and the match donations should not be credited to the original donor.

## How Has This Been Tested?
Tested on a local WordPress installation.

## Types of changes

New feature: adds a new filter

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/payments/class-give-payment.php
+++ b/includes/payments/class-give-payment.php
@@ -747,14 +747,16 @@ final class Give_Payment {
 			 */
 			$donor = apply_filters( 'give_update_donor_information_post_create', $donor, $payment_id, $payment_data, $args );
 
-			// Update Donor Meta once donor is created.
-			$donor->update_meta( '_give_donor_first_name', $this->first_name );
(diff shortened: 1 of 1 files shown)


andrewminion-luminfire:
After writing this, I thought of perhaps a cleaner way of accomplishing the same thing for my use case and submitted PR https://github.com/impress-org/give/pull/4017.

nciske:
@mathetos / @DevinWalker 

We'd appreciate if ya'll could give this some attention so we don't have to write an iffy workaround in our plugin. Thanks!

nciske:
Or this alternative: https://github.com/impress-org/give/pull/4017

kevinwhoffman:
Closing. See https://github.com/impress-org/give/pull/4017#issuecomment-470595748 for rationale.

PR #4017, opened 2019-02-21
add filter just after retrieving logged-in donor

andrewminion-luminfire:
## Description
Adds a filter to modify the `$donor` object after retrieving it from the logged-in user.
Similar to https://github.com/impress-org/give/pull/4016 but seems cleaner/more logical since this can be used in tandem with the `give_update_donor_information` filter to do the same thing.

Use case: we’re building a plugin to add matching donations and the match donations should not be credited to the original donor.

## How Has This Been Tested?
Tested on a local WordPress installation.

## Types of changes

New feature: adds a new filter

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/payments/class-give-payment.php
+++ b/includes/payments/class-give-payment.php
@@ -713,6 +713,18 @@ final class Give_Payment {
 			if ( did_action( 'give_pre_process_donation' ) && is_user_logged_in() ) {
 				$donor = new Give_Donor( get_current_user_id(), true );
 
+				/**
+				 * Filter donor class after the donation is completed and after retrieving the logged-in user and before customer table is updated.
+				 *
+				 * @since 2.5.0
+				 *
+				 * @param Give_Donor $donor        Donor object.
+				 * @param int        $payment_id   Payment ID.
+				 * @param array      $payment_data Payment data array.
+				 * @param array      $args         Payment args.
+				 */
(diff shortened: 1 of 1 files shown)


kevinwhoffman:
@ravinderk Please review this PR and https://github.com/impress-org/give/pull/4016 and leave your recommendation for which approach is best. We can review on this week's dev team call.

kevinwhoffman:
Hi @andrewminion-luminfire and @nciske. We are preparing Give 2.4.2 for release and I wanted to update you on how we've resolved this scenario for you.

After reviewing our own code, we realized that the `give_update_donor_information` filter was not in a very useful location to begin with. It was basically filtering an empty object. As you suggested, a filter would be much more helpful if it occurred after the donor object was hydrated.

So rather than add a new filter, we have moved the location of the `give_update_donor_information` filter to appear later where it is more useful. This location allows the donor object to be filtered after it is hydrated (either from an existing WordPress user or via email).

See this PR for full details:
https://github.com/impress-org/give/pull/4035/files

Thanks for taking the time to contribute. I'm closing out your PRs, but if the solution above doesn't do what you're looking for, let us know and we'll make sure to address it.

andrewminion-luminfire:
Thanks @kevinwhoffman—moving that filter works perfectly for what we need!

PR #4046, opened 2019-03-14
Issue/4045

DevinWalker:
## Description

This PR ensures that when "Guest Donations" is set to `enabled` that the log in is not required to process a donation.

Resolves https://github.com/impress-org/give/issues/4045

## How Has This Been Tested?

- Manually

## Types of changes

- Bug Fix

## Checklist:
- [x] My code is tested.


kevinwhoffman:
I confirmed the following for a form where `Registration` setting is set to `Login`:

- Login is optional when guest donations are enabled.
- Login is required when guest donations are disabled.

Behavior seems to accurately reflect settings now. 👍

PR #4057, opened 2019-03-22
update TCPDF to address security vulnerability

andrewminion-luminfire:
## Description
Update the bundled `includes/libraries/tcpdf` to 6.2.26, the same version as included via `composer`.

This should resolve this security vulnerability: https://nakedsecurity.sophos.com/2019/03/21/flaw-in-popular-pdf-creation-library-enabled-remote-code-execution/

## How Has This Been Tested?
~I have not tested this—assume that the `composer` version should be fine since you're already using it when possible.~ I removed the `composer`-installed version and exported PDF reports.

## Types of changes
Bug fix: update TCPDF library.

## Checklist:
- [x] My code is tested.
- [x] My code follows the WordPress code style.
- [x] My code follows has proper inline documentation.

PR #4058, opened 2019-03-22
Remove local tcpdf

andrewminion-luminfire:
## Description
Removes the included TCPDF library, relying on `composer` to install it instead. Alternative to #4057 

## How Has This Been Tested?
I removed the included TCPDF library, ran `composer install` to install dependencies, and used it in a local WordPress installation. I exported a PDF of donations/income and it worked fine.

## Types of changes
Bugfix: removes old TCPDF library.

## Checklist:
- [x] My code is tested.


DevinWalker:
Unfortunately we have many users still using old add-ons referencing the `includes/library` version. In the future we're going to remove this copy though.

PR #4131, opened 2019-06-07
fix: update Nonce verification error message

### LINKED ISSUE #4129: fix: remove the word "nonce" from all front end error messages.
# Bug Report

## User Story

As a donor or site admin, I want error messages that are helpful and inoffensive so that I can take action to resolve the problem.

In the UK (at least) the word `nonce` is slang associated with concept of pedophilia.

## Current Behavior
There are several places in the Give Core Code as well as add-ons that the word `nonce` is used in the front-end output of an error. So in addition to it not being helpful (a donor has no idea what `Nonce verification failed` means or the steps to resolve the problem), it can be offensive to see a slang term displayed.

## Expected Behavior
I expect any error message to point to what specifically is going wrong, and how the reader of the error message can resolve the issue. 

I expect that a site administrator or support technician would know exactly where in the code the error message is coming from to be able to resolve it quickly.

I expect there to be no offensive terminology in error messages.
## Bug Type

## Steps to Reproduce
1. Browse the code searching for the word `nonce` in translatable/echoed strings. 
2. See that many of them are similar to one another, but slightly different in terms of precisely the same string.

## Possible Solution
Make a list of each error message, and explain in layman's terms what is actually happening to cause the error message.
Create error messages that allow the Donor, User, and Support technician to all take necessary action.

## Related
https://secure.helpscout.net/conversation/873431370/41330/

## Acceptance Criteria
- [ ] the word `nonce` is never output on the front end or back end of the site.
- [ ] All error messages have a unique identifier, eg `Error 38: session issues. Please contact _______`
- [ ] All error messages have a next step for the reader of the message.

mathetos:
## Description

Fixes #4129 to remove the word "nonce" from appearing. This word is considered vulgar in the UK. Additionally, we have formatted these error messages in four different ways. This consolidates them all into one consistent string format to make the job of translating easier for translators. 

## How Has This Been Tested?

Visual confirmation.

## Types of changes

Various strings updated. *NOTE:* `npm run dev` was NOT run, so the `POT` files will be generated later.

## Checklist:
- [x] My code is tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ Give relies on several npm commands to get you started:
 * `npm run dev` - Runs a one time build for development. No production files are created.
 * `npm run production` - Builds the minified production files for release.
 
+##### Note only compatible with Node 10
+
 ### Development Notes
 
 * Ensure that you have `SCRIPT_DEBUG` enabled within your wp-config.php file. Here's a good example of wp-config.php for debugging:
--- a/includes/admin/donors/donor-actions.php
+++ b/includes/admin/donors/donor-actions.php
@@ -386,7 +386,7 @@ function give_add_donor_email( $args ) {
 	} elseif ( ! wp_verify_nonce( $args['_wpnonce'], 'give_add_donor_email' ) ) {
 		$output = array(
 			'success' => false,
-			'message' => __( 'Nonce verification failed.', 'give' ),
+			'message' => __( 'We\'re unable to recognize your session. Please refresh the screen to try again; otherwise contact your website administrator for assistance.', 'give' ),
 		);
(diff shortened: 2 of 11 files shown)


Benunc:
This still doesn't tell me as a support tech where the problem is occurring. Since the fix is different for nonce errors on different spots, I'd love something distinguishing them from one another in the error message, or at least in a log.

mathetos:
From my quick look at the context of each string, most of these are generic error helper functions; meaning the same error is used in many different places -- but most often for only slightly different reasons. Truthfully, I don't think there's much more actionable information to be provided, if a session fails, whether it's in the admin or frontend it's simply that the session couldn't identify the user, there's not much nuance to it.

PR #4170, opened 2019-07-11
fix: only show nag when Stripe Premium is lower than 2.2.0

### LINKED ISSUE #4169: fix: recurring + stripe notice should only display if Stripe is active
## Issue Overview

The "Attention: Give 2.5.0+ requires the latest version of the Recurring Donations add-on to process payments properly with Stripe." license notice displays even if I do not have Stripe active.

### Reported

https://wordpress.org/support/topic/persistent-nag-message-due-to-expired-add-on-license

I have confirmed the issue on my side as well.

DevinWalker:
## Description

Only displays the Stripe + Recurring version notice check if Stripe is active and less than version `2.2.0`.

Resolves #4169 

## How Has This Been Tested?

- Manually

## Types of changes

- Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/give.php
+++ b/give.php
@@ -700,10 +700,12 @@ if ( ! class_exists( 'Give' ) ) :
 
 				Give()->notices->register_notice(
 					array(
-						'id'          => 'give-compatibility-with-old-recurring',
-						'type'        => 'error',
-						'description' => $message,
-						'show'        => true,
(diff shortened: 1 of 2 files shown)


mehul0810:
Moving this issue and PR to my Sprint backlog as it still needs some improvement

PR #4179, opened 2019-07-15
fix: the stripe connect banner has appearance issues and the temporary disabling of it no longer works #4175

### LINKED ISSUE #4175: fix: the stripe connect banner has appearance issues and the temporary disabling of it no longer works
## Issue Overview

The Stripe connect banner does not have the same font size as it used to and the "Not right now" text doesn't work when clicked.

Related to #4157 

## Acceptance Criteria
- [x] Reproduce the issue
- [x] Ensure that the connect banner has proper font size
- [x] Ensure that the connect banner don't have 2 dismiss icons when Stripe add-on is active
- [x] Ensure that the connect banner is dismissble for 24 hours irrespective of Stripe add-on is active or not.
- [x] Testing

mehul0810:
## Description
This PR resolves #4175 

## How Has This Been Tested?
I've tested this PR as per the acceptance criteria on the issue. Also, worked on a fatal error which can be caused in future Stripe versions which is fixed within this PR.

## Types of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/give.php
+++ b/give.php
@@ -583,7 +583,7 @@ if ( ! class_exists( 'Give' ) ) :
 				! defined( 'GIVE_STRIPE_VERSION' ) ||
 				(
 					defined( 'GIVE_STRIPE_VERSION' ) &&
-					version_compare( '2.2.0', GIVE_STRIPE_VERSION, '>=' )
+					version_compare( GIVE_STRIPE_VERSION, '2.2.0', '>=' )
 				)
 			) {
 				require_once GIVE_PLUGIN_DIR . 'includes/gateways/stripe/class-give-stripe.php';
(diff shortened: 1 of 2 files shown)


give.php
```diff
@@ -583,7 +583,7 @@ private function includes() {
 				! defined( 'GIVE_STRIPE_VERSION' ) ||
 				(
 					defined( 'GIVE_STRIPE_VERSION' ) &&
-					version_compare( '2.2.0', GIVE_STRIPE_VERSION, '>=' )
+					version_compare( GIVE_STRIPE_VERSION, '2.2.0', '>=' )
```
DevinWalker: Oops, good catch here @mehul0810

PR #4255, opened 2019-09-20
fix: avoid fatal error with recurring donation using GiveWP core patch

mehul0810:
## Description
This PR resolves the fatal error issue when Stripe is not enabled payment gateway in GiveWP core and recurring `1.9.3` is used.

## How Has This Been Tested?
**Please check the testing video:** https://www.loom.com/share/e65239b333124cd195e219a61afdcf70

## Types of changes
Bugfix (non-breaking change which fixes an issue)

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/gateways/stripe/class-give-stripe.php
+++ b/includes/gateways/stripe/class-give-stripe.php
@@ -66,7 +66,7 @@ if ( ! class_exists( 'Give_Stripe' ) ) {
 			// Load files which are necessary for front as well as admin end.
 			require_once GIVE_PLUGIN_DIR . 'includes/gateways/stripe/includes/give-stripe-helpers.php';
 
-			// Bailout, if any of the Stripe gateway is not active.
+			// Bailout, if any of the Stripe gateways are not active.
 			if ( ! give_stripe_is_any_payment_method_active() ) {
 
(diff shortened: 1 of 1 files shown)


includes/gateways/stripe/class-give-stripe.php
```diff
@@ -68,6 +68,41 @@ public function includes() {
...
 			if ( ! give_stripe_is_any_payment_method_active() ) {
+
+				// If `get_plugin_data` fn not exists then include the file.
+				if ( ! function_exists( 'get_plugin_data' ) ) {
+					require_once( ABSPATH . 'wp-admin/includes/plugin.php' );
+				}
+
+				// Hardcoded recurring plugin basename to show notice even when recurring addon is deactivated.
+				$recurring_plugin_basename = 'give-recurring/give-recurring.php';
+				$recurring_plugin_data     = get_plugin_data( WP_CONTENT_DIR . '/plugins/' . $recurring_plugin_basename );
```
DevinWalker: @mehul0810 you should use https://developer.wordpress.org/reference/functions/plugin_dir_path/ or similar in case this isn't the standard path.

PR #4277, opened 2019-10-10
fix: prevent fatal error on upgrade #4272

### LINKED ISSUE #4272: fix: prevent fatal error on upgrade
# Bug Report

## User Story

As an admin, I want to be able to update my plugins without any errors so that I can take advantage of enhancements and security patches.

## Current Behavior

Under these conditions: 
- Give: <= 2.5.4
- Recurring: <= 1.9.3
- No Stripe gateway enabled

When the user updates Give on the WordPress admin Plugins page, it will appear to update correctly, but the user will get a `Fatal Error` on the next page load because the check to disable Recurring 1.9.3 is triggered by `plugins_loaded` which is not called during the AJAX plugin update. 

The user will see one of the following errors: 

```
Fatal error: 
Uncaught Error: Class 'Give_Stripe_Gateway' not found in /home/nerdru7/http://caveofapelles.com/wp-content/plugins/give-recurring/includes/gateways/stripe/class-give-recurring-stripe-gateway.php:118 

Stack trace: 
#0 ****/wp-content/plugins/give-recurring/includes/gateways/give-recurring-gateway.php(82): Give_Recurring_Stripe_Gateway->init() 
#1 ****/wp-content/plugins/give-recurring/includes/gateways/stripe/class-give-recurring-stripe-webhooks.php(46): Give_Recurring_Gateway->__construct() 
#2 ****/wp-content/plugins/give-recurring/includes/gateways/stripe/class-give-recurring-stripe-webhooks.php(236): Give_Recurring_Stripe_Webhooks->__construct() 
#3 ****/wp-content/plugins/give-recurring/give-recurring.php(262): require_once('/home/nerdru7/c...') 
#4 ****/wp-content/plugins/give-recurring/give-recurring.php(189): Give_Recurring->includes_global() 
#5 ****/wp-content/plugins/give-recurring/includes/gateways/stripe/class-give-recurring-stripe-gateway.php on line 118
```

or 

```
There was a problem with the Recurring Donations add-on. It triggered a fatal error. See details -- "An error of type E_ERROR was caused in line 296 of the file /home/melaniej/public_html/wp-content/plugins/give-recurring/give-recurring.php. Error message: Call to undefined function give_stripe_is_any_payment_method_active()".
```

Reloading the page, or navigating to another page after getting the error will correct the error by properly deactivating the incompatible version of Recurring as expected. 

## Expected Behavior

Admin update of Give plugin will not cause a fatal error. 

While the error does appear to go away on any additional page load, the customer will not know that and will try to recover their site. 

## Bug Type

- [x] This bug describes functionality that once worked as expected in version 2.5.4.

## Steps to Reproduce

1.Install & Activate Give 2.5.4
2. Install & Activate Recurring 1.9.3
3. Disable Stripe Payment Gateway
4. Update Give in WP Admin
5. Reload page

Reload the page again, or any other page, and the fatal error will go away, and recurring will be disabled. 

## Visuals

https://www.loom.com/share/55cabbfc3bd841bd80c7fbb6b5105c32

## Possible Solutions

- Create a "dummy" Give_Stripe Gateway class to prevent the error
- Find a way to do the check currently done after `plugins_loaded` during the WP admin ajax plugin update

## Acceptance Criteria
- [x] Updating Give to 2.5.8 from 2.5.4 with Recurring 1.9.3 or less should not result in a fatal error
- [x] Don't deactivate recurring add-on when Give is updated to 2.5.8 with no fatal error
- [x] Give can be updated without causing a fatal error

mehul0810:
## Description
This PR resolves #4272 

## How Has This Been Tested?
I've tested this PR as per the acceptance criteria on the issue description.

Find the below-detailed testing scenario, I've handled for Give + Recurring when Stripe gateway is disabled:
> When a customer have Give 2.5.4 or lower and Recurring 1.9.3 or lower, then updating to Give 2.5.8 which usually resulted in a fatal error and then on additional refresh deactivates the recurring add-on to get rid of a fatal error.
- I've improved this scenario. So, now when customer upgrades to Give 2.5.8 having Recurring 1.9.3 or lower, then the recurring add-on is not deactivated and there will be no fatal error. Donations will also process smoothly. But, still, we have a notice in place to let customer update recurring to 1.9.4+ to ensure that any edge case doesn't affect donation processing.

Let me know, if you have any questions related to the testing scenarios.

## Types of changes
Bugfix (non-breaking change which fixes an issue)

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/gateways/stripe/class-give-stripe.php
+++ b/includes/gateways/stripe/class-give-stripe.php
@@ -103,7 +103,7 @@ if ( ! class_exists( 'Give_Stripe' ) ) {
 									'id'          => 'give-recurring-fatal-error',
 									'type'        => 'error',
 									'description' => sprintf(
-										__( '<strong>Activation Error:</strong> Please update the Recurring Donations add-on to version <strong>1.9.4+</strong> in order to be compatible with GiveWP <strong>2.5.5+</strong>. If you are experiencing any issues please rollback GiveWP to 2.5.4 or below using the <a href="%s" target="_blank">WP Rollback</a> plugin and <a href="%s" target="_blank">contact support</a> for prompt assistance.', 'give'),
+										__( '<strong>Action Needed:</strong> Please update the Recurring Donations add-on to version <strong>1.9.4+</strong> in order to be compatible with GiveWP <strong>2.5.5+</strong>. If you are experiencing any issues please rollback GiveWP to 2.5.4 or below using the <a href="%s" target="_blank">WP Rollback</a> plugin and <a href="%s" target="_blank">contact support</a> for prompt assistance.', 'give'),
 										'https://wordpress.org/plugins/wp-rollback/',
 										'https://givewp.com/support/'
 									),
--- a/includes/gateways/stripe/class-give-stripe.php
+++ b/includes/gateways/stripe/class-give-stripe.php
@@ -88,6 +88,13 @@ if ( ! class_exists( 'Give_Stripe' ) ) {
 						isset( $recurring_plugin_data['Version'] ) &&
 						version_compare( '1.9.3', $recurring_plugin_data['Version'], '>=' )
 					) {
+
+						// Load Stripe SDK.
+						give_stripe_load_stripe_sdk();
+
+						// Include frontend files.
+						$this->include_frontend_files();
+
 						add_action('admin_notices', function() {
 
 							// Register error notice.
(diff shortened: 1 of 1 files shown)


mikejhale:
Worked with @mehul0810 to sort out the issue I was having in testing and confirm the the fix prevents the fatal error when upgrading from 2.5.4.

PR #4317, opened 2019-12-03
fix: improve shortcode button styles

henryholtgeerts:
## Description

- surround shortcode button text with span of class "give-shortcode-button-text"
- create new style rule w/ media query in admin/forms.scss to hide button text at certain resolutions (prevents overlapping)
- update button icon style to inherit button text color

## How Has This Been Tested?
Tested locally with, and checked that the button text is hidden at appropriate breakpoints. Also tested button icon color by testing different WP admin color palettes, and the icon color matches button text color.

[Button w text shown](https://drive.google.com/file/d/1OqRR6MGRxvXpSJB8H0IEEVaYT8e6bYmS/view?usp=sharing
)[Button w text hidden](https://drive.google.com/file/d/1l3W7CdLiUvuFpmpIrkI2j2e8wIK_f8uZ/view?usp=sharing)
[Button at smaller with, w text shown](https://drive.google.com/file/d/1B_-MN7dVV4nqJCSEORaiHfUFqJhLItbN/view?usp=sharing)
[Button styling with different admin color palette](https://drive.google.com/file/d/1XxHQFgGps562dKU966pjmheGNxasbnTb/view?usp=sharing)

## Types of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/assets/src/js/admin/button.js
+++ b/assets/src/js/admin/button.js
@@ -41,11 +41,9 @@ class GiveButton {
 	updateIcon( className, position = 'before' ) {
 		const icon = `<span class="${className}"></span>`;
 
-		if ( 'after' === position ) {
-			this.root.innerHTML = `${this._buttonTextWithWrapper} ${icon}`;
-		} else {
-			this.root.innerHTML = `${icon} ${this._buttonTextWithWrapper}`;
-		}
+		this.root.innerHTML = 'after' === position
+			? `${this._buttonTextWithWrapper} ${icon}`
+			: `${icon} ${this._buttonTextWithWrapper}`;
 
 		this.iconClassName = className;
 		this.iconPosition  = position;
--- a/assets/src/js/admin/button.js
+++ b/assets/src/js/admin/button.js
@@ -3,6 +3,7 @@ class GiveButton {
 		this.root          = element;
 		this.buttonText    = this.root.textContent.trim();
 		this.iconPosition  = 'before';
+		this.text = this.root.querySelector('.give-button-text') ? `<span class="give-button-text">${this.buttonText}</span>` : this.buttonText;
 
 		const icon = this.root.querySelector( '.dashicons' );
 
@@ -21,8 +22,7 @@ class GiveButton {
 
 	setBusyState() {
 		const busyIcon      = '<span class="dashicons dashicons-marker"></span>';
-		const text = this.root.querySelector('.give-button-text') ? `<span class="${this.buttonText}"></span>` : this.buttonText;
-		this.root.innerHTML = `${busyIcon} ${text}`;
+		this.root.innerHTML = `${busyIcon} ${this.text}`;
 		this.disable();
 	}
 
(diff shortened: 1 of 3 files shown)


ravinderk:
@helgatheviking I found the following issue when reviewing this pr:
- [ ] How did you decide the maximum screen size to hide button text because I can still see the overlap
image
- [ ] The text becomes visible on small screens and overlaps with others.
ovvvZuvpb5
- [ ] Can you confirm that hidden text will be readable on a screen reader because otherwise, they will not get an idea of what this button does?

@henryholtgeerts  you can upload images directly to GitHub instead of uploading it to google drive (blur image parts which contain private information)

henryholtgeerts:
Ah! I hadn't accounted for how button.js is reprinting the innerHTML for the button. I'll hop on a fix for that. I determined the width with some math taking into account the table columns, wp admin sidebar and margins. When tested on my own machine it hides at the right point. Wondering if this problems persists prior to having copied a shortcode?

helgatheviking:
@henryholtgeerts errrr.... I think you tagged me by mistake. :)

henryholtgeerts:
Sorry @helgatheviking ! Oops

@ravinderk In order to fix the problem with button.js removing the span tag as it reprinted the innerHTML, I modified the setBusyState and updateIcon functions of button.js. Those functions now first define the text of the button in a constant (mirroring what they did for icons) and use a ternary operator to determine text should be wrapped in a span if one is already present.

This could be moved to shortcode-button.js to limit the changes scope, but it seems useful to have someway of isolating text inside these buttons, and the new code shouldn't impact existing functionality.

ravinderk:
@henryholtgeerts Did you test your pr when taxonomies are active:
image

Note: you can enable taxonomies from:
image

assets/src/js/admin/button.js
```diff
@@ -21,7 +21,8 @@ class GiveButton {
 
 	setBusyState() {
 		const busyIcon      = '<span class="dashicons dashicons-marker"></span>';
-		this.root.innerHTML = `${busyIcon} ${this.buttonText}`;
+		const text = this.root.querySelector('.give-button-text') ? `<span class="${this.buttonText}"></span>` : this.buttonText;
```
ravinderk: instead of this why we can not use `this.buttonText` in the constructor.
```
this.buttonText    = this.root.textContent.trim();

// after update
this.buttonText    = this.root.querySelector('.give-button-text') ? `<span class="${this.buttonText}"></span>` : this.root.textContent.trim();
```

by the above way, we do not have to change code at multiple places.

henryholtgeerts: This was something I explored, but realized it adds more complexity to the updateText() function, where as this implementation does not impact the updateText() function at all. I just committed one possible solution, refactor both querySelectors into the constructor as you suggested but under a new property called "text" which refers to either the buttonText string of the full text span, depending on what is present.

DevinWalker:
Works well. Here's my testing video: https://www.loom.com/share/b22212affb27424dbf7f9ecea6f24605

Thanks @henryholtgeerts

PR #4350, opened 2019-12-18
feature: add list component

### LINKED ISSUE #4332: feature: add list component
# Feature Request

## User Story
As a developer, I want a simple list component so that I can quickly build an effective reports UI that only loads data as it is needed.

The list component should solely act as a list container. It should accept an onApproachScrollEnd prop and children which it formats as list items.

It should **not** be responsible for the content of the list items.

## Visuals
Example 1:
Screen Shot 2019-12-13 at 10 18 03 AM

Example 2:
Screen Shot 2019-12-13 at 10 18 16 AM

See Figma: 
[reports ui boards](https://www.figma.com/file/w38WexhmfOkcXRvFOTNgcO/GiveWP-Updated-Admin-Screens?node-id=0%3A10592)

## Acceptance Criteria
- [x] Styled to match Figma
- [x] Does not manage content of list items
- [x] Accepts onApproachScrollEnd callback prop

henryholtgeerts:
## Description
Resolves #4332 
Added list component for Reports UI. The component accepts children and an onScrollEnd prop which fires as a calllback function when the component is scrolled all the way to the bottom. 

## How Has This Been Tested?
Tested locally and with no errors logged in console.

Screen Shot 2019-12-19 at 11 24 27 AM

## Types of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/assets/src/js/admin/reports/components/list/index.js
+++ b/assets/src/js/admin/reports/components/list/index.js
@@ -1,12 +1,9 @@
 import { useEffect, createRef } from 'react'
+import './style.scss';
 
 const List = ({onScrollEnd, children}) => {
 
     const list = createRef()
-    const listStyle = {
-        height: '432px',
-        overflow: 'scroll'
-    }
 
     useEffect(() => {
 
@@ -30,7 +27,7 @@ const List = ({onScrollEnd, children}) => {
     }, [])
 
     return (
-        <div ref={list} style={listStyle}>
+        <div ref={list} class='list'>
             {children}
         </div>
     )
--- a/assets/src/js/admin/reports/components/list/index.js
+++ b/assets/src/js/admin/reports/components/list/index.js
@@ -6,7 +6,7 @@ const List = ({onScrollEnd, children}) => {
     
     const list = createRef()
 
-    useEffect((onScrollEnd) => {
+    useEffect(() => {
 
         function checkScroll (evt) {
 
@@ -26,7 +26,7 @@ const List = ({onScrollEnd, children}) => {
             }
         }
 
-    }, [])
+    }, [onScrollEnd])
 
     return (
         <div ref={list} className='list'>
--- a/assets/src/js/admin/reports/components/list/index.js
+++ b/assets/src/js/admin/reports/components/list/index.js
@@ -1,6 +1,6 @@
 import { useEffect } from 'react'
 
-const List = ({onApproachScrolleEnd, children}) => {
+const List = ({onApproachScrollEnd, children}) => {
 
     const list = createRef()
 
--- a/assets/src/js/admin/reports/components/list/index.js
+++ b/assets/src/js/admin/reports/components/list/index.js
@@ -1,6 +1,6 @@
 import { useEffect, createRef } from 'react'
 
-const List = ({onApproachScrollEnd, children}) => {
+const List = ({onScrollEnd, children}) => {
 
     const list = createRef()
     const listStyle = {
@@ -16,7 +16,7 @@ const List = ({onApproachScrollEnd, children}) => {
             const height = evt.target.offsetHeight
 
             if (remaining <= height) {
-                onApproachScrollEnd()
+                onScrollEnd()
             }
 
         }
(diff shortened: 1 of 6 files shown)


JasonTheAdams:
@henryholtgeerts Quick question on this: Is the scroll-to-end trigger something you plan on using?

henryholtgeerts:
@JasonTheAdams yes! planning on using onScrollEnd to trigger loading more list items from the rest api.

assets/src/js/admin/reports/components/list/index.js
```diff
@@ -1,6 +1,35 @@
...
+                onScrollEnd()
+            }
+
+        }
+        
+        list.current.addEventListener('scroll', checkScroll)
+    
+        return function cleanup () {
+            list.current.removeEventListener('scroll', checkScroll)
+        }
```
JasonTheAdams: I suggest allowing `onScrollEnd` to be optional. As such, only construct and destruct the event listener if that variable is set.

assets/src/js/admin/reports/components/list/index.js
```diff
@@ -1,6 +1,35 @@
...
+
+    }, [])
```
JasonTheAdams: Since you're referring to the `onScrollEnd` prop you'll want to make it a dependency of `useEffect`.

assets/src/js/admin/reports/components/list/index.js
```diff
@@ -1,6 +1,35 @@
...
+
     return (
-        <h1>List goes here!</h1>
+        <div ref={list} class='list'>
+            {children}
+        </div>
     )
 }
 export default List
```
JasonTheAdams: Please add `List.propTypes` and `List.defaultProps`.

henryholtgeerts: done! also added styles for "default-item" in the event that a list is empty, to display an error message.

assets/src/js/admin/reports/components/list/index.js
henryholtgeerts: Done! onScrollEnd is now passed to useEffect

assets/src/js/admin/reports/components/list/index.js
henryholtgeerts: done! I added a conditional check around this code to first check that onScrollEnd is defined

assets/src/js/admin/reports/components/list/index.js
JasonTheAdams: Sorry, that's not what I meant. 😄 

Parameters aren't used with `useEffect` as it's just a generic callback. When I said "make it a dependency", I meant:
```js
useEffect(() => {
  // code stuff
}, [onScrollEnd]);
```

assets/src/js/admin/reports/components/list/index.js
```diff
@@ -1,6 +1,52 @@
...
-        <h1>List goes here!</h1>
+        <div ref={list} className='list'>
+            {children}
+        </div>
     )
 }
+
+List.propTypes = {
+    onScrollEnd: PropTypes.func,
+    children: PropTypes.node
```
JasonTheAdams: Please add comments to document the props:
```
/** Callback triggered when the list is scrolled to its end */
onScrollEnd: PropTypes.func

/** Elements to render within the list **/
children: PropTypes.node

File: assets/src/js/admin/reports/components/list/index.js (hunk 3 above)
Comment (JasonTheAdams) [reply assets/src/js/admin/reports/components/list/index.js]: If it's a required element, just mark the prop type as required: `PropTypes.node.isRequired`. As this error shouldn't ever really display as it's not a runtime error, but a failure to utilize this component properly.

File: assets/src/js/admin/reports/components/list/index.js [hunk 5]
```diff
@@ -1,6 +1,50 @@
...
+List.propTypes = {
+    /** Callback triggered when the list is scrolled to its end */
+    onScrollEnd: PropTypes.func,
+    /** Elements to render within the list **/
+    children: PropTypes.node.isRequired
+}
+
+List.defaultProps = {
+    onScrollEnd: null,
+    children: null
```
JasonTheAdams: You don't need to provide a default if a prop is required.

Issue #4592, opened 2020-03-27
fix: prevent Report related errors when accessing wp-cli


henryholtgeerts:
# Bug Report

## User Story
As a user, I want to use wp-cli with running into errors so that I am able to run all my wp-cli scripts.

## Current Behavior
When accessed via wp-cli, a site running givewp can generate this error: 
`PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function Give\API\Endpoints\Reports\Endpoint::get_report_schema(), 0 passed in [stripped]/wp-includes/rest-api/class-wp-rest-server.php on line 1165 and exactly 1 expected in...`

## Expected Behavior
No Reports related errors should appear.

## Bug Type
- [ ] This bug describes functionality that once worked as expected in version X.X.X.
- [x] This bug describes functionality that never worked as expected.
- [ ] I am not sure whether this functionality ever worked as expected.

## Possible Solution
Remove the $request parameter being passed to get_report_schema() as it is not needed.

## Acceptance Criteria
- [x] Fatal error relating to get_report_schema() fn does not appear when using wp-cli

PR #4593, opened 2020-03-27
fix: prevent Reports related issue when using wp-cli

henryholtgeerts:
## Description
This PR resolves issue #4592. In the original issue, a Fatal Error was thrown when accessing a site through wp-cli with Give installed because get_report_shema() received too few arguments. In this PR, the $request parameter is no longer expected by the get_report_schema() fn, solving the issue.

## How Has This Been Tested?
Tested locally, passed PHPUnit tests. Functional when using wp-cli.

## Types of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/src/API/Endpoints/Reports/Endpoint.php
+++ b/src/API/Endpoints/Reports/Endpoint.php
@@ -178,10 +178,8 @@ abstract class Endpoint {
 
 	/**
 	 * Get our sample schema for a report
-	 *
-	 * @param WP_REST_Request $request Current request.
 	 */
-	public function get_report_schema( $request ) {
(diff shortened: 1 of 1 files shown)


JasonTheAdams:
@henryholtgeerts Even though a potential solution is described in the connected issue, please take the time to briefly summarize the issue, explain _what_ your solution is, and _how_ it resolves the issue.

JasonTheAdams:
@henryholtgeerts Pushing on this a little more, just for the learning opportunity.

Why was the parameter there to begin with? Why is it no longer necessary? Why is this the right way to solve it?

PR #4603, opened 2020-03-31
feature: add sliding animation to steps in sequoia template

### LINKED ISSUE #4548: feature: add sliding animation to steps in Sequoia template
# Feature Request

## User Story
As a user, I want a sliding animaiton to appear between steps of the Sequoia template so that I have smooth user experience.

## Acceptance Criteria
- [ ] When Sequoia template is activated, a sliding animation is utilized to move from one step to the next
- [ ] Sliding animation is accomplished in a performant way that works well across devices/browsers

henryholtgeerts:
## Description
Resolves #4548 
This PR uses CSS and JS to introduce a sliding effect for the Sequoia template, so that it slides from step to step. 

The effect is accomplished by ensuring that each "step" in encapsulated by a separate div, which is then positioned absolute, then animated with a combination of css transforms and css display settings in the goToStep function of the navigator object (referencing state stored in the currentStep property of the navigator object). 

Notably, this PR also moves 'setup' of each step out of the `navigator.goToStep` function, and into the new `navigator.init` function. This ensures that the height of each step can be accurately calculated within the goToStep function.

## Affects
This PR affects the markup, JS and CSS of the Sequoia template. Most of these changes live in the Sequoia assets directory, but changes to markup exist in Actions.php.

## What to test
Check that all steps of the Sequoia form template are functional. When you expand payment gateways are they visible?

Check that margins above the footer area are correct. This PR animated the footer's top margin to make vertical room for each step. 

## Checklist:
- [x] My code is tested.


henryholtgeerts:
> Presently, this PR runs into issues with how the iframe resizer package handles updates to iframe height. I believe that the resizer package may be using a setTimeOut method to handle listening to changes in height, rather than requestAnimationFrame (which would be smoother). I'm planning on investigating this in a separate issue, since it may involve significant work and was becoming something of a blocker on this PR, which is otherwise ready for merge.

This iframe resizing issue is now being tracked in a new, recently opened issue: https://github.com/impress-org/givewp/issues/4616

With that issue now open, this PR is set to merge.

PR #4650, opened 2020-04-17
feature: add redirecting and verifying views

### LINKED ISSUE #4646: feature: add redirecting and verifying views for Sequoia template
# Feature Request

## User Story
As a user, I want to see redirecting, and verifying payment screens that match the loading screen so that I am never left wondering what is going on, or presented with a blank or ugly screen in the donation process.

## Details
When a user first loads a donation form, they see a loading screen before content is loaded. A similar screen should be presented while the user is being redirected to an offsite payment gateway, and while the payment is being verified after a user returns from an offsite payment gateway.

## Visuals
See Figma: https://www.figma.com/file/jaVXDA4wjQD9OqTi0l9KQe/Form-Templates?node-id=1882%3A44&viewport=-259%2C1586%2C0.5

## Possible Solution
Introduce getRedirectingView and getVerifyingView methods to the Template base class, then provide appropriate views (based on loading view) to the Sequoia template.

## Acceptance Criteria
- [ ] User is presented with redirecting view prior to being redirected to offsite payment gateway
- [ ] User is presented with verifying view after returning from offsite payment gateway, before receipt view
- [ ] Redirecting and verifying views match those in the Figma design, and recycle the loader element from the loading view
- [ ] Height of redirecting view matches height of payment information step, and height of verifying view matches height of receipt view

henryholtgeerts:
## Description
Resolves #4646 
This PR introduces the getRedirectingView method, and defines specific redirecting and verifying views for the Sequoia template. It also improves the loading functionality with the introduction of the showLoader action, which allows templates to, using javascript, instruct the parent iframe to display the loading view while the iframe loads a new view underneath.

## Affects
This feature affects the iframe content utilities, Sequoia assets and views as well as the base template class.

## What to test
Test different payment gateway, both onsite and offsite. Are you shown a redirecting view when expected? Do you see the verifying view when the payment is being processed? Doe sthe loader appear to ease transitions between views?

## Screenshots:
Currently n/a

## Checklist:
- [x] My code is tested.


henryholtgeerts:
> @henryholtgeerts can we use `window.load` to hide loader and `beforeulnoad` event to show loader?

After some testing, I've found that the order/cycle events are fired is this:
iframe load -> iframe ready -> iframe beforeunload

Because the iframe can't send messages to a parentwindow until the onReady event is fired, I've decided to rely on onReady, and in onReady set an event listener for before unload. Ref: https://github.com/impress-org/givewp/commit/8b5eb35c9876d7e8e75d50f0e2dc84845377bfd6

henryholtgeerts:
> @henryholtgeerts As discussed, please show loader till receipt load completely.

Done, ref: https://github.com/impress-org/givewp/commit/15c7e996571184b40db24a7a95c5a522b012e640

PR #4691, opened 2020-04-29
fix: Sequoia no longer shows tooltips for unlabeled amounts

### LINKED ISSUE #4685: fix: do not show tooltips in new form template when no custom text is provided
# Bug Report

## User Story

As a donor, I want tooltips that add context so that I'm motivated to give.

## Details

Right now the tooltips when you hover over display amounts display even when there's no content. It currently defaults to the donation amount, which is already displayed and not useful.

## Expected Behavior

Tooltips only display when content is added to the field.

## Acceptance Criteria

- [ ] Tooltips do not display when no content is added to a donation level
- [ ] Tooltips appear properly when content is added to a level

henryholtgeerts:
## Description
Resolves #4685 
This PR resolves an issue that caused amounts without labels to display tooltips in the Sequoia form template. Previously, when a user added an amount without a label, a tooltip would appearing over the amount button that simply displayed the amount again (eg: When a $25 level button was hovered, the tooltip also ready $25). The issue is caused by the way that GiveWP core renders multilevel amount donation forms, where a donation level left blank would simply render with the donation amount. 

This behavior makes sense in the context of legacy forms, but with Sequoia we are hacking the frontend to display the level's amount, and this "label" as a tooltip. With this in mind, I reached for a frontend solution that would be unique to Sequoia, that doesn't touch core logic, keeping in line with how we have handled transforming these amount buttons previously. Touching only frontend scripts, this PR checks if the provided label matches the amounts value, and if so, refrains from setting up a tooltip. 

## Affects
This PR only affects the frontend JS of the Sequoia template.

## What to test
Do tooltips appear above donation levels without labels? Are tooltips impacted by changing the currency sign or position? Do tooltips render correctly when a label is provided?

## Screenshots:
PRDemo

## Checklist:
- [x] My code is tested.

### DIFF
--- a/src/Views/Form/Templates/Sequoia/assets/css/form.scss
+++ b/src/Views/Form/Templates/Sequoia/assets/css/form.scss
@@ -733,7 +733,13 @@ fieldset {
 		word-wrap: normal;
 		white-space: normal;
 		width: fit-content;
-		max-width: 120px;
+		max-width: 136px;
+	}
+
+	&.wide {
+		&::after {
+			width: 136px;
+		}
 	}
 }
 
--- a/src/Views/Form/Templates/Sequoia/assets/css/form.scss
+++ b/src/Views/Form/Templates/Sequoia/assets/css/form.scss
@@ -719,17 +719,20 @@ fieldset {
 	margin: 0 !important;
 
 	&::after {
+		display: block;
+		padding: 12px 18px;
 		background: #383838;
 		color: #fff;
-		padding: 12px 18px;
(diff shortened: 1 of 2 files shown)


henryholtgeerts:
Changes requested in @DevinWalker's view review
- [ ] Ensure that currency symbol is formatted on donation levels that do not have label
- [ ] Provide better styling for tooltips with long copy (use ellipses to limit text, use max-width to prevent tooltip going beyond level button width)

Issue #4684, opened 2020-04-27
feature: when only one gateway is enabled, Sequoia should not display it as a radio selection


henryholtgeerts:
# Feature Request

## User Story
As a user, if only one gateway is enabled, I want Sequoia to render the gateways section without a radio selection so that I am not shown a radio/tab interface for only one option.

## Details
If only one payment gateway is enabled, instead of displaying payment gateways as radio selections, display only the enabled gateway.

## Possible Solution
If only one gateway is present, set radio related elements to display: none.

## Acceptance Criteria
- [ ] When multiple gateways are enabled, it behaves as expected (radio selection)
- [ ] With only one gateway enabled, no radio selection interface is displayed
- [ ] Solution only impacts Sequoia template, and not legacy template rendering

PR #4693, opened 2020-04-30
feature: when a single gateway is enabled, Sequoia renders that gateway without radio selectors

henryholtgeerts:
## Description
Resolves: #4684 
Pervasively, when only one gateway was enabled, the gateway would not appear in the Sequoia form template. Now, when only one gateway is enabled, it is rendered by itself, and not as part of any radio selection.

The solve this problem I made sure that the moveFieldsUnderPaymentGateway function was only called when multiple gateways are enabled, so that it would not be moved into the radio selection (which is set to display none when only a single gateway is enabled). 

## Affects
This fix only impacts Sequoia template frontend assets, specifically JS.

## What to test
Try enabling multiple payment gateways, then test enabling only one payment gateway. Do gateways appear as expected? When only one is enabled, does it load by itself rather than in a radio selection interface?

## Screenshots:
Screen Shot 2020-04-30 at 1 55 56 PM

## Checklist:
- [x] My code is tested.

### DIFF
--- a/src/Views/Form/Templates/Sequoia/assets/js/form.js
+++ b/src/Views/Form/Templates/Sequoia/assets/js/form.js
@@ -232,6 +232,7 @@
 
 	navigator.init();
 
+	// Check if only a single gateway is enabled
 	if ( $( '#give-payment-mode-select' ).css( 'display' ) !== 'none' ) {
 		// Move payment information section when document load.
 		moveFieldsUnderPaymentGateway( true );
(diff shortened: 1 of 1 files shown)


DevinWalker:
This is worked well after testing one gateway enabled and multiple enabled. We'll need to open a new issue to address the styling of CC elements for Stripe and Square because they're not looking good in the current state but that's a separate issue:

PR #4701, opened 2020-05-01
Hotfix/no need for white emojis

bradp:
The 👉️ in the README has a skintone modifier in it, no need for that.


bradp:
Oops targetting wrong branch

PR #4702, opened 2020-05-01
Emoji fix

bradp:
The :point_right: in the README has a skintone modifier in it, no need for that.


bradp:
There we go, not brining extra commits now. I can git

DevinWalker:
Nice! Thanks

PR #4733, opened 2020-05-13
Fix: Clean up phpcs linting and code standard testing

bradp:
## Description
Cleans up our usage of phpcs config files, so that pre-commits are checked with the same standard that is in `phpcs.ruleset.xml`. 

Also adds a `npm run php:lint` to run a lint manually without having to run the automated fix.

This also gives us a much cleaner `phpcs.ruleset.xml` that can be extended to cover our specific code standards.

## What to test
Making changes in the codebase that will flag issues or will not flag issues will be affected by this and can be tested that way.

## Checklist:
- [x] My code is tested.


bradp:
> @bradp Any reason why the package-lock.json file has ~37k changes?

Most of it is npm package updates, that happened while adding some new deps

phpcs.ruleset.xml
```diff
@@ -1,196 +1,30 @@
...
+	<rule ref="PHPCompatibilityWP" />
+
+	<!-- Exclude directories we don't want checked. -->
+	<exclude-pattern>node_modules/*</exclude-pattern>
+	<exclude-pattern>vendor/*</exclude-pattern>
+
+	<!-- Bring in WP rules. -->
+	<rule ref="WordPress-Core">
+			<exclude name="WordPress.Files.FileName.InvalidClassFileName" />
+			<exclude name="WordPress.WP.I18n.MissingTranslatorsComment" />
```
ravinderk, line 20: @bradp Can we remove the extra tab from here but I am wondering why it is not removed by linters

phpcs.ruleset.xml
```diff
@@ -1,196 +1,30 @@
...
-		<!-- <exclude name="WordPress.WhiteSpace.ControlStructureSpacing.SpaceBeforeFunctionOpenParenthesis" /> -->
-		<!-- <exclude name="WordPress.WhiteSpace.OperatorSpacing.NoSpaceAfter" /> -->
-		<!-- <exclude name="WordPress.WhiteSpace.OperatorSpacing.NoSpaceBefore" /> -->
-		<!-- <exclude name="WordPress.WhiteSpace.OperatorSpacing.SpacingAfter" /> -->
-		<!-- <exclude name="WordPress.WhiteSpace.OperatorSpacing.SpacingBefore" /> -->
-		<exclude name="WordPress.XSS.EscapeOutput.OutputNotEscaped" />
-		<exclude name="WordPress.XSS.EscapeOutput.UnsafePrintingFunction" />
-		<exclude name="WordPress.Files.FileName.InvalidClassFileName" />
+	<!-- Check for PHP cross-version compatibility. -->
+	<config name="testVersion" value="7.1-" />
```
ravinderk, line 10: @bradp we support minimum PHP `5.6` will it make any difference?

phpcs.ruleset.xml
```diff
@@ -1,196 +1,30 @@
 <?xml version="1.0"?>
-<ruleset name="WordPress Coding Standards">
+<ruleset name="GiveWP Coding Standards">
 	<!-- See https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-ruleset.xml -->
 	<!-- See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress-Core/ruleset.xml -->
 
 	<!-- Set a description for this ruleset. -->
 	<description>A custom set of code standard rules to check for WordPress themes and plugins.</description>
```
ravinderk, line 7: @bradp do we have to update this description?

phpcs.ruleset.xml
```diff
@@ -1,196 +1,30 @@
...
-		<exclude name="WordPress.XSS.EscapeOutput.OutputNotEscaped" />
-		<exclude name="WordPress.XSS.EscapeOutput.UnsafePrintingFunction" />
-		<exclude name="WordPress.Files.FileName.InvalidClassFileName" />
+	<!-- Check for PHP cross-version compatibility. -->
+	<config name="testVersion" value="7.1-" />
+	<rule ref="PHPCompatibilityWP" />
+
+	<!-- Exclude directories we don't want checked. -->
+	<exclude-pattern>node_modules/*</exclude-pattern>
+	<exclude-pattern>vendor/*</exclude-pattern>
```
ravinderk, line 15: @bradp I think we can also exclude `sample-data`, `language` directory?

phpcs.ruleset.xml
bradp: Extra tab removed. Wasn't hit by the linters, because nothing is linting xml files at the moment.

phpcs.ruleset.xml
ravinderk: what do you think about the `assets` directory?

phpcs.ruleset.xml
bradp: Commented out for now.

composer.json
```diff
@@ -1,38 +1,44 @@
...
-      "\"vendor/bin/phpcs\" --config-set installed_paths vendor/wp-coding-standards/wpcs || true"
-    ]
-  },
-  "autoload": {
-	"psr-4": {
-	  "Give\\": "src/"
-	}
-  }
+    "support": {
+        "issues": "https://github.com/impress-org/give/issues"
```
ravinderk, line 25: @bradp Github repo URL is: `https://github.com/impress-org/givewp`

package.json
```diff
@@ -9,11 +9,12 @@
         "production": "cross-env NODE_ENV=production webpack",
         "build": "npm run production",
         "watch": "cross-env webpack --watch",
+        "installcodestandards": "composer run-script install-codestandards",
```
ravinderk, line 12: @bradp what is use of this command?

bradp: Will let you manually run the composer script to install the code standards into the /vendor folder.

ravinderk: @bradp Can you tell me the use case where the developer has to run this manually?

phpcs.ruleset.xml
ravinderk: @bradp did you check this ☝️

package.json
bradp: Anyone who has a copy of the repo can easily run it to install the scripts without having to reset their local install. Mostly it's there to be a helper.

ravinderk:
@bradp I find out that WordPress force long array syntax but can we add a rule to force short array syntax:
https://github.com/WordPress/WordPress-Coding-Standards/blob/1f525cca6c6124c2bdee6e7bb9fd79d7e96df176/WordPress-Core/ruleset.xml#L109

package.json
ravinderk: @JasonTheAdams Do you have any thoughts on this?

.lintstagedrc.json
```diff
@@ -1,6 +1,6 @@
 {
     "*.json": ["prettier --write", "git add"],
-    "*.php": ["npm run lint:textdomain", "npm run lint:php:fix", "git add"],
+    "*.php": ["npm run lint:textdomain", "npm run lint:php", "npm run lint:php:fix", "git add"],
```
ravinderk, line 3: @bradp when I added `npm run lint:php`  then it was blocking us most of the time and few conditions were not easy to pass. I will suggest to remove it.

@JasonTheAdams any thought?

DevinWalker: @ravinderk I'm not sure what you're saying above. Can you expand so it's clear to me?

ravinderk: image

@DevinWalker WPCS forces function naming with `_(underscore)` but now we started function naming in camel case also in `src` directory.
If we will autorun `npm run lint:php` command on every commit then it will be failing with error as mentioned in the above screenshot and we have to use `--no-verify` on most of the commits to bypass checks.

Let me know if it makes sense.

JasonTheAdams:
@bradp The more I think about it the more I don't care for the massive `package-lock.json` update. Updating all of the package dependencies is beyond the scope of this PR, and it doesn't even add/update any of the packages in the `package.json` file. Please revert that file to it's original state.

package.json
JasonTheAdams: I don't personally see any value in running `npm run installcodestandards` versus `composer run-script install-codestandards`. This is really just a pre-defined npm script to run a composer script.

.lintstagedrc.json
JasonTheAdams: This would need to work on _only_ the `includes` directory. Is that possible, @bradp?

bradp: We can do that for sure, we should also put together the standards for the `src` directory so that we can get that in place at the same time.

@JasonTheAdams

bradp:
Based on convo here & in Slack, let's postpone merging until we can get the coding standards for `src` in as well.

@JasonTheAdams Thoughts on specifics for that?

.lintstagedrc.json
ravinderk: @JasonTheAdams I removed `npm run lint:php` command for now which prevents showing warnings when committing code instead of that it will format code and commit code.

JasonTheAdams:
I know I'm late to the party, but this looks good to me.

Issue #948, opened 2016-08-24
a11y: review row action links accessibility


ramiy:
WordPress actions links are accessible. Give action links are not.

Screen:
- [x] **Forms** - uses build-in post-type functionality.
- [x] **Donations** - custom actions links.
- [x] **Donors** - custom actions links.
- [x] **Logs** - no actions links.

PR #4764, opened 2020-05-22
fix: unable to process recurring donations via onpage stripe credit card fields #948

mehul0810:
## Description
This PR resolves https://github.com/impress-org/give-recurring/issues/948 and supports PR https://github.com/impress-org/give-recurring/pull/949

## Affects
We deprecated a function which is no longer in use `give_stripe_is_checkout_enabled`

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/gateways/stripe/includes/admin/admin-helpers.php
+++ b/includes/gateways/stripe/includes/admin/admin-helpers.php
@@ -123,17 +123,6 @@ function give_stripe_is_connected() {
 }
 
 /**
- * Is Stripe Checkout Enabled?
- *
- * @since 2.5.0
- *
- * @return bool
- */
-function give_stripe_is_checkout_enabled() {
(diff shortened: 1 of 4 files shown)


mehul0810:
@ravinderk I have removed all the deprecated function instances from PHP files. But due to less time I have I am unable to check for the JS file usages. Also, the JS files deprecated code will not be called as those settings are not available. So, if both the PRs of this issue needs to be urgently merged and released. Than, we can release and create a separate issue to remove dead code from JS files. Otherwise, I will update the PR on Monday.

CC: @DevinWalker

mehul0810:
@ravinderk You can review this PR as the requested changes related to JS are used for admin settings as a Stripe Checkout type (i.e. Legacy or Checkout 2.0). So, there is no dead code within this PR now. Let me know if you have any questions.

CC: @DevinWalker

PR #4771, opened 2020-05-26
fix: newsletter fields are now styled in Sequoia form template

### LINKED ISSUE #4754: feature: newsletter fields should be styled consistently in Sequoia form template
# Feature Request

## User Story
As a user, I want newsletter fields styled consistently with other inputs in the Sequoia form template so that I have a cohesive donation experience.

## Details
Newsletter addons (ConstantContact, Mailchimp, etc) introduce a checkbox in the donation process so that users can opt into communication from the website admin. These inputs should be styled consistently with other inputs in the Sequoia form template. 

## Visuals
See Figma: https://www.figma.com/file/jaVXDA4wjQD9OqTi0l9KQe/Form-Templates?node-id=1882%3A44&viewport=-259%2C1586%2C0.5

## Possible Solution
Use existing checkbox styling to also target checkboxes introduced by newsletter addons. Then, introduce specific newsletter input styling to center/resize the fields to match Figma designs.

## Acceptance Criteria
- [x] All newsletter addons are supported by Sequoia form template (ConstantContact, MailChimp)
- [x] Newsletter inputs match Figma design

henryholtgeerts:
## Description
Resolves #4754 
Previously, when newsletter addons (Constant Contact, Mailchimp)  were enabled, the checkbox field used browser native styling, and did not match the Sequoia form template. This PR introduces styles and some frontend logic to ensure that newsletter optin fields match other prominent checkboxes aesthetically and functionally. 

## Affects
This PR affects Sequoia form template assets (CSS and JS) as well as the Sequoia form template class. 

## What to test
When different newsletter addons activated, try viewing the form from the frontend. Do the optin checkboxes appear above the "Donate Now" button as expected? Do they match other prominent checkboxes in the form? Is the border persisted when the checkbox is active, and removed when inactive?

## Screenshots:
Screen Shot 2020-05-26 at 4 58 30 PM

## Checklist:
- [x] My code is tested.


src/Views/Form/Templates/Sequoia/assets/js/form.js
```diff
@@ -196,6 +196,26 @@
 					$( '.give-fee-recovery-donors-choice' ).toggleClass( 'active' );
 				} );
 
+				// If mailchimp is opted in by default, add border on load
+				if ( $( 'input[name="give_mailchimp_signup"]' ).prop( 'checked' ) === true ) {
```
DevinWalker, line 200: @henryholtgeerts can this be written dynamically so there's no repeated code?

src/Views/Form/Templates/Sequoia/assets/js/form.js
```diff
@@ -196,6 +196,26 @@
...
+					$( '.give-mailchimp-fieldset' ).addClass( 'active' );
+				}
+
+				// Persist mailchimp input border when selected
+				$( '.give-mc-message-text' ).on( 'click touchend', function() {
+					$( '.give-mailchimp-fieldset' ).toggleClass( 'active' );
+				} );
+
+				// If mailchimp is opted in by default, add border on load
```
DevinWalker, line 209: Typo: your comment says mailchimp but this has to do with CC.

src/Views/Form/Templates/Sequoia/assets/js/form.js
```diff
@@ -379,6 +399,11 @@
 	 * @param {number} formID Form ID
 	 */
 	function refreshPaymentInformationSection( ev, response, formID ) {
+		// If mailchimp is opted in by default, add border on load
```
DevinWalker, line 402: What about constant contact? Is that not an option?

src/Views/Form/Templates/Sequoia/Sequoia.php
```diff
@@ -160,6 +160,28 @@ public function loadScripts() {
...
+		$mailchimpDynamicCss = "
+			.give-mailchimp-fieldset:hover,
+			.give-mailchimp-fieldset.active {
+				border: 1px solid {$primaryColor} !important;
+			}
+			.give-mailchimp-fieldset input[type='checkbox'] + span::after {
+				background-image: url(\"data:image/svg+xml,%3Csvg width='15' height='11' viewBox='0 0 15 11' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M5.73047 10.7812C6.00391 11.0547 6.46875 11.0547 6.74219 10.7812L14.7812 2.74219C15.0547 2.46875 15.0547 2.00391 14.7812 1.73047L13.7969 0.746094C13.5234 0.472656 13.0859 0.472656 12.8125 0.746094L6.25 7.30859L3.16016 4.24609C2.88672 3.97266 2.44922 3.97266 2.17578 4.24609L1.19141 5.23047C0.917969 5.50391 0.917969 5.96875 1.19141 6.24219L5.73047 10.7812Z' fill='%23{$rawColor}'/%3E%3C/svg%3E%0A\") !important;
+			}
+		";
+		wp_add_inline_style( 'give-sequoia-template-css', $mailchimpDynamicCss );
```
DevinWalker, line 172: Is this added regardless of whether the add-on is active or not?

src/Views/Form/Templates/Sequoia/Sequoia.php
```diff
@@ -160,6 +160,28 @@ public function loadScripts() {
...
+		$constantContactDynamicCss = "
+			.give-constant-contact-fieldset:hover,
+			.give-constant-contact-fieldset.active {
+				border: 1px solid {$primaryColor} !important;
+			}
+			.give-constant-contact-fieldset input[type='checkbox'] + span::after {
+				background-image: url(\"data:image/svg+xml,%3Csvg width='15' height='11' viewBox='0 0 15 11' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M5.73047 10.7812C6.00391 11.0547 6.46875 11.0547 6.74219 10.7812L14.7812 2.74219C15.0547 2.46875 15.0547 2.00391 14.7812 1.73047L13.7969 0.746094C13.5234 0.472656 13.0859 0.472656 12.8125 0.746094L6.25 7.30859L3.16016 4.24609C2.88672 3.97266 2.44922 3.97266 2.17578 4.24609L1.19141 5.23047C0.917969 5.50391 0.917969 5.96875 1.19141 6.24219L5.73047 10.7812Z' fill='%23{$rawColor}'/%3E%3C/svg%3E%0A\") !important;
+			}
+		";
+		wp_add_inline_style( 'give-sequoia-template-css', $constantContactDynamicCss );
```
DevinWalker, line 183: Same question here.

src/Views/Form/Templates/Sequoia/Sequoia.php
henryholtgeerts: For now, yes. The dynamic css is grouped by addon so that as we introduce functionality to reliably check whether an addon is enabled (@ravinderk has been working on this as part of the receipt compatibility) we will be able to load addon specific CSS only when appropriate.

src/Views/Form/Templates/Sequoia/Sequoia.php
henryholtgeerts: See related comment: https://github.com/impress-org/givewp/pull/4771#discussion_r431467642

src/Views/Form/Templates/Sequoia/assets/js/form.js
henryholtgeerts: Good catch, fixed: https://github.com/impress-org/givewp/commit/bddd8865e2ee87773bb1bb9884a0a47fa5be120c

src/Views/Form/Templates/Sequoia/assets/js/form.js
henryholtgeerts: Added support for constant contact: https://github.com/impress-org/givewp/commit/bddd8865e2ee87773bb1bb9884a0a47fa5be120c

src/Views/Form/Templates/Sequoia/assets/js/form.js
henryholtgeerts: Added a setupCheckbox function to handle listeners/setup of these kind of prominent checkboxes, ref: https://github.com/impress-org/givewp/commit/7b33419eb8aeea604596678716660c0d6b0a1ce5

henryholtgeerts:
Just noticed a bug where the checkbox is always reset to the form default setting whenever gateway is changed. This should not happen, a possible solution would be to store checkbox state in the window, then check this state in the setupCheckbox function.

src/Views/Form/Templates/Sequoia/Sequoia.php
ravinderk: @DevinWalker we have to add conditions to check if the addon is active or not to limit dynamic CSS size.
@henryholtgeerts can you explain what you are expecting?

@JasonTheAdams Can you add your thoughts on this because I think adding conditions to output only addon related dynamic style will be overdoing.

henryholtgeerts:
Bug related to persisting checked state despite selecting new gateways has been resolved! Addressing this bug also resulted a quick fix to ensure that the fee recovery notice appears in the correct place. 

Screen Shot 2020-05-28 at 10 54 39 AM

henryholtgeerts:
Also tested: what happens when admin sets a long default label. 

Screen Shot 2020-05-28 at 11 00 09 AM

src/Views/Form/Templates/Sequoia/Sequoia.php
JasonTheAdams: I agree that we should limit assets to loading only when needed, otherwise this is a slippery slope. Perhaps at some point we'll come up with a more formal system for checking for "features" versus addons, but in the meantime is there any reason we couldn't just use [is_plugin_active](https://developer.wordpress.org/reference/functions/is_plugin_active/) here?

ravinderk: @JasonTheAdams `class_exists` will be quicker than `is_plugin_active` and this function also not available on the frontend, so we have to load the extra file.

can we use `class_exists` on `defined`?

JasonTheAdams: I'm reluctant to use `class_exists` as we're not _actually_ using any kind of addon class, here. And we'd be coupling GiveWP core to the shape of the addon, ambiguously. The `is_plugin_active` function is fairly performant as the `active_plugins` option will certainly be autoloaded and cached, so it's just checking in an array.

To make it available we'd just need to do
`include_once ABSPATH . 'wp-admin/includes/plugin.php';`

ravinderk: @JasonTheAdams, In this case, I will suggest creating a wrapper function around it.

```
namespace Give\Helpers\Utils

class Utils{
     publish static function isPluginActive( $plugin ){
         if( ! function_exists( 'is_plugin_active')  ){
             include_once ABSPATH . 'wp-admin/includes/plugin.php';
         }

         return is_plugin_active( $plugin );
     }
}
```

JasonTheAdams: Yeah, I'm cool with that helper. Go for it.

henryholtgeerts: @ravinderk Are you implementing this in this PR?

ravinderk: @JasonTheAdams one more thought on this because we have to hardcode addon file path in `isPluginActive`, I want to suggest one more abstraction. Can we create a helper class for each addon?

```
namespace Give\Helpers\Addon\Recurring;

class Recurring{
    private static $plugin = 'give-recurring/give-recurring.php';

    public static function isPluginActive(){
        return Utils::isPluginActive( self::plugin )
    }
}
```

so instead of using `Utils::isPluginActive( 'give-recurring/give-recurring.php' )` we can use `RecurringAddon::isPluginActive()`.

JasonTheAdams: @ravinderk I like where your head is at, I just don't have as much context. Do those addon classes already exist? Do they do something other than check if the addon is active?

Also, slight tweak:
```php
<?php
namespace Give\Helpers\Addon\Recurring;

class Recurring{
    const PLUGIN_FILE = 'give-recurring/give-recurring.php';

    public static function isActive(){
        return Utils::isPluginActive( self::PLUGIN_FILE );
    }
}
```

That said, I don't want to abstract the crap out of this right now. If creating a single helper keeps this PR moving forward, I'd prefer that. We can optimize more, later.

henryholtgeerts: I'm gonna move forward with implementing the Helper method in this PR, and suggest we have a more robust conversation about these more abstracted classes on a GiveWP dev call/in a separate issue. I think it will ultimately be the way to go but I'm hesitant to add a bunch of new classes as part of this PR, especially if they all serve the same basic utility starting out.

ravinderk: @henryholtgeerts I implemented function in same pr: https://github.com/impress-org/givewp/compare/4df865e3a942...d970a0e0981f

henryholtgeerts: The helper method has been implemented in the PR, ref: https://github.com/impress-org/givewp/pull/4771/commits/e6c90b439cfd88c9375308c8c33256abaef121e6

src/Views/Form/Templates/Sequoia/Sequoia.php
```diff
@@ -122,37 +123,69 @@ public function loadScripts() {
...
-			.give-recurring-donors-choice:hover,
-			.give-recurring-donors-choice.active {
-				border: 1px solid {$primaryColor};
-			}
-			.give-recurring-donors-choice input[type='checkbox'] + label::after {
-				background-image: url(\"data:image/svg+xml,%3Csvg width='15' height='11' viewBox='0 0 15 11' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M5.73047 10.7812C6.00391 11.0547 6.46875 11.0547 6.74219 10.7812L14.7812 2.74219C15.0547 2.46875 15.0547 2.00391 14.7812 1.73047L13.7969 0.746094C13.5234 0.472656 13.0859 0.472656 12.8125 0.746094L6.25 7.30859L3.16016 4.24609C2.88672 3.97266 2.44922 3.97266 2.17578 4.24609L1.19141 5.23047C0.917969 5.50391 0.917969 5.96875 1.19141 6.24219L5.73047 10.7812Z' fill='%23{$rawColor}'/%3E%3C/svg%3E%0A\");
-			}
-		";
-		wp_add_inline_style( 'give-sequoia-template-css', $recurringDynamicCss );
+		if ( Utils::isPluginActive( 'give-recurring/give-recurring.php' ) ) {
```
ravinderk, line 126: @henryholtgeerts What do you think using only one property to collect all addon CSS and use only one

```php
<?php
if( $addonDynamicCss ){
    wp_add_inline_style( 'give-sequoia-template-css', $addonDynamicCss );
}
```

to register style

henryholtgeerts: Done, ref: https://github.com/impress-org/givewp/commit/b53771eecf80e58cc4c0dcfb061815702f8c7d2f

PR #4799, opened 2020-06-13
Update class-give-email-tags.php

petertwise:
Now that an anonymous checkbox is part of the core plugin, add an email tag {anonymous_status} to output if the donation has been marked as anonymous or not

## Description
Added the `anonymous_status` email tag using the same code style as the rest of the default email tags.

## Affects
Provides the anonymous_staus email tag by default instead of needing to use a snippet or plugin.

## What to test
Test the functionality of using the tag in a donor or admin email.
Test with the box checked or unchecked.
Ensure it does not interfere with other email tags.

## Checklist:
- [X] My code is tested.

### DIFF
--- a/includes/emails/class-give-email-tags.php
+++ b/includes/emails/class-give-email-tags.php
@@ -413,6 +413,12 @@ function give_setup_email_tags() {
 			'func'    => 'give_email_tag_donor_note',
 			'context' => 'donation',
 		),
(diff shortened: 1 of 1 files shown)


petertwise:
My pleasure - thanks for including it! This is a feature we will use on all of the sites we implement GiveWP on.

PR #4803, opened 2020-06-16
hotfix: Fee Recovery option consistently appears below donation amount buttons

henryholtgeerts:
## Description
Previously, when an admin setup different fee recovery options for specific gateways, the Fee Recovery checkbox loaded within each individual gateway. This hotfix ensures that every time a gateway is loaded, if a Fee Recovery checkbox is present, it is moved below the donation amount buttons in the "choose amount" step. Additionally, checked/unchecked status of the Fee Recovery option is persisted even as a user changes their selected gateway.

## Affects
This PR affects how the Multi-Step form template is rendered, specifically moving the Fee Recovery opt-in checkbox so that it appears consistently beneath the Choose Amount step.

## What to test

_The following conditions have been tested in my local install. Use this list for reference of known possible configurations this PR impacts._

With the Fee Recovery addon enabled, try setting up custom fee recovery options on a per gateway basis. 
- [x] Are you shown the option to opt-into Fee Recovery beneath the donation amount buttons? 
- [x] When you change gateways, and return to that page does opt-in message reflect the appropriate fee for the chosen gateway? 
- [x] If you select Fee Recovery, and then change gateways, is your selection persisted? 
- [x] Likewise, if you opt-out of Fee Recovery, and change gateways, is your decision persisted across gateways?
- [x] If you disable Fee Recovery for some gateways, and enable it for others, is the option and fee recovery message hidden as you would expect?

## Screenshots:
Screen Shot 2020-06-16 at 3 58 34 PM

## Checklist:
- [x] My code is tested.

### DIFF
--- a/src/Views/Form/Templates/Sequoia/assets/js/form.js
+++ b/src/Views/Form/Templates/Sequoia/assets/js/form.js
@@ -420,7 +420,7 @@
 			$( '#give-payment-mode-select' ).after( '<fieldset id="donate-fieldset"></fieldset>' );
 		}
 
-		// Elements to move into donate field`set (located at bottom of form)
+		// Elements to move into donate fieldset (located at bottom of form)
 		// The elements will appear in order of array
 		const donateFieldsetElements = [
 			'.give-constant-contact-fieldset',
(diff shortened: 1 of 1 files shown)


DevinWalker:
Looks good. Passed the tests and no JS errors.

PR #4813, opened 2020-06-17
Fixes the default headline value for the new form template in 2.7.0

### LINKED ISSUE #4812: 2.7.0 should not save form template headline as "Auto Draft"
# Bug Report

When I create a new donation form quickly and don't set the headline it should default to "Support Our Cause" rather than "Auto Draft".

## Steps to Reproduce

1. Create a new donation form
2. Publish or Create a Draft
3. See the Headline is "Auto Draft".

DevinWalker:
Resolves #4812

## Description

I removed the default field value for the "Headline" field because we have a fallback in place should one not be provided.

## What to Test

- [x] Publishing a donation form without a title or headline uses the text "Support Our Cause"
- [x] Customizing the donation form title replaces "Support Our Cause" text
- [x] Customizing the Form Template's heading replaces the text from the donation form's post title.

### DIFF
--- a/src/Views/Form/Templates/Sequoia/optionConfig.php
+++ b/src/Views/Form/Templates/Sequoia/optionConfig.php
@@ -6,7 +6,7 @@ use Give\Helpers\Form\Template\Utils\Frontend as FrontendFormTemplateUtils;
 $formInfo = get_post( FrontendFormTemplateUtils::getFormId() );
 
 // Setup dynamic defaults
-$introHeadline    = $formInfo->post_title ? $formInfo->post_title : __( 'Campaign Heading', 'give' );
+$introHeadline    = ($formInfo->post_title && 'Auto Draft' !== $formInfo->post_title ) ? $formInfo->post_title : __( 'Support Our Cause', 'give' );
 $introDescription = $formInfo->post_excerpt ? $formInfo->post_excerpt : __( 'Help make a difference today! All donations go directly to making a difference for our cause.', 'give' );
 
 return [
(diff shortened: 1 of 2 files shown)


src/Views/Form/Templates/Sequoia/optionConfig.php
```diff
@@ -6,7 +6,7 @@
 $formInfo = get_post( FrontendFormTemplateUtils::getFormId() );
 
 // Setup dynamic defaults
-$introHeadline    = $formInfo->post_title ? $formInfo->post_title : __( 'Campaign Heading', 'give' );
+$introHeadline    = ($formInfo->post_title && 'Auto Draft' !== $formInfo->post_title ) ? $formInfo->post_title : __( 'Support Our Cause', 'give' );
```
ravinderk, line 9: @DevinWalker This logic will break for non-english sites.

image

Please check if admin creating a new form or not instead of comparing the title.

src/Views/Form/Templates/Sequoia/sections/introduction.php
```diff
@@ -12,17 +12,23 @@
 
 <div class="give-section introduction">
 	<h2 class="headline">
-		<?php echo $headline; ?>
+		<?php
+		if ( empty( $headline ) || $headline === __( 'Auto Draft' ) ) {
+			_e( 'Support Our Cause', 'give' );
+		} else {
+			esc_html_e( $headline, 'give' );
```
ravinderk, line 19: @henryholtgeerts Dynamic string is not allowed in translation function. I will take care of this.

ravinderk: fixed: https://github.com/impress-org/givewp/compare/cf5510050926...0ca6b84e582c

PR #4816, opened 2020-06-17
fix: recurring optin border is now visible when set to checked by default

henryholtgeerts:
## Description
This PR improves logic for setting up recurring and fee recovery opt-in checkboxes, so that if the box is checked by default, the highlighted border is present by default. The PR uses the existing setupCheckbox functionality to make sure these checkboxes function consistently with other prominent checkboxes in forms (newsletter checkboxes, etc). 

Previously, if an admin setup a checkbox to be checked by default, the checkbox would load in a checked state as expected, but lacking the highlighted border that should accompany the checkmark. Because the "active" class was not being applied on load, it also meant that when a user unchecked the box it would cause the highlighted border appear. 

## Affects
This PR affects frontend logic related to rendering the form, specifically the fee recovery and recurring donation opt-in checkboxes. It ensures that these checkboxes function consistent with other prominent checkboxes in the form and use the correct styling on form load. 

## What to test
- [x] Try enabling the recurring opt-in setting by default, does the check box appear checked and with a highlighted border when the form first loads?
- [x] Try disabling the recurring opt-in setting by default, does the check box appear unchecked and without a highlighted border when the form first loads?
- [x] Try enabling the fee recovery opt-in setting by default, does the check box appear checked and with a highlighted border when the form first loads?
- [x] Try disabling the fee recovery opt-in setting by default, does the check box appear unchecked and without a highlighted border when the form first loads?

## Screenshots:
Screen Shot 2020-06-17 at 12 56 30 PM

## Checklist:
- [x] My code is tested.

### DIFF
--- a/src/Views/Form/Templates/Sequoia/assets/js/form.js
+++ b/src/Views/Form/Templates/Sequoia/assets/js/form.js
@@ -224,24 +224,28 @@
 					moveErrorNotice( $( this ) );
 				} );
 
+				// Setup recurring donations opt-in event listeners
 				setupCheckbox( {
 					container: '.give-recurring-donors-choice',
 					label: 'label[for*="give-admin_choice"]',
 					input: 'input[name="give-recurring-period"]',
 				} );
 
+				// Setup fee recovery opt-in event listeners
 				setupCheckbox( {
 					container: '.give-fee-recovery-donors-choice',
 					label: '.give-fee-message-label-text',
 					input: 'input[name="give_fee_mode_checkbox"]',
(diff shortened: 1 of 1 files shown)


DevinWalker:
@henryholtgeerts I found an issue: when you click on the opt-in the border appears and progress to the next step then go back it's removed.

henryholtgeerts:
@DevinWalker I tested this out locally and am having trouble reproducing this bug. When I follow the steps you described/illustrated in the GIF the border persists as expected.

PR #4839, opened 2020-06-18
fix: terms and conditions checkbox is now styled in Multi-Step form template

### LINKED ISSUE #4838: fix: terms and conditions checkbox should be above donate now button in Multi-Step form
# Bug Report

## User Story
As a user, I want the terms and conditions checkbox to be above the donate button so that I do not have to agree to terms and conditions every time I change the payment gateway.

# Current Behavior
When the terms and conditions setting is enabled, the Multi-Step form places the checkbox inside of each individual payment gateway's accordion area. This is not ideal because it means that every time a user changes gateway, they have to once again agree to the terms and conditions. Furthermore, the terms and conditions checkbox is currently unstyled, and the checkbox itself is hidden.

## Expected Behavior
The terms and conditions check box and description should be moved to beneath all payment gateways, and before the donate now button. It should appear after all other notices. It should also be styled consistent with other checkbox inputs in the Multi-Step form.

## Bug Type

## Steps to Reproduce
1. Create a new Multi-Step form
2. Enable terms and conditions

## Acceptance Criteria
- [x] Terms and conditions checkbox is moved beneath payment gateways
- [x] Terms and conditions checkbox is the last input before the donate now button
- [x] Terms and conditions checkbox is styled consistent with other checkboxes  that appear in the Multi-Step form
- [x] Terms and conditions frontend styles are appropriate for a very long terms and conditions message, and also able to handle long "Agree to terms?" label

henryholtgeerts:
## Description
Resolves #4838 
This PR introduces styles for the Terms and Conditions agremeent checkbox, and moves the checkbox below all payment gateways. Previously, when the term and conditions setting was enabled on a form using the Multi-Step form template, the checkbox was unstyled and loaded inside each individual payment gateway. 

## Affects
This PR affects frontend styles and form rendering logic of the Multi-Step form tempalte, specifically regarding the presentation of the "Terms and Conditions" checkbox.

## What to test
Enable Terms and Conditions for a form using the Multi-Step form template.
- [x] Does the terms and conditions checkbox appear beneath payment gateways?
- [x] If you attempt to submit without checking the terms and conditions does a validation notice appear?
- [x] Does the "Show Terms" button work as expected?
- [x] Does the terms and conditions checkbox use the form's primary color?
- [x] If a user sets a long string as the terms and conditions label, does it adjust accordingly?

## Screenshots:
PRDemo

## Checklist:
- [x] My code is tested.

### DIFF
--- a/src/Views/Form/Templates/Sequoia/Sequoia.php
+++ b/src/Views/Form/Templates/Sequoia/Sequoia.php
@@ -159,6 +159,17 @@ class Sequoia extends Template implements Hookable, Scriptable {
 			.payment [id*='give-create-account-wrap-'] label::after {
 				background-image: url(\"data:image/svg+xml,%3Csvg width='15' height='11' viewBox='0 0 15 11' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M5.73047 10.7812C6.00391 11.0547 6.46875 11.0547 6.74219 10.7812L14.7812 2.74219C15.0547 2.46875 15.0547 2.00391 14.7812 1.73047L13.7969 0.746094C13.5234 0.472656 13.0859 0.472656 12.8125 0.746094L6.25 7.30859L3.16016 4.24609C2.88672 3.97266 2.44922 3.97266 2.17578 4.24609L1.19141 5.23047C0.917969 5.50391 0.917969 5.96875 1.19141 6.24219L5.73047 10.7812Z' fill='%23{$rawColor}'/%3E%3C/svg%3E%0A\");
 			}
+			#give_terms_agreement:hover,
+			#give_terms_agreement:focus-within,
+			#give_terms_agreement.active {
+				border: 1px solid {$primaryColor} !important;
+			}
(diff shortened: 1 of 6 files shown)


henryholtgeerts:
@ravinderk I've taken care of the pointer issue, and made it consistent across all checkboxes of this style in the form. Ref: https://github.com/impress-org/givewp/pull/4839/commits/f89ac2360e35f97c96590678aaa51dadfcfad3b4

Investigating the shadow line that appears it seems to a rendering artifact specifically tied to Chrome. It appears when the fee breakdown message is set to 'display: none', while it is not actually in view of the browser (bc the message is being hidden while the user is looking step 2). You can confirm it is a rendering issue because the line disappears if you zoom in on the form in the browser (tested using command +). 

When using other browsers (tested in Safari and Firefox) the line does not appear.

DevinWalker:
Thanks @henryholtgeerts for the explanation about Chrome, which seems pretty edge case. Let's hold off investigating that further. I think we're good to merge this now as it satisfies all the requirements.

PR #4875, opened 2020-06-29
fix: trend percentage rounding is now based on significant digits

### LINKED ISSUE #4766: fix: trend percentages in Reports mini-charts should be more informative
# Bug Report

## User Story
As an admin, I want percentages rendered with mini-charts on the Reports page to be more informative/intuitive so that I can more quickly get an accurate idea of donation trends on my site.

## Current Behavior
Currently, these trend percentages compare the queried period with the previous period of equal length. This is intuitive and works well enough for smaller periods, but is significantly less helpful when exploring larger periods (especially if it is close to an all-time query).

## Expected Behavior
These percentages should be helpful to users, and when entering a range where that may not be the case, should possibly be hidden (rather than confuse users more).

## Bug Type

## Acceptance Criteria
- [ ] Reports mini-chart trend percentages are predictable
- [ ] Trend percentages are intuitive/informative
- [ ] Users gain real value from trend percentages when they are present

henryholtgeerts:
## Description
Resolves #4766
This PR introduces frontend logic to ensure that trend percentages are rounded appropriately, based on the size of the number and a maximum number of significant digits. Previously, when the server provided a large number as the potential trend percentage (sometimes in the thousands) it was still displayed with point values, despite adding little tangible value to users and further cluttering the mini charts at the top of the Reports page. Now, when a large value is returned, it will include no point values, while small values do maintain appropriate point values. This improvement draws on work from the GiveIO reports project to tackle the same problem, and utilizes the numberToDigits function. 

## Affects
This PR impacts frontend rendering logic, specifically how trend percentages are rounded and displayed on mini charts at the top of the Reports page. Now, large percentages are rounded to their whole number, while small percentages maintain up to two decimal places. 

## What to test
With sample donation data imported, go to the Reports page. Test out a few different reporting periods.
- [ ] Do the percent trends (found in mini charts beneath the income over time chart) round as expected?
- [ ] Are large percentages rounded to whole numbers?
- [ ] Are smaller, more specific percentages extended to two decimal places?

## Screenshots:
Screen Shot 2020-06-29 at 3 35 44 PM

## Checklist:
- [x] My code is tested.

### DIFF
--- a/assets/src/js/admin/reports/components/mini-chart/index.js
+++ b/assets/src/js/admin/reports/components/mini-chart/index.js
@@ -47,7 +47,7 @@ const MiniChart = ( { title, data } ) => {
 
 	useEffect( () => {
 		const newHighlightValue = getHighlightValue( data );
-		const newTrend = getTrend( data ) === 0 ? 0 : numberToDigits( getTrend( data ), 2, 2 );
+		const newTrend = getTrend( data ) === 0 ? 0 : numberToDigits( getTrend( data ), 3, 2 );
 		const newTooltipText = getTooltipText( data );
 		let newIndicator;
 
--- a/assets/src/js/admin/reports/components/mini-chart/index.js
+++ b/assets/src/js/admin/reports/components/mini-chart/index.js
@@ -47,7 +47,7 @@ const MiniChart = ( { title, data } ) => {
 
 	useEffect( () => {
 		const newHighlightValue = getHighlightValue( data );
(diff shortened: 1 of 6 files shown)


henryholtgeerts:
For reference, here's a video detailing my own process of testing this PRs functionality. 
https://www.loom.com/share/acb6589328ee4e64b55ca8a0faab6cc2

cc: @DevinWalker

PR #4939, opened 2020-07-20
Add Hooks helper for creating auto-instantiated hooks

### LINKED ISSUE #4938: Hooking into WordPress should be efficient and object-oriented
## Details

Pasted in from the Slack discussion:

@givewpdevteam I'm going to process a thought here in case someone else has a thought. Also, I'm annoyed with WordPress and want to vent this.

A frustration I have with WordPress hooks is that they're entirely functional and do not support OOP. Here is a very common pattern seen in WP themes/plugins:
```php
class DoSomeThing {
    public function hookStuffUp() {
        add_action('init', [$this, 'doTheActualIntendedThing']);
    }

    public function doTheActualIntendedThing() {
        // Do something important
    }
}

$class = new DoSomething();
$class->hookStuffUp();
```

A popular example of this is an Assets class that hooks into the parts of WP where you're intended to enqueue stuff.

Here's what's annoying about this pattern: The only way to hook up the class methods is to instantiate the class. Think about it. The only time you should actually need the Assets class to load is when the corresponding hooks fire. But, in this case, we're loading the instance into memory in case the hook fires. The ends up loading a lot more class instances into memory than are actually used.

What would be very cool would be the ability to do something like:
```php
class Assets {
    public function loadFrontEndAssets() {
        // load them assets
    }
}

add_action('wp_enqueue_scripts', 'Assets@loadFrontEndAssets');
```

In this case, the Assets the class would be instantiated when the action fires, and then method name after the "@" would be called with the hook parameters included. As a bonus, WordPress would provide the ability for a plugin or something to handle the object instantiation so it can use something like a Service Container.

I'm tempted to create something like the following in the Give class:
```php
public function addAction($tag, $class, $method = '__invoke', $priority = 10, $acceptedArgs = 1) {
    add_action($tag, function() use($class, $method) {
        $instance = $this->make($class);
        call_user_func_array([$instance, $method], func_get_args());
    }, $priority, $acceptedArgs);
}

// usage
give()->addAction('wp_enqueue_scripts, Assets::class, 'loadFrontEndAssets');
```

Basically, this would register an anonymous function (Closure) that gets called, which retrieves the class from the Service Container, and finally calls the method. This would be awesome.

Here's the downside... WordPress does not provide a good way of handling Closures within hooks. They add and fire just fine, but it's not at all easy to remove them later. I haven't cared much about this with themes, but it's more important for plugins.

That said, I personally believe that not every hooked function within a plugin should be removable, as the plugin can end up doing some super unpredictable things (then support gets contacted because some dev went off the rails). This is the dark side of a publicly mutational API like hooks.

I'm 70% in favor of adding this method and using it for critical bootstrapped methods within Give. But I'm open to what others think. :smile:

## Acceptance Criteria

-   [x] Hooks are not set up in the same class they're applied to (Single Responsibility)
-   [x] Hooks are efficiently handled within GiveWP (subjective)

JasonTheAdams:
Resolves #4938 

## Description

The targeted purpose for this PR is to provide a more memory efficient means of using WordPress hooks that also helps to organize our code.

Right now, every class that has hooks that need to be applied needs to be instantiated in order to have its hooks set up. This would be like leaving your house with everything you own _just in case_ you need anything while you're out. This provides a means of specifying the class and method to be called _if_ the action/filter ends up being used. So if the hook isn't used, the class isn't instantiated — which uses less memory.

This also gives us the ability to better centralize our hooks. So instead of hooks being scattered through hundreds of classes, they can instead be primarily set up within the Service Providers. So this...
```php
class FrontEndAssets {
    public function init() {
        add_action('wp_enqueue_scripts', [$this, 'loadScripts']);
    }

    public function loadScript() {
        // load scripts
    }
}

class BackEndAssets {
    public function init() {
        add_action('admin_enqueue_scripts', [$this, 'loadScripts']);
    }

    public function loadScript() {
        // load scripts
    }
}

$frontEnd = new FrontEndAssets();
$frontEnd->init();

$backEnd = new BackEndAssets();
$backEnd->init();
```

becomes this

```php
class AssetsServiceProvider {
    public function boot() {
        Hooks::add_action('wp_enqueue_scripts', FrontEndAssets::class, 'loadScript');
        Hooks::add_action('admin_enqueue_scripts', BackEndAssets::class, 'loadScript');
    }
}

class FrontEndAssets {
    public function loadScript() {
        // load scripts
    }
}

class BackEndAssets {
    public function loadScript() {
        // load scripts
    }
}
```

As you can see it's less code, easier to follow, and more memory efficient. Huzzah! 🎊 

## Affects

This does not affect any existing code. It's just new utilities.

## Pre-review Checklist

-   [ ] Tests included

### DIFF
--- a/src/Helpers/Hooks.php
+++ b/src/Helpers/Hooks.php
@@ -1,0 +1,79 @@
+<?php
+
+namespace Give\Helpers;
+
+use BadMethodCallException;
+
+class Hooks {
+	/**
+	 * A function which extends the WordPress add_action method to handle the instantiation of a class
+	 * once the action is fired. This prevents the need to instantiate a class before adding it to hook.
+	 *
+	 * @since 2.8.0
+	 *
+	 * @param string $tag
+	 * @param string $class
+	 * @param string $method
+	 * @param int    $priority
+	 * @param int    $acceptedArgs
+	 *
+	 * @return void
+	 */
+	public static function addAction( $tag, $class, $method = '__invoke', $priority = 10, $acceptedArgs = 1 ) {
+		add_action(
+			$tag,
+			static function () use ( $tag, $class, $method ) {
+				// Provide a way of disabling the hook
+				if ( apply_filters( "give_disable_hook-{$tag}", false ) || apply_filters( "give_disable_hook-{$tag}:{$class}@{$method}", false ) ) {
+					return;
+				}
+
+				$instance = give( $class );
+
+				if ( ! method_exists( $instance, $method ) ) {
+					throw new BadMethodCallException( "The method $method does not exist on $class" );
+				}
+
+				call_user_func_array( [ $instance, $method ], func_get_args() );
+			},
+			$priority,
+			$acceptedArgs
+		);
+	}
+
+	/**
+	 * A function which extends the WordPress add_filter method to handle the instantiation of a class
+	 * once the filter is fired. This prevents the need to instantiate a class before adding it to hook.
+	 *
+	 * @since 2.8.0
+	 *
+	 * @param string $tag
+	 * @param string $class
+	 * @param string $method
+	 * @param int    $priority
+	 * @param int    $acceptedArgs
+	 *
+	 * @return void
+	 */
+	public static function addFilter( $tag, $class, $method = '__invoke', $priority = 10, $acceptedArgs = 1 ) {
+		add_filter(
+			$tag,
+			static function () use ( $tag, $class, $method ) {
+				// Provide a way of disabling the hook
+				if ( apply_filters( "give_disable_hook-{$tag}", false ) || apply_filters( "give_disable_hook-{$tag}:{$class}@{$method}", false ) ) {
+					return func_get_arg( 0 );
+				}
+
+				$instance = give()->make( $class );
+
+				if ( ! method_exists( $instance, $method ) ) {
+					throw new BadMethodCallException( "The method $method does not exist on $class" );
+				}
+
+				return call_user_func_array( [ $instance, $method ], func_get_args() );
+			},
+			$priority,
+			$acceptedArgs
+		);
+	}
+}
--- a/src/Helpers/Hooks.php
+++ b/src/Helpers/Hooks.php
@@ -2,7 +2,7 @@
 
 namespace Give\Helpers;
 
-use BadMethodCallException;
+use InvalidArgumentException;
 
 class Hooks {
 	/**
@@ -20,6 +20,10 @@ class Hooks {
 	 * @return void
 	 */
 	public static function addAction( $tag, $class, $method = '__invoke', $priority = 10, $acceptedArgs = 1 ) {
+		if ( ! method_exists( $class, $method ) ) {
+			throw new InvalidArgumentException( "The method $method does not exist on $class" );
+		}
+
 		add_action(
 			$tag,
 			static function () use ( $tag, $class, $method ) {
@@ -29,10 +33,6 @@ class Hooks {
 				}
 
 				$instance = give( $class );
-
-				if ( ! method_exists( $instance, $method ) ) {
-					throw new BadMethodCallException( "The method $method does not exist on $class" );
-				}
 
 				call_user_func_array( [ $instance, $method ], func_get_args() );
 			},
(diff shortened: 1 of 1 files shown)


JasonTheAdams:
Please review this when you get a chance, @kjohnson. I ended up going with a static facade to keep with our other utilities as well as make use of auto-loading.

src/Helpers/Hooks.php
```diff
@@ -0,0 +1,79 @@
...
+			$tag,
+			static function () use ( $tag, $class, $method ) {
+				// Provide a way of disabling the hook
+				if ( apply_filters( "give_disable_hook-{$tag}", false ) || apply_filters( "give_disable_hook-{$tag}:{$class}@{$method}", false ) ) {
+					return;
+				}
+
+				$instance = give( $class );
+
+				if ( ! method_exists( $instance, $method ) ) {
```
kjohnson, line 33: `$class` can be passed directly to [method_exists()](https://www.php.net/manual/en/function.method-exists.php) which further defers the object instantiation.

```
mixed $object An object instance or a class name 
```

src/Helpers/Hooks.php
```diff
@@ -0,0 +1,79 @@
...
+	public static function addFilter( $tag, $class, $method = '__invoke', $priority = 10, $acceptedArgs = 1 ) {
+		add_filter(
+			$tag,
+			static function () use ( $tag, $class, $method ) {
+				// Provide a way of disabling the hook
+				if ( apply_filters( "give_disable_hook-{$tag}", false ) || apply_filters( "give_disable_hook-{$tag}:{$class}@{$method}", false ) ) {
+					return func_get_arg( 0 );
+				}
+
+				$instance = give()->make( $class );
```
kjohnson, line 67: In `addAction()` the syntax is `give( $class )`, which is inconsistent with calling `give()->make( $class )` in `addFilter()`.

src/Helpers/Hooks.php
```diff
@@ -0,0 +1,79 @@
...
+	 * @param int    $acceptedArgs
+	 *
+	 * @return void
+	 */
+	public static function addAction( $tag, $class, $method = '__invoke', $priority = 10, $acceptedArgs = 1 ) {
+		add_action(
+			$tag,
+			static function () use ( $tag, $class, $method ) {
+				// Provide a way of disabling the hook
+				if ( apply_filters( "give_disable_hook-{$tag}", false ) || apply_filters( "give_disable_hook-{$tag}:{$class}@{$method}", false ) ) {
```
kjohnson, line 27: Would it be more object oriented to expose a `removeAction( $tag, $class, $method )` method, as opposed to an inline filter?

src/Helpers/Hooks.php
JasonTheAdams: Yes! Great observation!

src/Helpers/Hooks.php
JasonTheAdams: @kjohnson Yeah, I thought about this. I have a couple reservations.
1. Folks are used to the hook functions. I felt like documenting a special hook was perhaps simpler than telling them how to pull in our Hooks class and call a method. But I could be convinced others.
2. The bigger challenge is that for this to work we'd have to register the hooks within the class. The reason being that we're using Closures within the WP hooks, so even we don't have a good way of removing them. As such, we'd just have to prevent action within the Closure ourselves. My first thought to do this would be to add a `private $registeredHooks` class property, and the following methods:

```php
	public function removeHook( $tag, $class, $method ) {
		if ( ! $this->isHookRegistered( $tag, $class, $method ) ) {
			return;
		}

		$index = array_search( "$class@method", $this->registeredHooks[ $tag ], true );

		unset( $this->registeredHooks[ $tag ][ $index ] );
	}

	private function registerHook( $tag, $class, $method ) {
		if ( ! isset( $this->registeredHooks[ $tag ] ) ) {
			$this->registeredHooks[ $tag ] = [];
		}

		$this->registeredHooks[ $tag ] = "$class@$method";
	}

	private function isHookRegistered( $tag, $class, $method ) {
		return isset( $this->registeredHooks[ $tag ] ) && in_array( "$class@method", $this->registeredHooks[ $tag ], true );
	}

```

We'd then use these to register the hook, check if the hook is still registered (if not, do not instantiate and fire), and remove hooks.

Thoughts?

kjohnson: Your reservations make sense. Also, going back to our previous conversation, the job to be done is memory management and not solving OOP issues in WordPress.

JasonTheAdams: Yeah, that's my thought. This route ends up laying our own hooks system on top of WP's. It's still less memory than instantiating, but a bit of a step backwards. I really think it's going to be quite rare for folks to remove our bootstrapping hooks.

So what do you think? Keep the removal system as is?

JasonTheAdams:
@kjohnson Ready for review, again.

PR #5012, opened 2020-08-05
Form Preview step now includes working donation form

### LINKED ISSUE #5008: Form Preview step should include a working donation form
## User Story

As a user, I want the form preview step of the onboarding wizard to include a working donation form so that I can see what a GiveWP donation form looks like.

## Details
Currently, the donation form step includes a placeholder where a working donation form belongs. This issue aims to replace the placeholder with any working multi-step donation form.

## Additional Context
Separate, related issues address creating a donation form on starting the setup wizard, and modifying that donation form based on the feature needs expressed in step 3 of the wizard, but this issue simply intends to display any working multi-step donation form on step 4 of the wizard.

## Acceptance Criteria
-   [x] Step 4 of onboarding wizard includes working multi-step donation form
-   [x] Donation form preview simply displays any available multi-step form, and is not concerned with creating one
-   [x] Multi-Step form focus is handled in an expected fashion, so that the onboarding wizard is still entirely keyboard accessible

henryholtgeerts:
Resolves #5008 

## Description

This PR introduces frontend and backend logic required to display a working donation form preview in the Form Preview step of the onboarding wizard. Previously, the step featured a placeholder stand in, rather than an actual working multi-step donation form. This PR replaces the placeholder with an iframe that loads a Multi-Step form. 

_**Note:** This PR does not setup the preview form itself. Instead, it is hardcoded to use a form with the id `18`. To modify which form is displayed (for your own testing) you can change what is returned by the `get_preview_form_id` method in the `Onboarding\Wizard\FormPreview` class._

## Affects

This PR affects frontend logic used to render the donation form preview (DonationForm component), some frontend layout styles, and introduces backend logic to render a form preview at the URL `wp-admin/?page=give-form-preview`.

The PR also includes some minor keyboard accessibility improvements, which came into play as work was done to ensure the the Form Preview step is keyboard accessible (users focus first on the Continue button, then can tab into the form preview, then into onboarding wizard step navigation). 

## Visuals
FormPreview

## Pre-review Checklist
-   [ ] Tests included
-   [x] Keyboard accessible
-   [x] Screen reader accessible
-   [x] Changelog updated

## Testing Instructions
Setup a Multi-Step form, with default settings (and using only the Test gateway). Make sure a featured image has been set for the form. Get the form's ID, then modify the `get_preview_form_id` method in the `Onboarding\Wizard\FormPreview` class to return the desired form ID. 

Navigate to the Onboarding Wizard at `wp-admin/?page=give-onboarding-wizard`, and advance to the Donation Form step. Check that the Wizard is still fully keyboard accessible, and that, on step 4, you can tab into the Form Preview after focus is initially set on the continue button.


assets/src/js/admin/onboarding-wizard/components/donation-form/index.js
```diff
@@ -1,20 +1,36 @@
...
-		alignItems: 'center',
-		justifyContent: 'center',
-		fontSize: '24px',
-		lineHeight: '32px',
-		textAlign: 'center',
+	const formPreviewUrl = getWindowData( 'formPreviewUrl' );
+	const [ iframeLoaded, setIframeLoaded ] = useState( false );
+
+	const iframeStyle = {
+		opacity: iframeLoaded === false ? '0' : '1',
```
kjohnson, line 19: Is there a particular reason that you are toggling between the elements by swapping the opacity, as opposed to conditionally rendering the elements?

assets/src/js/admin/onboarding-wizard/utils/index.js
```diff
@@ -9,6 +9,32 @@ export const getWindowData = ( value ) => {
...
+	const stepInputs = document.querySelectorAll( '.give-obw-step button, .give-obw-step input, .give-obw-step select' );
+	const navButtons = document.querySelectorAll( '.give-obw-step-navigation button' );
+	const formPreviewIframe = document.querySelector( '.give-obw-step iframe' );
+
+	stepInputs.forEach( ( input ) => {
+		input.setAttribute( 'tabindex', 1 );
+	} );
+
+	if ( formPreviewIframe !== null ) {
+		formPreviewIframe.setAttribute( 'tabindex', 2 );
```
kjohnson, line 28: We might want to consider setting focus to the container first (even if the preview isn't loaded) and allow the user to tab to the "Continue" button and back to the container/preview. As it stands, the tab order is visually backwards: starting at the "Continue" button and then tabbing to the Preview.

Also, I think there are accessibility issues with setting a tab index greater than 1, so the _proper_ way is to semantically order the elements in the DOM and let the tab order sort itself.

src/ServiceProviders/Onboarding.php
```diff
@@ -7,6 +7,7 @@
 use Give\Onboarding\SettingsRepositoryFactory;
 use Give\Onboarding\Setup\Page as SetupPage;
 use Give\Onboarding\Wizard\Page as WizardPage;
+use Give\Onboarding\Wizard\FormPreview as FormPreview;
```
kjohnson, line 10: The use of `as` isn't necessary here, since it is the same as the class name.

assets/src/js/admin/onboarding-wizard/components/donation-form/index.js
henryholtgeerts: In this case it's just a visual thing. Wanted to avoid jarring transitions between these states, and the fade between states is already present in the Multi-Step form template, so it felt right to mirror that here.

src/ServiceProviders/Onboarding.php
henryholtgeerts: Good catch! Missed this in some refactoring. Done, ref: https://github.com/impress-org/givewp/commit/af37134f5b1db7df4897ca4d84afcadc8d77e3a2

assets/src/js/admin/onboarding-wizard/utils/index.js
henryholtgeerts: Good thinking, I played with the focus for this step both ways and couldn't decide which felt more useful - but I trust your instinct here. 

I've gone ahead and reworked the Wizard component so that step navigation appears below the steps in the DOM tree (but using flex-direction: column-reverse they are visually displayed along the top as originally intended). This fix also made the setupTabIndex utility unnecessary, so I've replaced it with a more simple utility (setStepFocus) to set the correct focus when a user advances to a new a step. These changes mean no elements in the Onboarding Wizard app have a specific tabindex attribute set, and that now, on step 4, the iframe is focused first, then users can tab out of it to the continue button if they wish.

Ref: https://github.com/impress-org/givewp/commit/3fc9450da98a90dbd720c3d3d672604e907ae0c9

Issue #5015, opened 2020-08-05
Setup Page should provide better instructions for configuring Stripe webhooks.


kjohnson:
## User Story

As a site owner, I want understand how to configure my Strip webhooks so that I can accept donations.

## Details

Instructions for configuring Stripe webhooks do not include steps for selecting what events to send, which is required to configure the webhooks.

## Additional Context

> I also noticed that when you actually go to add the endpoint in Stripe, you have to manually select which events to listen for (or optionally, add all of them). This was sorta confusing to me as a user, since I didn't get any heads up about it in the paragraph on the Setup Page. Maybe clarifying this in the copy would be helpful, especially since our check to see if the Webhook is configured is based on listening for a specific event. @henryholtgeerts 

> I've also noticed that we tell the user to enable all events, and Stripe explicitly suggests not doing this. @JasonTheAdams 

## Acceptance Criteria

-   [ ] The user understands how to configure Stripe webhooks.

kevinwhoffman:
If we do decide to provide instruction for specific events, let's make sure we've reviewed our [current webhooks documentation](https://givewp.com/documentation/add-ons/stripe-gateway/#webhooks) and keep the docs team informed so that the web doc matches the contextual doc .

> Note: It is recommended to send all events over the webhook so that if Give is updated in the future to use any events that it currently does not, those will send as well. Unless your site is processing thousands of donations per day or more, you should have no issue with sending all events, and you can ignore the warnings from Stripe.

kevinwhoffman:
@kjohnson Can you share a screenshot of what the new Setup Page currently looks like after connecting to Stripe but before the webhook is configured? It's important to see what we are starting with so we can improve it.

kjohnson:
@kevinwhoffman 

Setup Page
![image](https://user-images.githubusercontent.com/10858303/89934741-0ec1e700-dbdf-11ea-8bce-9d0805106f9f.png)

Stripe Settings
![image](https://user-images.githubusercontent.com/10858303/89934851-36b14a80-dbdf-11ea-9424-43e50b35f975.png)

kjohnson:
I'm thinking of hiding the PayPal row if Stripe is configured (and vica versa) to focus on the configured gateway. This would simplify the step and afford Stripe more space to communicate the webhooks requirements.

![image](https://user-images.githubusercontent.com/10858303/89935567-5b59f200-dbe0-11ea-87d3-e537dd56a08d.png)

kevinwhoffman:
[...]

PR #5052, opened 2020-08-11
Setup shows only the gateway being configured.

kjohnson:
#5015

## Description

If Stripe is configured, hides PayPal (and vica versa).

This is based on the assumption that the new user is only setting up one payment gateway, which is a decision made to focus and simplify the Setup Page.

## Visuals

image

## Pre-review Checklist

-   [ ] Tests included
-   [ ] Keyboard accessible
-   [ ] Changelog updated

### DIFF
--- a/src/Onboarding/Setup/PageView.php
+++ b/src/Onboarding/Setup/PageView.php
@@ -49,6 +49,24 @@ class PageView {
 	}
 
 	/**
+	 * @return bool
+	 *
+	 * @since 2.8.0
+	 */
+	public function isStripeSetup() {
+		return \Give\Helpers\Gateways\Stripe::isAccountConfigured();
+	}
+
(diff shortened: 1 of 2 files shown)


src/Onboarding/Setup/PageView.php
```diff
@@ -48,6 +48,24 @@ public function render_template( $template, $data = [] ) {
...
+		return \Give\Helpers\Gateways\Stripe::isAccountConfigured();
+	}
+
+	/**
+	 * @return bool
+	 *
+	 * @since 2.8.0
+	 */
+	public function isPayPalSetup() {
+		return false;
```
kjohnson, line 66: This will be implemented when PayPal is available, otherwise we can add a check for PayPal Standard as a fallback.

PR #5098, opened 2020-08-13
Setup Stripe Webhooks URL is now based on test mode setting.

### LINKED ISSUE #5095: Configure Webhooks button should send user to Stripe dashboard in whichever mode their site is currently in
## User Story

As an admin, I want to be able to quickly configure Stripe webhooks so that I have an easy onboarding experience.

## Details

Currently, the Configure Webhooks button sends users to the Stripe dashboard in Live mode. Naturally, users then setup their GiveWP endpoint in live mode. However, their actual GiveWP site is most likely in Test mode, which means it won't hit the Live mode endpoint in Stripe.

To correct this, send users to the Stripe dashboard in whichever mode their Site is currently in.

## Steps to Reproduce
On a Fresh install:
1. The Setup page (`wp-admin/?page=give-setup-page`)
2. Copy endpoint URL from Stripe setup section
2. Click Configure Webhooks
3. Paste endpoint/setup new webhook
4. Return to the Setup Page, and see if Stripe Webhook spinner has updated
5. Check Stripe dashboard, and see if any Webhook attempts have been made

## Acceptance Criteria
-   [x] Configure Webhooks button directs users to the Stripe dashboard in the correct mode, so that configuration can happen without needing to know about Test/Live mode settings

kjohnson:
Resolves #5095

## Description

The Stripe URL to configure webhooks is now based on whether or not test mode is enabled.

## Pre-review Checklist

-   [ ] Tests included
-   [ ] Keyboard accessible
-   [ ] Changelog updated

### DIFF
--- a/src/Onboarding/Setup/PageView.php
+++ b/src/Onboarding/Setup/PageView.php
@@ -97,6 +97,20 @@ class PageView {
 	}
 
 	/**
+	 * @return string
+	 *
+	 * @since 2.8.0
+	 */
+	public function getStripeWebhooksURL() {
+		$testMode = give_get_option( 'test_mode' );
+		$url      = ( give_is_setting_enabled( $testMode ) )
+			? 'https://dashboard.stripe.com/test/webhooks'
+			: 'https://dashboard.stripe.com/webhooks';
+
+		return esc_url_raw( $url );
+	}
+
+	/**
 	 * Returns a qualified image URL.
 	 *
 	 * @param string $src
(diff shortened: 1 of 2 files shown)


kevinwhoffman:
@DevinWalker said:

> I think we may want to just highlight both URLs in the message with Live being the most prominent. We were talking about the best way to educate folks about the difference and I think explaining there are two separate areas in Stripe depending if you're testing or not is a good idea. The way this works here could lead to more confusion and the "I thought I already set the webhook" problem when going live.

Let's leave this PR open and outside of the alpha release for now since we need to come to a consensus on the best way to communicate Live/Test mode.

kjohnson:
@kevinwhoffman yeah, sounds like there isn't a solid direction here. As it stands, the Setup Page matches the Stripe Settings and that is a good place to start until we decide on a possible new direction.

kjohnson:
Closing in favor of #5149

PR #5115, opened 2020-08-17
WP Errors are no longer logged when using the block editor

### LINKED ISSUE #5114: GiveWP should not display "register_rest_route" notice with WP 5.5
## User Story

As an admin, I want GiveWP not to through errors when registering custom REST API routes so that I can edit pages without encountering PHP notices.

## Details
Currently, in WP 5.5+ some users encounter the following error when creating a new page:
```Notice: register_rest_route was called incorrectly. The REST API route definition for give-api/v2/form/(?P[\d]+) is missing the required permission_callback argument. For REST API routes that are intended to be public, use __return_true as the permission callback. Please see Debugging in WordPress for more information. (This message was added in version 5.5.0.) in /var/www/wp-includes/functions.php on line 5225```

A complete account of various user experiences with this error can be found at this Canny post: https://givewp.canny.io/admin/board/bug-reports/p/givewp-should-not-display-registerrestroute-notice-with-wp-55

## Expected Behavior
With WP Debug Log enabled, users should not encounter errors related to registering rest routes when using the block editor.

## Steps to Reproduce
In a WP 5.5 Install, running GiveWP 2.7.5:
1. Enable WP Debug log
2. Create a new page
3. View errors in log

## Acceptance Criteria
-   [x] Errors relating to register rest route are no longer logged when viewing the block editor

### PR DESCRIPTION
**Pull request by henryholtgeerts**
Resolves #5114 

## Description
This PR resolves WP Errors which were logged for users when loading the block editor in WP 5.5. Specifically, the errors made reference to a failure on GiveWP's part to clarify a `permission_callback` when registering rest routes. These errors were first introduced in WP 5.5. In accordance with WP recommendations, these endpoints which are meant to be public now include `permission_callback`s which simply return true. 

## Affects
This PR affects API route registration in the `Give_API_v2` class. In particular, it addresses the lack of a defined permission callback for a number of GiveWP core routes.

## Pre-review Checklist
-   [ ] Tests included
-   [ ] Keyboard accessible
-   [ ] Changelog updated

## Testing Instructions
In a WP 5.5 Install, running GiveWP 2.7.5:
- Enable WP Debug log
- Create a new page
- Check for errors in log

### DIFF
--- a/includes/api/class-give-api-v2.php
+++ b/includes/api/class-give-api-v2.php
@@ -79,9 +79,9 @@ class Give_API_V2 {
 	 */
 	private function init() {
 		// Setup hooks.
-		add_action( 'rest_api_init', array( $this, 'register_routes' ) );
-		add_action( 'wp_enqueue_scripts', array( $this, 'localize_script' ), 999 );
-		add_action( 'admin_enqueue_scripts', array( $this, 'localize_script' ), 999 );
+		add_action( 'rest_api_init', [ $this, 'register_routes' ] );
+		add_action( 'wp_enqueue_scripts', [ $this, 'localize_script' ], 999 );
+		add_action( 'admin_enqueue_scripts', [ $this, 'localize_script' ], 999 );
 	}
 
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by kevinwhoffman** [comment]
@henryholtgeerts From the Canny feedback, it sounds like users were seeing this notice in their browser. I suspect this is because the users had `WP_DEBUG` set to `true`, which would cause notices to render in the browser.

I'm trying to gauge the urgency of this fix, so please let me know under what conditions a user would see this notice in the browser.

File: includes/api/class-give-api-v2.php [hunk 1]
```diff
@@ -98,28 +98,31 @@ public function register_routes() {
...
 			$this->rest_base,
 			'/form/(?P<id>[\d]+)',
-			array(
-				'methods'  => 'GET',
-				'callback' => array( $this, 'get_forms_data' ),
-			)
+			[
+				'methods'             => 'GET',
+				'callback'            => [ $this, 'get_forms_data' ],
+				'permission_callback' => __return_true(),
```
kjohnson, line 104: @henryholtgeerts being a callback, shouldn't this be assigned `'__return_true'` (string), where as the above would execute and pass `true` (boolean).

henryholtgeerts: Ah yes - good call! Replaced with `__return_true`: https://github.com/impress-org/givewp/pull/5115/commits/557c09ec08eb3e902cd9be8ceacf160cfb64c6b3

PR #5139, opened 2020-08-18
Stripe is no longer enabled by default.

### LINKED ISSUE #5138: Stripe gateway should be disabled by default if not configured.
If a payment gateway is not configured, it shouldn't show in the list of options on the "sample" form at page-load time.

Currently: With a fresh install, if I go straight into the wizard, Stripe Credit card is enabled, and appears on the sample form. But it doesn't display any kind of warning it just has a light gray box.

kjohnson:
Resolves #5138

## Description

Removes Stripe as a default enabled gateway.

## Pre-review Checklist

-   [ ] Tests included
-   [ ] Keyboard accessible
-   [ ] Changelog updated

### DIFF
--- a/includes/install.php
+++ b/includes/install.php
@@ -367,7 +367,6 @@ function give_get_default_settings() {
 		'gateways'                                    => [
 			'manual'  => 1,
 			'offline' => 1,
-			'stripe'  => 1,
 		],
 		'default_gateway'                             => 'manual',
 


kevinwhoffman:
@kjohnson I think it makes sense _not_ to have Stripe active on a fresh install since it will not have been configured. However, if we don't have Stripe enabled, does that mean that a user would have to take the additional step of visiting the Payment Gateways settings in order to enable Stripe?

kjohnson:
@kevinwhoffman Good question. There is already code to ensure that Stripe is enabled when connected via the Setup Page.

```php
$this->ensureStripeEnabled();
```

https://github.com/impress-org/givewp/blob/develop/src/Onboarding/Routes/StripeWebhookRecievedRoute.php#L40

kevinwhoffman:
> There is already code to ensure that Stripe is enabled when connected via the Setup Page.

Nice, good decision. Just to clarify, if the user connects to Stripe through `Settings > Payment Gateways > Stripe`, then they will still have to manually enable Stripe from the `Payment Gateway` settings page?

henryholtgeerts:
Hmm, testing it out locally I'm running into some weird behavior - likely indicative of bugs with code not specifically introduced in this PR (since it's so light). 

On a fresh install, going through the Onboarding Wizard:
- If you do not select/modify the "features" selected by default in anyway, the Form Preview will include Offline Donations
- If you go back to the "features" step, and select any of the features (does not have to be Offline Donations), then the Form Preview will correctly reflect user's selection regarding Offline Donations
- On a fresh install, Stripe is disabled (as this PR introduces), and does not show up in the Form Preview
- However, once the user has visited the Setup Page (does not have to interact with it in any way), Stripe become enabled (test this on a fresh install by once proceeding through the Onboarding Wizard to the Form Preview step - no Stripe visible, but on a second fresh install try skipping the onboarding wizard, then opening it up - Stripe appears in the Form Preview)

These bugs likely have gone undetected because, until now, we didn't care if Stripe was enabled or not.

To answer @kevinwhoffman 's question - as it currently stands even if a user does not see Stripe in the form preview, it will be enabled as soon as they leave the Onboarding Wizard, which means that any attempt the Stripe gateway would inadvertently ensure that the Stripe gateway is enabled. 

cc: @kjohnson

kevinwhoffman:
@henryholtgeerts Thanks for the report.

Here is my suggestion:

1. Disable `Stripe - Credit Card` on a fresh install so as to prevent a nonfunctional Credit Card option from appearing in the wizard's form preview.
2. On the Setup Page, if a user successfully connects to Stripe, configures the webhook, and we can confirm the webhook test event was received, then at that point the `Stripe - Credit Card` gateway should be enabled and the copy should communicate this to the user.

@DevinWalker Can you please review this suggestion and sign off or propose changes?

I'd like to get @DevinWalker's opinion on the following:

DevinWalker:
@kevinwhoffman the one thing I would like is that on the donation form preview in the onboarding wizard that the admin is provided with information that they can accept credit cards directly on site once they connect a payment gateway like Stripe or PayPal.

kjohnson:
@kevinwhoffman technical note: Enabling Stripe is required to listen for the webhook, which is why I enabled the gateway via the webhooks polling endpoint (after they connect the account).

kevinwhoffman:
> the one thing I would like is that on the donation form preview in the onboarding wizard that the admin is provided with information that they can accept credit cards directly on site once they connect a payment gateway like Stripe or PayPal.

I'm hearing that it's important to convey to the user that GiveWP allows them to accept credit cards on site, and I agree with that 100%. However we should also consider that showing them a credit card field in the form preview does not accurately reflect how the form works at that point, and may not actually be possible depending on the gateway they choose (especially with PayPal Standard still being a promoted gateway until 2.9.0).

Another way to accomplish this goal would be in the messaging beside the form preview. We are already mentioning the ability to `Add or remove payment options`. Can we do more to emphasize the possibility of on-site credit card donations in this space through either design or messaging?

image

DevinWalker:
@kevinwhoffman overall I'm not too worried about it. I think that text does it's job in conveying the message that you can add credit card on site payment options.

kevinwhoffman:
> I think that text does it's job in conveying the message that you can add credit card on site payment options.

With that said, I believe this is good to merge. I'll instruct the team to focus testing on the Stripe connection worfklow in the next round of testing.

PR #5152, opened 2020-08-19
Optional Last Name is denoted in the Multi-Step Template

### LINKED ISSUE #5112: Multi-Step form should clarify required vs. optional fields
## User Story

As a user, I want the Multi-Step for to indicate required feilds so that I know which fields I must fill out, rather than simply relying on validation notices.

## Details
Currently, the Multi-Step form does not indicate what fields are required vs those that are optional.

## Acceptance Criteria
-   [x] Required fields are differentiated from optional fields.

kjohnson:
Resolves #5112

## Description

The "Last Name" field is now denoted as optional with in-line placeholder text, matching the design of the Multi-Step Template.

## Visuals

**Multi-Step Template** is updated.
Screenshot_2020-08-19 Custom Donation Form – WordPress

**Legacy Template** is not changed.
Screenshot_2020-08-19 Custom Donation Form – WordPress(1)

## Pre-review Checklist

### DIFF
--- a/includes/forms/template.php
+++ b/includes/forms/template.php
@@ -22,7 +22,7 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @return string Donation form.
  * @since 1.0
  */
-function give_get_donation_form( $args = array() ) {
+function give_get_donation_form( $args = [] ) {
 
 	global $post;
 	static $count = 1;
@@ -62,9 +62,9 @@ function give_get_donation_form( $args = array() ) {
 	$form_action = add_query_arg(
 		apply_filters(
 			'give_form_action_args',
-			array(
+			[
 				'payment-mode' => $payment_mode,
-			)
+			]
 		),
 		give_get_current_page_url()
 	);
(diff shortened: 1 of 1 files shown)


henryholtgeerts:
Hmm...I'm not entirely sure about this particular approach. Generally, with the Multi-Step form template these kinds of changes are handled by actually modifying markup on the frontend - and not touching existing PHP based rendering logic.
 
The route we've taken with handling Multi-Step forms isn't initially obvious, and I think your approach here works/makes sense, but for consistency's sake I think we should use JS to handle adding this "optional" line to all required fields when loaded inside the Multi-Step form. 

Worth noting: a JS approach which loops through input fields in the Multi-Step form would catch _any_ optional fields (ex: if an admin decided to set the Company field to enabled, but optional). This PR only addresses the Last name field, while not providing any coverage for other core (non FFM) that can be set to optional.

kjohnson:
@henryholtgeerts great feedback! I'll adjust my approach and open a new PR.

PR #5171, opened 2020-08-20
Remove Stripe connect and configure from the Setup Page.

### LINKED ISSUE #5170: Setup Page should not manage connection/configuration of Stripe accounts.
Defer the responsibility of connecting to and configuring Stripe to the existing Stripe payment gateway settings.

kjohnson:
Resolves #5170

## Description

Removes Stripe connect and configure from the Setup Page, in favor of the existing payment gateway settings.

## Affects

The "Connect to Stripe" button now links to the payment gateway settings.

## Pre-review Checklist

-   [ ] Changelog updated


kevinwhoffman:
@kjohnson One of the things we discussed was making sure that the new approach is not any less convenient than the old approach. The old approach actually allowed the user to start the Stripe Connect workflow from the old welcome page like so:

1. Activate GiveWP
2. Land on old Welcome page
3. Click "Connect to Stripe"
4. Agree to authorize Stripe
5. Land back on the Stripe settings page where the webhook modal is displayed

What do you think about still allowing the user to connect to Stripe Live mode from the Setup Page, but then redirecting them back to the Stripe settings page to communicate the webhook configuration?

kjohnson:
@kevinwhoffman My concern is that they leave their site on one page (Setup Page) and then return on another page (Settings), which means backwards navigation doesn't return to the Setup Page.

kevinwhoffman:
> My concern is that they leave their site on one page (Setup Page) and then return on another page (Settings), which means backwards navigation doesn't return to the Setup Page.

I just tested going through the process, and backwards navigation still took me to the Setup page.

My bigger concern is trying to avoid having the user pass through this settings screen just to find the Connect to Stripe button if we could at least initialize the process from the Setup page. I would rather the user not have to process the intricacies of Stripe Checkout, SEPA, and BECS if they don't have to.

image

kjohnson:
Yeah, I can work with that. I'll update the PR.

kjohnson:
@kevinwhoffman changes pushed to update the Stripe connect flow from the Setup Page.

PR #5186, opened 2020-08-24
License activation is restored by resolving JS translation error.

### LINKED ISSUE #5185: Localization should support translations in JavaScript
WordPress v5.4 dropped the localized global variable `commonL10n` in favor of transaltions in JavaScript.

See [Changeset 48267](https://core.trac.wordpress.org/changeset/48267), [Ticket #51123](https://core.trac.wordpress.org/ticket/51123)

Reported via [Canny Feeddback]((https://feedback.givewp.com/bug-reports/p/adding-license-should-work-on-wp-55)

> When navigating to Donations > Settings > Licenses, I should be able to add the license without problems.
For now, when I add the license, nothing happens and the page freezes. The console shows a JS error.
>
> image

kjohnson:
Resolves #5185

## Description

Implemented `wp.i18n.__` for translating common text in WordPress `>=5.5` with backwards compatibility for `commonL10n` for WordPress `<=5.4.2`.

## Pre-review Checklist

-   [x] Changelog updated

## Testing Instructions

Navigate to Donations -> Settings -> Licenses, add a license key and click "Activate License".

### DIFF
--- a/assets/src/js/plugins/give-api/notice.js
+++ b/assets/src/js/plugins/give-api/notice.js
@@ -90,6 +90,8 @@ export default {
 		 * Note: use only in WP Backend
 		 *
 		 * @since 2.5.0
+		 * @since 2.8.0 Localization updated to support changes in WordPress 5.5.
+		 *
 		 * @param {string} notice Notice description.
 		 * @param {string} type   Notice type.
 		 * @param {object} args   Notice type.
@@ -97,7 +99,13 @@ export default {
 		 * @return {string} Notice HTML.
 		 */
 		getAdminNoticeHTML: function( notice, type = 'info', args = { dismissible: true } ) {
-			const btnText = commonL10n.dismiss || '';
+			/**
+			 * WordPress 5.5 removed the localized `commonL10n` in favor of translation in JavaScript.
+			 */
+			const btnText = ( 'undefined' !== typeof commonL10n ) ?
+				commonL10n.dismiss :
+				wp.i18n.__( 'Dismiss this notice.' );
+
 			return `<div class="give-notice notice notice-${ type }${ args.dismissible ? ' is-dismissible' : '' }"><p>${ notice }${ args.dismissible ? ` <button type="button" class="notice-dismiss"><span class="screen-reader-text">${ btnText }</span></button>` : '' }</p</div>`;
 		},
 	},
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Set a specific placeholder for the base country select setting, no longer reads "Select a form". (#5163)
 -   Form preview within the Onboarding Wizard now remains centered on larger viewports. (#5180)
+-   Fixed translation of common text to support WordPress 5.5, with backwards compatibility for `commonL10n`. (#5186)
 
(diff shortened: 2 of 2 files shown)


assets/src/js/plugins/give-api/notice.js
```diff
@@ -90,14 +90,22 @@ export default {
...
 		 * @param {object} args   Notice type.
 		 *
 		 * @return {string} Notice HTML.
 		 */
 		getAdminNoticeHTML: function( notice, type = 'info', args = { dismissible: true } ) {
-			const btnText = commonL10n.dismiss || '';
+			/**
+			 * WordPress 5.5 removed the localized `commonL10n` in favor of translation in JavaScript.
+			 */
+			const btnText = ( 'undefined' !== typeof commonL10n ) ?
```
ravinderk, line 105: @kjohnson This text comes from PHP which is already translated, do we have to retranslate it in js?

image

assets/src/js/plugins/give-api/notice.js
```diff
@@ -90,14 +90,22 @@ export default {
...
+				commonL10n.dismiss :
+				wp.i18n.__( 'Dismiss this notice.' );
```
ravinderk, line 107: @kjohnson I will recommend to import `__` from `wp.i18n` and then use `__` in logic.

image

kjohnson: @ravinderk good suggestion. In this case, I'm referencing `wp.i18n` inline behind a conditional tag because it isn't available for this file in WordPress `5.4` - (console error `wp is not defined`).

ravinderk: 👍

kevinwhoffman:
@kjohnson Thanks for the speedy fix. While the the PR description is technically correct, it's not clear that this PR restores license activation, which was the source of the urgency. A PR title like the following could address the urgent aspect of the issue and still remain accurate:

> License activation is restored by resolving JS translation error

Also please remember to include `Testing Instructions` for every PR that has user-facing consequences.

kevinwhoffman:
@Benunc This was originally deemed urgent and therefore a hotfix, but given Kyle's explanation in https://github.com/impress-org/givewp/issues/5185#issuecomment-679214050 where he states that license activation still technically works, we should revisit whether this warrants a hotfix.

> ...license activation does work, but the UI is broken which makes it look like license activation is broken. Refreshing the page does show that the license was activate.

@kjohnson Please hold off on merging this until we determine the urgency, because if it's a true hotfix then it needs to be merged to `master`, not `develop`.

Benunc:
So the current workaround is to refresh the page after you get that error, and it will show as licensed correctly?

If we can't make it show as activated, I think it's still urgent. But if a page refresh fixes it, then I'm fine waiting until 2.8.0

kjohnson:
To clarify, my testing shows that this does **not** prevent the license from being activated.

Rather this prevents the confirmation notice from displaying.

Refreshing the page does show that the license was activated as expected.

kevinwhoffman:
We discussed this on the GiveWP Development Call. Since the license does refresh and the workaround is to refresh the page, we are going to merge this into the `release/2.8.0` branch to ensure it gets shipped alongside 2.8.0 this week.

kevinwhoffman:
@kjohnson If this is ready to merge, go ahead and do so, and I will include it in the next beta.

PR #5205, opened 2020-08-26
fix: fee recovery checkbox position

### LINKED ISSUE #5204: Fee Recovery checkbox position should work in multi-step form template
## User Story

As a site admin, I want that Fee recovery checkbox appears in the selected place ( i.e Below the credit card fields ).

## Details

For some strange reason, CSS styles for the Fee Recovery add-on are also defined in GiveWP.

Related issues:
https://github.com/impress-org/give-fee-recovery/issues/281
https://github.com/impress-org/give-fee-recovery/issues/278

## Expected Behavior

The fee recovery box should appear in the selected place.

## Steps to Reproduce

1. Install Fee recovery add-on
2. Setup a new donation form and enable fee recovery option per form
3. Set Checkbox location to "Below the credit card fields"
4. View form on the front-end

## Visuals

| Backend  | Front-end |
| ------------- | ------------- |
|  image | image |

## Acceptance Criteria

-   [ ] Fee recovery box should appear in the selected place ( i.e Below the credit card fields ).

alaca:
Resolves #5204 

## Description

As a site admin, I want that Fee recovery checkbox appears in the selected place ( i.e Below the credit card fields ).
This PR solves the problem by removing the CSS hard-coded order property.

## Affects

Fee Recovery add-on

## Pre-review Checklist

## Testing Instructions

1. Install Fee recovery add-on
2. Setup a new donation form and enable fee recovery option per form
3. Set Checkbox location to "Below the credit card fields"
4. View form on the front-end
5. Change Checkbox location again and then view form on the front-end


henryholtgeerts:
For reference, I just performed a quick audit of each location setting to see what currently works and what doesn't:

- [x] Below donation level fields
- [x] Below top donation amount input
- [ ] Above payment options
- [ ] Below payment options
- [x] Above personal info fields 
- [x] Below personal info fields
- [ ] Above credit card fields
- [ ] Below credit card fields

Notably, above/below credit card fields have been identified as being misplaced by logic in the `moveFieldsUnderPaymentGateway()` fn, and should be an easy fix. Currently, they are moving Fee Recovery inputs to the Choose Amount step (in line with original designs for the Multi-Step form), but after user testing it has become apparent they they should both simply move the Fee Recovery input to directly above the "Donate" button.

The above/below payment options is a little trickier, but also not impossible to fix. Because in these locations the Fee Recovery input is placed on page load, we can make sure to move it as part of the init process for the multi-step form (rather than needing the mutation observer, since the input only loads once). In the case of `above payment options`, the input appears to be loading somehow in an absolute position. In the case of `below payment options` the input is consistently loading underneath the Donate button. Because each of these are so specific, it shouldn't be too challenging to reliably infer based on where they loaded, where they should actually end up. 

@alaca I'll keep working on this today, and let you know what more I run into! Thanks for pointing out the moveFieldsUnderPayment gateway fn on the call - definitely where a lot of these inconsistencies stem from. 👍

henryholtgeerts:
@alaca I now have gotten the input positions on the frontend to accurately reflect the setting on the back end! Feel free to test it out, but locally I just tested each setting at in now behaves as expected. There are still minor improvements to made to the `feerecovery.scss` to ensure that margins for the Fee Recovery input are consistent in each of its placements, but I'll leave that to you since it should be quick. Let me know if you run into more issue in your own testing of it!

PR #5209, opened 2020-08-26
On German installs, there is now enough room for all labels in the Stripe Checkout modal

### LINKED ISSUE #5206: On a German install, the checkout modal on a legacy form should not look bad
## Steps to Reproduce
1. On 2.8.0 create a legacy form.
1. enable stripe checkout modal.
1. change the site language to German.
1. Ensure that the translations all download (Dashboard > Updates > Translations)
1. view the form created in step one, specifically when the form pops up.

## Visuals
The Card name field specifically gets two wide, causing issues: 
image

### Acceptance Criteria
- [x] Stripe Checkout modal is wide enough to display all labels without wrapping

henryholtgeerts:
Resolves #5206 

## Description
This PR resolves a minor style issue that occurred when admin opted to use the Stripe modal in Legacy Forms, on a site with German translation. Previously, when translated to German the "Cardholder Name" label was so long that it wrapped to a new line, which caused issues with the associated tooltip and looked bad. This PR increases the maximum width of the Stripe checkout modal to `425px` and sets a width of `90%`, to ensure that the modal is still responsive when viewed on mobile devices or inside a Multi-Step form iframe.

## Affects
This PR affects styles for the Stripe Checkout modal. 

## Visuals
Screen Shot 2020-08-26 at 9 45 18 AM

## Pre-review Checklist
- [x] Changelog updated

## Testing Instructions
1. On 2.8.0 create a legacy form.
2. enable stripe checkout modal.
3. change the site language to German.
4. Ensure that the translations all download (Dashboard > Updates > Translations)
5. view the form created in step one, specifically when the form pops up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 -   Onboarding Form Preview default image has been updated. (#5203)
+-   Stripe Checkout modal max-width has been increased to 425px (#5209)
 
 ## [2.8.0-beta.2] - 2020-08-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 -   Onboarding Form Preview default image has been updated. (#5203)
--   Stripe Checkout modal max-width has been increased to 425px (#5209)
+-   Stripe Checkout modal max-width has been increased to fit-content (#5209)
 
 ## [2.8.0-beta.2] - 2020-08-25
 
--- a/assets/src/css/frontend/give-stripe.scss
+++ b/assets/src/css/frontend/give-stripe.scss
@@ -70,7 +70,8 @@
 		transform: translate(-50%, -50%);
 		background: rgb(255, 255, 255);
 		overflow: hidden;
-		width: 385px;
+		width: 90%;
+		max-width: 425px;
 		border-radius: 0.5rem;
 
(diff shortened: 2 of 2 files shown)


kevinwhoffman:
@henryholtgeerts This feedback just came in from @mathetos and I'm going to ask that you address it in this PR so that we can just review the Stripe Checkout modal once more prior to release.

> One small item I noticed with the Stripe Checkout is that if you tab from the Name it goes to the "Show Terms" link instead of the Expiration field. You can see it highlighted here:

image

> If the Exp field is filled, and you tab from the Name field it goes then to the Donate Now button. So there's some keyboard navigation issues there.

Benunc:
This does fix German, but I tried Dutch (Nederlands) and it breaks differently there. Is there a way to just make the overflow different, or to have the modal flex larger for longer languages?

Screen Shot 2020-08-26 at 1 16 01 PM

Benunc:
@kevinwhoffman @mathetos I believe the keyboard navigation issue exists in Stripe Credit card as well, not just checkout, and not just 2.8.0. It's one of those issues I keep meaning to bring up.

henryholtgeerts:
@Benunc I've modified the PR so that the max-width is based on the content inside the modal. Ref: https://github.com/impress-org/givewp/pull/5209/commits/de667b86e5f385d374f8ddb04195b703cd465e22

@kevinwhoffman After briefly looking into the issue you described, it is part of a larger probably with how the Multi-Step form handles keyboard accessibility. Currently, the form uses tab index attributes to tell the browser what order to tab with. Because the iframe element does not have a tab index, it gets skipped in favor of the next item in the DOM with one (in this case, the Terms and Agreements box). I'd suggest we create a separate issue that overhauls keyboard accessibility in the Multi-Step form so that it does not rely on the tabindex attribute.

PR #5267, opened 2020-09-15
Totals block is now polished

### LINKED ISSUE #5261: Give Totals block presentation should be polished
## User Story

As a admin, I want the Give Totals block frontend and admin appearance polished so that it fits nicely alongside other GiveWP elements.

## Details
The Give Totals block contains the entire feature set necessary to ship. Now there are only a few areas of polish necessary to finish out development.

## Additional Context
The Give Totals block started as the Milestone block, which had a set of design needs that was larger in scope than an a simple Give Totals block would require. 

## Acceptance Criteria
-   [x] Default message is updated to be more generic
-   [x] Admin controls all work/look cohesive in the block editor interface
-   [x] Users can control block colors

henryholtgeerts:
Resolves #5261 

## Description
This PR polishes the frontend display of the Totals block, and organizes inspector controls into panels, as well as introduces a color. palette control to select the goal color. 

## Affects
This PR primarily affects frontend display of the Totals block, and some block editor related rendering logic.

## Visuals
Screencast walking through block implementation:
https://www.loom.com/share/bd6c44112ce54a7ea87fdf9347185763

## Pre-review Checklist
-   [x] Changelog updated

## Testing Instructions
1. Create a new page or post
2. Add the Totals block
3. Test out various settings, as described in the screencast under Visuals

## User Documentation
This PR introduces a new block! It needs some documentation, which can be largely lifted from the existing give_totals documentation: https://givewp.com/documentation/core/shortcodes/give_totals/

Additionally, the block needs a better description (which appears in the block directory area and the inspector area).


kevinwhoffman (changes requested):
@henryholtgeerts Great work on the Totals block. I recorded a [screencast](https://cln.sh/xA1OUj) of myself putting a landing page together to mimic a real world use case. Here are a few of the changes I identified while testing:

## Requested Changes

### From Kevin
- [ ] The `Total` setting should be renamed `Filters` to more accurately describe the controls within it.
- [ ] The block settings should be in the following order: `Goal`, `Filters`, `Content`
- [ ] From top to bottom, the block should display the same as the shortcode: goal summary, progress bar, message (I changed my mind on this after testing it). This will also align the order of the block presentation with the order of the settings suggested above.
- [ ] Spacing around the block should be appropriate for most themes.
	- e.g. On Twenty Sixteen and a few other themes, the margins seem to collapse around the Totals block.
- [ ] Font size of the goal summary (`$360 of $1,000`) should be the same size as the theme's body text so that it doesn't appear too small in comparison.

### From Matt
- [ ] It's important to have a toggle to disable the Goal if they prefer. That's in the shortcode and it can be very useful and powerful particularly in blocks. I think a toggle in the "Goal" section above the "Goal Color" would do it, and then "Goal Color" could even be conditional based on whether the progress bar is displayed or not
- [ ] The "Goal" is an amount. But the progress bar is the visual, so the "Goal Color" should probably instead be "Progress Bar Color"
- [ ] The default goal amount should be set to $1,000 when the block first renders.

kevinwhoffman:
@henryholtgeerts To clarify the presentation order, I re-ordered the elements in dev tools and took a screenshot of what we're looking for (which is the same order as the shortcode presentation):

1. Goal summary
2. Progress bar
3. Message

image

kevinwhoffman:
@kjohnson I focused entirely on user testing in my review. Would you mind doing the code review?

src/Totals/resources/views/totals.php
```diff
@@ -7,28 +7,26 @@
...
 		<div class="give-totals__progress">
 			<?php $percent = ( $this->getTotal() / $this->getGoal() ) * 100; ?>
-			<div class="give-totals__progress-bar" style="width: <?php echo $percent < 100 ? $percent : 100; ?>%"></div>
-			<div class="give-totals__progress-text"> 
-				<?php
-					$total = $this->metric === 'revenue' ? $this->getFormattedTotal() : $this->getTotal();
-					echo sprintf( __( '%1$s of %2$s', 'give' ), $total, $this->getFormattedGoal() );
-				?>
-			</div>
+			<div class="give-totals__progress-bar" style="width: <?php echo $percent < 100 ? $percent : 100; ?>%; background: linear-gradient(180deg, <?php echo $this->getColor(); ?> 0%, <?php echo $this->getColor(); ?> 100%), linear-gradient(180deg, #fff 0%, #ccc 100%);"></div>
```
kjohnson, line 22: I would consider using [min([ percent, 100 ])](https://www.php.net/manual/en/function.min.php) instead of an inline ternary.

kjohnson (approved):
Minor comment, but otherwise looks good to me.

src/Totals/resources/views/totals.php
henryholtgeerts: Good thinking, done: https://github.com/impress-org/givewp/commit/8003d753c1d9439ba2ae8869d97fcbaf1dfbe281

henryholtgeerts:
@kevinwhoffman Changes outlined in your comment have all been incorporated! Ready for a second look whenever you have a free moment.

kevinwhoffman (changes requested):
@henryholtgeerts Coming along nicely! Two remaining issues:

- [ ] When inserting a new Give Totals block, the Goal section should be expanded since it is first. Currently the Content section is still expanded.
- [ ] The front-end spacing looks better, however I noticed that some editor styles are messing with the spacing of the Totals block within the editor. You might need to apply more defensive styling to prevent that from happening.

image

henryholtgeerts:
@kevinwhoffman Both issues you pointed out have been taken care of! See https://github.com/impress-org/givewp/pull/5267/commits/e4d05045cf1a0cc8b5804dd4e3b12bfd12d99c84 and https://github.com/impress-org/givewp/pull/5267/commits/165e1b73f433e3d20971c8688865423ec36d7dbf.

kjohnson:
@henryholtgeerts reading through the feedback here, I think I have some ideas to improve the UX of the Totals blocks. This also comes with some consideration of the upcoming Campaigns block, which is very similar.

I'm considering using `<InnerBlocks />`, which would offer two benefits:
1. Child blocks can be re-arranged for different layout options.
1. Child blocks introduce scope to the different settings.

With the Total Block as an example:

```
Totals Block
|
└─InnerBlocks
       |
       └─ Paragraph Block
       └─ Progress Bar Block
```

The Paragraph Block as a child of the Totals Block can still use the replacer tags in the context of the Totals Block and Server Side Rendering, while offering the ability to change the positioning in the layout.

The Progress Bar Block will have its own settings in its own scope (ex fill color), while having context of the selected forms inherited from the Totals Block.

Looking ahead to the Campaigns Block, we can use the context to set an inherited, static goal value as set by the Campaign, where as the Totals Block would support a custom goal. The implementation of the goal is the same, but we are only changing the entry point of the goal.

Another advantage is increased flexibility for advanced users to incorporate other layout and design blocks within the Totals Block.

This is the information that I'd like to discus in our call later today, but wanted to share it here in the context of the work on the block.

kevinwhoffman:
@henryholtgeerts Two additional changes requested from Matt and Devin today:

- [ ] `Show goal in block?` should be changed to `Show progress bar`.
    - They emphasized that the goal is the amount that is needed to be raised; the progress bar is the visual representation of the progress towards the goal.
- [ ] The loading animation that appears when the block is re-rendered should not change the height of the block container. Currently the loading state causes the entire block height to change which temporarily shifts any content that is under the block.

henryholtgeerts:
Update: The label change has been made and I am working on resolving the issue related to block loading state during re-render. After further investigation, this same issue appears on other GiveWP server side rendered blocks.

henryholtgeerts:
After some further testing, at this point I'm gonna suggest we explore a solution to the flickering issue separate from this PR. This problem is not unique to this block, appears in other GiveWP server-side rendered blocks, and has so far proven to be something of a rabbit hole owing to poor documentation of the ServerSideRender package.

kevinwhoffman (approved):
@henryholtgeerts Thanks for the update. If the loading animation is a general issue with our server-side rendered blocks, then I agree it should be resolved in a separate PR.

Great work here. This is good to merge. 🚀

PR #5270, opened 2020-09-16
Campaigns are now accessible via the WordPress Admin

### LINKED ISSUE #5265: Campaigns should be accessible via the WordPress Admin
## User Story

As a site owner, I want to access my campaigns so that I can organize my donation efforts.

## Details

Site owners organizing their donation efforts with campaigns should be able to add/remove donation forms from the WordPress admin.

## Acceptance Criteria

-   [ ] A campaign overview can be accessed by administrators.

kjohnson:
Resolves #5265  

## Description

Adds a new admin submenu for Campaigns, along with some React app scaffolding and components.

## Visuals

image

## Pre-review Checklist

-   [ ] Changelog updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,8 @@
         "wp-textdomain": "^1.0.1"
     },
     "dependencies": {
+        "@wordpress/components": "^10.1.0",
+        "@wordpress/scripts": "^12.2.0",
         "accounting": "latest",
         "chart.js": "^2.9.3",
         "chartjs-plugin-crosshair": "^1.1.4",
--- a/package.json
+++ b/package.json
@@ -84,7 +84,6 @@
     },
     "dependencies": {
         "@wordpress/components": "^10.1.0",
-        "@wordpress/scripts": "^12.2.0",
         "accounting": "latest",
         "chart.js": "^2.9.3",
         "chartjs-plugin-crosshair": "^1.1.4",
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -83,6 +83,7 @@
         "wp-textdomain": "^1.0.1"
     },
     "dependencies": {
+        "@wordpress/components": "^10.1.0",
```
henryholtgeerts, line 86: In other areas of the codebase we have opted to enqueue needed WP scripts and grab wp packages from the window. I'm open to doing things the way you have here, but curious if you can offer an insight into why this would be better practice?

henryholtgeerts (approved):
Happy with this as a scaffolding PR! Tested locally, worked as expected. I left one question I'd like you to answer before merging, but otherwise this all looks good for a foundation to refine and build on. 👍

package.json
kjohnson: At the time, I was using the `@wordpress/components` `<Icon />` component in Storybook, which isn't connected to WordPress. In which case it should actually be a `dev` dependency.

henryholtgeerts: Make sense, I'd say keep it as you have it then!

kjohnson: Hm... Thinking through it, if we are going to be using Storybook, which doesn't have localized WordPress we may need to import from these dependencies.

Alternatively, we could localize WordPress components to Storybook, but my concern is that we would be overly coupling Storybook and WordPress.

henryholtgeerts: I feel like importing them as dev dependencies makes the most sense, given the need to use components within Storyboook.

PR #5275, opened 2020-09-18
Improve country handling in onboarding

### LINKED ISSUE #5140: Onboarding with country should be simple and intuitive
## User Story

As an admin, I want to connect my PayPal account with the right country so that it works properly.

## Details

Presently, there are two fields in the PayPal Commerce settings: the button to connect, and the country of the account. When the button is clicked it immediately starts the onboarding, using the country. In order to change the country, however, the user needs to change it in the dropdown and then click "Save changes". It's very easy to imagine a setting (in fact I've made this mistake myself while testing) where the user selects the country the in the dropdown and then clicks "Connect" button. It _feels_ like the country is selected, but it hasn't been applied.

An important note here, is that we only need the country when the user is connecting. They really shouldn't be able to change the country after being connected. In fact it does nothing. Perhaps we can come up with a way of asking for the country when the user goes to connect?

Video: https://share.getcloudapp.com/6queQZ2K

## Acceptance Criteria

-   [x] User is able to reliably connect with the correct country

JasonTheAdams:
Resolves #5140 

## Description

The goal of this PR is to simplify the onboarding process such that it is no longer necessary to save the settings for the country code to be applied. Instead, pass the country code via ajax for the partner link.

Ultimately it should perform like the following:

### When the account is not connected
- Display the country field
- Hide all other fields
- When the "Connect with PayPal" button is clicked, pass the country so it's not necessary to "save settings"

### When the account is connected
- Hide the country field
- Show other settings (when there are any)
- Show the connected account details

## Affects

PayPal Commerce onboarding process

## Pre-review Checklist

## Testing Instructions

Will fill out once completed

### DIFF
--- a/src/PaymentGateways/PayPalCommerce/AdminSettingFields.php
+++ b/src/PaymentGateways/PayPalCommerce/AdminSettingFields.php
@@ -38,7 +38,7 @@ class AdminSettingFields {
 		/* @var MerchantDetails $merchantRepository */
 		$merchantRepository = give( MerchantDetails::class );
 		?>
-		<tr valign="top" class="<?php echo $merchantRepository->accountIsConnected() ? 'give-hidden' : ''; ?>">
+		<tr valign="top" class="js-fields-has-custom-saving-logic<?php echo $merchantRepository->accountIsConnected() ? ' give-hidden' : ''; ?>">
 			<th scope="row" class="titledesc">
 				<label for="give_paypal_commerce_country"><?php esc_html_e( 'Account Country', 'give' ); ?></label>
 			</th>
--- a/src/PaymentGateways/PayPalCommerce/AdminSettingFields.php
+++ b/src/PaymentGateways/PayPalCommerce/AdminSettingFields.php
@@ -33,54 +33,52 @@ class AdminSettingFields {
 		$recurringAddonInfo     = Give_License::get_plugin_by_slug( 'give-recurring' );
 		$isRecurringAddonActive = isset( $recurringAddonInfo['Status'] ) && 'active' === $recurringAddonInfo['Status'];
 		?>
-		<table class="form-table give-setting-tab-body give-setting-tab-body-gateways">
-			<tbody>
-				<tr valign="top">
-					<th scope="row" class="titledesc">
-						<label for="give_paypal_commerce_country"><?php esc_html_e( 'PayPal Connection', 'give' ); ?></label>
-					</th>
-					<td class="give-forminp give-forminp-select">
-						<div id="give-paypal-commerce-account-manager-field-wrap">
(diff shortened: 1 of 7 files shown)


JasonTheAdams:
@ravinderk I could use your help on this. I _sort of_ have it working, but there are a few issues I feel like you'd know a faster solution for as you put together the settings for this:
- If the country is changed and the button clicked (without saving), then an "Are you sure?" message pops up after connecting via PayPal as the form thinks it's changed and the page is being refreshed
- After disconnecting the country does not reflect the country applied
- After disconnecting the country field does not reappear until the page is refreshed

ravinderk:
@JasonTheAdams It is complex issue when show/hide element which does not exist on setting page. To keep it simple `Account Country` setting will output on the setting page and it will show/hide on basis of PayPal account connection status.

I also noticed that we want to hide fields conditionally: `Account Country`, `Connect/Disconnect PayPal Button` and rest setting fields. Because there are only two fields, I added logic for those field only. In my opinion we can update logic in future when we will add more setting to this setting page.

src/PaymentGateways/PayPalCommerce/AdminSettingFields.php
```diff
@@ -21,9 +23,64 @@ class AdminSettingFields {
...
+				'id'               => 'paypal_commerce_account_country',
+				'options'          => give_get_country_list(),
+				'chosen'           => true,
+				'placeholder'      => esc_html__( 'Choose a country', 'give' ),
+				'show_option_all'  => false,
+				'show_option_none' => false,
+				'data'             => [
+					'search-type' => 'no_ajax',
+				],
+				'selected'         => $merchantModel->accountCountry ?: $settingRepository->getAccountCountry(),
```
JasonTheAdams, line 59: @ravinderk Is there any reason to keep the country being stored in `wp_options`? I would think the selected should either be what's in the `MerchantDetail` or the GiveWP country. But setting the country here shouldn't affect the GiveWP country setting. Make sense?

ravinderk: @JasonTheAdams I noticed that you add you function to save account country in `wp_options` [`Give\PaymentGateways\PayPalCommerce\Repositories\Settings::updateAccountCountry`](https://github.com/impress-org/givewp/pull/5275/files#diff-2f4056164e4a949b1a5fecaf044f75bfR56) to maintain admin selection even after PayPal account disconnected.

If you want to get ride of this then we can store account country as temporary data and remove it after onboarding admin. Only draw back of this is admin will not able to see last selected country as `selected` in account country setting after disconnecting PayPal account.

JasonTheAdams:
Great work, @ravinderk! Thanks for the alley oop!
alley oop

PR #5339, opened 2020-10-03
Feature: extend reports page menu

### LINKED ISSUE #5332: Add support for adding custom menu items on the Reports page
## User Story

As a developer, I want to extend the current menu on the Reports page so I can add a custom menu item required by the new add-on we are building.  

## Details
image

## Acceptance Criteria

-   [x] Developer is able to extend the current Reports menu by using JavaScript

alaca:
Resolves #5332 

## Description
This PR adds support to extend the current menu on the Reports page using `wp.hooks`.

## Affects
Reports page menu.

## Visuals

image

## Pre-review Checklist

## Testing Instructions
The easiest way to test it would be to create a new Add-on by using the new Add-on boilerplate

1. Generate new Add-on.

2. Create a new JS file called `menu-items.js` and insert the following code:
```js
const { __ } = wp.i18n;
const { addFilter } = wp.hooks;

addFilter( 'givewp-reports-page-menu-items', 'Give', ( data ) => {
	data.push( {
		href: 'url',
		text: __( 'Inserted with Add-on', 'give' ),
	} );

	return data;
} );
```

3. Include `menu-items.js` file in webpack.mix.js

```js

mix
	.setPublicPath( 'public' )
	.sourceMaps( false )

	// admin assets
	.js( 'src/Addon/resources/js/admin/menu-items.js', 'public/js/' )
```

4. Enqueue  `menu-items.js`
```php
wp_enqueue_script(
    'give-menu-filter',
    ADDON_URL . 'public/js/menu-items.js',
    [ 'wp-i18n', 'wp-hooks' ],
   ADDON_VERSION,
);
```

5. Run `npm run production`


assets/src/js/admin/reports/components/tabs/index.js
```diff
@@ -8,11 +9,22 @@ import Tab from '../tab';
...
 
+	const renderItems = () => {
+		const items = applyFilters( 'givewp-reports-page-menu-items', [] );
+
+		if ( Array.isArray( items ) && items.length ) {
+			return items
+				.filter( ( item ) => item.hasOwnProperty( 'href' ) && item.hasOwnProperty( 'text' ) )
+				.map( ( item, i ) => <a key={ i } className="nav-tab" href={ item.href }>{ item.text }</a> );
+		}
+	};
```
JasonTheAdams, line 20: Couple suggestions:
- Make use of `useMemo` here to memoize the `items` variable, that way the items are only rendered when they change
- Change the hook to `givewp-reports-page-menu-links`, as I suspect we'll ultimately want to distinguish the links (i.e. anchors) from new pages, which will likely have its own register.
- Instead of filtering, do a check when mapping through the items and use `console.warn` to indicate improperly shaped items. Right now this would just silently ignore malformed items with no explanation. Guide devs in the right way to use something.

JasonTheAdams (changes requested):
Good work, @alaca! Few suggestions!

assets/src/js/admin/reports/components/tabs/index.js
alaca: @JasonTheAdams I did everything except making use of the useMemo function. The reason is very simple, `items` are not component props, so component won't be re-rendered each time React updates the DOM. If they were passed as component props, then we could use `React.memo` to prevent re-rendering, but in this case, it will be redundant.

JasonTheAdams: You are right, though I wish changes due to WordPress hooks did actually re-render the component. Oh well!

JasonTheAdams (approved):
Good to go! 👍

PR #5341, opened 2020-10-05
Buttons are now allowed in the content area

### LINKED ISSUE #5340: Buttons should be allowed in the content area
## User Story

As an admin, I want to be able to add buttons the Multi-Form Goal block content area so that I can easily create a call to action element if my site needs it.

## Details

I am able to add buttons to the core Media & Text block, so I think we should be able to allow buttons in our block. If that's possible, we could display the block with some starter content and buttons that would do more to show the user what is possible with this block.

## Acceptance Criteria

-   [x] Admin can optionally add Button Group blocks to the content area of a Multi-Form Goal block

henryholtgeerts:
Resolves #5340 

## Description
This PR introduces the ability for admin to add Buttons to the Multi-Form Goal block. It does so through a new child block called Multi-Form Goal Content, which has uses the same block template that Multi-Form Goals previously did, but also includes an "allowed blocks" attribute which allows users to add core paragraph and buttons blocks.

## Affects
This PR affects the editor interface for Multi-Form Goal blocks.

## Visuals
Screen Shot 2020-10-05 at 12 01 02 PM

## Pre-review Checklist
-   [x] Changelog updated

## Testing Instructions

1. Add new page or post
2. Add Multi-Form goal block
3. Add buttons to Multi-Form Goal block
4. Check how editor appearance compares with frontend appearance

### DIFF
--- a/src/MultiFormGoals/resources/js/blocks/multi-form-goal-content/index.js
+++ b/src/MultiFormGoals/resources/js/blocks/multi-form-goal-content/index.js
@@ -20,41 +20,43 @@ import '../../../css/editor.scss';
  * Register Block
  */
 const blockTemplate = [
-	[ 'core/media-text', {
-		imageFill: true,
-	}, [
-		[ 'give/multi-form-goal-content', {} ],
-	] ],
-	[ 'give/progress-bar', {} ],
+	[ 'core/heading', {
+		placeholder: __( 'Heading', 'give' ),
+	} ],
+	[ 'core/paragraph', {
+		placeholder: __( 'Summary', 'give' ),
+	} ],
 ];
 
-export default registerBlockType( 'give/multi-form-goal', {
-	title: __( 'Multi-Form Goal', 'give' ),
+const allowedBlocks = [
+	'core/paragraph',
+	'core/buttons',
+];
+
+export default registerBlockType( 'give/multi-form-goal-content', {
+	title: __( 'Multi-Form Goal Content', 'give' ),
 	description: __( 'The Multi-Form Goals block displays progress made across donation forms towards a common goal.', 'give' ),
 	category: 'give',
 	icon: <GiveLogo color="grey" />,
+	parent: [ 'give/multi-form-goal' ],
 	keywords: [
 		__( 'donation', 'give' ),
 		__( 'multi form goals', 'give' ),
 	],
-	supports: {
-		align: [
-			'wide',
-		],
-	},
 	edit: () => {
 		return (
-			<div className="give-multi-form-goal-block">
+			<div className="give-multi-form-goal-content-block">
 				<InnerBlocks
 					template={ blockTemplate }
-					templateLock="all"
+					allowedBlocks={ allowedBlocks }
+					templateLock={ false }
 				/>
 			</div>
 		);
 	},
 	save: () => {
 		return (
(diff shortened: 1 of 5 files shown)


src/MultiFormGoals/resources/js/blocks/multi-form-goal/index.js
```diff
@@ -23,12 +23,7 @@ const blockTemplate = [
 	[ 'core/media-text', {
 		imageFill: true,
 	}, [
-		[ 'core/heading', {
-			placeholder: __( 'Heading', 'give' ),
-		} ],
-		[ 'core/paragraph', {
-			placeholder: __( 'Summary', 'give' ),
-		} ],
+		[ 'give/multi-form-goal-content', {} ],
```
kjohnson, line 26: @henryholtgeerts should the new "content" block wrap the `core/media-text` block as well? Or are we being intentionally opinionated with having an image?

src/MultiFormGoals/resources/js/blocks/multi-form-goal-content/index.js
```diff
@@ -0,0 +1,64 @@
...
+const blockTemplate = [
+	[ 'core/heading', {
+		placeholder: __( 'Heading', 'give' ),
+	} ],
+	[ 'core/paragraph', {
+		placeholder: __( 'Summary', 'give' ),
+	} ],
+];
+
+const allowedBlocks = [
```
kjohnson, line 31: We should probably include `core/heading` in the `allowedBlocks` list.

kevinwhoffman:
@henryholtgeerts @kjohnson I just tested the buttons integration. It definitely complicates the experience, and I'm not sure the end result is worth that complexity.

Specifically, I keep getting stuck inside the content block with this extra space at the bottom where you would insert another block. Sometimes it goes away when the block is unfocused, however other times the gap stays in place.

image

What do you think about the user experience?

henryholtgeerts:
@kevinwhoffman @kjohnson Thanks for taking a look at it. I'm also not entirely sold on this feature for the same reasons Kevin mentioned. I think it adds significant complexity, so an experience that otherwise is very straight forward. In addition to the issues with the appender message Kevin pointed out, there are also inconsistencies in how themes handle margins for nested InnerBlocks (like we are using here) which are challenging to anticipate. 

I am personally in favor of scrapping this feature, and revisiting it down the road if there's significant user demand. If it were deemed extremely necessary, I could see how we might justify forking the media & text block altogether to achieve a more cohesive editing experience (than I have in this PR), but to me the ability to add buttons alone doesn't offset the technical debt we would be assuming with that approach.

kjohnson:
Those are good points. Especially for users that are new to blocks the complexity can be counter productive.

@henryholtgeerts are you suggesting scrapping Media & Text in favor of a more static layout?

kevinwhoffman:
> I am personally in favor of scrapping this feature, and revisiting it down the road if there's significant user demand.

I agree. This was a good spike that taught us the limits of the Media & Text block when used as an inner block.

I'm in favor of closing out this PR and returning to the previous behavior that supported an image, heading, and text above the progress bar. I would note that the image controls, including the focal point control, afford a lot of flexibility, so let's make sure not to lose those features going forward.

henryholtgeerts:
Per discussion here, I am closing out this PR.

PR #5346, opened 2020-10-06
Add hash to export files to avoid collisions

JasonTheAdams:
## Description

When export files are created they have the same name, something like `give-payments.csv`. The problem is that due to the fact that exports are built via AJAX in multiple runs, it's possible for more than one export (from different users) to be happening simultaneously. In this case, the different exports would actually write to the same file.

This adds a hash to the file name on a per export basis to avoid this conflict. Note that it does not affect the final file name as downloaded by the user, as this is defined in the request header and isn't an issue. This only affects the file temporarily on the server for download.

## Affects

Exporting donations via Tools

## Pre-review Checklist

-   [x] Changelog updated

## Testing Instructions

Run an export and watch the `wp-contents/uploads` directory to see if the file temporarily generated has a hash in it. The downloaded file name should be as before and not contain a hash.


includes/admin/upgrades/upgrade-functions.php
```diff
@@ -156,6 +156,10 @@ function give_do_automatic_upgrades() {
 			give_v270_upgrades();
 
 			$did_upgrade = true;
+
+		case version_compare( $give_version, '2.9.0', '<' ):
+			give_v290_remove_old_export_files();
+			$did_upgrade = true;
```
JasonTheAdams, line 162: @ravinderk Am I adding this automatic upgrade correctly? I want this to run once someone upgrades to 2.9.0.

JasonTheAdams:
@DevinWalker I've added an automatic upgrade to remove any leftover export files. This is ready for review.

JasonTheAdams:
@ravinderk I've added you as a reviewer. Please contact me before doing the review.

DevinWalker (approved):

includes/admin/upgrades/upgrade-functions.php
ravinderk: This is right way.

includes/admin/upgrades/upgrade-functions.php
```diff
@@ -3815,3 +3819,15 @@ function give_v270_store_stripe_account_for_donation_callback() {
...
+/**
+ * Removes any leftover export files that should've been deleted
+ *
+ * @since 2.9.0
+ */
+function give_v290_remove_old_export_files() {
+	@unlink( WP_CONTENT_DIR . '/uploads/give-payments.csv' );
+	@unlink( WP_CONTENT_DIR . '/uploads/give-donors.csv' );
+
+	give_set_upgrade_complete( 'v290_remove_old_export_files' );
```
ravinderk, line 3832: @JasonTheAdams For automatically upgrade you do not have to store id in the database if not any logic depends upon it. I will remove it.

ravinderk: fixed: https://github.com/impress-org/givewp/compare/e96b9571c95b...a3c608f60b0c

ravinderk:
@JasonTheAdams Did you test this pr against a large database? I exported donations for a donation form which has around 100 donations. During export six files generated and only the sixth file was downloaded which does not contain all data. I was expecting 100 donations but I got only twenty in the export file.

Let me know if you have any questions.

image

image

ravinderk (changes requested):
Added my review in [this comment](https://github.com/impress-org/givewp/pull/5346#issuecomment-709850742)

alaca:
As the main problem with the export is that multiple users can export the same reports at the same time which will cause collisions because the same file name is being used for all users. The solution is to add a hash which is user related ( User ID ) so we can know for sure which user is exporting what.

JasonTheAdams:
Thanks for the help on this, @alaca! It's working great!

PR #5348, opened 2020-10-07
Multi-Form Goal Progress Bar styles are encapsulated via a Shadow DOM

### LINKED ISSUE #5347: Multi-Form Goal Progress Bar styles should be encapsulated
## Acceptance Criteria

- [ ] Styles are not inherited from the DOM

kjohnson:
Resolves #5347

## Description

Encapsulates progress bar HTML/CSS using the Shadow DOM. Styles are loaded via a `<style />` tag in the HTML which are carried over into the Shadow DOM.

**Note:** the Shadow DOM is only implemented at the time of render and not within the block editor.

## Affects

- Adds a separate build for styles.

## Visuals

image
image

## Pre-review Checklist

-   [x] Changelog updated\

## Testing Instructions

- Enqueue extra styles on the page
- Progress Bar component should not inherit styles

### DIFF
--- a/src/MultiFormGoals/resources/views/progressbar.php
+++ b/src/MultiFormGoals/resources/views/progressbar.php
@@ -42,7 +42,7 @@ $uniqid = uniqid();
 	(function() {
 		const container = document.getElementById('<?php echo $uniqid; ?>')
 		const content = container.innerHTML
-		const shadow = container.attachShadow({mode: 'closed'});
+		const shadow = container.attachShadow({mode: 'open'});
 		shadow.innerHTML = content
 	})()
 </script>
--- a/src/MultiFormGoals/resources/views/progressbar.php
+++ b/src/MultiFormGoals/resources/views/progressbar.php
@@ -2,11 +2,15 @@
 /**
  * Multi-Form Goals block/shortcode template
  * Styles for this template are defined in 'blocks/multi-form-goals/common.scss'
- *
  */
+
+$uniqid = uniqid();
 ?>
 
-<div class="give-progress-bar-block">
+<div id="<?php echo $uniqid; ?>" class="give-progress-bar-block">
+	<style>
+		<?php echo file_get_contents( GIVE_PLUGIN_DIR . 'assets/dist/css/multi-form-goal-block.css' ); ?>
+	</style>
 	<div class="give-progress-bar-block__goal">
 		<div class="give-progress-bar-block__progress">
 			<?php $percent = ( $this->getTotal() / $this->getGoal() ) * 100; ?>
@@ -34,3 +38,9 @@
 		<?php endif; ?>
 	</div>
 </div>
+<script>
+	const container = document.getElementById('<?php echo $uniqid; ?>')
+	const content = container.innerHTML
+	let shadow = container.attachShadow({mode: 'closed'});
+	shadow.innerHTML = content
+</script>
--- a/src/MultiFormGoals/resources/views/progressbar.php
+++ b/src/MultiFormGoals/resources/views/progressbar.php
@@ -39,8 +39,10 @@ $uniqid = uniqid();
 	</div>
 </div>
 <script>
-	const container = document.getElementById('<?php echo $uniqid; ?>')
-	const content = container.innerHTML
-	let shadow = container.attachShadow({mode: 'closed'});
-	shadow.innerHTML = content
+	(function() {
+		const container = document.getElementById('<?php echo $uniqid; ?>')
+		const content = container.innerHTML
+		const shadow = container.attachShadow({mode: 'closed'});
+		shadow.innerHTML = content
+	})()
 </script>
--- a/src/MultiFormGoals/resources/views/progressbar.php
+++ b/src/MultiFormGoals/resources/views/progressbar.php
@@ -11,29 +11,30 @@ $uniqid = uniqid();
 	<style>
 		<?php echo file_get_contents( GIVE_PLUGIN_DIR . 'assets/dist/css/multi-form-goal-block.css' ); ?>
 	</style>
-	<div class="give-progress-bar-block__goal">
-		<div class="give-progress-bar-block__progress">
+	<!-- Target the Progress Bar Block elements using CSS "parts", see https://developer.mozilla.org/en-US/docs/Web/CSS/::part -->
+	<div part="goal" class="give-progress-bar-block__goal">
+		<div part="progress" class="give-progress-bar-block__progress">
 			<?php $percent = ( $this->getTotal() / $this->getGoal() ) * 100; ?>
-			<div class="give-progress-bar-block__progress-bar" style="width: <?php echo min( [ $percent, 100 ] ); ?>%; background: linear-gradient(180deg, <?php echo $this->getColor(); ?> 0%, <?php echo $this->getColor(); ?> 100%), linear-gradient(180deg, #fff 0%, #ccc 100%);"></div>
+			<div part="progress-bar" class="give-progress-bar-block__progress-bar" style="width: <?php echo min( [ $percent, 100 ] ); ?>%; background: linear-gradient(180deg, <?php echo $this->getColor(); ?> 0%, <?php echo $this->getColor(); ?> 100%), linear-gradient(180deg, #fff 0%, #ccc 100%);"></div>
 		</div>
 	</div>
-	<div class="give-progress-bar-block__stats">
-		<div class="give-progress-bar-block__stat">
-			<div><?php echo $this->getFormattedTotal(); ?></div>
-			<div><?php echo __( 'raised', 'give' ); ?></div>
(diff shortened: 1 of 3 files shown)


src/MultiFormGoals/resources/views/progressbar.php
```diff
@@ -34,3 +38,9 @@
 		<?php endif; ?>
 	</div>
 </div>
+<script>
+	const container = document.getElementById('<?php echo $uniqid; ?>')
+	const content = container.innerHTML
+	let shadow = container.attachShadow({mode: 'closed'});
```
JasonTheAdams, line 44: I believe we want this to be open. Open doesn't mean it will suddenly be affect by parent styles, but it means the the shadow contents can be affected, which we want — they just have to be very intentionally extended.

src/MultiFormGoals/resources/views/progressbar.php
```diff
@@ -34,3 +38,9 @@
 		<?php endif; ?>
 	</div>
 </div>
+<script>
+	const container = document.getElementById('<?php echo $uniqid; ?>')
```
henryholtgeerts, line 42: Declaring this as a `const` means that the Progress Bar block can only be rendered once per page. If you add multiple Multi-Form Goal blocks to a page, you'll see this error in console: `Uncaught SyntaxError: Identifier 'container' has already been declared`

henryholtgeerts (changes requested):
In my own testing, this implementation was effective at shielding the first instance of the Progress Bar from outside styles, but failed to do so for any additional instances.

Screen Shot 2020-10-07 at 1 34 39 PM

src/MultiFormGoals/resources/views/progressbar.php
kjohnson: Ah, yeah - that makes sense. I'll update the implementation.

src/MultiFormGoals/resources/views/progressbar.php
kjohnson: I remember now - updated.

src/MultiFormGoals/resources/views/progressbar.php
kjohnson: @henryholtgeerts I pushed an update to encapsulate the scripting.

JasonTheAdams:
Good note, @henryholtgeerts. That's part of this experiment. Please use this PR to test _how_ folks would be able to extend/adjust the styles of the component. Don't merge this PR until we have a clear understanding of that as it's critical to this methodology.

kjohnson:
@henryholtgeerts good question. Currently, you would need to use JS to add styles to the Shadow DOM, which is probably more complex than just adding a action hook.

That said, I'm thinking of adding `textarea` setting for Custom CSS, which can be injected in the proper place at render.

Thoughts? /cc @JasonTheAdams @kevinwhoffman @DevinWalker

JasonTheAdams:
If we did actually register this as a web component then we could make use of [::part](https://developer.mozilla.org/en-US/docs/Web/CSS/::part). 🤔

JasonTheAdams:
To be clear, I don't mind the idea of a css text area, but that does mean someone would have to provide the same CSS for _every_ block on their site via that text area. Ideally, they'll have a way of setting styles in one place that affect the component everwhere.

kjohnson:
@JasonTheAdams looking at the [::part()](https://developer.mozilla.org/en-US/docs/Web/CSS/::part) documentation, I don't think it requires a web component:

> The ::part CSS pseudo-element represents any element within **a shadow tree** that has a matching part attribute.

kjohnson:
I do like `::part()` for supporting shared styles and one-less setting.

A quick local test seems to work!

kjohnson:
I also like that the parts, like classes, are discover-able and self-documenting for developers via the HTML.

kevinwhoffman:
I like that the shadow DOM approach protects our components from unwanted styling conflicts, but I still think it is important for developers to be able to style our components through traditional CSS that they can write in their editor of choice and commit to their project's source control. If `::part()` let's developers write CSS in this way, then I'm all for it.

Having to add CSS through some form of text field in a settings panel is far from ideal. This should be a last resort IMO.

henryholtgeerts:
Chiming in here with some info about browser support for ::part selector: https://caniuse.com/?search=%3A%3Apart

I was also able to reproduce the styling workflow @kjohnson described with screenshots, it makes sense and works as expected locally.

JasonTheAdams:
Thanks, @henryholtgeerts! The most concerning part about `::part` (snicker) is the market share for legacy iOS Safari versions. The good news is that it's already legacy, so it should continue to decrease. Also, it is an all-or-nothing type of deal — that is, the component won't look partially styled, but will still look exactly as we designed it, which isn't what they want, but doesn't look terrible. 🤔

kjohnson:
I think I'm comfortable with calling custom styling a progressive enhancement.

kevinwhoffman:
We should also consider using `<template>` as an alternative method of allowing customization of our components. This approach would allow the developer to customize not only the styles but also the markup used in the template (I'm less concerned about allowing custom markup but it is worth mentioning).

My understanding is that when a template is used in combination with a Web Component, the template can include classes and styles from the light DOM.

- https://css-tricks.com/crafting-reusable-html-templates/
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template
- https://caniuse.com/template

kevinwhoffman:
The end of this article reviews a few strategies for styling web components that are relevant to this discussion: https://css-tricks.com/encapsulating-style-and-structure-with-shadow-dom/#strategies-for-styling-the-shadow-dom

kjohnson:
What I like about the `part` attributes approach is that it offers discover-able CSS selectors in the browsers developer console, which is probably where developers will look first. By contrast, exposing an API via CSS Custom Properties would rely on external documentation.

kevinwhoffman:
@kjohnson This came up on a call with @JasonTheAdams @DevinWalker @mathetos and we agree with you that `part` feels like the best path forward here.

Am I able to pull down this PR and test the styling of the component using `part`?

kjohnson:
@kevinwhoffman yes - have at it.

kevinwhoffman:
After spending some time exploring this approach ([screencast](https://drive.google.com/file/d/1MPerO6kRh1hYH2wK-hDCAmlJpCNei6UB/view)), I think `part` provides the best balance of protection for our component styles and customization for end users.

There are a couple of items we should address before this PR merges:

1. As mentioned in the screencast, we should revisit how the parts are named and consider reusing the same name for related parts such as the stat labels and values.
2. Decide whether the `give-` prefix is necessary for CSS classes printed within the shadow DOM. For example, can we simply use `progress-bar` instead of `give-progress-bar-block` since the component is encapsulated?

mathetos:
@kevinwhoffman Since we discovered that you can have multiple names for parts, I think rather than reusing the same name for related parts, we should use a good naming convention that allows for multiple parts. So using your example in the screencast, instead of reusing `stat-value` we could use 

* `stat stat-count-value`
* `stat stat-total-value`
* `stat stat-goal-value`

Then targeting just `part(stat)` would get all of those, and also be able to use `part(stat-count-value)` to target the individual element as well.

henryholtgeerts:
We should also take some time to determine how this style encapsulation will work in the block editor and other page builders. I understand that this PR is not aimed at addressing/exploring this question so much as the frontend, but it will certainly be important that we deliver users consistently styled components both in their editors and on the page.

kevinwhoffman:
@henryholtgeerts That's a good point regarding styling in the editor versus the front end. However I'm not sure the shadow DOM approach makes that challenge much different than any other custom CSS on blocks.

For example, if I use the Customizer to add custom CSS to a page on the front end, those styles are not applied in the editor. If an end user wanted absolute parity between styles in the front end and editor, then they'd have to enqueue those custom styles in both places. This is obviously not ideal, but it's a WordPress problem, not a problem related to this specific approach.

henryholtgeerts:
@kevinwhoffman I agree, that there is not much of an issue regarding UX for how custom styles are loaded here - that paradigm is already well established. My concern is more so with how we ensure that our components are shielded from theme styles within the block editor the same way they are on the frontend.

My worry is more so that a theme might define styles which, if we don't make sure the shadow DOM approach works for blocks in the editor, might affect our components in the block editor, but not on the frontend. Users could potentially see a block which inherits theme styles within the block editor, but find that it is uses GiveWP styles on the frontend. I'm hoping to ensure consistency across editors and the frontend.

kevinwhoffman:
> My concern is more so with how we ensure that our components are shielded from theme styles within the block editor the same way they are on the frontend.

Ah, I get your concern now. Thanks for clarifying. I agree that's worth looking into.

kjohnson:
@henryholtgeerts that is certainly something to consider.

As it stands, the Shadow DOM implementation is a part of the shortcode rendering, which isn't integrated with the `react` environment of the Block Editor.

PR #5353, opened 2020-10-07
Added filter for reports to support additional post status

mehul0810:
This PR is a hotfix so it doesn't have any issue related to it. Please find the Slack thread here: https://givewp.slack.com/archives/C0FAGC83C/p1602078781075300

## Description
This PR adds a filter to extend the functionality so that users who wants to customize the post status for donations can add their registered post status support to reports in admin.

## Affects
https://github.com/impress-org/givewp/blob/develop/src/API/Endpoints/Reports/Endpoint.php#L364

## Pre-review Checklist
-   [ ] Tests included
-   [ ] Keyboard accessible
-   [ ] Changelog updated

### DIFF
--- a/src/API/Endpoints/Reports/Endpoint.php
+++ b/src/API/Endpoints/Reports/Endpoint.php
@@ -361,10 +361,13 @@ abstract class Endpoint implements RestRoute {
 		$gateway = array_keys( $gatewayObjects );
 
 		$args = [
-			'post_status' => [
-				'publish',
-				'give_subscription',
-			],
+			'post_status' => apply_filters(
+				'give_reports_add_post_status',
+				[
+					'publish',
+					'give_subscription',
+				]
+			),
 			'number'      => $number,
 			'paged'       => 1,
 			'orderby'     => $orderBy,


src/API/Endpoints/Reports/Endpoint.php
```diff
@@ -361,10 +361,13 @@ public function getPayments( $startStr, $endStr, $orderBy = 'date', $number = -
...
-				'publish',
-				'give_subscription',
-			],
+			'post_status' => apply_filters(
+				'give_reports_add_post_status',
+				[
+					'publish',
+					'give_subscription',
+				]
+			),
```
JasonTheAdams, line 370: @mehul0810 Thanks for the PR! Instead of just filtering the status, let's apply a `give_reports_get_payments_query_args` filter to the whole `$args` array.

mehul0810: @JasonTheAdams Just wondering whether as this is admin reports will filtering the whole `$args` array would be useful? 🤔  I thought about it and it seems to me that if I filter the whole `$args` incorrectly then it can break the reports completely.

Let me know what you think.

mehul0810: @JasonTheAdams I think the suggestion you provided is good to implement as filtering `$args` will help specially if anyone doing major customizations with GiveWP. So, I am doing the change just now.

JasonTheAdams: If people are going to break things, this filter won't stop them. 😉

mehul0810: @JasonTheAdams Additionally, I noticed that just adding the filter is not working out as condition for `publish` and `give_subscription` for reports is scattered across smaller modules present within the reports page as in the screenshot. I believe I can achieve it using a helper fn which will implement the conditional logic and the function used instead of the conditional check everywhere but this can take time to implement and test the functionality. So, I will revert back by tomorrow.

image

mehul0810:
@JasonTheAdams It seems that the change which I need is not a quick fix. Hence, closing this PR for now. I will create a PR again when I have something working.

PR #5357, opened 2020-10-08
Multi-Form Goal Block uses the Revenue table to calculate progress towards a goal.

### LINKED ISSUE #5356: Multi-Form Goal block should leverage the Revenue Table
Currently, the Multi-Form Block uses a direct query which joins the `posts` table with `give_donationmeta`. This does not account for multiple currencies as compatibility for the Currency Switcher add-on.

Instead, the new Revenue table as introduced by @ravinderk should be the source of truth.

## Acceptance Criteria
- [ ] A donation in EUR (with a base currency of USD) is calculated towards the goal in the base currency (USD).

kjohnson:
Closes #5356 

## Description

- Replaces the post/meta join with a post/revenue join.

## Affects

- `Give\MultiFormGoals\ProgressBar/Model::getTotal()` now accounts for the amount as stored in the lowest unit of currency.

## Pre-review Checklist

-   [x] Changelog updated

## Testing Instructions

- Make a donation in USD (base currency).
- Make a donation in EUR.
- Create a Multi-Form Goal block.
- Total reflects the two donations in the base currency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   PayPal Donations is a new payment gateway (#5079)
 -   PayPal Donations supports currency switcher (#5335)
 -   PayPal Donations supports subscriptions (#5173, #5221, #5308)
+
 ### Changed
 
 -   Reports page main menu is now extendable ( #5339 )
 -   Multi-Form Goal Progress Bar only shows "time to go" if the end date has not passed (#5350)
+-   Multi-Form Goal Block uses the Revenue table to calculate progress towards a goal. (#5357)
 
 ### Fixed
 
--- a/src/MultiFormGoals/ProgressBar/Model.php
+++ b/src/MultiFormGoals/ProgressBar/Model.php
@@ -115,7 +115,7 @@ class Model {
 	protected function getTotal() {
 		$query   = new Query( $this->getForms() );
 		$results = $query->getResults();
-		return $results->total;
+		return $results->total / 100;
 	}
 
 	/**
--- a/src/MultiFormGoals/ProgressBar/Query.php
+++ b/src/MultiFormGoals/ProgressBar/Query.php
@@ -25,29 +25,21 @@ class Query {
 	public function getSQL() {
 		$sql = "
             SELECT
-                sum( meta_value ) as total,
-                count( meta_value ) as count,
-                avg( meta_value ) as average
+                sum( revenue.amount ) as total,
+                count( payment.ID ) as count
             FROM wp_posts as payment
-                JOIN wp_give_donationmeta as meta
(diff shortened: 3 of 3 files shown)


henryholtgeerts:
@kjohnson Testing this out locally, I'm seeing some inconsistent totals between the Multi-Form Goal and progress measured on forms themselves. When I filter the Multi-Form Goal to a single form, I'm seeing a smaller total in the Multi-Form Goal block than on the form's itself. This likely has to do with how we are accounting for fee recovery?

Progress shown in the Multi-Form Goal block (restricted to a single form):
Screen Shot 2020-10-09 at 9 00 55 AM

Progress shown on the actual donation form:
Screen Shot 2020-10-09 at 9 01 48 AM

I think that the logic you've implemented does not count recovered fees towards a goal. In my opinion, that's the correct behavior, but it is inconsistent with existing implementations of goals in GiveWP. There was some discussion about this in the original Canny post, and if I remember correctly the decision made was to emulate existing Goals logic, even if it is not ideal.

Canny post: https://givewp.canny.io/admin/feedback/qa-multi-form-goals/p/multi-form-goal-block-should-be-compatible-with-other-add-ons?boards=qa-multi-form-goals

kevinwhoffman:
@henryholtgeerts You're correct that goal progress should include the donation + covered fees in order to be consistent with how form goals are currently calculated.

kjohnson:
@henryholtgeerts I'm looking into the inconsistency between the block progress and the form progress. That said, this pulls from the revenue table, so any changes would likely need to be made there.

kjohnson:
| _ | Fee Recovery |
| --- | --- |
| Block | image |
@henryholtgeerts I'm showing that Fee Recovery is the same for the Block and the Form, but not the List Table.

| Form | image |
| Table | image |

henryholtgeerts:
@kjohnson Hmm...maybe I'm missing something on my end. Here's what I'm seeing in testing:

| _ | Fee Recovery |
| --- | --- |
| Block | image |
| Form | image |
| Table | image |

kjohnson:
@henryholtgeerts hm... those are all different numbers... uh oh. Any idea what the discrepancy is?

kjohnson:
@henryholtgeerts I already know that the form and the list table show different numbers, so let's focus on the block and the form. The obvious difference is the number of donations. Can you take a look at the seven donations on the form? Maybe this is related to payment status?

henryholtgeerts:
@kjohnson I also am noticing now that the "Revenue" column of the list table differs from the progress presented in the "Goal" column for a given form. This maybe points to broader issues in the revenue table that need to be addressed for all this to be resolved.

kjohnson:
@henryholtgeerts I'm going to get these changes into `develop`, because this change itself does not calculate the revenue, but rather leverages the Revenue table. From there we can further investigate any discrepancies.

PR #5371, opened 2020-10-13
Multi-Form Goal query results are cached

### LINKED ISSUE #5365: Multi-Form Goal query results should be cached
The Donation Revenue query should only run once per request cycle to avoid duplicate queries.

## Acceptance Criteria
- [x] Query only runs once.
- [x] With multiple goals on the page, the query only runs once per goal block.

kjohnson:
Resolves #5365

## Description

The `MultiFormGoal/ProgressBar/Query` now only runs once per block instance per page load. This prevents running duplicate queries on page load while allowing for real-time results as donations are made.

I originally was looking into `Give_Cache` class. However, I changed direction for the following reasons:
1. Results are unique to each block instance.
1. Progress should be updated in real time as donations are made.

Therefor, the results are cached within the request cycle to avoid duplicate queries and each instance of the block model has its own cache (because the configuration will be different).

## Pre-review Checklist

-   [x] Changelog updated

## Testing Instructions

- Add a Multi-Form Goal Block to a page.
- Check that the Revenue table is only queried once


JasonTheAdams:
@kjohnson I'm confused. This is caching the result for the quest, which _does_ limit it to only once per request. I was talking about caching it for a timeframe. Ideally, this would cache for, say, one hour, and then invalidate upon a new donation. Do you feel like that's too much for this PR?

kjohnson:
Invalidating the cache on a new donate would work. The trade off is complexity vs benefit. This would introduce a dependency of the Cache and an integration with the Donation process. Being that the query is so light, I opted for simplicity. That said, im happy to implement it if requested.

henryholtgeerts (approved):
@kjohnson @JasonTheAdams Looking over the PR and thinking it through this morning, I feel like Kyle's approach here makes the most sense. I was also wondering how we could use Give_Cache here to store some data until a new donation comes through, but that solution runs into challenges when you consider that Multi-Form Goals aren't really persisted anywhere outside the page's markup, and so any checks to invalidate a cache would happen on page load anyway and be likely be resource heavy unto themselves. 

For that reason, I think Kyle's solution works well, and we can re-evaluate if we run into big problems during testing or after release.

Issue #5399, opened 2020-10-22
Reports should correctly display "All Time" revenue without having to back up a day.


kjohnson:
## Details

> When I click on Reports, and then switch the data to "All Time" the number is different than if I then back up the start date by one day. @Benunc 

## Expected Behavior

Switching data to "All Time" should not require manually modifying the date parameters

## Steps to Reproduce

Assuming a fresh install:

1. Create a donation
1. Go to Donations -> Reports
1. Click "All Time"

The donation is not included in report data.

## Visuals

![image](https://user-images.githubusercontent.com/10858303/96923997-08be7100-1480-11eb-9a1c-0f8e4060a11a.png)

Donation is not included in the report data.

![image](https://user-images.githubusercontent.com/10858303/96923922-f3e1dd80-147f-11eb-94b7-3931fd07d968.png)

## Acceptance Criteria

-   [ ] Clicking "All Time" includes all donations.

kjohnson:
It looks like the `sanitizeDate` is appending a time (`24:00:00`) to **both** the start date and the end date, there by excluding the first day's donations.

PR #5401, opened 2020-10-22
Renewal payments are now included in reports.

### LINKED ISSUE #5398: Renewals should appear in the donation report
## User Story

## Steps to Reproduce

1. Go to...
2. Select...
3. Scroll down to...
4. View error

## System Information

</details>

## Acceptance Criteria

-   [ ] Renewal donations are included in reports data.
-   [ ] Renewal donations are filtered by `mode` (live or test).

kjohnson:
Resolves #5398

## Description

Renewal payments weren't being included in the results of the `Give_Payments_Query` because they do not include a value for the `_give_payment_mode` meta.

Ideally, the renewal system would be updated to include this meta key/value, but that would require a database update to correct historical data.

Instead, I opted to add an additional query for renewal payments and combine the results.

## Affects

Updates the shared `Endpoint::getPayments()` to combine one-time and renewal payments.

## Pre-review Checklist

-   [x] Changelog updated

## Testing Instructions

1. Create a new donation form
1. Create a donation
1. Manually create a renewal
1. See Donations > Reports

Note: Due to #5399, you may need to manually adjust the start date to include first day's payments (for a fresh install).

The renewal payment is now listed in the reports.


kevinwhoffman:
> Ideally, the renewal system would be updated to include this meta key/value, but that would require a database update to correct historical data. Instead, I opted to add an additional query for renewal payments and combine the results.

@kjohnson @JasonTheAdams  I understand the rationale behind the solution in this PR, but I want to check on a couple of things before merge...

Given that:

- We are migrating revenue from all donations (including renewals) to the revenue table in 2.9.0
- AND We bet on a pitch that will update the reporting screen to be powered by the revenue table in the next cycle.

I have the following questions:

- Would renewal payments be automatically included once reporting is powered by the revenue table, thus solving the problem in which renewal payments are currently excluded?
- Should we be distinguishing whether revenue is one-time ore recurring in the revenue table to make aggregate reporting more performant? I failed to ask about this in the reporting pitch but it's worth thinking about now.

kjohnson:
To support renewals from the revenue table we will either need to back fill payment mode for renewal payments or introduce another layer of joins (to include the payment mode from the parent payment).

That is something to consider and would like to discuss with @JasonTheAdams.

JasonTheAdams:
Are renewals a unique post type? Do they also store their amounts and such in the donationmeta table using the same keys?

Also, what's the "payment mode"? Is that the payment gateway?

kjohnson:
@JasonTheAdams renewals use the same `give_payment` post type, but leverage the post status (`give_subscription` instead of `publish`) to differentiate.

Also, `payment mode` is `live` or `test`, which is marked as meta on the `give_payment`, but renewals don't seem to have that meta set. Presumably because the mode is inherited from the parent, but `WP_Query::meta_query` doesn't support meta lookups on the parent post.

kevinwhoffman:
@JasonTheAdams Can you bring this to a resolution based on Kyle's answer?

kjohnson:
Closing in favor of #5409

PR #5403, opened 2020-10-23
Fix PayPal Donation webhooks in live mode

### LINKED ISSUE #5402: PayPal Donation webhooks should work in live mode
## Details

After some testing, it was brought to light that webhooks are working for sandbox mode, but not live mode. This is obviously a problem and must be fixed before launch.

## Expected Behavior

Webhooks should work in live mode.

## Steps to Reproduce

These steps must be done on a site that's reachable by PayPal (not local).

1. Make a one-time donation
2. Pull up the transaction for the donation within PayPal
3. Refund the transaction
4. Check the donation in GiveWP — it should be marked as refunded

## Acceptance Criteria

-   [x] Webhooks work in sandbox (test) mode
-   [x] Webhooks work in live mode

JasonTheAdams:
Resolves #5402 

## Description

So a while back I found that there was an error in the [PayPal Event signature verification docs](https://developer.paypal.com/docs/api/webhooks/v1/#verify-webhook-signature-post-request-body). It says to grab the values from the headers. The `auth_algo` parameter, for example, comes from the `PAYPAL-AUTH-ALGO` header in the event request. I pointed out that the header is not actually in all caps, it's in studly case: `Paypal-Auth-Algo`.

I turns out I was half right. The headers are in studly case _in sandbox mode_, but apparently they are in all caps _in live mode_. In other words, the header shape sent by PayPal is inconsistent. This is not ideal, but it is the way it is.

This PR checks for both shapes to handle either situation.

## Affects

PayPal Donations webhook event handling

## Pre-review Checklist

-   [x] Changelog updated

## Testing Instructions

These steps must be done on a site that's reachable by PayPal (not local).

1. Make a one-time donation
2. Pull up the transaction for the donation within PayPal
3. Refund the transaction
4. Wait a few minutes to give the webhook time
5. Check the donation in GiveWP — it should be marked as refunded


JasonTheAdams:
@ravinderk @kevinwhoffman I've tested both sandbox and live mode on my non-profit website by making a donation and then refunding it from PayPal. In both cases the webhook went through and the corresponding donation was marked as refunded in GiveWP.

Ravinder, it would be great if you could test it again on your sandbox server, just to be safe. Otherwise, please do a code review.

src/PaymentGateways/PayPalCommerce/DataTransferObjects/PayPalWebhookHeaders.php
```diff
@@ -0,0 +1,91 @@
...
+		];
+
+		$payPalHeaders = new self();
+		$missingKeys   = [];
+		foreach ( $headerKeys as $property => $key ) {
+			if ( ! isset( $headers[ $key ] ) ) {
+				$key = strtoupper( $key );
+			}
+
+			if ( isset( $headers[ $key ] ) ) {
```
kevinwhoffman, line 67: This seems like it should be the else statement for the condition on line 63, since it is the opposite of that condition.

kevinwhoffman (approved):
@JasonTheAdams I read through the code and left one question. Otherwise approved. 👍

src/PaymentGateways/PayPalCommerce/DataTransferObjects/PayPalWebhookHeaders.php
JasonTheAdams: Hah! Yeah, I can see why it looks like that. The reason it's not the case is because `$key` is changing within the first condition. That is, if the `$key` is not found then change it to all uppercase and try again.

PR #5409, opened 2020-10-27
Renewal payments are now included in reports (the subsequent pull request)

### LINKED ISSUE #5398: Renewals should appear in the donation report

kjohnson:
Resolves #5398

## Description

Renewal payments weren't being included in the results of the `Give_Payments_Query` because they do not include a value for the `_give_payment_mode` meta.

Ideally, the renewal system would be updated to include this meta key/value, but that would require a database update to correct historical data.

Instead, I opted to add an additional query for renewal payments and combine the results.

## Affects

Updates the shared `Endpoint::getPayments()` to combine one-time and renewal payments.

## Visuals

image

## Pre-review Checklist

-   [x] Changelog updated

## Testing Instructions

1. Create a new donation form
1. Create a donation
1. Manually create a renewal
1. See Donations > Reports

Note: Due to #5399, you may need to manually adjust the start date to include first day's payments (for a fresh install).

The renewal payment is now listed in the reports.


src/API/Endpoints/Reports/Endpoint.php
```diff
@@ -410,13 +410,70 @@ public function getPayments( $startStr, $endStr, $orderBy = 'date', $number = -
...
+				( $this->testMode ) ? 'test' : 'live',
+				$this->currency
+			)
+		);
+		return array_map(
+			function( $renewalPaymentID ) {
+				return new Give_Payment( $renewalPaymentID );
+			},
+			$renewalPaymentIDs
+		);
```
JasonTheAdams, line 474: I believe this is going to cause an N + 1 problem wherein you are querying all of the payment IDs together, but then querying each payment object individually, therefore for every ID you have 1 additional query. Couldn't you instead query `renewals.*` and then generate the payment (or at least the `WP_Post`) from the data? That way it's all done in a single query.

src/API/Endpoints/Reports/Endpoint.php
```diff
@@ -410,13 +410,70 @@ public function getPayments( $startStr, $endStr, $orderBy = 'date', $number = -
...
+		/**
+		 * Query renewal payments based on parent payments.
+		 * Because the renewal does not have a payment mode stored
+		 * the results are not included in the payment query,
+		 * so we have to run a second query and combine results.
+		*/
+		$payments = array_merge(
+			$payments,
+			$this->getRenewalPayments( $startStr, $endStr, $gateway )
+		);
```
JasonTheAdams, line 422: I would imagine both queries should be sorted by `DATE DESC`, which case after merging these we would need to do a `usort()` to put them in order.

JasonTheAdams (changes requested):
This approach seems solid. I have a couple notes, however, on what I think could be improved. Open to any feedback, however, if I'm misunderstanding anything.

src/API/Endpoints/Reports/Endpoint.php
kjohnson: Ah, yup. I see the nested `get_post()` inside the `Give_Payment` definition.

kjohnson: @JasonTheAdams the query is quickly getting out of hand, given all the granular data points that the reporting page uses for each payment. I may need to change course and query all donationmeta and then sort in PHP, as opposed to trying to join each of these different meta items in a single query.

kevinwhoffman:
@JasonTheAdams @kjohnson We're close to the 2.9.0 release, and it sounds like we're still figuring out the best approach here, so I'm going to suggest we target the `develop` branch and address this fix in a patch release.

kevinwhoffman:
Per Slack discussion, I changed the base to `develop` and we will target this fix in a patch release following GiveWP 2.9.0.

kjohnson:
I'm going to open a new PR off of `develop` since this branch has `release/2.9.0` changes.

PR #5410, opened 2020-10-27
Importing donations are not affecting form suggestions anymore

### LINKED ISSUE #5408: Importing donation should not affect form suggestions
## User Story

When importing donations, it seems to have affect on the forms the donations were related to. 
For example, I imported donations of $35, $50, and $100, and now you can see the form has those suggestions.

## Expected Behavior
Importing donations should not affect form suggestions.

## Steps to Reproduce

1. Make a few donations.
2. Export those donations.
3. Delete the donations in GiveWP.
4. Import the donations.
5. View the form on the front end the donations were made in.

## Visuals

image

## Acceptance Criteria

-   [x] Imported donations should not affect form suggestions.

alaca:
Resolves #5408 

## Description

This PR fixes the issue where imported donations were affecting the form amount suggestions.  The problem was in the code logic, where the check for Donation Level texts was made and if the Level Title column from the imported file is different than already defined form levels, it will add imported donation amount as another donation level. The issue is solved by removing the piece of code which was doing that.

## Affects

Donations import.

## Pre-review Checklist

## Testing Instructions

1. Make a few donations.
2. Export those donations.
3. Delete the donations in GiveWP.
4. Import the donations.
5. View the form on the front end the donations were made in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 -   Imported donations now store the correct revenue amount (#5407)
--   Importing donations are not affecting form suggestions anymore (#5410)
+-   Imported donations do not affect form suggestions anymore (#5410)
 
 ## 2.9.0-beta.4 - 2020-10-26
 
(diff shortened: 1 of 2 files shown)


JasonTheAdams:
So apparently when importing donations, it will create or modify the necessary forms. In doing so, it will set up the form donation levels and set the lowest as the default. This seems strange, as it's definitely just guessing as to what the donation levels based on existing donations. I feel like this shouldn't be a part of importing donations.

That said, I don't think this PR needs to deal with that. I just wanted to make a note of it.

JasonTheAdams (approved):
This resolves the issue! Thanks!

PR #5425, opened 2020-10-31
Create a filter for enhancement plugin worth.

meetawahab:
Create a few filters to customize the Give dashboard menu items that will be worthy for developers or/and web Apps CMS.
#hacktoberfest


ravinderk (changes requested):
@meetawahab Thank you for your contribution.

@JasonTheAdams I think the new filter hooks we are adding in this PR to rename menu names are unnecessary and we do not want to merge them.

kjohnson:
@ravinderk I agree with you.

While I see the benefit of adding filters, I would prefer not to make this customizable (aside from translation) to ensure that the product reflects the documentation.

@meetawahab we appreciate you contribution and your interest in GiveWP.

PR #5436, opened 2020-11-05
Fix loader color when using shortcode or block #5435

### LINKED ISSUE #5435: Multi-step form loader color is incorrectly the default green color when embedded
## User Story

As an admin I want the loader color on my embedded donation forms to be the correct color so my brand stays consistent.

## Details

It doesn't matter if the browser is incognito or not... the color is the default when embedded.

## Acceptance Criteria

-   [ ] Embedded forms have the correct color when embedded

DevinWalker:
Resolves #5435

## Description

IFrameView passes the `$formId` now so the loader can properly pull the color from template options.

## Affects

Shortcodes / Embeds

## Pre-review Checklist


JasonTheAdams (approved):
@DevinWalker I made a couple tweaks as I was afraid the way we did it was not at all clear, as someone may remove the variable thinking it was just a leftover.

But it works and is good to go. 👍

PR #5486, opened 2020-12-04
Edit Profile tab now persists data

### LINKED ISSUE #5464: Edit Profile controls should persist changes
## User Story

As a user, I want controls on the Edit Profile tab of the Donor Profiles app to persist changes I make so that I can update profile settings.

## Details
Previously, work was done to introduce UI controls for the Edit Profile tab. Currently, these controls do not display or persist actual donor information. This PR should introduce logic via a REST endpoint that allows logged in donors to securely make changes to their donor information from the Donor Profile app.

## Acceptance Criteria
-   [x] Core Edit Profile fields load existing donor information
-   [x] Core Edit Profile fields can be used to persist changes to donor information
-   [x] Donor Profile settings can only be saved by logged in donors

henryholtgeerts:
Resolves #5464 

## Description
This PR introduces logic to persist changes made via the "Edit Profile" tab of the new Donor Profile area. Building on UI implementation work of a previous PR, this PR makes some changes to frontend logic but is primarily focused on implementation of the new Profile REST endpoint which interacts with frontend fields.

## Affects
This PR affects frontend logic of the Edit Profile tab, introducing UI support for adding/removing additional emails and addresses, as well as a workable dropzone for Avatar controls.

On the backend, the PR introduces the ProfileRoute and builds on the Profile model, so that it adequately supports all necessary changes a user might make to their donor profile.

## Pre-review Checklist


src/DonorProfiles/Helpers/FormatList.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+namespace Give\DonorProfiles\Helpers;
+
+/**
+ * Formats an associative array into a JS parsable array of objects.
+ *
+ * @since 2.8.0
+ */
+class FormatList {
```
kjohnson, line 10: If no changes have been made, then we can probably extract this for re-use - I'll look into that.

src/DonorProfiles/Helpers/SanitizeProfileData.php
```diff
@@ -0,0 +1,71 @@
...
+	 * Sanitize array of addresses passed
+	 *
+	 * @param array $addresses
+	 *
+	 * @return array
+	 *
+	 * @since 2.11.0
+	 */
+	public static function sanitizeAdditionalAddresses( $addresses ) {
+		foreach ( $addresses as $key => $value ) {
```
kjohnson, line 34: Since this wraps `sanitizeAddress`, we could probably map over it, instead of using a loop.

```php
return array_map( $this->sanitizeAddress, $addresses );
```

src/DonorProfiles/Profile.php
```diff
@@ -0,0 +1,279 @@
...
+use \Give_Donor as DonorModel;
+
+class Profile {
+
+	protected $donor;
+	protected $id;
+
+	public function __construct( int $donorId ) {
+		$this->id    = $donorId;
+		$this->donor = new DonorModel( $donorId );
```
kjohnson, line 14: I would suggest creating a factory that accepts a `$donorId` and then inject the `DonorModel` instance via the constructor, as opposed to creating new objects here.

src/DonorProfiles/Profile.php
```diff
@@ -0,0 +1,279 @@
...
+	 * @param object $data Object representing profile data to update
+	 *
+	 * @return array
+	 *
+	 * @since 2.11.0
+	 */
+	public function update( $data ) {
+
+		// Handle updates to the donor meta table
+		$this->updateDonorMetaDB( $data );
```
kjohnson, line 29: Since a number of things are happening here in turn, I would suggest refactoring this out using a "pipeline" pattern, which accepts multiple process classes (each handling a specific part of the update, thus a single responsibility ).

For example:
```php
( new ProfileUpdatePipeline( $data ) )
    ->update( new UpdateDonorMeta )
    ->update( new UpdateAdditionalEmails )
    ->update( new UpdateDonor )
    ->update( new UpdateDonorAddressMeta )
;
```

kjohnson: This would also simplify the responsibility of the Profile class, which can focus on formatting and mapping Donor data.

src/DonorProfiles/Repositories/Donations.php
```diff
@@ -196,7 +196,7 @@ protected function getPaymentInfo( $payment ) {
 		return [
 			'amount'   => $payment->subtotal,
 			'currency' => $payment->currency,
-			'fee'      => $payment->fees,
+			'fee'      => ( $payment->total - $payment->subtotal ),
```
kjohnson, line 199: For my edification, why are you favouring a calculation over an existing property? I don't know the benefit/requirement here.

src/DonorProfiles/Routes/ProfileRoute.php
```diff
@@ -0,0 +1,144 @@
...
+	 * @return object
+	 *
+	 * @since 2.11.0
+	 */
+	public function sanitizeData( $value, $request, $param ) {
+
+		$sanitizeHelper = '\Give\DonorProfiles\Helpers\SanitizeProfileData';
+		$sanitizedValue = json_decode( $value );
+
+		$attributesMap = [
```
kjohnson, line 63: Since so many have unique sanitizing strategies., would it be simpler here to procedurally sanitize each key/value? I think the attribute mapping requires a greater mental overhead in this case.

kjohnson: This would also reduce the `[ class, method ]` array syntax in favour of an object-oriented syntax, which can be highlighted.

src/DonorProfiles/resources/js/app/components/text-control/index.js
```diff
@@ -12,7 +12,7 @@ const TextControl = ( { label, value, onChange, icon, type } ) => {
 					<FontAwesomeIcon icon={ icon } />
 				) }
 				<input
-					id={ label && toKebabCase( label ) }
+					id={ label && toKebabCase( `${ label }-${ Math.floor( Math.random() * Math.floor( 1000 ) ) }` ) }
```
kjohnson, line 15: This logic is duplicated in `select-control`, if these lines need to change for the same reason then they should probably be extracted to a helper method. Otherwise, if each may change independently of the other then the duplicate is the preferred method.

kjohnson (changes requested):

kjohnson:
@henryholtgeerts I just dropped a lot of feedback, but it is a big change 😄 

I made some suggestions that I think will improve maintenance.

src/DonorProfiles/Helpers/FormatList.php
kjohnson: @henryholtgeerts https://packagist.org/packages/kjohnson/format-object-list

includes/class-give-donor.php
```diff
@@ -1238,6 +1238,9 @@ public function add_email( $email = '', $primary = false ) {
 	 * @return bool          If the email was removed successfully.
 	 */
 	public function remove_email( $email = '' ) {
+
+		error_log( $email );
```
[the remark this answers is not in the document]
henryholtgeerts: Done, ref: https://github.com/impress-org/givewp/pull/5486/commits/4fde99d90f1f1dd3d8e72870cfe846240f128cd5

src/DonorProfiles/Helpers/FormatList.php
henryholtgeerts: Thanks for making that package available! Done, ref: https://github.com/impress-org/givewp/pull/5486/commits/1e56795ca28d972bcf2539e48bcae9c6b8626ee5

src/DonorProfiles/Helpers/SanitizeProfileData.php
henryholtgeerts: Done, ref: https://github.com/impress-org/givewp/pull/5486/commits/7c6f3e6f97dacf4928535707691bc239c5bc8903

src/DonorProfiles/Helpers/LocationList.php
```diff
@@ -11,12 +13,12 @@ class LocationList {
 	public static function getCountries() {
 		$countries = give_get_country_list();
 		unset( $countries[''] );
-		return FormatList::fromKeyValue( $countries );
+		return FormatObjectList\Factory::fromKeyValue( $countries );
```
kjohnson, line 16: @henryholtgeerts quick note: I updated the implementation a little, so you'll need to call `->format()` to get the formatted list. The `Factory` returns an instance of the formatter class.

src/DonorProfiles/Profile.php
henryholtgeerts: Great thinking, really cleared up the Profile class! Done, ref: https://github.com/impress-org/givewp/pull/5486/commits/d1793daf8e0073de0a31fe3f989d7d04eef1f429

src/DonorProfiles/Helpers/LocationList.php
henryholtgeerts: Ah - I see! Tried that implementation first but IDE was yelling at me because the doc blocks for FormatObjectList/Factory indicate that it would return an array (not an object). Could you update doc blocks your package real quick to reflect this and I'll make the adjustment here as well?

src/DonorProfiles/Profile.php
henryholtgeerts: Done, ref: https://github.com/impress-org/givewp/pull/5486/commits/fb9d2b8aa942807130229eae836108dc3715dd69

src/DonorProfiles/Repositories/Donations.php
henryholtgeerts: I'm performing a calculation here because, despite being documented, a `fees` property is not actually accessible on the Give_Payment class. A quick search for `->fees` or even `fees` reveals that this logic is not centralized in the codebase as I was originally led to believe.

src/DonorProfiles/Routes/ProfileRoute.php
henryholtgeerts: This definitely cleared up the function some -- I think my initial hesitancy about a procedural approach related to concerns about setting defaults. But after implementing once, I realize that the default are predictable and I can just set them inside the sanitization functions. Done, ref: https://github.com/impress-org/givewp/pull/5486/commits/66b56f62ca81580ab279ca8df7afa086af7bd1be

src/DonorProfiles/resources/js/app/components/text-control/index.js
henryholtgeerts: Good thought -- I've gone ahead and extracted this into a `toUniqueId` utility. Ref: https://github.com/impress-org/givewp/pull/5486/commits/91a8870d8841d21c0db9a0907a0ac3deb9e70ae4

src/DonorProfiles/Helpers/LocationList.php
henryholtgeerts: I've corrected the implementation here, but still suggest updating the doc blocks in your package to correctly reflect the values returned by Factory methods. Ref: https://github.com/impress-org/givewp/pull/5486/commits/5e7f868d222d1c107e1bf0d37d3bd0da7278e51c

kjohnson (changes requested):
For the refactor to the `FormatObjectList` package, I'd rather separate the changes to `Onboarding` into a separate PR, as opposed to intermingling with the Donor Profile.

Also, the `FormatListTest` is now redundant, as the package is tested.

henryholtgeerts:
@kjohnson Sounds, good - I'll revert the changes to the Onboarding domain and revert the test back to its prior implementation (because it is still relevant as long as Onboarding is not reliant on the package -- even if that is in the works)!

kjohnson (approved):
That'll do it - good to go!

PR #5532, opened 2020-12-27
Hide state field for South Korea

### LINKED ISSUE #5536: state field not required for South Korea
## Details

state field is not required for korean addresses.

jgreys:
Resolves #5536

## Description

The status field is not required and not used for Korean addresses. In order to fix this, the "hidden" parameter should be set to "True" when South Korea is selected as country.

## Pre-review Checklist

## Testing Instructions

1. Select "South Korea" as country.
2. State field should be hidden.

### DIFF
--- a/includes/country-functions.php
+++ b/includes/country-functions.php
@@ -688,6 +688,7 @@ function give_get_country_locale() {
 			'KR' => [
 				'state' => [
 					'required' => false,
+					'hidden'   => true,
 				],
 			],
 			'KW' => [


ravinderk:
@jgreys
- Can you confirm that South Korea has `Province` instead of state?
- Do you have to mention the `Province` in the address?
   if yes that we can rename the `state` field label `State` to `Province` to resolve the confusion as we are doing in the `give_get_country_locale` function for other countries. For example `CA`

ravinderk (changes requested):
@jgreys
- Can you confirm that South Korea has `Province` instead of state?
- Do you have to mention the `Province` in the address?
   if yes that we can rename the `state` field label `State` to `Province` to resolve the confusion as we are doing in the `give_get_country_locale` function for other countries. For example `CA`

ravinderk (approved):
@DevinWalker @kjohnson I reviewed the Woocommerce plugin and find out that the state field is not necessary when the donor country set to `South Korea`. Changes in this pull request look good to me but I am open to a second opinion.

image

jgreys:
I'm living in Korea for a little bit more than 7 years now and I've rarely seen the "Province" field in Korean websites.

Koreans use an address autocompletion system similar to what Google offers but based on the postal code. (http://postcode.map.daum.net/guide)

I integrated this solution to WooCommerce with my plugin "Korea for WooCommerce" (https://github.com/jgreys/korea-for-woocommerce). Also, I'm working on a similar solution for Give these days. :)

Lastly, here are 2 screenshots of extremely popular sites in Korea (Naver & CJmall) that support my point. Naver is like the Korean Google here.

Screenshot 2021-01-05 at 16 56 18

Screenshot 2021-01-05 at 17 03 48

PR #5543, opened 2021-01-07
Sample onboarding tests are now implemented

henryholtgeerts:
Resolves # N/A

## Description
This PR introduces some onboarding e2e tests (and additional supporting test functions), to demonstrate e2e testing implementation. Notably, it adds specs for testing the setup page and onboarding wizard, as well as support for a `data-givewp-test` attribute used to target specific elements for testing. To query those targets, tests can make use of the `cy.getTest( "id" )` command. 

## Affects
This PR affects some frontend rendering logic for the setup page and onboarding wizard, and introducing new testing specs to the tests/e2e/onboarding directory.

## Pre-review Checklist

## Testing Instructions

1. Install and setup all necessary e2e testing dependencies
2. Use `npm run env start` to start the wp-env for testing
3. Use `npm run test:e2e` to open Cypress tests
4. Select any spec under onboarding, check that it passes
5. Close Cypress and use `npm run env stop` to stop the wp-env when finished testing

### DIFF
--- a/assets/src/js/admin/onboarding-wizard/components/button/index.js
+++ b/assets/src/js/admin/onboarding-wizard/components/button/index.js
@@ -4,9 +4,9 @@ import PropTypes from 'prop-types';
 // Import styles
 import './style.scss';
 
-const Button = ( { onClick, children } ) => {
+const Button = ( { onClick, testId, children } ) => {
 	return (
-		<button className="give-obw-button" onClick={ onClick }>
+		<button className="give-obw-button" data-givewp-test={ testId } onClick={ onClick }>
 			{ children }
 		</button>
 	);
@@ -14,11 +14,13 @@ const Button = ( { onClick, children } ) => {
 
 Button.propTypes = {
 	onClick: PropTypes.func,
+	testId: PropTypes.string,
 	children: PropTypes.node,
 };
 
 Button.defaultProps = {
 	onClick: null,
+	testId: null,
 	children: null,
 };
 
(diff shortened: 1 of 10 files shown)


assets/src/js/admin/onboarding-wizard/components/button/index.js
```diff
@@ -4,21 +4,23 @@ import PropTypes from 'prop-types';
 // Import styles
 import './style.scss';
 
-const Button = ( { onClick, children } ) => {
+const Button = ( { onClick, testId, children } ) => {
```
kjohnson, line 7: @henryholtgeerts Instead of modifying the component itself, what if we use composition to wrap testable components (which could add the test attribute)?

henryholtgeerts: Do you mean to introduce a testable version of components (eg <TestableButton/>) or a generic <Test/> that would wrap the inputs we aim to test? The specific direction I took was for simplicity and to mirror the GIveIO implementation (see their [Checkbox](https://github.com/impress-org/giveio/blob/develop/resources/js/components/ControlCheckBox/index.jsx) component as an example).

kjohnson: Yeah... hm.

I was hoping for something like

```
<Testable id="test1">
    <MyComponent />
</Testable>
```

But that doesn't seem to work the way that I had hoped.

henryholtgeerts: Yeah, I was just now tryina work out something along those lines locally. Ultimately seems simplest to do the directly applied attribute approach - though I definitely see the appeal in composition here.

kjohnson: I think we are on the same page.

PR #5554, opened 2021-01-12
Unit test structure is refactored

### LINKED ISSUE #5552: Unit test structure should be refactored
## Details

Automated tests should be re-organized to clearly define the different types of tests: unit, integration, e2e.

kjohnson:
Resolves #5552

## Description

Combines WP unit and PHP unit tests as `unit` tests. WP unit tests blur the lines of unit and integration, but this is a compromise between _purity_ and practicality.

## Pre-review Checklist


JasonTheAdams (commented):
Looks good, @kjohnson. We'll probably want some kind of `RefreshDatabase` trait that can be applied to a test Class to start a transaction before each test and then rollback after each test. Otherwise, the database will bloat (and likely make conflicts) as more tests run.

.github/workflows/wordpress.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         php: [ 5.6, 7.2 ]
-        wordpress: [ '4.9', '5.0', '5.1', '5.2', '5.3', '5.4', '5.5', latest ]
+        wordpress: [ 4.9, 5.6, latest ]
```
JasonTheAdams, line 25: We can probably just do:
```suggestion
        wordpress: [ 4.9, latest ]
```

kjohnson: Ah, sure, because `latest` would cover the 5* range 👍🏻

kjohnson:
@JasonTheAdams Re: `RefreshDatabase`: It looks like `WP_UnitTestCase` (via `WP_UnitTestCase_Base`) already handles the database cleanup, see [WP_UnitTestCase_Base#145](https://core.trac.wordpress.org/browser/trunk/tests/phpunit/includes/abstract-testcase.php#L145).

```php
/**
 * After a test method runs, reset any state in WordPress the test method might have changed.
 */
public function tearDown() {
    global $wpdb, $wp_query, $wp;
    $wpdb->query( 'ROLLBACK' );
    // ...
}
```

.github/workflows/wordpress.yml
kjohnson: Resolved via 7d767ec

JasonTheAdams:
@kjohnson Well looky there! Today I learned WordPress already does that.
Ah shucks

.github/workflows/wordpress.yml
```diff
@@ -54,7 +54,7 @@ jobs:
 
       - name: Setup tests
         run: |
-            bash tests/bin/install.sh wordpress_test root root localhost ${{ matrix.wordpress }}
+            bash tests/unit/bin/install.sh wordpress_test root root localhost ${{ matrix.wordpress }}
 
       -   name: Execute tests
           run: vendor/bin/phpunit
```
JasonTheAdams, line 60: @kjohnson Just because I happened to run into this on GiveIO, I suggest the following:
```suggestion
          run: php -d memory_limit=-1 vendor/bin/phpunit
```

kjohnson: Sounds good! Suggestion merged.

For reference: https://github.com/impress-org/giveio/commit/b401c1a4dc794877a4e3362b544d06408fc5fed0

JasonTheAdams (approved):
I made a couple changes to the `install.sh` file to allow for spaces in the socket path, but I was able to get everything set up and successfully ran all of the tests. 👍

JasonTheAdams:
Hmm... something isn't working right with the tests on Github Actions. Don't merge.

PR #5577, opened 2021-02-01
Donor Profiles are now extensible

### LINKED ISSUE #5434: Donor Profiles should be extensible
## User Story

As a developer, I want the Donor Profile area to be extensible so that addons can easily add relevant tabs and controls.

## Details
The Donor Profile frontend app needs to easily extended both within GiveWP core and by addons. To achieve this, two abstract classes should be created, which other addons can use to help in registering new Donor Profile tabs and endpoints. Additional donor profile tabs should include logic for enqueuing scripts used to render the tab content. One source of inspiration for this API could be the block registration API, where blocks are registered both in the backend and through JS.

## Additional Context
Ultimately, donor Profile tabs also need to be surfaced in the Donor Profile block and shortcode, where admin can selectively enable or disable certain tabs. This functionality will be addressed in a future PR and is not a concern currently.

## Acceptance Criteria
-   [x] New Donor Profile tabs can be easily registered via PHP
-   [x] New Donor Profile endpoints can be easily registered via PHP
-   [x] New Donor Profile tabs and tab content can be easily registered via JS
-   [x] Addon informed Donor Profile tab routes should be removed from core
-   [x] Core Donor Profile tabs register through the same Tab register that addons do

henryholtgeerts:
Resolves #5434 

## Description

This PR introduces Tab registration logic needed to make the Donor Profiles app extensible by addons. Specifically, it introduces the Tabs register, as well as Tab and Route abstracts to help addons easily extend the Donor Profiles app. 

## Affects
In addition to introducing the Tabs register, this PR refactors existing core Tab logic so that all tabs are registered through the Tab register consistently. 

## Pre-review Checklist

## Testing Instructions

1. Add a new Donor Profile block to an empty page
2. Check that all tabs load as expected


src/DonorProfiles/Tabs/DonationHistoryTab/DonationsRoute.php
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	/**
+	 * @return array
+	 *
+	 * @since 2.10.0
+	 */
+	protected function getData() {
+
+		$repository = new DonationsRepository();
```
kjohnson, line 41: This could be injected as a dependency and resolved from the container. One benefit would be to mock the repository for unit testing the return format of the `getData()` method.

src/DonorProfiles/Tabs/TabsRegister.php
```diff
@@ -0,0 +1,117 @@
...
+	 * @param string $id
+	 *
+	 * @return string
+	 */
+	public function getTab( $id ) {
+		if ( ! isset( $this->tabs[ $id ] ) ) {
+			throw new \InvalidArgumentException( "No migration exists with the ID {$id}" );
+		}
+
+		return $this->migrations[ $id ];
```
kjohnson, line 56: I'm assuming that this would be returning an item from the `$this->tabs` property?

src/DonorProfiles/Tabs/TabsRegister.php
```diff
@@ -0,0 +1,117 @@
...
+
+	/**
+	 * Add a tab to the list of tabs
+	 *
+	 * @since 2.11.0
+	 *
+	 * @param string $tabClass FQCN of the Tab Class
+	 */
+	public function addTab( $tabClass ) {
+		if ( ! is_subclass_of( $tabClass, Tab::class ) ) {
```
kjohnson, line 78: I would suggest type hinting the method parameter and let PHP enforce the sub-classing contract.

src/DonorProfiles/Tabs/TabsRegister.php
```diff
@@ -0,0 +1,117 @@
...
+			throw new \InvalidArgumentException( 'Class must extend the ' . Tab::class . ' class' );
+		}
+
+		$tabId = $tabClass::id();
+
+		if ( isset( $this->tabs[ $tabId ] ) ) {
```
kjohnson, line 84: I would suggest re-using the `hasTab` method here, instead of duplicating the `isset` logic.

src/DonorProfiles/Tabs/TabsRegister.php
```diff
@@ -0,0 +1,117 @@
...
+
+	public function registerTabRoutes() {
+		foreach ( give()->donorProfileTabs->tabs as $tabClass ) {
+			$tab = new $tabClass;
+			$tab->registerRoutes();
+		}
+	}
+
+	public function enqueueTabAssets() {
+		foreach ( $this->tabs as $tabClass ) {
```
kjohnson, line 113: It looks like `registerTabRoutes` and `enqueueTabAssets` are doing something similar, in that they are iterating over the same tabs, but the former is referencing off of a global, while the later is referencing a class property. Is there a need for this difference, or can the two be more consistent?

kjohnson (changes requested):
A couple of small changes and a couple of comments to consider.

kjohnson (changes requested):
A couple of small changes and a couple of comments to consider.

src/DonorProfiles/Tabs/TabsRegister.php
henryholtgeerts: Ah, good catch! Yes - meant to be returning from the tabs property. Done, ref: https://github.com/impress-org/givewp/pull/5577/commits/56179f86fb9de23248a6bf0f5b2d68f8aca4e554

src/DonorProfiles/Tabs/TabsRegister.php
henryholtgeerts: Here, the addTab method is actually expecting a string, since what is being stored in the tabs property is just an array of fully qualified class names (as strings). This avoids constructing the classes themselves until they are needed. The manual check is required since we're really checking if the FQCN provided represents a class that does in fact extend the Tab class.

src/DonorProfiles/Tabs/TabsRegister.php
henryholtgeerts: Good thinking! Done, ref: https://github.com/impress-org/givewp/pull/5577/commits/8b485cfb89418896ff7c91c70fb6574e2c83ae63

src/DonorProfiles/Tabs/TabsRegister.php
henryholtgeerts: Good catch here -- they should both be referencing off of the global. Done, ref: https://github.com/impress-org/givewp/pull/5577/commits/cef17075fb33483fee71fdaf5c6052b6335c46f5

src/DonorProfiles/Tabs/DonationHistoryTab/DonationsRoute.php
henryholtgeerts: I've gone ahead and changed this so that the DonationsRepository is injected as a dependency. I'm holding off on registering the repository as a global, since for the time being this is the only place it is referenced. Done, ref: https://github.com/impress-org/givewp/pull/5577/commits/4677b3f9d697f820499e91800c719f076ebac52e

src/DonorProfiles/Tabs/TabsRegister.php
```diff
@@ -49,11 +49,11 @@ public function hasTab( $id ) {
 	 * @return string
 	 */
 	public function getTab( $id ) {
-		if ( ! isset( $this->tabs[ $id ] ) ) {
-			throw new \InvalidArgumentException( "No migration exists with the ID {$id}" );
+		if ( ! $this->hasTab( $id ) ) {
+			throw new \InvalidArgumentException( "No tab exists with the ID {$id}" );
```
kjohnson, line 53: There is probably a better exception that we can throw, or throw a custom exception.

`InvalidArgumentException` is thrown "if an argument is not of the expected type."

src/DonorProfiles/Tabs/TabsRegister.php
```diff
@@ -81,7 +81,7 @@ public function addTab( $tabClass ) {
 
 		$tabId = $tabClass::id();
 
-		if ( isset( $this->tabs[ $tabId ] ) ) {
+		if ( $this->hasTab( $tabId ) ) {
 			throw new \InvalidArgumentException( 'A tab can only be added once. Make sure there are not id conflicts.' );
```
kjohnson, line 85: Likewise, we should throw a different exception type here. The above use of `InvalidArgumentException` in `addTab()` makes sense, but on line 84-85 the argument is correct and valid (type) but there is a conflict.

Also, throwing a custom exception is easy to document for what to catch.

kjohnson (changes requested):

src/DonorProfiles/Tabs/DonationHistoryTab/DonationsRoute.php
kjohnson: To clarify, I was suggesting that `DonorRepository` be resolved from the container via the class constructor, which eliminates the need for manually instantiating a class.

src/DonorProfiles/Tabs/TabsRegister.php
kjohnson: Sounds good!

src/DonorProfiles/Tabs/TabsRegister.php
henryholtgeerts: Done, ref: https://github.com/impress-org/givewp/pull/5577/commits/fdaa13fd9a0dd47ee65e037b271fa2e7e60caa8c

src/DonorProfiles/Tabs/TabsRegister.php
henryholtgeerts: Done, ref: https://github.com/impress-org/givewp/pull/5577/commits/fdaa13fd9a0dd47ee65e037b271fa2e7e60caa8c

kjohnson (approved):

PR #5591, opened 2021-02-10
Logs UI is now implemented

alaca:
Resolves #5583

## Description

This PR implements the new Logs UI in the Give Admin.

## Affects

Logs UI screen;

## Visuals

image

## Pre-review Checklist

-   [x] Acceptance criteria satisfied and marked in related issue
-   [x] Relevant `@since` tags included in DocBlocks
-   [x] Changes logged to the `Unreleased` section of `CHANGELOG.md`
-   [x] [Self Review](https://give.gitbook.io/development-manual/devops/github/code-reviews#self-review) of code and UX completed

## Testing Instructions

**Navigate to**

`Donations -> Tools -> Logs`

If you don't have any logs in the database, you can use this snippet to add them.

```php
add_action( 'admin_init', function() {
  if ( ! get_option( 'insert_test_logs' ) ) {
    for ( $i = 0; $i <= 100; $i++ ) {
      foreach( [ 'error', 'warning', 'info', 'success' ] as $type ) {
        Log::$type( 'This is an ' . $type, [
	  'source' => 'Test ' . $type,
	  'additional' => $i . $type
        ] );
      }
    }
    
    update_option( 'insert_test_logs', true );
  }
} );
```

PR #5598, opened 2021-02-12
The new Migrations UI is now implemented

### LINKED ISSUE #5578: Migration Log UI should be implemented
## User Story

As a developer, I want the Migration Log UI to be implemented.

## Details
Currently, there is no UI for the new Logging system. There is no option for the site admin to see which migrations have run or to perform actions on failed migrations. 

## Additional Context

The new Migration log UI should show all migration logs in a list table which will be sortable by migration status and date. The UI should also have an option to perform actions on the failed migrations ( re-run migration ).

## Acceptance Criteria

-   [x] Migration Log UI implemented
-   [x] MIgration list table is sortable by `Status` and `Date`
-   [x] MIgration list table pagination implemented
-   [x] Option to re-run failed migrations

alaca:
Resolves #5578

## Description

This PR implements the new Migration UI in the Give Admin.

## Affects
`Donations -> Tools -> Data` Screen

## Visuals
image

## Pre-review Checklist

## Testing Instructions

Go to the `Donations -> Tools -> Data` Screen


henryholtgeerts (approved):
Tested locally, works as expected! Good work here. This PR faces the same minor polish improvements I noted in the logs PR (https://github.com/impress-org/givewp/pull/5591#pullrequestreview-590818481), but I am recommending that polish be addressed in a separate issue/PR, since it is otherwise ready to be sent to the rest of the team for testing. 👍

JasonTheAdams (commented):
The whole "run order" concept is interesting. It's currently being set as a property of the migration itself. If you think about it, a migration wouldn't know it's "run order" as it's relative to the migrations around it. Run order is something derived by a collection of migrations.

As such, instead of storing the run order, when all of the migrations are together and you want to sort them, use the `MigrationRegister::getMigration($migration->getId())` method. You can then use the `Migration::timestamp()` method across all migrations (whether they come from the database or not) in a [usort](https://www.php.net/manual/en/function.usort.php) callback to sort the migrations. Then the index is the numeric Run Order to be used on the front-end.

I could be missing something. Let me know if that's helpful. 😄

alaca:
@JasonTheAdams since we are storing all migrations in the database ( except pending migrations ), the problem of not having a run order position ( if we don't store the timestamp ) will show up if the user activates an add-on that has migrations and then deactivates it. The migration will be stored in the database and it will be shown in the list table. Having run order only in memory could work if we show only currently registered migrations in the Migrations list table. That being said, I'm thinking that hiding migrations that are not in the register might be a good idea.

JasonTheAdams:
Ah, I see what you're saying. Yeah, I'm actually OK with migrations not appearing for deactivated add-ons. I think it's a pretty niche case and we'll see if it comes up as a problem. I doubt it, though.

alaca:
@JasonTheAdams I refactored how migrations are sorted and how they are displayed, and I'm not using the database anymore to sort them. I'm using PHP array functions instead. This method allows me to sort all migration, including pending migrations, by all columns. This is ok I guess as we probably never gonna have more than 100 migrations.

src/MigrationLog/MigrationLogModel.php
```diff
@@ -29,6 +29,11 @@ class MigrationLogModel {
 	 */
 	private $error;
 
+	/**
+	 * @var int
+	 */
+	private $run_order;
+
```
JasonTheAdams, line 36: Is this still used?

JasonTheAdams (approved):
Great work, @alaca! I had one quick question, but in any case merge when ready!

src/MigrationLog/MigrationLogModel.php
alaca: No, that one was left behind. I'll remove that.

alaca: Actually, it's already removed 
14a4c7e

src/API/Endpoints/Migrations/RunMigration.php
```diff
@@ -0,0 +1,129 @@
...
+	public function handleRequest( WP_REST_Request $request ) {
+		global $wpdb;
+		$migrationId  = $request->get_param( 'id' );
+		$migrationLog = $this->migrationLogFactory->make( $migrationId );
+
+		// Begin transaction
+		$wpdb->query( 'START TRANSACTION' );
+
+		try {
+			$migrationClass = $this->migrationRegister->getMigration( $migrationId );
```
kjohnson, line 103: We are assuming here that the migration (by `id`) is going to exist. While this is inside of a try/catch, should we be specifically checking if the migration exists?

alaca: No need to check that as the `getMigration` method will throw an exception anyway if migration doesn't exist.

kjohnson: Cool - just checking 👍🏻

src/MigrationLog/MigrationLogModel.php
JasonTheAdams: Looking at that commit, it looks like you removed it's usage but not the property definition.

alaca: Now its removed for real 7e2013c3

PR #5601, opened 2021-02-12
Field API enforces unique names

### LINKED ISSUE #5594: Fields API should enforce unique names
Each node in the AST should have a unique identifier, which should be enforced when a new node is added to the tree.

kjohnson:
Resolves #5594

## Description

Adds name collision detection to `FieldsAPI\FieldCollection`.

## Affects

Updated `insertBefore()` and `insertAfter()` to first check for name collisions.

## Pre-review Checklist

### DIFF
--- a/src/Framework/FieldsAPI/FieldCollection.php
+++ b/src/Framework/FieldsAPI/FieldCollection.php
@@ -7,6 +7,7 @@ use Give\Framework\FieldsAPI\FieldCollection\Contract\GroupNode;
 
 class FieldCollection implements GroupNode {
 
+    use FieldCollection\NameCollision;
     use FieldCollection\InsertNode;
     use FieldCollection\MoveNode;
     use FieldCollection\RemoveNode;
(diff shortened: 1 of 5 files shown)


JasonTheAdams (approved):
Very well done. I really like how you've broken these all up into testable traits. You're a module monster. 💪 

One observation, but outside of the scope of this PR, is that I noticed that when inserting before or after another node we fail silently if the sibling node isn't found. It simply won't insert the node. I'm thinking we should throw a `NodeNotFound` exception in these occasions. Folks can, of course, catch the exception and do nothing, but we should give devs a way of knowing it didn't work.

kjohnson:
> One observation, but outside of the scope of this PR, is that I noticed that when inserting before or after another node we fail silently if the sibling node isn't found. It simply won't insert the node. I'm thinking we should throw a `NodeNotFound` exception in these occasions. Folks can, of course, catch the exception and do nothing, but we should give devs a way of knowing it didn't work.

@JasonTheAdams yeah, that sounds good. I created a separate issue to address the new exception.

PR #5640, opened 2021-02-23
Fields API throws an exception when inserting relative to a node that does not exist

### LINKED ISSUE #5635: Fields API should throw an exception when inserting relative to a node that does not exist
## Details

> [...] I noticed that when inserting before or after another node we fail silently if the sibling node isn't found. It simply won't insert the node. I'm thinking we should throw a NodeNotFound exception in these occasions. Folks can, of course, catch the exception and do nothing, but we should give devs a way of knowing it didn't work. @JasonTheAdams https://github.com/impress-org/givewp/pull/5601#pullrequestreview-595857157

## Acceptance Criteria

-   [x] When inserting a node relative to another node that does not exist an exception is thrown.

kjohnson:
Resolves #5635

## Description

- Adds the `ReferenceNodeNotFoundException` exception, which is thrown when inserting before or after a node that does not exist.
- Optimizes `InsertNode` methods `insertBefore` and `insertAfter` by avoiding unecessary name collision checks on each nesting level, since the collision check is already recursive.
- Moves `InsertNode` trait tests to a separate `InsertNodeTest` test file.

## Pre-review Checklist

## Testing Instructions

Testing instructions are outlined in `tests/unit/tests/Framework/FieldsAPI/FieldCollection/InsertNodeTest.php`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+-   Fields API throws an exception when inserting relative to a node that does not exist (#5640)
+
 ## 2.9.6 - 2021-01-13
 
 ### New
--- a/src/Framework/FieldsAPI/FieldCollection/Exception/ReferenceNodeNotFoundException.php
+++ b/src/Framework/FieldsAPI/FieldCollection/Exception/ReferenceNodeNotFoundException.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Give\Framework\FieldsAPI\FieldCollection\Exception;
+
+/**
+ * @unreleased
+ */
+class ReferenceNodeNotFoundException extends \Exception {
+	public function __construct( $name, $code = 0, Exception $previous = null ) {
+		$message = "Reference node with the name \"$name\" not found - cannot insert new node.";
+		parent::__construct( $message, $code, $previous );
+	}
+}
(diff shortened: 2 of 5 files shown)


JasonTheAdams (changes requested):
Works exactly as I'd hoped. Couple dev notes:
- In < PHP 7.2 please make sure to use docblocks to indicate parameter and return types
- Can we please rename or do something with the `_insertAfter` methods? The underscore prefix is a convention that's by no means clear. Really, these could be a single function:
```php
	protected function insertAfter( $siblingName, Node $node ) {
		$this->checkNameCollisionDeep( $node );

		$siblingIndex = $this->getNodeIndexByName( $siblingName );
		if ( false !== $siblingIndex ) {
			$this->insertAtIndex(
				$siblingIndex + 1,
				$node
			);
			return $this;
		} elseif ( $this->nodes ) {
			foreach ( $this->nodes as $childNode ) {
				if ( $childNode instanceof GroupNode ) {
					$childNode->_insertAfter( $siblingName, $node );
				}
			}
			return $this;
		}
		throw new ReferenceNodeNotFoundException( $siblingName );
	}
```
I noticed the protected methods were returning either null or `$this->insertAtInex` — which doesn't actually return anything and is therefore also null. So might as well just return `$this`.

kjohnson:
@JasonTheAdams I separated out separate functions to avoid re-checking for collision with each recursive call.

The `_` prefix convention I thought was common for non-public methods, but I updated to a "Recursive" postfix for clarity).

JasonTheAdams:
Ah, that does make sense. Good call on avoiding the re-checking.

As far as the underscore goes, I believe the convention started in JS because there was no actual scope for functions, so it was used to indicate scope in its absence. Still, it's not meant to replace having a descriptive name. Even you changing the name to add `Recursive` made a big difference in helping me understand why you had two functions. 👍

JasonTheAdams (approved):
Great! Thanks for the tweaks! 🚀

PR #5643, opened 2021-02-23
Form Templates now support custom checkbox fields

### LINKED ISSUE #5587: Multi-Step Form Template should style checkbox inputs in core
Currently, adding a custom checkbox input to the Multi-Step Form Template renders the input hidden, showing just the label.

## Acceptance Criteria

- [ ] A custom checkbox input is visible

kjohnson:
Resolves #5587

## Description

- Adds custom data attributes for style selection.

## Affects

- Updates the Sequoia JS to call `setupCheckbox` on custom checkbox fields.
- Updates the Sequoia CSS to handle input replacement for custom checkbox fields.
- Updates the Legacy CSS to align custom checkbox fields.

## Visuals

NOTE: The Sequoia template still needs some work with regards to spacing, but that will be addressed in another change. The spacing styles are more general than the checkbox field and apply to all custom fields based on template location. I'll need to inspect each section and tweak the styles. For example, placing the fields before the gateways doesn't need any tweaks, but the insert location for after donation amount is flush with the side.

Screenshot_2021-02-23 Donation Form – WordPress(1)

Screenshot_2021-02-23 Donation Form – WordPress(2)

## Pre-review Checklist

## Testing Instructions

- Add a custom checkbox field

```php
add_action( 'give_fields_after_donation_levels', function( $fieldCollection, $formID ) {
	$fieldCollection->append(
		give_field( 'checkbox', 'my-custom-field' )
			->label( 'My Custom Field' )
	);
}, 10, 2 );
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 -   Fields API throws an exception when inserting relative to a node that does not exist (#5640)
+-   Form Templates now support custom checkbox fields. (#5643)
 
 ## 2.9.6 - 2021-01-13
 
--- a/assets/src/css/frontend/forms.scss
+++ b/assets/src/css/frontend/forms.scss
@@ -492,6 +492,7 @@ form[id*='give-form'] {
 		margin: 0 0.25em 0 0;
 	}
 }
+// Align checkbox/label for custom checkbox fields.
 div[data-field-type='checkbox'] {
 	display: flex;
 
--- a/assets/src/css/frontend/forms.scss
+++ b/assets/src/css/frontend/forms.scss
@@ -492,6 +492,17 @@ form[id*='give-form'] {
 		margin: 0 0.25em 0 0;
 	}
 }
+div[data-field-type='checkbox'] {
+	display: flex;
+
+	label {
+		order: 2;
+	}
+
+	input[type='checkbox'] {
+		order: 1;
+	}
+}
 
 /*---------------------------------
 Donation Level: Dropdown/Select
--- a/src/Form/LegacyConsumer/FieldView.php
+++ b/src/Form/LegacyConsumer/FieldView.php
@@ -5,13 +5,12 @@ namespace Give\Form\LegacyConsumer;
 use Give\Framework\FieldsAPI\FormField;
 
 class FieldView {
-    public static function render( FormField $field ) {
-        echo $field->getType();
-        echo '<div class="form-row form-row-wide">';
-            ob_start();
-            include plugin_dir_path( __FILE__ ) . '/templates/label.html.php';
-            include plugin_dir_path( __FILE__ ) . '/templates/' . $field->getType() . '.html.php';
-            echo ob_get_clean();
-        echo '</div>';
-    }
+	public static function render( FormField $field ) {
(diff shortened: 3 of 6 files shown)


JasonTheAdams:
@kjohnson Friendly note on this PR description: Issues are for describing problems/needs and PRs are for describing solutions. As such, the visualizations here are perfect for the Issue, whereas these pictures should reflect what the working solution looks like. 🙂👍

JasonTheAdams:
Actually, now that I read your image description more carefully, I think you're saying this _is_ the solution in action, and you'll do more later. Out of curiosity, why use another PR to fix the Sequoia spacing styles?

kjohnson:
@JasonTheAdams the spacing styles are more general than the checkbox field and apply to all custom fields based on template location. I'll need to inspect each section and tweak the styles. For example, placing the fields before the gateways doesn't need any tweaks, but the insert location for after donation amount is flush with the side.

JasonTheAdams:
Roger that. Thanks for helping me understand!

JasonTheAdams:
@kjohnson Can you please provide testing instructions? I'm not sure what field to use. I was expecting to see a Checkbox field, but I don't see one.

kjohnson:
@JasonTheAdams I updated the PR with testing instructions for adding a checkbox field.

JasonTheAdams (approved):
I just noticed in this PR that all fields are actually the same class: FormField. I would really rather have different classes for the different types, making types a first-class citizen of the Fields API. For example: Text, Select, Checkbox, and so forth.

Right now, for example, you could do `$textField->options()`. That doesn't really make any sense. I'd also like to have things such as `$checkbox->checkedByDefault()` as a simple way of setting the default.

It's probably not in the scope of this PR, but it is something important to me I think we should have done before releasing this to the wild. Happy to chat about this!

JasonTheAdams:
This will make type checking better in cases such as:
```php
$collection->append(
    Field::checkBoxGroup(
        Field::checkBox('Do you want to...')->checked(),
        Field::checkBox('Would you like...'),
    );
);
```
In this case the `Field::checkBoxGroup` is a collection which can quickly verify (and even via types in PHP 7) that it's being passed only CheckBox fields.

kjohnson:
@JasonTheAdams happy to talk about it but I've come across a lot of problems trying to use different classes for different fields types. I've found much better success with type being a property.

Also, I already have type checking for features - for example:

```php
    public function supportsOptions() {
        return in_array( $this->getType(), [
            FieldTypes::TYPE_SELECT,
            FieldTypes::TYPE_RADIO,
        ] );
    }
```

JasonTheAdams:
Just finished a call with @kjohnson wherein we discussed this and he won. 😉 

Really, he made some excellent points wherein the question became whether we have a minimal data structure which extends out in different directions — e.g. a base Field class which is extended by a Select class. The pro of this is that is only stores what data is needed, but the con is that it's inherently limiting and puts a lot more type checking down the line when processing fields. The method he's proposing means there's a uniform data model — that is, _all_ fields have the same properties, even if one or two aren't used when rendered. This eliminates the need for type checking and keeps things nice and flexible.

We believe that flexibility is the greater goal in our case, and classes can be added later if we want to add syntactical sugar (e.g. `CheckBox::checkedByDefault()`), but these classes wouldn't change the underlying data model.

So we're proceeding as is! 👍

PR #5660, opened 2021-03-01
Admin are now shown Donor Profiles upgrade notice and Donor Profile page is generated on new installs

### LINKED ISSUE #5630: Donor Profile page should be generated on new installs
## User Story

As an admin, I want a Donor Profile page to be generated on fresh installs so that I do not need to manually create one.

## Details

Currently, the Donor Profile block must be manually introduced to a fresh GiveWP install. Ideally, a Donor Profile page would be automatically generated on install, with the Donor Profile block added, or the Donor Profile shortcode added if the block editor is disabled.

## Acceptance Criteria

-   [x] On fresh installs, a Donor Profile page is automatically created
-   [x] If the block editor is enabled, the page is created with the Donor Profile block
-   [x] If the block editor is disabled, the page is created with the Donor Profile shortcode

### LINKED ISSUE #5638: Updating to GiveWP 2.10.0 should prompt users to update their legacy donation management pages
## User Story

As an admin, after adding a Donor Profile block to a page, I want to be prompted to easily update my legacy donation management page settings so that donors can be sent to the Donor Profile from their donation receipt.

## Details

Currently, even when the Donor Profile block is added, admin must remember to manually update their legacy donation management page settings, so that their new Donor Profile page is respected (in place of Donation History, Subscription Management, etc). Ideally, admin will be easily prompted to update these settings when they update to GiveWP 2.10.0.
## Additional Context

This solution addresses an issue raised in this comment: https://github.com/impress-org/givewp/issues/5628#issuecomment-783974582

## Acceptance Criteria

-   [x] Updating to GiveWP 2.10.0 displays a prompt for Admin to create a Donor Profile page and udpate their legacy donation management pages
-   [x] Admin can create a Donor Profile page and update their legacy donation management pages from the setting page, if they dismissed the prompt previously

henryholtgeerts:
Resolves #5638 
Resolves #5630 

## Description
This PR introduces logic to display an upgrade notice to explain the Donor Profiles feature to admin who have an existing GiveWP install. Additionally, it introduces logic to generate and setup a Donor Profile page when requested.

## Affects
This PR affects the admin interface, specifically pages with GiveWP related controls.

## Visuals
Screen Shot 2021-03-01 at 5 06 26 PM

Screen Shot 2021-03-01 at 5 06 47 PM

## Pre-review Checklist

## Testing Instructions

On an existing GiveWP install:
1) Check that the new Donor Profiles notice appears on GiveWP related admin pages
2) Click "Setup Donor Profiles", check that you are directed to the settings page
3) Find the Donor Profile Page setting. In the description text, click the link that says "Let us generate one for you."
4) The setting page should reload, now with the Donor Profile setting setup (with a page title Donor Profile), and the Override Legacy Donation Management Pages setting enabled. 
5) Check that the Donor Profile upgrade notice is no longer visible.
6) Go to Pages, and edit the new Donor Profile page.
7) Check that the Donor Profile block appears as expected, with accent color controls available.
8) Return to the Pages view, and delete the Donor Profile page (permanently delete the page).
9) Check that the upgrade notice re-appears on GiveWP related admin pages.
10) Click the "x" to dismiss the upgrade notice.
11) Refresh the page to ensure that the Upgrade Notice does not reappear.

On a fresh install:
1) Check that a Donor Profile page was created, and set as the History Page.
2) Check that the Donor Profile page includes a Donor Profile block.
3) In GiveWP settings, check that the Donor Profile page is set, and that the Override Setting is enabled.

### DIFF
--- a/src/DonorProfiles/Admin/UpgradeNotice.php
+++ b/src/DonorProfiles/Admin/UpgradeNotice.php
@@ -14,8 +14,11 @@ class UpgradeNotice {
 
 		// Give Admin Only.
 		if ( give_is_admin_page() ) {
-			$donorProfilePageIsSet = empty( give_get_option( 'donor_profile_page' ) ) ? false : true;
-			if ( $donorProfilePageIsSet === false ) {
+
+			$donorProfilePageIsSet = empty( give_get_option( 'donor_profile_page' ) ) || get_post_status( give_get_option( 'donor_profile_page' ) ) === false ? false : true;
+			$historyPageIsSet      = empty( give_get_option( 'history_page' ) ) ? false : true;
+
+			if ( $donorProfilePageIsSet === false && $historyPageIsSet === true ) {
 				return true;
 			} else {
 				return false;
--- a/src/DonorProfiles/Admin/UpgradeNotice.php
+++ b/src/DonorProfiles/Admin/UpgradeNotice.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Give\DonorProfiles\Admin;
+
+class UpgradeNotice {
+
+	public function register() {
+		if ( $this->shouldRenderOutput() ) {
+			$this->renderOutput();
+		}
+	}
+
+	protected function shouldRenderOutput() {
+		return true;
+	}
+
+	protected function renderOutput() {
+		echo $this->getOutput();
+	}
+
+	protected function getOutput() {
+		ob_start();
+		$output = '';
+		require $this->getTemplatePath();
+		$output = ob_get_contents();
+		ob_end_clean();
+
+		return $output;
+	}
+
+	protected function getTemplatePath() {
+		return GIVE_PLUGIN_DIR . '/src/DonorProfiles/resources/views/upgradenotice.php';
+	}
+}
--- a/src/DonorProfiles/Admin/UpgradeNotice.php
+++ b/src/DonorProfiles/Admin/UpgradeNotice.php
@@ -25,22 +25,13 @@ class UpgradeNotice {
 	 * @since 2.10.0
 	 */
 	protected function shouldRenderOutput() {
-
-		// Give Admin Only.
-		if ( give_is_admin_page() ) {
-
-			$donorProfilePageIsSet = empty( give_get_option( 'donor_profile_page' ) ) || get_post_status( give_get_option( 'donor_profile_page' ) ) === false ? false : true;
-			$historyPageIsSet      = empty( give_get_option( 'history_page' ) ) ? false : true;
-
-			if ( $donorProfilePageIsSet === false && $historyPageIsSet === true ) {
-				return true;
-			} else {
-				return false;
-			}
-			return true;
-		} else {
+		if ( ! give_is_admin_page() ) {
 			return false;
 		}
+
+		$donorProfilePageIsSet = ! empty( give_get_option( 'donor_profile_page' ) ) && get_post_status( give_get_option( 'donor_profile_page' ) );
+		$historyPageIsSet      = ! empty( give_get_option( 'history_page' ) );
+		return ! $donorProfilePageIsSet && $historyPageIsSet;
 	}
 
 	/**
--- a/src/DonorProfiles/Admin/UpgradeNotice.php
+++ b/src/DonorProfiles/Admin/UpgradeNotice.php
@@ -11,7 +11,19 @@ class UpgradeNotice {
 	}
 
 	protected function shouldRenderOutput() {
-		return true;
+
+		// Give Admin Only.
+		if ( give_is_admin_page() ) {
+			$donorProfilePageIsSet = empty( give_get_option( 'donor_profile_page' ) ) ? false : true;
+			if ( $donorProfilePageIsSet === false ) {
+				return true;
+			} else {
+				return false;
+			}
+			return true;
+		} else {
+			return false;
+		}
 	}
 
 	protected function renderOutput() {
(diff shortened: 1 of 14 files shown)


henryholtgeerts:
@JasonTheAdams I've opened this draft PR to show what is in place, and ask for suggestions on 2 last items that are proving tricky. 

1st issue) In the description for the Donor Profile page setting, I include a link to generate a new donor profile page. My thinking was to add a `give-generate-donor-profile-page` query param to the URL, and use the `parse_request` hook to check for that specific query param to trigger a method which generates a donor profile page with the correct content, and save the donor profile page setting with that ID. All of that functionality is in place and works as expected -- with one issue! For whatever reason, `parse_request` does not fire in the WP admin context where the GiveWP settings live. 

So this doesn't work:
http://give.test/wp-admin/edit.php?post_type=give_forms&page=give-settings&give-donor-profile-action=generate-donor-profile-page

But this does:
http://give.test/wp-admin/edit.php?give-donor-profile-action=generate-donor-profile-page 

Obviously, the later isn't gonna cut it, if we intend to provide a link an admin can click to generate a donor profile page, which successfully generates a donor profile page, and returns them to the same page they were on when they clicked it.

One possible solution here:
- Go with AJAX instead of an actual anchor tag. On click, generate the donor profile page through an AJAX request, then reload the page when that request succeeds. 

Mostly -- I'm curious if I'm just missing something straightforward, or if my original approach is way off and a better solution exists.

2nd issue) This one is fairly simple but I believe merits a broader discussion. Should the "override legacy donation management pages" toggle update the aforementioned settings directly (run give_update_option on Donation History, Edit Profile, Subscriptions page settings, etc) or instead use hooks to intercept requests for those pages. 

I initially implemented the former solution (because it's easy!) but realized there may be value in making sure that an admin's legacy donation management page settings are preserved, even if the Donor Profile override setting is enabled.

JasonTheAdams:
Great work, Henry! Let's see if I can help give some direction, here. 😃

### Issue # 1
I don't recommend adding query strings to pages and then reloading the page. The reason for this is because we have no idea what context the user is in and what effect the current query strings have. They also could have done something like arrived with `$_POST` data, so refreshing could do strange things. All that to say, it's just not a good idea.

My suggestion would be one of two options:
- AJAX (as you mentioned)
- Redirect the user to a fixed URL

Going the AJAX route is fine, but does require adding some JS and having to deal with that whole thing. The simpler route would be to just redirect the user to some page (could be the admin dashboard) with a query string in place: `https://example.com/wp-admin?give-generate-donor-profile-page=1`. At that point you could use the [wp_loaded](https://developer.wordpress.org/reference/hooks/wp_loaded/) action to create the new page, and then use [wp_safe_redirect](https://developer.wordpress.org/reference/functions/wp_safe_redirect/) to redirect them to the page on the front-end. In this case the user would experience clicking the "Setup Donor Profiles" link, some non-visual redirecting, and then finally looking at the shiny new page.

I could be talked out of taking the user to the Donor Profile page. I just thought it might be cool for them to see it. The downside is that they probably wouldn't have much to look at as they likely haven't made donations.

The real pro of the AJAX method is if we want to keep the user exactly where they are and not interrupt their flow. So to me if we want to take them somewhere then we should go with the redirect route; if we want to let them stay where they are then AJAX.

What do you think, @DevinWalker?

JasonTheAdams:
### Issue # 2

Since the settings are easily set back, I'm inclined to just keep things simple as you have: Change the settings to what they ought to be.

henryholtgeerts:
@JasonTheAdams Ah this is incredibly helpful! Thanks for you suggestions. I was feeling dubious of the query strings approach (especially as it became more convoluted) and the redirect approach you describe sounds good to me. I'll take a stab at implementing that and the override functionality you suggested this AM.

henryholtgeerts:
@JasonTheAdams I've incorporated your suggestions here, now just adding some logic to let admin dismiss the notice, and trying to resolve issues with adding a block by setting content `insert_post`! 

The Donor Profile insets fine to the post, but when you open the block editor, it displays it within a classic editor block (rather than displaying the actual donor block). I suspect this is a result of the block being server side rendered, but I'll keep posted on what I figure out.

src/DonorProfiles/Admin/UpgradeNotice.php
```diff
@@ -0,0 +1,49 @@
...
+
+			if ( $donorProfilePageIsSet === false && $historyPageIsSet === true ) {
+				return true;
+			} else {
+				return false;
+			}
+			return true;
+		} else {
+			return false;
+		}
```
JasonTheAdams, line 29: @henryholtgeerts I suggest a return early method, in general. It helps avoid nesting. Also, the final `return true` inside the conditional block is unreachable.

Finally, `empty($foo) ? false : true` and `if ($truthy) { return true; } else { return false; }` are redundant. The `empty()` function returns a boolean, and conditional expressions return booleans — so just used the returned booelan.
```suggestion
		if ( ! give_is_admin_page() ) {
			return false;
		}

		$donorProfilePageIsSet = ! empty( give_get_option( 'donor_profile_page' ) ) && get_post_status( give_get_option( 'donor_profile_page' ) );
		$historyPageIsSet      = ! empty( give_get_option( 'history_page' ) );

		return ! $donorProfilePageIsSet && $historyPageIsSet;
```

henryholtgeerts: Ah, good thinking! Much clearer. I've gone ahead and reflected this change in other areas where I attempt something similar: https://github.com/impress-org/givewp/pull/5660/commits/3f2ad24b7e1c6428160f9b16518d93379e7f38ec

JasonTheAdams (approved):
Excellent work, @henryholtgeerts! I'm excited for new and existing users to enjoy the new donor experience! 🚀

PR #5674, opened 2021-03-05
Donor Profile tests that Receipt Info contains donation data

kjohnson:
## Description

Asserts that the array returned by the Donations Repository contains the correct amount and user email of the donation.


henryholtgeerts (approved):
Looks great! Good to see an example mocking a payment with that helper class.

PR #5706, opened 2021-03-11
Donor Dashboard errors are now logged

### LINKED ISSUE #5701: Donor Dashboard errors should be logged
## User Story

As an admin, I want errors in the donor dashboard to be logged so that I can more easily work with support to diagnose the underlying problem.

## Details

The Error notice that appears in the donor dashboard does not currently log the error. It should log the underlying error and make itself more clear to donors what is going on.

From Canny:

Whenever this error shows on the front end, it needs to either 
- be actionable there (give the donor next steps to see their donation history) or
- log an error to the logs with something more actionable.

Option 1 is preferrable, since it's clear that the donor has been verified (in this case by email access) we should show them what specifically is preventing their history from showing.

If we go with option 2, the error message should be changed to "An error occurred while retrieving your donation records. Contact a site administrator and have them search the logs at Donations > Tools > Logs for a more specific cause of the problem."

303df2763b628d4d518a37b49d4a9b32

## Acceptance Criteria

-   [ ] Donor Dashboard error is logged
-   [ ] Donor Dashboard DB error is made more clear

alaca:
Resolves #5701

## Description

This PR adds error logging to the Donor Dashboard. 

## Affects
 Donor Dashboard

## Visuals

|Donor Dashboard | Logs |
|---|---|
| image | image |

## Testing Instructions
The easiest way to trigger an error would be to make changes to one of the SQL queries in `src/DonorProfiles/Repositories/Donations.php`, e.g changing the table name or table column name will throw an error for sure. 

## Pre-review Checklist

### DIFF
--- a/src/DonorProfiles/resources/js/app/tabs/donation-history/content.js
+++ b/src/DonorProfiles/resources/js/app/tabs/donation-history/content.js
@@ -38,6 +38,9 @@ const Content = () => {
 				<p style={ { color: '#6b6b6b' } }>
 					{ error }
 				</p>
+				<p style={ { color: '#6b6b6b' } }>
+					{ __( 'Contact a site administrator and have them search the logs at Donations > Tools > Logs for a more specific cause of the problem.', 'give' ) }
+				</p>
 			</Fragment>
 		);
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Donors with no donations now see a "No Donations" notice in Donor Dashboard (#5694)
 -   Donor Dashboard iframe now resizes when the parent window resizes (#5693)
 -   Migration table id column length is now increased (#5698)
+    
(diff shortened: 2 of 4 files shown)


src/DonorProfiles/resources/js/app/tabs/donation-history/content.js
```diff
@@ -38,6 +38,9 @@ const Content = () => {
 				<p style={ { color: '#6b6b6b' } }>
 					{ error }
 				</p>
+				<p style={ { color: '#6b6b6b' } }>
```
alaca, line 41: @henryholtgeerts I hardcoded this additional info, hope this approach is ok? I was also thinking to wrap everything in one `Error` component, but I'm not sure about that. Let me know what you think.

henryholtgeerts: I think this is fine, since it's the only place an error message occurs right now. 👍

henryholtgeerts (approved):
These changes look good! Thanks for setting this up 👍

PR #5718, opened 2021-03-15
Onboarding Settings Repository Factory no longer triggers a fatal error for corrupted option values

### LINKED ISSUE #5717: Onboarding Settings Repository should catch invalid arguments
## Details

Reference: https://wordpress.org/support/topic/fatal-error-uncaught-typeerror-8/

`Fatal error: Uncaught TypeError: Argument 1 passed to Give\Onboarding\SettingsRepository::__construct() must be of the type array, bool given`

<details>
Fatal error: Uncaught TypeError: Argument 1 passed to Give\Onboarding\SettingsRepository::__construct() must be of the type array, bool given, called in /home/customer/www/mysite.org/public_html/wp-content/plugins/give/src/Onboarding/SettingsRepositoryFactory.php on line 23 and defined in /home/customer/www/mysite.org/public_html/wp-content/plugins/give/src/Onboarding/SettingsRepository.php:22 Stack trace: #0 /home/customer/www/mysite.org/public_html/wp-content/plugins/give/src/Onboarding/SettingsRepositoryFactory.php(23): Give\Onboarding\SettingsRepository->__construct(false, Object(Closure)) #1 /home/customer/www/mysite.org/public_html/wp-content/plugins/give/src/Onboarding/Wizard/Page.php(49): Give\Onboarding\SettingsRepositoryFactory->make(‘give_settings’) #2 [internal function]: Give\Onboarding\Wizard\Page->__construct(Object(Give\Onboarding\FormRepository), Object(Give\Onboarding\SettingsRepositoryFactory), Object(Give\Onboarding\LocaleCollection)) #3 /home/customer/www/mysite.org/public_htm in /home/customer/www/mysite.org/public_html/wp-content/plugins/give/src/Onboarding/SettingsRepository.php on line 22
</details>

## Expected Behavior

An invalid argument type should not cause a fatal error.

## Steps to Reproduce

1. Make a new Settings Repository with a non-array value as the first parameter.

## Acceptance Criteria

-   [ ] Invalid arguments are caught and throw an exception.
-   [ ] Invalid arguments do not cause a fatal error.

kjohnson:
Resolves #5717 

## Description

If for some reason, the value of a "settings" option is update and not an array format the `SettingsRepository` will throw a `TypeError` exception.

This prevents a fatal error by casting the value of the option injected by the `SettingsRepositoryFactory`.

## Testing Instructions

See `tests/unit/tests/Onboarding/SettingsRepositoryFactoryTest.php`.

Overwrite a "settings" option so that the value is not an array (and does not trigger a default value of an array, ie `false` or an empty string). When a SettingsRepository is created with a non-array option value it should no longer throw a `TypeError` exceptionm thus nnot triggering a fatal error.

## Pre-review Checklist


JasonTheAdams (commented):
I'm not sure this is the best way to go about it. I can see that the option default is an empty array, so if it's storing an array then why is the value being stored as a boolean? This is preventing the one exception to happen, but if it's now being passed `[false]` will that just create another exception elsewhere? This feels like a band-aid to a problem we don't yet understand.

Please correct me if I'm wrong and I just don't understand how this fixes things.

kjohnson:
@JasonTheAdams that is a fair criticism. I see this more as preventing a fatal error on customer sites. The original report suggests that there is something else going on that causes the setting value to be malformed, but in that case a customer is getting a fatal error on the login screen making it difficult to recover from the issue.

In practice, `[ false ]` wouldn't throw an exception later on because the getter handles the missing key gracefully.

That said, it might help to add some logging when the option value is cast to an array so that we can have a record to which we can refer.

JasonTheAdams:
This is what worries me:
```php
$settings = $factory->make('my_setting'); // option returns false which stores as [false]

$expectedValue = $settings->get('value'); // returns null as the internal $settings is [0 => false]

doSomething($expectedValue);
```

So we're creating a `SettingsRepository` that contains a completely unexpected value. It's designed for an associative array but got a numeric array. The implementing code has no idea there's a problem and tries to grab the value by key. It returns null simply because it checks for the key. Now we `doSomething()` with the value, which may or may not work with a null value. This all stems from the fact that the data coming from the database is malformed.

If the `SettingsRepository` is expecting an associative array for it to do its job, then it _should_ break when handed something else. If we really wanted to prevent an exception at some point, it should happen via `try...catch` in the implementing code:
```php
try {
  $settings = $factory->make('my_setting');
} catch(InvalidSettingsValueException $exception) {
  // do something
}
```
Applying the "fix" in the factory means intentionally passing along malformed data, which just means the fatal exception will likely occur downstream.

kjohnson:
Ok, so we could have a `try...catch` to prevent the fatal error, but in turn throw a non-fatal error so that it breaks in a manageable way.

JasonTheAdams:
I'm fine with that. Really, here, we don't know what's actually going on. But I do agree it shouldn't block users from being able to log in. So where that's happening we can catch it, log it, and display an error that doesn't prevent login.

kjohnson:
Yeah, we need something so that the user can access logs.

kjohnson:
@JasonTheAdams I updated the solution to try/catch and throw a non-fatal error.

JasonTheAdams:
I don't think inside the factory is the right place for the `try...catch`. It's too generic of a location as this class is _very_ generic, wrapping the options. We ought to figure out what specific instance of the factory being used is preventing the login from working, and catch that. We don't want to generically prevent the exception as we want squelching exceptions to be the exception and not the rule (pun intended).

kjohnson:
@JasonTheAdams looking at the Call Stack, the specific instance is a call to the `give_settings` on the Wizard Page construct.

JasonTheAdams:
Let's do a quick call on this tomorrow. For the life of me I cannot reproduce the exception, and I'm really trying to break this.

kjohnson:
@JasonTheAdams sounds good. I'll admit that I didn't do my due diligence on recreating the error. I received a Stack Trace that showed a Fatal Error that prevented site access and I jumped at it.

The more I look into it the more I realize that this is very unlikely and the cause is unknown and not directly related to the "fix" here.

JasonTheAdams:
It makes me wonder if something like site caching is involved. For this to fail the entire `give_settings` option would have to be missing. I have no idea how that would be possible. I even tried a non-existent key and it still allowed me to log in.

kjohnson:
Closing this until we get a clearer direction.

PR #5739, opened 2021-03-19
The Donor Dashboard shortcode now works on 4.9.x 

### LINKED ISSUE #5738: The Donor Dashboard shortcode should work on 4.9.x
## User Story

As an admin, I want the donor Dashboard shortcode to work on 4.9.x so that I can make use of the new feature without updating WP core.

## Details

From Canny:
Currently on WordPress 4.9.15 the donor dashboard shortcode flashes like it's about to render and then disappears. There are no console errors, and no other functionality seems broken, but the dashboard does not appear.

Additionally, the page generated for the Donor Dashboard should use a shortcode whenever the block editor is not available/enabled.

## Expected Behavior

The Donor Dashboard shortcode should work as expected.

## Steps to Reproduce

1. WordPress 4.9.15
2. Page with the donor dashboard shortcode of [give_donor_dashboard]

## Visuals

N/A

## Acceptance Criteria

-   [x] Donor Dashboard shortcode works in 4.9.x
-   [x] Donor Dashboard page is generated with shortcode in contexts where block editor is disabled/unavailable

henryholtgeerts:
Resolves #5738 

## Description

This PR introduces changes to ensure that the Donor Dashboard shortcode works as expected with WP core 4.9.x, and that the generated Donor Dashboard page uses a shortcode when the block editor is not enabled.

## Affects

This PR affects frontend rendering logic of the Donor Dashboard application, removing all WP 5.0 dependencies. Additionally, this PR affects the Donor Dashboard page generation logic, specifically adding conditions to check for whether or not the block editor is enabled at the point the Donor Dashboard page is being generated.

## Visuals

N/A

## Testing Instructions

On WP 4.9.x
1. Generate a Donor Dashboard page
2. Check that it works as expected on the frontend
3. Check that the page uses a shortcode in the editor to display the Donor Dashboard

On WP 5.x with Classic Editor enabled
1. Generate a Donor Dashboard page
2. Check that it works as expected on the frontend
3. Check that the page uses a shortcode in the editor to display the Donor Dashboard

On WP 4.9.x without Gutenberg plugin enabled
1. Generate a Donor Dashboard page
2. Check that it works as expected on the frontend
3. Check that the page uses a block in the editor to display the Donor Dashboard

On WP 5.x without Classic Editor enabled
1. Generate a Donor Dashboard page
2. Check that it works as expected on the frontend
3. Check that the page uses a block in the editor to display the Donor Dashboard
4. 
## Pre-review Checklist


src/DonorDashboards/Admin/Settings.php
```diff
@@ -151,26 +151,57 @@ public function generateDonorDashboardPage() {
...
+	}
+
+	/**
+	 * Determine whether the block editor is active
+	 *
+	 * @return bool
+	 *
+	 * @since 2.10.0
+	 */
+	private function isBlockEditorActive() {
```
kjohnson, line 177: It looks like this function is doing more than one thing - could you separate out the logic so that each function does "one thing"? This would also clarify what is specifically being checked.

I see:

- Is Gutenberg Active (thus also installed)
- Is WP version greater than 5.0-beta.
- Is the Classic Editor active (thus also installed)

kjohnson (changes requested):

src/DonorDashboards/Admin/Settings.php
henryholtgeerts: Done! Ref: https://github.com/impress-org/givewp/pull/5739/commits/56f856aa047246146112138ccd378a06041277c9

Benunc: I personally use the Disable Gutenberg plugin on client sites (because the Classic editor plugin puts "upsells" to try gutenberg that I don't want my clients to see in some cases). 

Also, classic editor has lots of conditional logic where the plugin itself can be active, but only restricting use on Pages, not posts, etc. 

What I found was that the front-end still parses the block markup even with Gutenberg disabled.

kjohnson: @henryholtgeerts it sounds like this changes our algorithm, so let's consider @Benunc's feedback for a refactor.

kjohnson (commented):
@henryholtgeerts I'm deferring my review until we can consider @Benunc's feedback. If you need, I'll be available for a quick call after the betting table meeting.

src/DonorDashboards/Admin/Settings.php
henryholtgeerts: I've added a check for the Disable Gutenberg plugin as well, so that the Donor Dashboard page is genretated with shortcode if that plugin is enabled.

I do not believe it is worth our time to account for conditional setting inside either the Classic Editor plugin or Disable Gutenberg, since
1. They each use their own unique methods/logic for storing these conditions and our development for this edge case should not be bound up in keeping current with their decision making.
2. Admins who have these plugins installed are likely familiar with shortcodes, and will not be confused to see one appear in the block editor, if they have indeed disabled the block editor for certain post types but not all.

I think adding a check for the Disable Gutenberg plugin is warranted, since itself and Classic Editor are by a wide margin the two most popular methods of disabling the block editor. 

Currently, the algorithm works like this:

If a site does not have Classic Editor activated AND it does not have Disable Gutenberg activated AND ( it is running WP 5.0+ OR it has the Gutenberg plugin installed ), it will load the page via a block. Otherwise, it will load the page via a shortcode.

kjohnson (approved):

Benunc:
It all checks out for me. :shipit:

Benunc (approved):

PR #5759, opened 2021-03-26
 GiveWP Database management tools are now backward compatible with MySQL 5.6

### LINKED ISSUE #5758: GiveWP Database management tools should be backward compatible to MySQL 5.6
## User Story

As a  site admin, I want the GiveWP Database management tools to be compatible with MySQL 5.6. 

## Details

Currently, the migration log table creation process is failing on older MySQL versions. It shouldn't.

## Expected Behavior

 The migration log table creation process should work on servers that are using MySQL 5.6.

## Steps to Reproduce

1. Install Give 2.10.0 on a server that uses MySQL 5.6 

## Acceptance Criteria

-   [ ] The migration log table creation process is working with MySQL 5.6

alaca:
Resolves #5758 

## Description
This PR resolves the issue with the migration log table creation process which was failing on older MySQL versions. Table creation was failing because the default value for the `last_run` column was set to `CURRENT_TIMESTAMP` which is not supported for the `DATETIME` column type in MySQL versions <= 5.6. The issue is resolved by changing the column type to `TIMESTAMP`.

## Affects

Create migration table SQL. 

## Testing Instructions

Install Give on a server that has MySQL version <= 5.6

## Pre-review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 -   Fix broken link by correctly closing href tag (#5746)
+-   GiveWP Database management tools are now backward compatible with MySQL 5.6 (#5759)
 
 ## 2.10.0 - 2021-03-22
 
--- a/src/MigrationLog/MigrationLogRepository.php
+++ b/src/MigrationLog/MigrationLogRepository.php
@@ -39,8 +39,8 @@ class MigrationLogRepository {
 	 */
 	public function save( MigrationLogModel $model ) {
 		$query = "
-			INSERT INTO {$this->migration_table} (id, status, error ) 
-			VALUES (%s, %s, %s) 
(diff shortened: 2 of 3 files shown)


JasonTheAdams:
@alaca Are there any issues to consider for sites that have already run the migration and will have a datetime column instead of timestamp?

alaca:
@JasonTheAdams `TIMESTAMP` and `DATETIME` are almost identical, and there won't be any issues if we change the type. 
However, I revisited this issue again and came to the conclusion that we don't have to change the type of the column. The most straightforward approach to fix this issue  is to remove the redundant part `DEFAULT CURRENT_TIMESTAMP` because the `last_run` column value is already being set [here](https://github.com/impress-org/givewp/blob/3422b4bba60e653c53a66083dc085d7ee48be422/src/MigrationLog/MigrationLogRepository.php#L47).

JasonTheAdams (approved):
Looks to be working! Great job!

PR #5762, opened 2021-03-29
Exported CSVs now support Chinese/Japanese/Korean characters

### LINKED ISSUE #5761: Chinese/Japanese/Korean characters broken when generating/exporting CSV
## Details

When I export the data in CSV format ( Donations > Tools > Export ), if the data contains Chinese, Japanese, or Korean characters, they are broken.

## Expected Behavior

When exporting data, the data must be readable.

## Steps to Reproduce

1. Go to Donations > Tools > Export
2. Click on Generate CSV
3. Open the csv file generated
4. Characters broken

jgreys:
Resolves #5761 

## Description

The addition of byte order mark (BOM) before the datastream of a CSV file ensures proper display of Chinese characters, Japanese and Korean.

## Affects

Export / Generate CSV function

## Testing Instructions

1. Go to Donations > Tools > Export
2. Click on Generate CSV
3. Open the csv file generated
4. Characters not broken

## Pre-review Checklist

### DIFF
--- a/includes/admin/tools/export/class-export-earnings.php
+++ b/includes/admin/tools/export/class-export-earnings.php
@@ -46,6 +46,8 @@ class Give_Earnings_Export extends Give_Export {
 		header( 'Content-Disposition: attachment; filename=' . apply_filters( 'give_earnings_export_filename', 'give-export-' . $this->export_type . '-' . date( 'n' ) . '-' . date( 'Y' ) ) . '.csv' );
 		header( 'Expires: 0' );
 
+		// UTF-8 BOM
+		echo "\xEF\xBB\xBF";
 	}
 
 	/**
--- a/includes/admin/tools/export/class-export.php
+++ b/includes/admin/tools/export/class-export.php
@@ -67,6 +67,9 @@ class Give_Export {
 		header( 'Content-Type: text/csv; charset=utf-8' );
 		header( 'Content-Disposition: attachment; filename=' . $file_name . '.csv' );
 		header( 'Expires: 0' );
+
+		// UTF-8 BOM
+		echo "\xEF\xBB\xBF";
 	}
 
 	/**


jgreys:
Hello! Did I miss anything in my PR?

kjohnson:
@jgreys this did fall off of my radar, but I have now assigned it to myself for review.

kjohnson:
Trying to do my due diligence here, having encountered BOM related issues in the past.

Specifically, when working on Ninja Forms I remember introducing a filter to configure use of the BOM in CSV export. It was initially hard coded, but a filter was added to remove it as it can cause issues in some places (while it resolves issues in other places).

Additionally, I'll note that the below does not specify the `utf-8` character set in the `Content-Type` header.

See [1b8c76.patch](https://gist.github.com/kjohnson/4aafcd943ef77a2c9983265f50272521)
```patch
@@ -99,14 +99,20 @@ function ninja_forms_export_subs_to_csv( $sub_ids = '', $return = false ){
   header("Content-type: application/csv");
-  echo "\xEF\xBB\xBF"; // Byte Order Mark
+  echo apply_filters('ninja_forms_csv_bom',"\xEF\xBB\xBF") ; // Byte Order Mark
}
```

Citing the [Unicode standard](https://www.unicode.org/versions/Unicode13.0.0/ch02.pdf), I think we should maintain a default of not including a BOM, but perhaps allow one to be added for specific use cases. See Unicode v13, chapter 2, section 6 - emphasis added:

> "Use of a BOM is **neither required nor recommended for UTF-8**, but may be encountered in contexts where UTF-8 data is converted from other encoding forms that use  a  BOM  or  where  the  BOM  is  used  as  a  UTF-8  signature. "

@jgreys I would request, as a maintainer, that instead of hard-coding the BOM mark we add a hook for prepending an optional BOM to the CSV export.

jgreys:
Agree! That would be the best :)
How would you like to name it?

kjohnson:
@jgreys lets add an action after the headers are set called `give_export_headers` and pass `$this->export_type` as a parameter, that should cover both export files.

kjohnson:
Closing this due to inactivity. Happy to re-visit at a later time!

PR #5776, opened 2021-04-02
Logs table creation is now backward compatible with MySQL 5.6

### LINKED ISSUE #5775: Logs table creation should be backward compatible with MySQL 5.6
## User Story

As a site admin, I want the `give_log` table creation process not to cause fatal errors so that I can continue to use my website. 

## Details

Currently, the `give_log` table creation process is failing on older MySQL versions ( 5.6 ) which causes fatal errors. 

## Expected Behavior

The logs table creation process should work on servers that are using MySQL 5.6.

## Steps to Reproduce

1. Install Give 2.10.1 on a server that uses MySQL 5.6

## Acceptance Criteria

-   [x] The log table creation process is working with MySQL 5.6

alaca:
Resolves #5775

## Description

This PR resolves the issue with the logs table creation process which was failing on older MySQL versions. Table creation was failing because the default value for the `date` column was set to `CURRENT_TIMESTAMP` which is not supported for the `DATETIME` column type in MySQL versions <= 5.6. The issue is resolved by removing the default value and by setting the date with PHP.

## Affects

Create `give_logs` table SQL.

## Testing Instructions

Install GiveWP 2.10.1 on a server that has MySQL version <= 5.6

## Pre-review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+-   Logs table creation is now backward compatible with MySQL 5.6 (#5776) 
+
 ## 2.10.1 - 2021-03-30
 
 ### Fixed
--- a/src/Log/LogFactory.php
+++ b/src/Log/LogFactory.php
@@ -64,7 +64,7 @@ class LogFactory {
 			'source'   => esc_html__( 'Give Core', 'give' ),
 			'context'  => [],
 			'id'       => null,
-			'date'     => null,
+			'date'     => date( 'Y-m-d H:i:s' ),
 		];
 	}
 }
--- a/src/Log/LogModel.php
+++ b/src/Log/LogModel.php
@@ -233,6 +233,7 @@ class LogModel {
 		$repository = give( LogRepository::class );
 
 		if ( $this->getId() ) {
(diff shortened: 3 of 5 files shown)


JasonTheAdams:
@alaca Did you audit the other migrations to make sure there isn't anything else causing MySQL 5.5 issues?

alaca:
@JasonTheAdams I did a search for the `CURRENT_TIMESTAMP`, didn't find anything in the core files. However, the Funds add-on is using the same approach. I'll open an issue for that as well.

JasonTheAdams (changes requested):
When the `LogModel::save()` method is used it stores the id but not the new date. When a model is saved all insert-generated fields should be applied to the model.

alaca:
@JasonTheAdams resolved. 

8186fa14 e99c68ac

JasonTheAdams:
@alaca Can you please look into why the tests are failing?

alaca:
@JasonTheAdams the problem was in the `date` column not being set.

Anyway, the test is failing again, but now for different [reasons](https://github.com/impress-org/givewp/runs/2285023758?check_suite_focus=true).

JasonTheAdams:
Roger that. I'll take it from here. I'm trying to add it so all tests run on MySQL 5.6 and 8.0 so we catch these sorts of issues in the future.

JasonTheAdams (approved):
Working and we have all tests running for MySQL 5.6 and 8.0 now. 💪

PR #5787, opened 2021-04-09
Donors can no longer see other donors donations (on certain hosts)

### LINKED ISSUE #5780: Donor Dashboard should not be broken for donors logged in with WP accounts
## User Story

As a donor, I want the Donor Dashboard to work the same regardless of how I am authenticated so that I can log in with my WP account and see only my donations.

## Details

Full details about this bug were first reported by here:
https://github.com/impress-org/givewp/pull/5766#issuecomment-812803925

In short, by removing reliance on the X-WP-Nonce header, the Donor Dashboard was made workable for donors using email access on hosts that don't play nice with WP nonces, but made unuseable for donors attempting to log in via a WP account. 

More concerning, when logged in as a user with an associated WP account, all donations are made visible.

## Expected Behavior

The Donor Dashboard experience should be consistent, regardless of how you are authenticated.

## Steps to Reproduce

On a host with known caching conflicts:

1.  Log in as a donor with an associated WP account
2. Navigate to the Donor Dashboard
3. See that it is broken

## Visuals

See original report: https://github.com/impress-org/givewp/pull/5766#issuecomment-812803925

## Acceptance Criteria

-   [x] Under no circumstance is a Donor able to see donations made by other donors via the Donor Dashboard

henryholtgeerts:
Resolves #5780 

## Description

This PR prevents donors from every being shown donations that do not belong to them.. Previously, in certain caching settings, if the system was unable to identify the current donor ID, then all donations would be returned on the Donations endpoint.

## Affects

This PR affects backend logic for the Donations endpoint of the Donor Dashboard.

## Visuals

N/A

## Testing Instructions

This is a very difficult bug to test. I have so far been unable to replicate the issue.

## Pre-review Checklist

### DIFF
--- a/src/DonorDashboards/Tabs/DonationHistoryTab/DonationsRoute.php
+++ b/src/DonorDashboards/Tabs/DonationHistoryTab/DonationsRoute.php
@@ -53,28 +53,6 @@ class DonationsRoute extends RouteAbstract {
 	 */
 	protected function getData( DonationsRepository $repository, $donorId ) {
 
-		// Check that the provided donor ID is valid
-		if ( Give()->donors->get_donor_by( 'id', $donorId ) !== false ) {
-			Log::error(
-				esc_html__( 'An error occurred while retrieving donation records', 'give' ),
-				[
-					'source'   => 'Donor Dashboard',
-					'Donor ID' => $donorId,
-					'Error'    => __( 'An invalid Donor ID was provided.', 'give' ),
-				]
-			);
-
-			return new WP_REST_Response(
-				[
-					'status'        => 400,
-					'response'      => 'database_error',
-					'body_response' => [
-						'message' => esc_html__( 'An error occurred while retrieving your donation records. Contact the site administrator for assistance.', 'give' ),
-					],
-				]
-			);
-		}
-
 		// If the provided donor ID is valid, attempt to query data
 		try {
 			$donations = $repository->getDonations( $donorId );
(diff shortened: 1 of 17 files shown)


JasonTheAdams:
@henryholtgeerts I see that this function is used in `DonationsRoute::getData()`. What happens with those other functions if the donor id is invalid?

henryholtgeerts:
@JasonTheAdams That's a good point. Considering this is a hotfix, I had initially opted to be as specific as possible in addressing the problem (to avoid unexpected bugs), but looking things over it makes more sense to perform the validity check in the donations route directly. Ref: https://github.com/impress-org/givewp/pull/5787/commits/034d000618a102ab2f3100690ed6d39e7af5383e

src/DonorDashboards/Tabs/DonationHistoryTab/DonationsRoute.php
```diff
@@ -52,6 +52,30 @@ public function handleRequest( WP_REST_Request $request ) {
...
+					'Error'    => __( 'An invalid Donor ID was provided.', 'give' ),
+				]
+			);
+
+			return new WP_REST_Response(
+				[
+					'status'        => 400,
+					'response'      => 'database_error',
+					'body_response' => [
+						'message' => esc_html__( 'An error occurred while retrieving your donation records.', 'give' ),
```
JasonTheAdams, line 72: @henryholtgeerts For end-user facing errors like this, please give the end-user an actionable next step. Even if it's to reach out to the website owner and let them know.

henryholtgeerts: Good call! Done, ref: fix: add actionable item to message

henryholtgeerts:
@JasonTheAdams This PR has been tested to check for that existing functionality works as expected, and the Log item generated has been expanded to include any session information that might be relevant. When you can, take a look at it (maybe test it out on a staging site for good. measure).

JasonTheAdams (approved):
Great! This does not resolve the underlying authentication issues, but it does resolve the issue of donors seeing others' donations and provides better guidance in the case of errors.

PR #5788, opened 2021-04-09
Prevent legacy logging from throwing unhandled exceptions

### LINKED ISSUE #5786: Logging should not throw unhandled exceptions
## Details

Our logging system is used to stored information about the application in various states. It isn't a part of the business logic of the application, but is meta to the application itself. As such, it should never throw unhandled exceptions itself, otherwise we risk the logging system breaking the application.

## Acceptance Criteria

-   [x] Exceptions thrown when logging are caught and handled
-   [x] Logging exceptions do not break the application

JasonTheAdams:
Resolves #5786 

## Description

The new `Log` facade does not allow unhandled exceptions, but the legacy system still does as it's using the `LegacyFactory` directly. We should just have the various systems use the facade unless they have a specific reason to generate the log (such as the migration runner which handles the exception itself).

## Affects

The legacy migration system

## Testing Instructions

Add a line to throw an exception in the new logging repository and then trigger the legacy logging system to insert a log.

## Pre-review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 -   Logs table creation is now backward compatible with MySQL 5.6 (#5776) 
+-   Logs will no longer cause an exception (#5788)
 
 ## 2.10.1 - 2021-03-30
 
--- a/includes/class-give-logging.php
+++ b/includes/class-give-logging.php
@@ -1,6 +1,7 @@
 <?php
 
 use Give\Log\Log;
+use Give\Log\LogFactory;
 use Give\Log\LogRepository;
 use Give\Log\ValueObjects\LogType;
 use Give\Log\Helpers\LogTypeHelper;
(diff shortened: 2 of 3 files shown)


JasonTheAdams:
@alaca Can you please look this over? I'm almost certain I didn't use the new logging system correct. I tried a number of times to understand how to use it, but it's pretty confusing. Can we please update the `Log` class to include some examples in the docblock on how the class should be used?

alaca:
@JasonTheAdams  in 99% of situations we should use `Log::LOG_TYPE( $message, $context = [] );` but in this particular case we are hooking into the existing logging system,  therefore it's much convenient to use `LogFactory` than `Log` class. My mistake was I didn't wrap it in a `try...catch` block.

alaca (approved):

JasonTheAdams:
Thank you, @alaca. I would still like to update the Log class docblock to include an example of how to use the class. The context array is confusing as I know it has certain keys that are "special" and others that get added to the actual context area of the log. We need to document this.

alaca:
@JasonTheAdams yes, the context array is not very intuitive. I'll update the docblock later today.

PR #5800, opened 2021-04-20
Donor Dashboard Logout now works while in the same session

### LINKED ISSUE #5799: Donor Dashboard Logout should work while in the same session
## User Story

## Details

From Canny: 

> Currently if you create a donation on a form without WordPress account creation, and then navigate to the donor dashboard, it's impossible to log out of the donor dashboard. The logout button essentially does nothing.
> The good news is that if a subsequent donation is made in the same browser, those two donations remain separate and attached to a different donor. But the logout interface is confusing, and should be fixed.

## Expected Behavior

Logout button should destroy the Give Session, and not require that a user is logged in with a WP account to have an affect.

## Steps to Reproduce

In an incognito window, to avoid existing cookies
1. Create a donation under a new email (without an associated WP account)
2. Navigate to the Donor Dashboard page
3. See that the Donor Dashboard loads as expected
4. Click "Logout"
5. Notie that the dashboard does not log out

## Visuals

N/A

## Acceptance Criteria

-   [x] Logout button successfully logs donor out of Donor Dashboard (regardless of if they are logged in via a session, or WP account)

henryholtgeerts:
Resolves #5799 

## Description

This PR introduces logic to ensure that the current Give Session is consistently destroyed when a Donor clicks the "Logout" button in the Donor Dashboard. Previously, the current Give Session was only destroyed for Donors logged in as WP users, now the Give Session is destroyed even when a donor is not logged in as a WP user.

## Affects

This PR affects logout route logic for the Donor Dashboard.

## Visuals

N/A

## Testing Instructions

1. Create a donation under a new email (without an associated WP account)
2. Navigate to the Donor Dashboard page
3. See that the Donor Dashboard loads as expected
4. Click "Logout"
5. Notice that the dashboard logs out correctly

## Pre-review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+-   Donor Dashboard Logout now works while in the same session (#5800)
+
 ## 2.10.2 - 2021-04-14
 
 ### Changed
--- a/includes/class-give-session.php
+++ b/includes/class-give-session.php
@@ -515,6 +515,7 @@ class Give_Session {
 	 */
 	public function destroy_session() {
 
+		give_setcookie( 'give_nl', '', time() - YEAR_IN_SECONDS, apply_filters( 'give_session_use_secure_cookie', false ) );
 		give_setcookie( $this->cookie_name, '', time() - YEAR_IN_SECONDS, apply_filters( 'give_session_use_secure_cookie', false ) );
 		give_setcookie( $this->nonce_cookie_name, '', time() - YEAR_IN_SECONDS, apply_filters( 'give_session_use_secure_cookie', false ) );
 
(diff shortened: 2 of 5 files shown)


JasonTheAdams (changes requested):
Great work, @henryholtgeerts! It's now logging out properly when accessed via email. I noticed an odd bug, however, when logged in as a WP user. When I go to log out, I actually have to do it twice. The first time it refreshes the page but I'm still logged in, but if I log out again then it works. Sounds like something in the Session or whatever isn't being fully cleared the first time around.

henryholtgeerts:
@JasonTheAdams Ah, this is looks like it is the result of some mis-handled merge conflict (my fault). There is a line that had been removing all actions before firing wp_logout, which prevented the destroy_session call from being made at the right time. I've removed that line, and it is working again as expected. Ref: https://github.com/impress-org/givewp/pull/5800/commits/5248af67dccba5923376bccaef7bf9e7cf3ea092 👍

JasonTheAdams (approved):
Sweet! Logging out is fun again!
Fun again

PR #5805, opened 2021-04-27
Add context to PhpStorm to resolve objects to their proper class types

### LINKED ISSUE #5806: PhpStorm should be able to know give( The_Class::class ) is an object of The_Class::class
## Details

PhpStorm has problems knowing what generalized container objects are when they are returned.

## Acceptance Criteria

-   [ ] When I type `give( The_Class::class )->` in PhpStorm IDE
-   [ ] Then I should start to see the autocomplete based on the properties/methods available in `The_Class`.

sc0ttkclark:
Resolves #5806

## Description

This follows the work done in https://github.com/the-events-calendar/tribe-common/pull/1579 which implements this for our containers in Tribe Common, thought this would be handy for Give since you also use containers!

Allow PHPStorm Editor to resolve the different types when using the container to get an instance to an object, this works only when the container is called with a string that is either the full qualified class of the object or when the name is used on the container such as:

```php
give( Object_Type::class )->resolves_names_of_method_or_properties();
give( 'Object_Type' )->resolves_names_of_method_or_properties();
```

## Affects

Dev-only improvements for those using PhpStorm to contribute.

## Testing Instructions

1. Type `give( The_Class::class )->` in PhpStorm IDE
2. It should start to show the autocomplete options based on the properties/methods available in The_Class

## Pre-review Checklist

-   [ ] ~~Relevant `@since` tags included in DocBlocks~~
-   [ ] ~~Includes unit and/or end-to-end tests~~

### DIFF
--- a/.phpstorm.meta.php
+++ b/.phpstorm.meta.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace PHPSTORM_META {
+	// Allow PHP Storm Editor to resolve return types when calling give( Object_Type::class ) or give( `Object_type` )
+	override(
+		\give( 0 ),
+		map( [
+			'' => '@',
+			'' => '@Class',
+		] )
+	);
+}
--- a/.phpstorm.meta.php
+++ b/.phpstorm.meta.php
@@ -1,7 +1,7 @@
 <?php
 
 namespace PHPSTORM_META {
-	// Allow PHP Storm Editor to resolve return types when calling give( Object_Type::class ) or give( `Object_type` )
+	// Allow PhpStorm IDE to resolve return types when calling give( Object_Type::class ) or give( `Object_Type` ).
 	override(
 		\give( 0 ),
 		map( [
--- a/give.php
+++ b/give.php
@@ -492,7 +492,7 @@ final class Give {
  *
  * @return object|Give
  */
-function Give( $abstract = null ) {
+function give( $abstract = null ) {
 	static $instance = null;
 
 	if ( $instance === null ) {
@@ -508,4 +508,4 @@ function Give( $abstract = null ) {
 
 require __DIR__ . '/vendor/autoload.php';
 
-Give()->boot();
+give()->boot();


JasonTheAdams:
Thanks, @sc0ttkclark! Out of curiosity, in The Events Calendar, is this something you're planning on including in releases? Or stripping? I'm more than happy to include this for GiveWP development, but on the fence as to whether I want to include PhpStorm meta with releases. 🤔

sc0ttkclark:
It's something I assume will be excluded from releases. I don't think it would make sense to release it in your case since the Give plugin itself doesn't include other dev centric tooling. But totally open to your thoughts here as it could sway my own position :)

kjohnson:
Very cool! I didn't know that was a possibility. 🙌🏻

JasonTheAdams:
@sc0ttkclark As far as I can tell, it comes down to this:

**Do** include it in releases:
👍 3rd party themes and plugins get the benefit of auto-complete without having to use the repo version of GiveWP
👎 We're starting to introduce dev files to releases, which we haven't done

**Do not** include it in releases:
👍 Releases are pure and simple, only including what's necessary
👎 No auto-complete for folks unless they use the repo version

For us, and this could be different for you, I'm not at this point really encouraging too much use of 3rd parties to utilize the `give()` method and underlying service container. In fact, I'll likely make it such that `give(InternalClass::class)` is intended for internal use, whereas public API's will be offered as `give()->publicThing` — at which point we can use the docblock for type-hinting.

Ok, I think I've convinced myself to _not_ include this in releases. Thanks, Scott! I'll test this out and merge it.

JasonTheAdams:
@sc0ttkclark Out of curiosity, did you have to flush your caches or something to get this to work? I'm trying to test this out and I can't seem to get it to work. 🤔

sc0ttkclark:
It worked almost immediately for me. I'm on a new machine now but I'll try it out on Monday if that's fine with you. I'll confirm steps if there's anything, but invalidating cache and reloading from disk would be a good idea.

kjohnson:
@JasonTheAdams per our discussion in Slack, I have updated the definition of the `give()` function to be lowercase, see https://github.com/impress-org/givewp/pull/5805/commits/7894557edc520768cfa97e04844e6618b6980590. It seems that the override function name must match case, even though user defined function names in PHP are supposed to be case-insensitive - either way updating the function name to lowercase matches convention.

JasonTheAdams (approved):
This will be great! Glad we figured out the config issue, and I've wanted to change the function name to `give` for a while to follow convention.

PR #5827, opened 2021-05-12
Support for the decimal amounts is now implemented

### LINKED ISSUE #5815: The multi-step donation form should allow decimal numbers for the custom donation amount
## User Story

As a donor, I want to use decimal numbers in my new multi-step donation form, such as `$14.70`

## Details

If a donor enters a custom donation amount `$14.70`, the donation amount will be rounded up to `$15`. This happens only when the multi-step form template is being used. The legacy form template is not affected by this bug. 

**UPDATE:**
This behavior of the donation amount input is created on purpose to satisfy the design requirements of the multi-step form template. 

## Expected Behavior

The donor should be able to make donations using decimal numbers for donation amounts when the multi-step form template is being used.

## Steps to Reproduce

1. Create a new donation form that uses the multi-step template
2. Navigate to that form on the front-end 
3. Enter 14.90 for the donation amount, or any other decimal number and click somewhere else outside the form. 
4. Donation amount should be rounded up or down

## Visuals

deepin-screen-recorder_Select area_20210505174355

## Acceptance Criteria

-   [x] Donor can use decimal numbers for the donation amount when donating

alaca:
Resolves #5815

## Description

This PR adds an option in the Donation Form Options section for the site admin to enable or disable decimal amounts when the Multi-step template is used. This option is disabled by default. 

## Affects

Multi-step form donation amount input field and donation level amounts.

## Visuals

| Option Location | Donation form - Option Disabled | Donation form - Option Enabled |
|---|---|---|
| image | image | image |

## Testing Instructions

1. Create a new donation form that uses multi-step form template
2. Enable/Disable _Decimal Amounts_ option under the _Step 2: Payment Amount_
3. Viewdonation  form on the frontend

**Note:** 
If wording used to describe the new option doesn't make sense, please update it! 

## Pre-review Checklist

### DIFF
--- a/assets/src/js/plugins/give-api/api.js
+++ b/assets/src/js/plugins/give-api/api.js
@@ -326,12 +326,12 @@ const Give = {
 		},
 
 		/**
-		 * Helper function used to determine if the given value is a float
+		 * Helper function used to determine if the given number has decimal
 		 * @param value
 		 * @returns {boolean}
 		 */
-		isFloat: function( value ) {
-			return Number( value ) === value && value % 1 !== 0;
+		numberHasDecimal: function( value ) {
+			return Math.floor( value ) !== Number( value );
 		}
 	},
 
--- a/assets/src/js/plugins/give-api/api.js
+++ b/assets/src/js/plugins/give-api/api.js
@@ -324,6 +324,15 @@ const Give = {
 			}
 			return url;
 		},
+
+		/**
+		 * Helper function used to determine if the given value is a float
+		 * @param value
+		 * @returns {boolean}
+		 */
+		isFloat: function( value ) {
+			return Number( value ) === value && value % 1 !== 0;
+		}
 	},
 
 	/**
--- a/src/Views/Form/Templates/Sequoia/assets/js/form.js
+++ b/src/Views/Form/Templates/Sequoia/assets/js/form.js
@@ -455,6 +455,10 @@
 	setupFFMInputs();
 	setupInputIcons();
 
+	if ( 'enabled' === templateOptions.payment_amount.decimals_enabled ) {
+		updateFormDonationLevelsLabels();
+	}
+
 	/**
 	 * Limited scope of optional input labels, specifically to User Info, see issue #5160.
 	 */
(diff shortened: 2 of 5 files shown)


assets/src/js/plugins/give-api/api.js
```diff
@@ -324,6 +324,15 @@ const Give = {
...
 		},
+
+		/**
+		 * Helper function used to determine if the given value is a float
+		 * @param value
+		 * @returns {boolean}
+		 */
+		isFloat: function( value ) {
+			return Number( value ) === value && value % 1 !== 0;
+		}
```
JasonTheAdams, line 335: I suggest coming about this slightly differently. In JS _all_ numbers are floats. That is, the only numeric primitive is Number, and Number is a 64-bit floating point value. So, really, all numbers in JS are floats. I believe what we're really checking for here is whether or not a Number has decimals:
```js
numberHasDecimals: function( value ) {
	return Math.floor( value ) === Number( value );
}
```

alaca: @JasonTheAdams you are right, all numbers in JS are floats, so this function doesn't make much sense. I'll update this as you suggested.

src/Views/Form/Templates/Sequoia/assets/js/form.js
```diff
@@ -849,4 +853,36 @@
...
+			const donationLevels = Give.form.fn.getVariablePrices( donationForm );
+
+			$.each( donationLevels, function( j, level ) {
+				if ( 'custom' === level.price_id ) {
+					return;
+				}
+
+				const symbol = Give.form.fn.getInfo( 'currency_symbol', donationForm );
+				const position = Give.form.fn.getInfo( 'currency_position', donationForm );
+				const precision = Give.form.fn.getInfo( 'number_decimals', donationForm );
```
kjohnson, line 874: @alaca These three variables could be defined outside of the loop and re-used for reach iteration.

src/Views/Form/Templates/Sequoia/optionConfig.php
```diff
@@ -104,6 +104,17 @@
 				],
 				'default'    => __( 'Continue', 'give' ),
 			],
+			[
+				'id'      => 'decimals_enabled',
+				'name'    => __( 'Decimal amounts', 'give' ),
+				'desc'    => __( 'Do you want to enable decimal amounts?', 'give' ),
```
kjohnson, line 110: I think we can make the description text more helpful. In this case, when the setting is disabled decimal values are rounded. That is probably helpful information.

src/Views/Form/Templates/Sequoia/assets/js/form.js
alaca: @kjohnson you are right! 

Resolved 7c5a377c

src/Views/Form/Templates/Sequoia/optionConfig.php
alaca: Resolved: 1647d254

kjohnson (approved):

PR #5837, opened 2021-05-17
Catch uncaught exceptions and log them

### LINKED ISSUE #5836: GiveWP should log uncaught exceptions
## Details

As we continue to increase our use of exceptions within our codebase, we run into the issue that, presently, our exceptions are only logged if we specifically catch them. We should come up with a way to catch uncaught exceptions such that we log all exceptions, even if they're not explicitly caught. This will help us to better troubleshoot customer scenarios.

## Acceptance Criteria

-   [x] Uncaught exceptions are logged

JasonTheAdams:
Resolves #5836 

## Description

This PR tackles the issue that exceptions in GiveWP are otherwise unlogged if not explicitly caught. It does this in a couple ways:

First, it registers with [set_exception_handler](https://php.net/manual/en/function.set-exception-handler.php) to handle any uncaught exceptions. If a previous handler was registered (by another plugin, theme, or whatever) then it passes the Exception along.

It introduces a `LoggableException` interface which helps to identify an exception as coming from GiveWP, and gets data from the exception in a way that gets logged nicely into our system. I also provided a `Loggable` trait that provides the methods as a quick applier. Since traits can't implement interfaces, this is by convention.

When an uncaught exception occurs, the handler checks to see if the Exception is native to GiveWP. If it is, then it logs the exception.

## Affects

Anywhere exceptions were thrown within GiveWP

## Testing Instructions

Just throw a random exception somewhere in the code that uses the `Loggable` trait or `LoggableException` interface.

## Pre-review Checklist

### DIFF
--- a/give.php
+++ b/give.php
@@ -40,6 +40,7 @@
  */
 
 use Give\Container\Container;
+use Give\Framework\Exceptions\LogUncaughtGiveExceptions;
 use Give\Framework\Migrations\MigrationsServiceProvider;
 use Give\Form\Templates;
 use Give\Revenue\RevenueServiceProvider;
@@ -223,6 +224,7 @@ final class Give {
 		$this->bindClasses();
 
 		$this->loadServiceProviders();
+		$this->setupExceptionHandler();
 
 		// Load form template
 		$this->templates->load();
@@ -249,6 +251,7 @@ final class Give {
 	private function bindClasses() {
 		$this->container->singleton( 'templates', Templates::class );
 		$this->container->singleton( 'routeForm', FormRoute::class );
+		$this->container->singleton( LogUncaughtGiveExceptions::class );
 	}
 
 	/**
@@ -472,6 +475,18 @@ final class Give {
 	 */
 	public function __call( $name, $arguments ) {
 		return call_user_func_array( [ $this->container, $name ], $arguments );
+	}
+
+	/**
+	 * Sets up the Exception Handler to catch and handle uncaught exceptions
+	 *
+	 * @unreleased
+	 */
+	private function setupExceptionHandler() {
+		/** @var LogUncaughtGiveExceptions $handler */
+		$handler = $this->container->make( LogUncaughtGiveExceptions::class );
+
+		$handler->setupExceptionHandler();
 	}
 }
 
--- a/src/Container/Exceptions/BindingResolutionException.php
+++ b/src/Container/Exceptions/BindingResolutionException.php
@@ -3,7 +3,8 @@
 namespace Give\Container\Exceptions;
 
 use Exception;
+use Give\Framework\Exceptions\Traits\Loggable;
 
 class BindingResolutionException extends Exception {
-	//
+	use Loggable;
 }
--- a/src/Framework/Exceptions/Contracts/LoggableException.php
+++ b/src/Framework/Exceptions/Contracts/LoggableException.php
@@ -1,0 +1,25 @@
+<?php
+
+declare( strict_types=1 );
+
+namespace Give\Framework\Exceptions\Contracts;
+
+interface LoggableException {
+	/**
+	 * Returns the human-readable message for the log
+	 *
+	 * @unreleased
+	 *
+	 * @return string
+	 */
+	public function getLogMessage();
+
+	/**
+	 * Returns an associated array with additional context for the log
+	 *
+	 * @unreleased
+	 *
+	 * @return array
+	 */
+	public function getLogContext();
+}
--- a/src/Container/Container.php
+++ b/src/Container/Container.php
@@ -6,8 +6,8 @@ use ArrayAccess;
 use Closure;
 use Exception;
 use Give\Container\Exceptions\BindingResolutionException;
-use InvalidArgumentException;
-use LogicException;
+use Give\Framework\Exceptions\Primitives\InvalidArgumentException;
+use Give\Framework\Exceptions\Primitives\LogicException;
 use ReflectionClass;
 use ReflectionException;
 use ReflectionParameter;
@@ -544,7 +544,7 @@ class Container implements ArrayAccess {
 	 * @param string $id Identifier of the entry to look for.
 	 *
 	 * @return mixed Entry.
-	 * @throws Exception
+	 * @throws InvalidArgumentException|BindingResolutionException
 	 */
 	public function get( $id ) {
 		try {
--- a/src/Controller/PayPalWebhooks.php
+++ b/src/Controller/PayPalWebhooks.php
@@ -2,7 +2,7 @@
 
 namespace Give\Controller;
 
-use Exception;
+use Give\Framework\Exceptions\Primitives\Exception;
 use Give\PaymentGateways\PayPalCommerce\Repositories\MerchantDetails;
 use Give\PaymentGateways\PayPalCommerce\Repositories\Webhooks;
 use Give\PaymentGateways\PayPalCommerce\DataTransferObjects\PayPalWebhookHeaders;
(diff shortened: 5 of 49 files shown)


JasonTheAdams:
@kjohnson This is ready for review! Let me know what you think!

give.php
```diff
@@ -249,6 +252,7 @@ public function init() {
 	private function bindClasses() {
 		$this->container->singleton( 'templates', Templates::class );
 		$this->container->singleton( 'routeForm', FormRoute::class );
+		$this->container->singleton( LogUncaughtGiveExceptions::class );
```
kjohnson, line 255: Is there a particular reason why this is named as a verb, instead of a noun, ie `UncaughtExceptionLogger`?

src/Framework/Exceptions/Contracts/LoggableException.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+declare( strict_types=1 );
+
+namespace Give\Framework\Exceptions\Contracts;
+
+interface LoggableException {
```
kjohnson, line 7: This seems like an unnecessary redundancy, if the `Loggable` trait is what is being used. The trait serves as a contract. It seems that we should have a preferred way instead of increasing the complexity.

The only application for a redundant interface that stands out is allowing a class to provide its own implementation. In which case, the class could override the trait methods.

kjohnson (commented):
@JasonTheAdams a couple of discussion points.

Also, I'm assuming that the unit tests are _in progress_ and not just missing, right? 😉

give.php
JasonTheAdams: Because when I wrote it I wasn't sure if I going making an Action class or not. It's not, so I'll rename. 😃

src/Framework/Exceptions/Contracts/LoggableException.php
JasonTheAdams: Yeah, I thought you might catch that. I really prefer Interfaces to signify contracts. I learned in the process of doing this that Traits can't implement Interfaces. I wanted the Interface to be the contract, and the Trait to be the helper. 

You can even see that when checking if a class is GiveWP's, I'm checking for both — this wasn't my intended behavior. I'm tempted to remove the trait check and add the Interface to the primitive exceptions and other native GiveWP exceptions.

What do you think?

kjohnson: I was also considering removing the trait (`composition`) in favor of the interface (`inheritance`).

The interface is better for enforcing the contract and code re-use can be achieved with a base (or abstract) class.

The trait would come into play to separate the concerns of the implementation itself, if such separation were deemed necessary.

JasonTheAdams: The biggest reason for the Trait is the Primitive exceptions. They _have_ to extend the base PHP exception classes they extend. So I make a base `LoggedException` class for them to use. This also comes into play if someone wants to extend one of those classes, as well. I'm ok with thinking of the Interface as the contract and the Trait as a compositional helper.

give.php
JasonTheAdams: Resolved in 0e71cc4

src/Framework/Exceptions/Contracts/LoggableException.php
JasonTheAdams: Tentative resolution in f83c8f6

give.php
```diff
@@ -249,6 +252,7 @@ public function init() {
 	private function bindClasses() {
 		$this->container->singleton( 'templates', Templates::class );
 		$this->container->singleton( 'routeForm', FormRoute::class );
+		$this->container->singleton( UncaughtExceptionLogger::class );
```
JasonTheAdams, line 255: @kjohnson I keep debating with myself as to whether or not I really want this to go through the Service Container. I set this whole thing up before the Service Providers are registered and booted, so we couldn't use DI in this class anyway. It's also meant to be a very simple, standalone class. I'm 80% convinced to just make this a good 'ol `new UncaughtExceptionLogger` case, unless you think otherwise.

kjohnson: Yeah, I'm agreeable to separating this out - it is essentially a "bolt on" for isn't intended to be re-used or otherwise integrated.

src/Framework/Exceptions/UncaughtExceptionLogger.php
```diff
@@ -53,15 +53,6 @@ public function handleException( Exception $exception ) {
...
-			return true;
-		}
-
-		$traits = class_uses( $exception );
-		if ( isset( $traits[ Loggable::class ] ) ) {
-			return true;
-		}
-
-		return false;
+		return $exception instanceof LoggableException;
```
kjohnson, line 56: As this is simplified, it may not be necessary to have a separate method. "If the exception is an instance of LoggableException" is self-explanatory - or at least as readable as `isGiveException` (which isn't necessarily accurate at this point, rather the point is that the exception is loggable - and now I'm rambling...).

src/Container/Exceptions/BindingResolutionException.php
```diff
@@ -3,8 +3,9 @@
 namespace Give\Container\Exceptions;
 
 use Exception;
+use Give\Framework\Exceptions\Contracts\LoggableException;
 use Give\Framework\Exceptions\Traits\Loggable;
 
-class BindingResolutionException extends Exception {
+class BindingResolutionException extends Exception implements LoggableException {
 	use Loggable;
```
kjohnson, line 10: Since we are using the `LoggableException` interface as the single type check, does the `Loggable` trait now represent added complexity in that each class is using a trait to implement an interface? Would this be better with `inheritance`, such as a base or abstract class?

JasonTheAdams: Please see my note, [here](https://github.com/impress-org/givewp/pull/5837#discussion_r634511688), as to why inheritance won't work.

src/Framework/Exceptions/Contracts/LoggableException.php
JasonTheAdams: As a note, @kjohnson, what I'm introducing here _is compositional_ by nature. Whether an exception is loggable or not is pluggable, and shouldn't affect it's inheritance.

kjohnson: Cool, yeah, I can get behind that 👍🏻

src/Framework/Exceptions/UncaughtExceptionLogger.php
JasonTheAdams: Cleaned up in b411819

give.php
JasonTheAdams: Great! Cleaned up in a42a80b

JasonTheAdams:
@kjohnson I tried to add some testing, but I'm running into an issue I could use your help with. As you can see, I adjusted the `Log` class such that it's now passing through the Service Container. This was done so I could mock it. The failure is that it's saying the `Log::error` method never gets called, but I can see that it is, leading me to believe I'm either not mocking it right or the mock isn't working with the `call_user_func_array` function... or something else I'm just not thinking of.

kjohnson:
@JasonTheAdams it looks like you need to set the expectation on the returned `$mock` and pass the expected parameters as individual arguments (as opposed to an array of arguments).

```php
public function testShouldLogException() {
	$logger = new UncaughtExceptionLogger();

	$mock = $this->mock( Log::class, function(PHPUnit_Framework_MockObject_MockBuilder $builder) {
		return $builder->setMethods(['error'])->getMock();
	} );

	$mock->expects( $this->once() )
		->method( 'error' )
		->with( '', [] );

	$logger->handleException( new ExceptionLogged() );
}
```

```
..                                                                  2 / 2 (100%)

Time: 4.15 seconds, Memory: 46.50MB

OK (2 tests, 1 assertion)

```

JasonTheAdams:
Baaaahh!! I knew it was something simple!! 😂

kjohnson (approved):
LGTM 👍🏻

PR #5843, opened 2021-05-25
Refactor Fields API

### LINKED ISSUE #5844: Field api should support all field input types
## User Story

As a developer, I want the field api to support all input types so that I can use it to build custom fields.

## Details

The field api only supports a few input types.  In order to implement the api into our add-on plugins like form field manager, it needs to support all the available input types we offer & their attributes.

## Additional Context

Here's a screenshot of our current input templates.

Screen Shot 2021-05-26 at 9 16 10 AM

## Acceptance Criteria

_All Field Types are supported with proper attributes_

- [x] text
- [x] hidden
- [x] textarea
- [x] select [dropdown]
- [x] select [multi select]
- [x] radio
- [x] checkbox
- [x] file_upload
- [x] url
- [x] email
- [x] repeat
- [x] date
- [x] phone

jonwaldstein:
Resolves #5844 

## Description

This pull request began as an introduction of new field types in the Fields API for form field manager. It has evolved into a refactor of the new Fields API. For the most part the underlying data structure has remained the same and since the work was previously broken into easily composed pieces, making changes to the API was quite easy. 

```php

use Give\Framwork\FieldsAPI\Checkbox;
use Give\Framwork\FieldsAPI\Email;
use Give\Framwork\FieldsAPI\Hidden;
use Give\Framwork\FieldsAPI\Form;
use Give\Framwork\FieldsAPI\Option;
use Give\Framwork\FieldsAPI\Select;
use Give\Framwork\FieldsAPI\Text;

$form = Form::make( 'donations' );

$form->append(

    Text::make( 'firstName' )
        ->label( 'First Name' )
        ->required(),
        
    Text::make( 'lastName' )
        ->label( 'Last Name' )
        ->required(),
        
    Email::make( 'email' )
        ->label( 'Email' )
        ->required(),
        
    Select::make( 'foods' )
        ->label( 'Foods' )
        ->options(
            Option::make( 'poutine', 'Poutine' ),
            [ 'fish-n-chips', 'Fish and Chips' ],
            'sushi',
            Option::make( 'pizza' )
        ),
        
    Checkbox::make( 'pineapple-on-pizza' )
        ->label( 'Pineapple on pizza is a bad idea.' )
        ->checked(),
        
    Hidden::make( 'honeypot' )
        ->defaultValue( fakeGenerateHoneypotValue() )
        ->required()
        
);

// Do something with all of the nodes in the form (e.g. render the form).
$form->all();

// Do something with just the Field nodes (e.g. persist data).
$form->getFields();

// The form can be serialized as JSON.
json_encode( $form );
```

## Affects

In `Give\Framework\FieldsAPI`

- `FormField` replaced with `Text`, `Email`, `Url`, `Phone`, `Select`, `Checkbox`, `Radio`, `File`, `Hidden`, `Date`, `Textarea`.
- `FieldCollection` replaced with `Form` and `Group`.
- This namespace has been re-organized so that the public API sits the top-most level.

## Testing Instructions

The Give core plugin doesn’t use the FieldsAPI yet, but Text to Give does.

## Pre-review Checklist


src/Form/LegacyConsumer/Renderer.php
```diff
@@ -0,0 +1,345 @@
...
+	/**
+	 * Render the field collection
+	 */
+	public function render() {
+		$this->fieldCollection->walk(
+			function ( $field ) {
+				// Determine the needs of the field based on the type
+				$config = static::deriveConfigFromType( $field->getType() );
+
+				if ( $field->getType() === FieldTypes::TYPE_SECTION ) {
```
knowler, line 25: I was thinking that instead of a "section" type that it could be a `GroupNode` in the field collection, but the `WalkNodes::walkCollection` method "walks past `GroupNode`s," so unless that implementation was changed, "section" or "fieldset" needs to remain a field type.

knowler:
https://github.com/impress-org/givewp/blob/a25023d6dc5740fa85dd3c79e9d879948cb7fd74/src/Form/LegacyConsumer/FieldView.php#L19-L32

I don’t think that it’s possible to render a "section" or "fieldset" field type if we constrict the renderer (i.e. `FieldView::render`) to a single field at a time. It is possible to track that we are within a section and even have new sections close out the previous section, but since the renderer does not have any awareness that the field might be the last one, it will never know to close out the final section.

For example, if I were to try (and fail) to implement sections in the current `FieldView` (the following is at the start of the method):

```php
class FieldView {
    public static function render( FormField $field ) {
        // Track if this is a fieldset
        static $isFieldset = false;
        
        if ( $field->getType() === 'section' ) {
	        // If this was previously rendering a fieldset, close it off.
	        if ( $isFieldset ) {
		        echo "</fieldset>";
	        }
        
	        // Start a new fieldset
	        echo "<fieldset>";
        
	        $isFieldset = true;
        
	        return;
        }
        
        // Render other field types
```

This all works, but there is no way to know when the current section ends unless there is a new section, so the final section would not have a closing `</fieldset>` tag.

It seems to me that a consumer that supports section types would need access to the entire collection as it was rendering individual fields (or at least whether or not it is the final node in the collection). For that reason, I think that the renderer should just be given the entire collection which it can walk through itself to render the fields.

For clarity, that all is based on section types being implemented as a field type. In this case (the easiest implementation), a collection with sections in it would look like this (notice how it's flat):

```
root (GroupNode)
    text
    radio
    section                  <fieldset>
    radio
    textarea
    section                  <fieldset>
    checkbox
    checkbox
    checkbox
```

If we made it so that section fields (in FFM) actually created new `GroupNode`s within the collection, then a consumer rendering per field would make more sense and IMO the API would be clearer to use in general. Here’s a example representation of what I mean by that:

```
root (GroupNode)
    text
    radio
    section (GroupNode)       <fieldset>
        radio
        textarea
    section (GroupNode)       <fieldset>
        checkbox
        checkbox
        checkbox
```

Sorry if this is a bit of brain dump, but FFM sections are most different field type and I think that we need to decide what we are going to do with them if we are going to successfully use the Fields API in FFM.

jonwaldstein:
@knowler sections in ffm are just codename for a heading that uses a legend element.

I think your thinking of the static fieldset element that currently wraps all the fields.  Don't worry about that for now, we're still deciding what to do with it.

Screen Shot 2021-06-03 at 4 21 04 PM

knowler:
@jonwaldstein Wait, so in FFM sections are `<legend>` elements without a `<fieldset>`? That’s not valid HTML (it works, it’s just breaking [rules](https://html.spec.whatwg.org/multipage/form-elements.html#the-legend-element)).

jonwaldstein:
@knowler yeah there's currently one fieldset that wraps all the fields. You can find all this and other field structures in ffm via `class-render-form.php`.

kjohnson:
> there's currently one fieldset that wraps all the fields.

Yes. And I don't think we need that to wrap the fields. What we **do** need to do is make sure that Currency Switcher is aware of the fields, which is stored in a data attribute. As for styling we can update the CSS.

knowler:
I think that regardless of what FFM previously rendered for its "section" field:

1. If a section field type is added to the Fields API, it should be a `GroupNode` since that would make it easier for consumers to manage the fields defined in the section. I think that a good expectation for how a consumer would render this type should be a `<fieldset>` with an optional `<legend>` as a label (what I mean by "expectation" is the same thing as if we give a consumer a field with the type of `select`: it probably will render a `<select>` element — of course, consumers can do whatever they want). (Sidenote: I think that `FieldGroup` or `group` or some variant of that is a better name than "section" for this type of field).
2. I still think that when FFM registers a "section" field it probably makes sense that it be whatever the Fields API provides for a section/fieldset type. I don’t think we should add a heading type for reasons which I’ll explain.
3. If we don’t want the legacy consumer to render fieldsets for section types, then we should make it render a `<div>` with the label content instead of a `<legend>` since as previously noted, that’s invalid HTML which I think we can fix even if this is legacy. (It shouldn't be a heading element since it’d most likely be invalid HTML and it would be too much work to try and make it valid.)

I’m sorry if it seems like I’m making a big deal of this, but since the legacy consumer is using the Fields API, it cannot render any field that doesn’t have a type defined by the Fields API and since the Fields API lacks the types that FFM needs to be able to use, then we need to define them in the Fields API and that’s not something we should do carelessly. 

Every field type we create in the Fields API should have some ideal expectation for how a consumer will (roughly) render it. Consumers are free to render them however they want — and that’s a good thing — but, if we make types too ambiguous or if we make types with only legacy markup needs in mind, then we are going to find ourselves in a place wanting better defined types since consumers will render fields in unexpected ways (i.e. the type was too ambiguous) or hyper-specific ways (i.e. the type was too specific), and that will make it harder to build on the API in the future, since it could become polluted with many types which are practically the same (i.e. types were created too specifically) or it could contain types which are consumed with wildly different results (i.e. types were too ambiguous). I totally get that the legacy consumer might not be the ideal consumer, but we can’t let how it consumes the API dictate what the API looks like. If push comes to shove, the Fields API should be prioritized and the legacy consumer and FFM (or other addons) should work around that (e.g. that might mean changing how it renders something or registers a field).

If we aren’t going to take the time to carefully make changes to the Fields API so that FFM can use it, I don’t think we should continue updating FFM to use the Fields API. It would be self-defeating.

kjohnson:
Looking at the FFM implementation of the `section` type, it appears to be _essentially_ a pre-build HTML field. So we can probably just treat it as an opinionated HTML field - the Fields API and Legacy Consumer will need an HTML field type anyway.

| Builder | Form |
| --- | --- |
| image | image |

 

/cc @knowler

knowler:
@kjohnson That makes sense as an alternative solution than adding a group field (which I think is what requires much more thought).

kjohnson:
>adding a group field to the Fields API - which I think is what requires much more thought

@knowler I agree. There is a lot to be considered with actually grouping fields together, but I don't think we are at that cross-roads, yet. Thanks for putting thought into this, it will come up again! Maybe sooner rather than later.

knowler:
@jonwaldstein @ravinderk All the fields register in FFM and display via the Legacy Consumer as discussed yesterday.

knowler:
Either there need to be first class support for field width inside the Fields API or we need to create a sort of backdoor which the legacy consumer can use to set stuff that there is no API for. For example, I am thinking of adding field width as `__unsafe_legacy_consumer_field_width` attribute which the legacy consumer can pop off the `$attributes` array and use for setting that.

I think that sort of back door seems fine right now. What do you think @JasonTheAdams @kjohnson?

knowler:
@kjohnson Can you review [these changes](https://github.com/impress-org/givewp/compare/0426a5b84daf...f4ba33a)? I believe this reflects all we discussed.

`FormField::options` now supports:

- Arrays, e.g. `['Pizza', 'Tacos', 'Sushi']`.
- Associative arrays, e.g. `['pizza' => 'Pizza', 'tacos' => 'Tacos', 'sushi' => 'Sushi']`.
- An array of entries/tuples (`$value, $label`), e.g. `[['pizza', 'Pizza'], ['tacos', 'Tacos'], ['sushi', 'Sushi']]`.
- Or an array of `FieldOption`s, e.g. `[new FieldOption('pizza', 'Pizza'), new FieldOption('tacos', 'Tacos'), new FieldOption('sushi', 'Sushi')]`.

All of these store on the object as `FieldOption[]`.

src/Framework/FieldsAPI/FormField/FieldOption.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace Give\Framework\FieldsAPI\FormField;
+
+/**
+ * Class FieldOption
+ *
+ * @unreleased
+ */
+class FieldOption {
```
JasonTheAdams, line 10: Implement the [JsonSerializable](https://www.php.net/manual/en/class.jsonserializable.php) interface and tweak the method name (camelcase), then it will be official. 😎

knowler: Can do! The `FormField` class is missing those as well.

knowler:
> @knowler Our intention is to implement Groups for columns.
> 
> For example, First Name and Last Name would be in a group (a nested field collection), which the Consumer can decide how to render, most likely as columns.
> 
> ```
> ├── Collection
> │   ├── Nested Collection
> │       ├── First Name
> │       └── Last Name
> │   ├── Email
> │   └── Donation Amount
> ```
> 
> I'd prefer to avoid setting width _specifically_ but rather rely on inferring the width organically.
> 
> That said, I see that FFM supports Full, Half, One-Third, Two-Thirds. To make that work, I think we can use the nested field collection method above and programmatically infer columns based on width by being _a little_ clever. I theory, people are building columns (as opposed to jagged layouts), so we should be able to segment fields based on width and add them to nested collections. (Does that make sense?)

@kjohnson This is clever, and I went back and forth on it, however, I don’t think it will work. Right now, FFM allows users to set the width per field which means that users are not constrained to strict layouts per row. For example, if a user creates a field with the width set to a third, but then sets the next field’s width to a half, and then the next to two thirds, how should the fields be registered? Those column widths fit into two rows. If we created collections based on the first field’s width setting, then the latter two field’s width setting would be ignored. Further, using the same example, say that the last field was omitted, then if the field collection had one less field, the consumer would infer a two column layout for the row.

With that being said, I still think the suggestions I made in https://github.com/impress-org/givewp/pull/5843#issuecomment-858958281 would work. I prefer the temporary solution for the task at hand though, since a change to the Fields API requires more thought. 

Another idea I had was being able to provide hints for what an attribute was provided for. For example, a _root_ `class` attribute could provide the consumer hints that these attributes should be placed on the root element of the field or an _input_ `class` attribute could hint that the attributes were for the input element.

src/Framework/FieldsAPI/FormField/FieldOptions.php
```diff
@@ -2,29 +2,107 @@
...
+
+		// Loop through the options and transform them to the proper format.
+		foreach ( $options as $maybeLabel => $fieldOptionOrArrayOrValue ) {
+			if ( $fieldOptionOrArrayOrValue instanceof FieldOption ) {
+				// Since the given format matches the proper format (i.e. an array of FieldOption),
+				// we can avoid additional cycles of the loop by just setting the value and breaking
+				// out of it early.
+				$this->options = $options;
+				break;
+			}
```
JasonTheAdams, line 50: Let's allow for a mixed bag. Just append this FieldOption and move on to the next.

src/Framework/FieldsAPI/FormField/FieldOption.php
knowler: Oh, I see that’s because `Node` implements it.

knowler: Resolved in 192196c

src/Framework/FieldsAPI/FormField/FieldOption.php
```diff
@@ -0,0 +1,55 @@
...
+	 */
+	public function getLabel() {
+		return $this->label;
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function jsonSerialize() {
+		return (object) [
```
JasonTheAdams, line 50: You don't need to cast as an object for this interface.

knowler: Resolved in 6933915

src/Framework/FieldsAPI/FormField/FieldOptions.php
knowler: Resolved in 7d94367

src/Framework/FieldsAPI/FormField/FieldOptions.php
```diff
@@ -2,29 +2,90 @@
...
+	 */
+	public function options( array $options ) {
+		// Reset options, since they are meant to be set immutably
+		$this->options = [];
+
+		// Determine if the provided options are an associative array before the loop to avoid extra work.
+		$areOptionsAnAssocArray = Arr::isAssoc( $options );
+
+		// Loop through the options and transform them to the proper format.
+		foreach ( $options as $maybeLabel => $fieldOptionOrArrayOrValue ) {
```
alaca, line 43: @knowler I'm implementing your changes into P2P, great work btw, but I have one question. Shouldn't the current option key be the option value instead of label? I'm asking because I will have to swap the [value with label](https://github.com/impress-org/give-peer-to-peer/blob/5bc87e8bd8fa3c53e34410fd53e321f17084c647/src/P2P/Admin/Settings/CampaignSettings.php#L80) in order to display the value and label correctly on P2P.

knowler: Nice catch. Yep, that's how it should be and even how I wrote the example. I'll fix that here.

alaca: Awesome, thanks!

knowler: Resolved in 8dd3b29

src/Framework/FieldsAPI/FormField/FieldOptions.php
```diff
@@ -2,29 +2,90 @@
...
+				// In this case, what is provided matches the proper format, so we can just append it.
+				$this->options[] = $value;
+			} elseif ( is_array( $value ) ) {
+				// In this case, what has been provided in the value is an array with a value then a label.
+				// This matches the constructor of `FieldOption`, so we can unpack it as arguments for a new instance.
+				$this->options[] = new FieldOption( ...$value );
+			} elseif ( $areOptionsAnAssocArray ) {
+				// In this case, the array is associative, therefore, we expect the key to be the option’s value
+				// and the value to be the option’s label.
+				$this->options[] = new FieldOption( $key, $value );
```
JasonTheAdams, line 54: @knowler Let's just drop this. I'd rather not have too many options for... options. The class method is obvious, the list method is quick and clear, and the tuple method is robust. The associate method doesn't add anything and has limitations.

knowler: Sounds good. I agree that though associative arrays _seem_ like a natural fit for options, their limitations create a number of issues, and ultimately they prove to be not a good fit.

knowler: Resolved by 99ce87f

kjohnson:
@knowler don't forget to update the tests for the Fields API Options feature, see `tests/unit/tests/Framework/FieldsAPI/FormField/FieldOptionsTest.php`

We should be sure to cover the different accepted argument types.

src/Framework/FieldsAPI/FieldCollection/Contract/GroupNode.php
```diff
@@ -4,4 +4,8 @@
 
 interface GroupNode extends Node {
 	public function getFields();
+	public function append( Node $node );
+	public function getNodeIndexByName( $name );
+	public function getNodeByName( $name );
+	public function count();
```
knowler, line 10: Here’s the code that inspired this change:

https://github.com/impress-org/givewp/blob/d0176e99b0cc46a5f65d9d62eb905aae5d2ed483/src/Framework/FieldsAPI/FieldCollection.php#L53-L55

Before, instances of `GroupNode`s which did not have a `getNodeByName` method would break here. The quickfix would have just been to change `GroupNode` to `FieldCollection` in the condition, however, I thought it would be more beneficial to add the method in question as well as a few other desirable methods to the `GroupNode` contract.

kjohnson: @knowler Yeah, I think this makes sense and was the original intent that `GroupNode`s be recursively searchable. Updating the check to `FieldCollection` would have been incorrect - good call.

knowler:
Jason, Kyle, and myself discussed some design challenges we were facing with the Fields API and decided on the following course of action in order to improve it. A number of these changes will be breaking.

We will be **creating a class for each field type in place of the generic `FormField` class which is currently implemented.** The reason for the prior implementation is still valid: inheritance _can_ create a nightmare, especially when things slightly differ. We will be keeping an eye to this wisdom as we move forward, however, we desire to replace this first implementation since we feel that the generic `FormField` class is too ambiguous and it can be used in undesirable ways (e.g. a text field can set options since `FormField` is using the `FieldOptions` trait). Luckily, Kyle implemented said class using traits which we can use to compose the new classes. These classes will only use the features they need which we believe will lead to more predictable feels as well as a better developer experience.

We will be **removing the `FieldAttributes` functionality** for a number of reasons that I will outline below:

First, it is unclear if the attributes are meant to be used as HTML attributes or if they are meant to be generic attributes for the field. This opens the door for consumers to use this property in varying and unexpected ways.

Second, if they are HTML attributes it is likely that the user will expect them to be set on the `input`, `textarea`, or `select` element. This is problematic for the `class` attribute since in many cases we expect consumers to not just be rendering a single element — the input might be nested in the layout markup that supports it. If the class attribute is expected to be placed on the input element like the rest of the attributes, then we greatly limit the user’s ability to style anything else that is rendered for the field (e.g. the wrapper, the label, help text, etc.). Conversely, if we separate this attribute from the rest, so that consumers can place it on the topmost element it is rendering while including the rest of the input element, then we betray the Fields API user’s expectation of setting attributes for the field (i.e. attributes are not just HTML attributes for the input element).

Third, the Fields API should be declarative as possible. We want to avoid including instructions for specific consumers types. Not all consumers are rendering HTML and, for those which are, we have no idea what that HTML looks like. We are not creating a subset of the DOM. It is better to clearly describe what a field is and then let the consumer translate that to its given medium (e.g. HTML, JSON, etc.).

Moving forward, we have found it helpful to separate attributes/properties into a few categories:

- **Meta**: properties which describe a field’s identity.
    - `name`, `value`, `type` (though see the last sub-point of the next point), `label`, `helpText`, `placeholder`.
- **Validation**: properties which describe constaints for a field’s value.
    - HTML form field attributes such `required`, `minlength`, `maxlength`, `min`, `max`, `multiple`, `pattern`, `accept`, etc. fit this category. Note that we might use more generic names for these in the actual API.
    - `options` fits this category since it describes the possible valid values.
    - I’d argue that `type` actually fits this category since types inherently constrain their value.
- **Visibility**: properties which describe how a field can be viewed.

These categories are not gospel for creating new properties, however, they are a helpful guide.

Being able to generically describe validation is crucial. HTML validation is not the only validation that occurs. Frontend form libraries will can use the validation rules to provide further validation which HTML cannot. The backend also will use validation rules for ensuring good input.

**TL;DR: What needs to be done?**

- [ ] Add classes for each field type and remove `FormField`.
- [ ] Implement custom JSON serialization for each field type.
    - Jason discussed exploring abstract methods in traits for this.
- [ ] Remove `FieldAttributes`.
- [ ] Add a root class name property to fields.

knowler:
Commits after c20a253 will be breaking if code relies on the Fields API. Keep in mind you are using this branch.

knowler:
(I might expand more on this... but submitting for now.)

Jason and I discussed the Fields API nomenclature and ultimately settled upon
the following.

Everything is a _node_. The things are _elements_, _fields_, and _groups_.

- Group: a node that has other nodes.
- Field: a node that has data.
- Element: a node that is not a field or a group.

A _form_ is a special group that is the root node.

All nodes have a unique type that ultimately indicate what it is. An _element_
node and a _field_ node cannot use the same type (e.g. a text element and a text
field — instead for this case consider a paragraph element and a text field).

```php
$form = Form::make()               // name: root
    ->label('Donations')
    ->append(
    
        // A paragraph element (not being implemented)
        Paragraph::make('instructions')
            ->content('Select how much to donate below. 😄'),
            
        // An amount field (not being implemented)
        Amount::make('donationAmount')
            ->suggestions(10, 25, 50, 100, 250, 500)
            ->required(),
            
        // A group field
        Group::make('personalInformation')
            ->label('Personal Information')
            ->append(
                Text::make('firstName')
                    ->label('First Name')
                    ->required(),
                Text::make('lastName')
                    ->label('Last Name')
                    ->required(),
                Email::make('email')
                    ->label('Email')
                    ->required(),
                Checkbox::make('newsletterOptIn')
                    ->label('Opt-in to receive weekly newsletter'),
            ),
            
    );
    
// Get all nodes (useful for frontend-rendering contexts)
$form->all();

// Get all $fields (useful for non-frontend-rendering contexts)
$form->getFields();

// Get a field by name
$form->getField('email');
```

We also discussed and concluded that the Fields API is not for styling at all.
Instead, it is recommended that a consumer provide their own API for styling
their markup. Node names in the Fields API are unique so it is possible to
"select" a node to style which then a consumer that is rendering can receive
an additional object with the instructions for styling that node.

knowler:
@JasonTheAdams this is ready for review. I will take care of outstanding documentation and testing needs in a separate PR.

JasonTheAdams (changes requested):
A _lot_ of excellent work here, @knowler! I left some comments, but the vast majority of this is awesome. Happy to talk over some items. Let's work together to get this merged tomorrow.

src/Form/LegacyConsumer/Commands/SetupFieldConfirmation.php
```diff
@@ -32,20 +32,20 @@ public function __invoke( $hook ) {
 
 		$formID = give_get_payment_meta( $this->payment->ID, '_give_payment_form_id' );
 
-		$fieldCollection = new FieldCollection( 'root' );
-		do_action( "give_fields_{$hook}", $fieldCollection, $formID );
+		$form = Form::make( 'root' );
```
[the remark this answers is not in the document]
knowler: Instead of `form` should we use `collection` or `node`? Will these always be collection types?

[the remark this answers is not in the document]
JasonTheAdams: Collection, as these are just collections that will ultimately be introduced to the Form in Give 3.0. Make sense?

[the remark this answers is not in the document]
knowler: Yep, that makes sense.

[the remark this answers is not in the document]
knowler: Should we still be using the `Form` class though? Or should we use `Group` here?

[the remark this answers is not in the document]
knowler: Resolved in 532c2d8.

src/Form/LegacyConsumer/FieldView.php
```diff
@@ -2,45 +2,74 @@
...
-			ob_start();
-			include plugin_dir_path( __FILE__ ) . '/templates/label.html.php';
-			include plugin_dir_path( __FILE__ ) . '/templates/' . $field->getType() . '.html.php';
-			echo self::mergeAttributes( ob_get_clean(), $field );
+	public static function render( Node $field ) {
+		$type = $field->getType();
+
+		//echo "<pre>";
+		//echo esc_attr( json_encode($field, JSON_PRETTY_PRINT) );
+		//echo "</pre>";
```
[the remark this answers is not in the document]
knowler: Resolved in 58c5da9.

src/Framework/FieldsAPI/Concerns/AppendNodes.php
```diff
@@ -0,0 +1,18 @@
...
+namespace Give\Framework\FieldsAPI\Concerns;
+
+use Give\Framework\FieldsAPI\Contracts\Node;
+
+trait AppendNodes {
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function append( Node ...$nodes ) {
```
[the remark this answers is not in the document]
knowler: No I don't think so. I moved it into the `InsertNode` trait: de8cc48

src/Framework/FieldsAPI/Concerns/AppendNodes.php
```diff
@@ -0,0 +1,18 @@
...
+		foreach ( $nodes as $node ) {
+			$this->insertAtIndex( $this->count(), $node );
```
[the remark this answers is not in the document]
knowler: Resolved by moving `AppendNodes::append` to `InsertNode`: de8cc48

src/Framework/FieldsAPI/Concerns/SerializeAsJson.php
```diff
@@ -0,0 +1,38 @@
...
+					}
+					return $value;
+				},
+				// Only serialize properties which are not hidden.
+				array_diff_key(
+					get_object_vars( $this ),
+					// Respect hidden, if set.
+					array_flip( property_exists( $this, 'hidden' ) ? $this->hidden : [] ),
+					// Ignore hidden property.
+					[ 'hidden' => 'hidden' ]
```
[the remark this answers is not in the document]
knowler: I don' think this is being used at all, but it was a way to hide attributes from serialization since right now every attribute on the class is being serialized. The inspiration was from Eloquent models and their `$hidden` array attribute.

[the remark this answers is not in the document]
knowler: I'm going to remove this for now, since it isn't being used.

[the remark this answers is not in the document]
knowler: Removed in 81ca7e5

knowler:
@JasonTheAdams I’ve addressed all of your comments. Please, re-review when you get a chance.

src/Framework/FieldsAPI/Concerns/SerializeAsJson.php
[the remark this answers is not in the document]
JasonTheAdams: Yeah, I like the idea. Let's add it when we need it.

src/Form/LegacyConsumer/Commands/SetupFieldConfirmation.php
```diff
@@ -32,20 +32,20 @@ public function __invoke( $hook ) {
 
 		$formID = give_get_payment_meta( $this->payment->ID, '_give_payment_form_id' );
 
-		$fieldCollection = new FieldCollection( 'root' );
-		do_action( "give_fields_{$hook}", $fieldCollection, $formID );
+		$collection = Group::make( 'root' );
```
JasonTheAdams, line 35: @knowler Does this mean we're making one Group per hook that has a name of "root"? I just want to make sure we're not getting into name collision in the future when these are all merged into the same Form.

knowler: Yep, do you have any suggestions for what we can use for the group names instead?

JasonTheAdams: How about the hook name?

knowler: That works. 2145b4d

knowler:
@JasonTheAdams This is ready for re-review btw.

src/Framework/FieldsAPI/Concerns/AppendNodes.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Give\Framework\FieldsAPI\Concerns;
+
+use Give\Framework\FieldsAPI\Contracts\Node;
+
+trait AppendNodes {
```
JasonTheAdams, line 7: So this can be deleted now, right?

knowler: Ya, I thought I had deleted it already, but I guess I just removed its usage from the `Group` class.

knowler:
@JasonTheAdams Ready for review again.

src/Framework/FieldsAPI/Concerns/AppendNodes.php
knowler: Resolved in 5b58566

knowler:
@JasonTheAdams I think we got everything that we discussed. This is ready for re-review.

JasonTheAdams (approved):
Awesome work, @knowler! I'm excited for the next evolution of the Fields API! 🚀

PR #5846, opened 2021-05-26
Uninstalling no longer checks for updates (which throws exception)

### LINKED ISSUE #5845: Uninstalling GiveWP should not throw an exception
## User Story

As a WordPress user, I want uninstall GiveWP so that I can move on without it (probably sadly).

## Details

Presently, uninstalling GiveWP on a fresh install throws an exception. The issue doesn't seem to occur when there are add-ons installed (not even activated). It's oddly niche, but it is easily reproducible.

## Expected Behavior

GiveWP should simply uninstall without any final fighting words.

## Steps to Reproduce

1. Set up a fresh WP installation
2. Install GiveWP from wp.org
3. Go to the plugins screen
4. Delete the GiveWP plugin
5. See the error thrown.

## Acceptance Criteria

-   [x] GiveWP peaceably uninstalls

JasonTheAdams:
Resolves #5845 

## Description

This fix removes a hook during uninstalling which is used to check for addon updates. Obviously, when uninstalling we don't care if there are any updates. Just uninstall.

This also sneaks in a fix wherein uncaught errors in PHP 7 could be an Error instead of an Exception, so typing the class is a problem.

## Affects

Uninstallation flow

## Testing Instructions

1. Install GiveWP
2. Uninstall GiveWP

Try installing both with and without add-ons present.

## Pre-review Checklist

### DIFF
--- a/uninstall.php
+++ b/uninstall.php
@@ -25,6 +25,9 @@ include_once( 'give.php' );
  */
 give()->init();
 
+// Prevent checking for plugin updates
+remove_filter( 'pre_set_site_transient_update_plugins', 'give_check_addon_updates', 999 );
+
 global $wpdb, $wp_roles;
 
 
--- a/uninstall.php
+++ b/uninstall.php
@@ -61,7 +61,19 @@ if ( give_is_setting_enabled( give_get_option( 'uninstall_on_delete' ) ) ) {
 	// Delete All the Terms & Taxonomies.
 	foreach ( array_unique( array_filter( $give_taxonomies ) ) as $taxonomy ) {
 
-		$terms = $wpdb->get_results( $wpdb->prepare( "SELECT t.*, tt.* FROM $wpdb->terms AS t INNER JOIN $wpdb->term_taxonomy AS tt ON t.term_id = tt.term_id WHERE tt.taxonomy IN ('%s') ORDER BY t.name ASC", $taxonomy ) );
+		$terms = $wpdb->get_results(
+			$wpdb->prepare(
+				"
+				SELECT t.*, tt.*
+				FROM $wpdb->terms AS t
+				INNER JOIN $wpdb->term_taxonomy AS tt
+					ON t.term_id = tt.term_id
+				WHERE tt.taxonomy IN ('%s')
+				ORDER BY t.name ASC
+				",
+				$taxonomy
+			)
+		);
 
 		// Delete Terms.
 		if ( $terms ) {
@@ -106,7 +118,11 @@ if ( give_is_setting_enabled( give_get_option( 'uninstall_on_delete' ) ) ) {
 	// Get all options.
 	$give_option_names = $wpdb->get_col(
 		$wpdb->prepare(
-			"SELECT option_name FROM {$wpdb->options} where option_name LIKE '%%%s%%'",
+			"
+			SELECT option_name
+			FROM {$wpdb->options}
+			WHERE option_name LIKE '%%%s%%'
+			",
 			'give'
 		)
 	);
--- a/uninstall.php
+++ b/uninstall.php
@@ -3,10 +3,10 @@
  * Uninstall Give
  *
  * @package     Give
- * @subpackage  Uninstall
+ * @since       1.0
  * @copyright   Copyright (c) 2016, GiveWP
  * @license     https://opensource.org/licenses/gpl-license GNU Public License
- * @since       1.0
+ * @subpackage  Uninstall
  */
 
 // Exit if accessed directly.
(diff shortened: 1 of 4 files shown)


JasonTheAdams:
@Benunc Would you be so kind as to test this out for me?

Benunc:
This PR fixes the issue of the fatal error, and on testing, I only find one issue, and it's that some DB tables are left behind: 
Screen Shot 2021-05-27 at 2 14 26 PM

JasonTheAdams:
Good call. I'm going to improve our table delete code to drop all tables with the "give_" prefix.

JasonTheAdams:
@Benunc I just shot you a new zip to try out which clears all GiveWP tables (including tables introduced by add-ons). If that works, please approve this and we'll get it out!

Benunc:
This all checks out. Here's the only lingering issues:

### Issue 1: The modal popup checkbox does not actually have any affect on deleting data.
In order to delete all GiveWP data, the setting in Donations > Settings > Advanced has to be set. In my tests, the checkbox on the deactivation survey has no affect on anything.

###Issue 2: GiveWP core + Recurring active.
If you have the recurring addon active and the settings set to delete all data in Donations > Settings > Advanced you have to first deactivate and delete recurring, and then deactivate and delete GiveWP core. It's not currently possible to delete GiveWP recurring without GiveWP core active, and throws a fatal error. Also, if you delete GiveWP core first, it does not clean up the Give Subscriber user role. If you delete GiveWP Recurring first (with the setting set to delete all data) it removes the give_subscriptions db table, but not the give_subscriptionmeta table, and it deletes the Give Subscriber user role. 

Issue 2 can likely be handled in a separate issue on the recurring repo.

JasonTheAdams:
@ravinderk Can you please wrap this up in time for the next hotfix release? The last thing remaining here is Issue 1 as Ben described above. Don't worry about Issue 2 for now. I started looking into this and I'm confused by a couple things:
1. Is this supposed to set the setting to delete everything upon plugin deletion?
2. Why is this [calling the export action](https://github.com/impress-org/givewp/blob/develop/assets/src/js/admin/plugin-deactivation-survey.js#L144)?

If you're not able to get this working by the next hotfix release go ahead and merge this. It's more important that this goes out to fix the exception thrown when deleting, than that this specific checked box is fixed at this time.

cc: @kjohnson

ravinderk:
@JasonTheAdams 

> Is this supposed to set the setting to delete everything upon plugin deletion?

Yes. I updated code to opt-in admin for plugin data deletion: https://github.com/impress-org/givewp/pull/5846/commits/c18a5b4d9fbb5b40eda59fbea00518a5b3a0ef1b

> Why is this calling the export action?

This ajax call reset all give stats.

kjohnson:
@ravinderk is this ready for re-review?

ravinderk:
> @ravinderk is this ready for re-review?


@kjohnson Yes. it is ready to re review.

uninstall.php
```diff
@@ -35,8 +38,11 @@
 	$give_post_types = [ 'give_forms', 'give_payment' ];
 	foreach ( $give_post_types as $post_type ) {
 
-		$give_taxonomies = array_merge( $give_taxonomies, get_object_taxonomies( $post_type ) );
-		$items           = get_posts(
+		foreach ( get_object_taxonomies( $post_type ) as $taxonomy ) {
```
kjohnson, line 41: Is the loop necessary since we are just assigning the values to a new array? How does the loop improve the `array_merge`?

uninstall.php
```diff
@@ -111,7 +118,11 @@
 	// Get all options.
 	$give_option_names = $wpdb->get_col(
 		$wpdb->prepare(
-			"SELECT option_name FROM {$wpdb->options} where option_name LIKE '%%%s%%'",
+			"
+			SELECT option_name
+			FROM {$wpdb->options}
+			WHERE option_name LIKE '%%%s%%'
```
kjohnson, line 124: Are there too many `%`s in the WHERE clause? It looks like we are using wildcards along side a merge syntax, so the merged value would be `%%give%%`, which would be redundant wildcards - correct?

kjohnson:
Submit and Deactivate seems to be working for me - the advanced setting is enabled correctly.

@Benunc I did notice that this only works for "Submit and Deactivate" and not "Skip and Deactivate" - is that the correct behavior?

kjohnson (commented):
This looks good to me - I left a couple of comments, but they don't holdup a hotfix. Pending @Benunc's review this is good to go.

Benunc:
I think we are all set here. I would anticipate "Skip and deactivate" to not take into account anything above.

In my tests, now the checkbox is working. The only thing that's left on deletion is a recurring role of `Give Subscriber`, but that is removed if you deactivate recurring first.

kjohnson (approved):
With @Benunc's confirmation I am approving this.

@ravinderk there are still a couple of open comments, but we can address those separately.

PR #5855, opened 2021-06-22
Add Facade class to give framework

jonwaldstein:
## Description

Th Facade pattern allows us to use helper classes as a singleton with flexibility of dependency injection or used directly.

The idea is to create your helper class as an instantiated class, then create a facade version that will extend `Facade` and returns the instance of the intended class that holds your logic.

## Visuals

Main Class
```
class UrlGenerator {
	public function createSignature()
}
```

Facade Class
```
/*
* @method static createSignature()
*/
class URL extends Facade  {

	protected static function getFacadeClass() {
		return UrlGenerator::class;
	}
}
```

Use it!

```
URL::createSignature();
```

## Pre-review Checklist


jonwaldstein:
@JasonTheAdams this is ready for review.  I included an initial unit test that should work in theory but wasn't able to run it properly, so should be revisited for more extensive testing.

JasonTheAdams:
@jonwaldstein What's up with all the weird formatting going on in the Changelog?

jonwaldstein:
> @jonwaldstein What's up with all the weird formatting going on in the Changelog?

i'm not sure, i just added one line and that happened.

JasonTheAdams:
Ah, it's auto-splitting long lines into multiple lines. Can you please remove that from PhpStorm and recover the previous formatting?

JasonTheAdams (approved):
@jonwaldstein I made a couple tweaks:
1. I got the tests passing
2. I made it so the class method _must_ be overridden by child classes, otherwise it throws an exception.

After thinking about the `static::class` default, I realized that is an extremely rare case wherein that would actually work — classes _only_ using magic methods. So I'd rather just have it always explicitly set.

Please look this over. If you agree, we can merge.

JasonTheAdams:
@jonwaldstein Please take a moment to review this when you get a chance so we can merge it.

src/Framework/Support/Facades/Facade.php
```diff
@@ -0,0 +1,51 @@
...
+	 *
+	 * @unreleased
+	 *
+	 * @return string
+	 * @throws Exception
+	 */
+	protected static function getFacadeClass() {
+		throw new Exception( 'This method must be overridden and return the class to decorate' );
+	}
+}
```
jonwaldstein, line 51: @JasonTheAdams is there any reason we couldn't make this an abstract class/method so this is forced to be used or does _static_ make it problematic?

JasonTheAdams: There is no such thing as an `abstract static` method, I'm afraid. We'd have to change how this works like so:
1. The `Facade::__callStatic` method is fired upon calling some method
2. The Facade class instantiates itself
3. The Facade uses an abstract instance method to get the class to be used
4. The class method is called

This is more closely to how Laravel works, but I opted for this direction for two reasons:
- To avoid potential method naming collision with the instance class
- To avoid the need to instantiate the Facade class, making two instances per Static Facade pair (that said, it's _very_ lightweight).

That said, I'm definitely open to it if you can think of a major disadvantage, here. The biggest that I can see — which puts me on the edge — is that this would not be caught by static analysis, wherein an abstract method would.

jonwaldstein: @JasonTheAdams gotchya, yeah honestly the only thing I was concerned about with the current direction is not having static analysis for the required method.

JasonTheAdams: Took at stab at this in 7d438eb

JasonTheAdams:
@jonwaldstein Give this another look for me. It took some head scratching, but I think I came up with a way of doing this that avoid redundant instantiation while taking advantage of an abstract method for static analysis.

The only downside to this method is that every method call checks to see if the `Facade` subclass and decorated class are stored in the service container. But I think this is a negligible cost as, under the hood, it's literally just an `isset()` on an array, which is about is cheap as it gets.

JasonTheAdams:
@jonwaldstein I would like to start using this in other PRs, can you please prioritize this last review? I think it's good to go.

jonwaldstein:
> @jonwaldstein I would like to start using this in other PRs, can you please prioritize this last review? I think it's good to go.

This works perfect, I just tested it with my URL facade and everything worked as expected including the static analysis.  Good to go (i can't approve my own PR) 👍

PR #5861, opened 2021-07-02
Exception Handler passes along exceptions and errors

### LINKED ISSUE #5860: Logged exceptions should still display
## Details

A while back we added a means of catching and logging all GiveWP exceptions. This works great. What doesn't work great is that once an exception is caught the system assumes our exception handler has, in fact, handled the exception and so it does nothing further. This results in blank pages like the following:

Image 2021-07-02 at 2 45 14 PM

We want to allow WordPress or whatever else to do something further with the exception, such as display it to the user. Blank pages are scary. Really, we're not trying to "handle" exceptions, but merely log them.

## Acceptance Criteria

-   [ ] GiveWP exceptions are logged
-   [ ] Exceptions, logged or not, are passed along to the system

JasonTheAdams:
Resolves #5860 

## Description

This resolves the issue wherein handled exceptions weren't being passed to the system for display (or whatever else), resulting in a demoralizing white screen.

A couple things to consider in this solution:
1. It seems that Errors (not Exceptions) work differently. Regardless of whether they're handled, they still get passed along and, in this case, are rendered in the browser. That's why I'm checking to see if the variable is, in fact, an Exception.
2. The solution seems to be to just throw the error. PHP is apparently smart enough to recognize the exception is being thrown within the handler, so it doesn't create a loop.
3. I figure if there's another handler in place, we can let them determine whether it should be re-thrown or not. This _may_ be problematic if multiple handlers are throwing, but I think we're good.

## Visuals

Image 2021-07-02 at 2 52 44 PM

## Testing Instructions

Try throwing Errors (`trigger_error()`) and Exceptions in a few places. Make sure that in all cases it works the same as if our handler code wasn't there at all. Try throwing the Exception, then disable our handler code, and try again. The goal is that there should be no difference.

## Pre-review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Uncaught exception handler no longer breaks on Errors (#5846)
 -   Uninstalling GiveWP no longer throws an exception (#5846)
+-   Caught GiveWP exceptions no longer display a white screen (#5861)
 
 ## 2.11.1 - 2021-05-24
 
--- a/src/Framework/Exceptions/UncaughtExceptionLogger.php
+++ b/src/Framework/Exceptions/UncaughtExceptionLogger.php
@@ -34,6 +34,8 @@ class UncaughtExceptionLogger {
 	 * @since 2.11.1
 	 *
 	 * @param Exception|Error $exception
+	 *
+	 * @throws LoggableException
 	 */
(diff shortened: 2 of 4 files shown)


JasonTheAdams:
@kjohnson Please review this when you get a chance.

kjohnson (approved):
Looks good to me. I tested locally with `trigger_error()` and throwing the `Give\Framework\Exceptions\Primitives/Exception` manually.

PR #5863, opened 2021-07-06
Pulling product data from givewp.com can no longer break site

### LINKED ISSUE #5862: GiveWP.com reliability should not affect GiveWP customer site reliability
## Details

GiveWP checks whether to see GiveWP add-ons are premium or free add-ons. The way it does this is by querying all the EDD products on givewp.com, and then checking to see if the add-on appears in that list. The current method was introduced in GiveWP 2.9.2.

The problem is that the code does not handle failures well. A couple things compound into a major problem when givewp.com is, for some reason, down:
1. The add-on information is _only_ cached if it receives good data
2. Every GiveWP add-on runs the function to check itself
3. Every failed call takes 5 seconds before it times out

So if you have 6 add-ons, that's 5 seconds per add-on, so the whole site hangs up for 6 x 5 = 30 seconds. Mosts hosts will simply timeout the whole request at this point. 💥😱🔥

GiveWP.com going down should not be able to take down customer sites.

## Acceptance Criteria

-   [x] Premium add-ons are successfully identified
-   [x] If GiveWP.com goes down the site using GiveWP stays up

JasonTheAdams:
Resolves #5862 

## Description

This PR tackles the problem that givewp.com going down can bring down other sites, as they're making requests to get EDD product data in order to determine whether add-ons are premium.

A couple layers of caching are applied:
1. The request times out after 3 seconds
2. If the request returns invalid for any reason, an empty array is cached for an hour
3. Multiple checks in the same request (i.e. multiple add-ons) will only query or check the cache one time.

With this in place, there should be only one possible query per hour to givewp.com that can hang up a single request for 3 seconds. Also, in any case, multiple queries to wp_options (transient) are reduced to 1, no matter how many add-ons are installed.

## Affects

Checking whether a given add-on is premium or not.

## Testing Instructions

1. Make a local site with an endpoint that times out
2. Change `PremiumAddonsListManager::PRODUCTS_LIST_API_URL` to point do that endpoint.

Also, check for other givewp.com requests to make sure we haven't missed anything that can have this problem.

Finally, make turn off your server and point to the right endpoint. Make sure that it still works as expected.

## Pre-review Checklist

### DIFF
--- a/give.php
+++ b/give.php
@@ -43,6 +43,7 @@ use Give\Container\Container;
 use Give\Framework\Exceptions\UncaughtExceptionLogger;
 use Give\Framework\Migrations\MigrationsServiceProvider;
 use Give\Form\Templates;
+use Give\License\LicenseServiceProvider;
 use Give\Revenue\RevenueServiceProvider;
 use Give\Route\Form as FormRoute;
 use Give\ServiceProviders\PaymentGateways;
@@ -155,6 +156,7 @@ final class Give {
 		LogServiceProvider::class,
 		FormLegacyConsumerServiceProvider::class,
 		ShimsServiceProvider::class,
+		LicenseServiceProvider::class,
 	];
 
 	/**
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -853,7 +853,7 @@ function give_get_plugins( $args = [] ) {
 	}
 
 	if ( ! empty( $args['only_premium_add_ons'] ) ) {
-		$premiumAddonsListManger = new PremiumAddonsListManager();
+		$premiumAddonsListManger = give( PremiumAddonsListManager::class );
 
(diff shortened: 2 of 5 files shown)


JasonTheAdams:
@kjohnson @Benunc Please test this ASAP so we can get this hotfix out.

kjohnson:
@JasonTheAdams In my testing, the `PremiumAddonsListManager` is instantiated three (3) times per request - does this need to be a singleton in order to preserve per request caching?

See `includes/misc-functions.php`

JasonTheAdams:
@kjohnson Hah! Well, before seeing your update I went ahead and made it a singleton. 😆

kjohnson (approved):

knowler (approved):
Works for me. 👍🏻

PR #5882, opened 2021-07-15
Creating a donation form programmatically no longer throws an undefined index error

kjohnson:
## Description

I ran across this while create a donation form programmatically for Peer-to-peer, which was causing an undefined index error in the `LegacyFormSettingsCompatibility` class. This was likely also happening when creating a donation for programmatically during Onboarding, although the error was likely suppressed because of the AJAX request context.

## Testing Instructions

```php
use Give\Onboarding\DefaultFormFactory;

give( DefaultFormFactory::class )->make();
```

## Pre-review Checklist

### DIFF
--- a/src/Form/Template/LegacyFormSettingCompatibility.php
+++ b/src/Form/Template/LegacyFormSettingCompatibility.php
@@ -70,6 +70,14 @@ class LegacyFormSettingCompatibility {
 					continue;
 				}
 
+				/**
+				 * Prevents an undefined index error of the field id is not set in the settings group.
+				 * @unreleased
+				 */
+				if ( ! isset( $settings[ $groupId ] ) || ! isset( $settings[ $groupId ][ $field['id'] ] ) ) {
+					continue;
+				}
+
 				Give()->form_meta->update_meta( $formId, $field[ self::$key ], $settings[ $groupId ][ $field['id'] ] );
 				$alreadySavedLegacySettings[] = $field[ self::$key ];
 			}
--- a/src/Form/Template/LegacyFormSettingCompatibility.php
+++ b/src/Form/Template/LegacyFormSettingCompatibility.php
@@ -72,6 +72,7 @@ class LegacyFormSettingCompatibility {
 
 				/**
 				 * Prevents an undefined index error of the field id is not set in the settings group.
+				 * @link https://github.com/impress-org/givewp/pull/5882
 				 * @unreleased
 				 */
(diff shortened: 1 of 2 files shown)


src/Form/Template/LegacyFormSettingCompatibility.php
```diff
@@ -70,6 +70,15 @@ public function save( $formId, $settings ) {
 					continue;
 				}
 
+				/**
+				 * Prevents an undefined index error of the field id is not set in the settings group.
+				 * @link https://github.com/impress-org/givewp/pull/5882
+				 * @unreleased
+				 */
+				if ( ! isset( $settings[ $groupId ] ) || ! isset( $settings[ $groupId ][ $field['id'] ] ) ) {
```
ravinderk, line 78: @kjohnson this function runs internally when you save the donation form on the form edit screen. If template-specific settings do not exist in `$_POST` then this function will not be called, so I am confused why this function is running in your case.

https://github.com/impress-org/givewp/blob/f9d39b1642a161d432c99729d107e87f45b209ca/includes/admin/forms/class-metabox-form-data.php#L66

Above filter cause of calling this function. I checked onboarding flow and I think the `$_POST[{templateId}]` setting will be empty, so we exit early in filter callback. Can you review it again?

ravinderk (changes requested):
@kjohnson I left a question for you.

src/Form/Template/LegacyFormSettingCompatibility.php
kjohnson: @ravinderk I think it is being triggered by a call to `\Give\Form\Template\Options::getCheckoutLabelField()`, see [P2P\DonationFormFactory](https://github.com/impress-org/give-peer-to-peer/blob/develop/src/P2P/Factories/DonationFormFactory.php#L53)

ravinderk: @kbjohnson90 you are adding field registration array to setting value array.

Please replace
```php
TemplateOptions::getCheckoutLabelField()
```

with
```php
'checkout_label' => ewc_html__( 'Donate Now', 'give-p2p ')
```

Edit Donation Form ‹ GiveWP — WordPress 2021-07-16 at 10 22 33 PM

kjohnson:
Thanks for the assist @ravinderk! 🦾

src/Form/Template/LegacyFormSettingCompatibility.php
kjohnson: Implemented in P2P.

PR #5883, opened 2021-07-15
Add opt-in to Onboarding Wizard

### LINKED ISSUE #5852: Add email opt-in into the onboarding wizard and settings area
## Details

We need to allow GiveWP core users to opt into our valuable email marketing. When we built the on-boarding wizard, we lost the ability for free users to opt-in, and that has been detrimental to our email list. 

Goals for implementing this:
* Surface the opportunity in an intuitive and informative way
* For CAN-SPAM concerns, and .org guidelines, ensure they are opted-out by default
* But make opting-in as easy as possible. 
* Do not cloud the general plugins area or the WP Dashboard in any way. Limit all of this to GiveWP-specific screens. 

## Implementation Details

The current recommendation is to surface the email marketing opt-in, in 3 different places:

### GiveWP Onboarding Wizard

New users of GiveWP are walked through an onboarding wizard. We ask them specific information about their organization and how they will be doing fundraising. This is a good place to add a little more information for email marketing purposes. 

After the wizard is done they won't see this area again, so it's important that there are additional areas for users to opt into our email list. 

If they opt-in here, then they will get a user_meta key called "givewp_marketing" == true.

**CRITERIA**
None. All new users will see this opportunity regardless.

**CONCEPT**
Wizard Step 1 with opt-in-2

**ADDITIONAL DETAILS**
You'll notice there is now also a "Your Cause Name" field in addition to the "Cause Type" field (which already exists in the wizard). The info from all three fields will be passed into our email marketing platform if they provide their email address. 

### GiveWP Set-up Screen
After the Onboarding wizard, they are dropped into a Setup screen. This should reflect whether they opted in or not during the wizard. If they did not opt-in, then they will have the chance to opt-in here as well. 

**CRITERIA**
This appears only if the "givewp_marketing" user meta is still "false".

**CONCEPT**
If they did not opt-in during the wizard, they'll see this:
image

If they did opt-in, or after they do directly from the setup screen, they'll see this:
image

### GiveWP Settings Screen
After going through the wizard and the setup screen, they will most likely land on the Settings screen in a short amount of time. So if they have not opted in at the Wizard, or at the Setup screen, then the Settings screen should have an opt-in ability with a stronger incentive, like a free premium add-on. 

**CRITERIA**
This appears only if:
1. They don't have any paid add-ons installed
2. AND the "givewp_marketing" user meta is still "false"

**CONCEPT**
If they did not opt-in at the wizard or the Setup screen, then on the Settings page they'll see this notice with a stronger incentive to opt-in:
image

If they opt-in on this CTA, then they'll see this confirmation message only while they are on the page. When they return to the settings page, no CTA will be there at all. 
image

## ACTIVECAMPAIGN API INFORMATION*

Here's docs on creating contacts:
https://developers.activecampaign.com/reference#create-or-update-contact-new

In the case of optin-in during the wizard, we want to pass the following:
* Contact Name (default contact field, populated from user first name)
* Contact email (default contact field, populated from email field)
* Account (default contact field, populated from "Organization Name")
* Organization Website (Custom contact field, auto-populated from website url)
* Type of Fundraising (Customer contact field, populated from the wizard "Cause Type" field) 

## Acceptance Criteria

-   [x] The wizard onboarding opt-in is present.
-   [ ] The Setup screen email opt-in status is present and it shows the correct status based on whether the user opted in during the wizard or not.
-   [ ] The settings screen CTA is present if there are no paid add-ons installed and the "givewp_marketing" user meta is set to `false`. 
-  [ ] If they opt-in at the Setup page, there is a confirmation message
-  [ ] If they opt-in at the Settings screen, there is a confirmation message
-  [ ] The contact is added to our marketing list at each opt-in form correctly. 
-  [ ] The contact custom meta is added in our marketing list correctly from all three opt-in forms

DevinWalker:
Resolves #5852

## Description

Adds an opt-in to the onboarding wizard that integrates with ActiveCampaign and our Telemetry server.

## Affects

- Onboarding UI / UX
- ActiveCampaign
- Telemetry Opt-in

## Visuals & Testing Instructions

Please see my testing video here: 

https://www.loom.com/share/7d09f820f9f14d8ea80eb0d38dbf5ff8

## Pre-review Checklist


JasonTheAdams:
@DevinWalker In testing this locally, I found that it was failing because my admin user does not have a first and last name specified. I'm inclined to make those optional — unless Active Campaign requires them — as it's very possible that many users won't have those. Thoughts?

DevinWalker:
> @DevinWalker In testing this locally, I found that it was failing because my admin user does not have a first and last name specified. I'm inclined to make those optional — unless Active Campaign requires them — as it's very possible that many users won't have those. Thoughts?

Good catch! Yeah let's definitely make those optional.

assets/src/js/admin/onboarding-wizard/app/steps/your-cause/index.js
```diff
@@ -14,40 +14,60 @@ import ContinueButton from '../../../components/continue-button';
...
+
+			<div className="give-obw-optin-field">
+				<h2>{__( 'What are you fundraising for?', 'give' )}</h2>
+				<span className="screen-reader-text">{__( 'What type of cause is yours?', 'give' )}</span>
+				<SelectInput testId="cause-select" value={causeType} onChange={( value ) => dispatch( setCauseType( value ) )} options={getCauseTypes()} />
+			</div>
+
+			<div className="give-obw-community-field">
+				<h2>{__( '🌱 Would you like to join the GiveWP Community?', 'give' )}</h2>
+				<p>{__( 'If you opt-in, some data will be shared with our teams (so they can work their butts off to improve). We will also share some helpful info on fundraising, WordPress and our products from time to time.  And if you skip this, that’s okay! GiveWP will still work just fine.', 'give' )}</p>
```
mathetos, line 58: Change here to "By opting-in, you allow some basic data about how you use GiveWP to be used for us to improve the plugin for others. You also will receive emails from us with fundraising tips and more (which you can always unsubscribe from if you need to). If you skip this step, that's okay! GiveWP will still be setup for you no problem."

mathetos (approved):
Just the one wording tweak and we're good to go!

assets/src/js/admin/onboarding-wizard/app/steps/your-cause/index.js
JasonTheAdams: Added in 27ea52c

PR #5884, opened 2021-07-16
Remove deprecation notices for hooks core is using

knowler:
Two commits in #5843 added deprecation notices for hooks that core was using. This removes the notices.

Reverts 59729b2c106434e30ac6fda23e4c888a4f1fe30f.
Reverts 2aff65a7f54f2367bcf562e741b361a0f835f51c.


kjohnson (commented):
@knowler I think this change isn't necessary. The added hooks aren't _necessarily_ causing the issue, but rather the LegacyFormConsumer wasn't accounting for these hooks.

See https://github.com/impress-org/givewp/commit/f9d39b1642a161d432c99729d107e87f45b209ca

knowler:
I agree.

PR #5885, opened 2021-07-16
Only walk fields for commands that deal with fields

knowler:
This fixes an issue where some Legacy Consumer commands were walking more than just fields. This updates the commands to use `walkFields` method instead of `walk`.


kjohnson (approved):

kjohnson:
@knowler could you update the `CHANGELOG`?

knowler:
@kjohnson Does it need to be updated if none of this has been included in a formal release (i.e. non-pre-release)?

JasonTheAdams:
Yes, always update the changelog for every PR. 😃

PR #5887, opened 2021-07-19
Fields API factory now reflects individual type classes

kjohnson:
## Description

Updates the Fields API Factory to reference the individual type classes.

## Affects

Specifically, the Factory is used by the `give_field( $type )` function for registering custom fields.

```php
use Give\Framework\FieldsAPI\Facades\Factory;

function give_field( $type, $name ) {
	return Factory::make( $type, $name );
}
```

## Pre-review Checklist


knowler (commented):
I wonder if we should call this factory `Node` since `Field` is now a category of node. I suspect that we will include all types of nodes in the `Types` class, regardless of what category they fall under. Thoughts, @JasonTheAdams?

kjohnson:
@knowler we could just call it `Factory` there is only one and it could be moved top level along side `Types`.

knowler:
@kjohnson I like that (I am biased towards flat file structures though). Let's do it.

tests/unit/tests/Framework/FieldsAPI/Factory/FieldTest.php
```diff
@@ -0,0 +1,13 @@
...
+
+use Give\Framework\FieldsAPI\Factory\Field;
+use Give_Unit_Test_Case;
+
+class FieldTest extends Give_Unit_Test_Case {
+
+	public function testMakeTextField() {
+		$field = Field::text('my-text-field');
+
+		$this->assertEquals( 'text', $field::TYPE );
```
JasonTheAdams, line 11: This is a bit of a weak check. Let's use `assertInstanceOf` instead, and then assert the name to be `my-tet-field`.

JasonTheAdams:
I like this. I would also like to use the new `Facade` class to make this into a static facade.

kjohnson:
@JasonTheAdams @knowler I've updated this PR to reflect the following:

- Rename Field Factory to Factory
- Access the Factory using the Factory Facade
- Use an instance of assertion in the unit test

@knowler also, I accidentally let a local change slip in - specifically adding `HasLabel` and `HasEmailTag` to the `Hidden` field type. This was previously causing an error with Peer-to-peer. I added the traits because I think that they are valid for a hidden field type. I can revert this change if needed, but I would have PRd the change anyway.

src/Framework/FieldsAPI/Factory.php
```diff
@@ -0,0 +1,27 @@
...
+class Factory {
+
+	/**
+	 * @unreleased
+	 * @param string $type
+	 * @param string $name
+	 * @return Node
+	 * @throws TypeNotSupported
+	 */
+	public function make( $type, $name ) {
```
JasonTheAdams, line 20: To make this more future proof, let's do this:
```php
public function make( $type, ...$args ) {
    return new $class( ...$args );
}
```

kjohnson: @JasonTheAdams resolved by 5cd1cfc3591747e0f196be86708ffa9ede73b107.

PR #5889, opened 2021-07-20
Fields API: do not extend Form from Group

knowler:
## Description

I realized that `Group` doesn’t have a `TYPE` constant set and I forgot that a sub-class in PHP could override constants, but I made the decision to not extend `Form` from `Group`. I still like this change because I didn’t feel good about extending `Form` from `Group` and this way people don't need to stumble on late static binding. Just seems less problematic.

## Affects

- `Give\Framework\FieldsAPI\Form`
- `Give\Framework\FieldsAPI\Group`

## Pre-review Checklist


knowler:
@JasonTheAdams This is ready for review. I think the most important thing here is just making sure that `Group` has the `TYPE` constant set. Otherwise, changing `Form` to not extend `Group` isn’t super important for this release.

JasonTheAdams:
@knowler Do you anticipate that the form will ever be less than a Group? Or is there another scenario you think this will avoid?

knowler:
> Do you anticipate that the form will ever be less than a Group?

@JasonTheAdams, I guess not, though I was worried that maybe stuff added to `Group` in the future could be undesirable for `Form`. I think what bothers me is that I don't want a single node type to become the base abstraction for others (at least for the node types that are in core). `Collection` is the better abstraction in my opinion — `Form` and `Group` are collections (they both implement `Contracts/Collection`) that happen to also be nodes (implementing `Contracts/Node`) and being nodes they have potentially different functionality.

I’d be fine with just adding the `TYPE` constant to `Group` and kicking this decision down the road if a real issue comes up.

knowler:
@JasonTheAdams I'll let you make the final decision on this one. I’m not dead-set on the change.

PR #5892, opened 2021-07-21
Fields API: set `File` validation rules using `ValidationRules`

knowler:
## Description

The custom validation rules set in `File` were not updated to use the `ValidationRules $validationRules` property on the `Field` class. This PR updates them to do that. This is important since otherwise the rules will appear top-level instead of with all the validation rules during serialization.

## Affects

`Give\Framework\FieldsAPI\File`

## Pre-review Checklist


knowler:
A note for reviewing: I think it's outside of the scope of this PR to change whatever default values are set for these validation rules.

knowler:

JasonTheAdams (approved):
This is nice! I like it!

PR #5877, opened 2021-07-13
Donation tooltips are now escaped properly

JasonTheAdams:
Resolves #5876 

## Description

This tackles the problem that tooltips weren't being escaped properly, allowing in side-effects. The issue was that the tooltip being rendered was simply grabbing the node text without escaping it, and applying it directly to the attribute as a string.

Now, the node element is created and constructed using the DOM functions, which handle escaping and such properly.

## Affects

Donation tooltip options

## Visuals

Image 2021-07-13 at 2 58 31 PM

Image 2021-07-13 at 2 59 04 PM

## Testing Instructions

Set the following as the Text for a Donation Option: `"onmouseover=alert(/OOPS/)//`, then hover over that donation option in the form. It **should now** open an alert.

## Pre-review Checklist

-   [ ] Acceptance criteria satisfied and marked in related issue
-   [ ] Relevant `@since` tags included in DocBlocks
-   [ ] Changes logged to the `Unreleased` section of `CHANGELOG.md`
-   [ ] Includes unit and/or end-to-end tests
-   [ ] [Self Review](https://give.gitbook.io/development-manual/devops/github/code-reviews#self-review) of code and UX completed

PR #5894, opened 2021-07-22
Filtered donation level text no longer breaks form HTML

### LINKED ISSUE #5893: Enabling Recurring Donations on individual donation levels causes HTML to show on the front end
Using GiveWP 2.12.0, when an individual donation level has recurring donations enabled there is HTML in the corresponding front end button.

This occurs on both Multi Step and Legacy forms.

https://givewp.canny.io/admin/board/bug-reports/p/enabling-recurring-donations-on-individual-donation-levels-causes-html-to-show-o

kjohnson:
Resolves #5893

## Description

There are two things going on here.

First, a class is conditionally added if the text is "short" (less than 50 characters) - otherwise the value is an empty string.. This is done with a ternary inside of `classList.add()`. When passing an empty class `classList` throws an error. Instead, this PR conditionally adds the class as a separate statement.

Second, the Recurring Donations add-on (or any use of the `give_form_level_text` that included HTML would show escaped HTML in the rendered text. This PR strips HTML tags when outputting the text.

Note: While originally reported as a conflict with the Recurring Donation add-on, the issue is replicated with a simple filter of the level text. Therefor all changes are made in GiveWP.

### Sequoia
| Before | After |
| --- | --- |
| Screenshot from 2021-07-22 14-27-38 | Screenshot from 2021-07-22 14-27-50 |

### Legacy
| Before | After |
| --- | --- |
| Screenshot from 2021-07-22 14-28-13 | Screenshot from 2021-07-22 14-28-24 |

| Buttons | Radio | Select |
| --- | --- | --- |
| Screenshot from 2021-07-22 14-39-18 | Screenshot from 2021-07-22 14-39-07 | Screenshot from 2021-07-22 14-38-52 |

### Reduced test case

```php
add_filter( 'give_form_level_text', function( $level_text, $form_id, $level ) {
    return $level_text . ' ' . "<span>Hello World</span>";
}, 10, 3);

```

## Pre-review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+-   Filtered donation level text no longer breaks form HTML (#5894)
+
 ## 2.12.0 - 2021-07-21
 
 ### Added
--- a/includes/forms/template.php
+++ b/includes/forms/template.php
@@ -551,7 +551,7 @@ function give_output_levels( $form_id ) {
 					$level_classes,
 					$formatted_amount,
 					array_key_exists( '_give_default', $price ) ? 1 : 0,
-					esc_attr( strip_tags( $level_text ) )
+					$level_text
 				);
 			}
 
@@ -593,7 +593,7 @@ function give_output_levels( $form_id ) {
 					$formatted_amount,
 					( give_is_default_level_id( $price ) ? 'checked="checked"' : '' ),
 					array_key_exists( '_give_default', $price ) ? 1 : 0,
-					strip_tags( $level_text )
+					$level_text
 				);
 			}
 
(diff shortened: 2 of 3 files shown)


JasonTheAdams:
@kjohnson Looks like this may be a regression from my previous PR: https://github.com/impress-org/givewp/pull/5877

I'm thinking about just removing the `esc_html` I added, as it wasn't really necessary — the big change was adding the attribute via the `Element.setAttribute()` method. It seems like `esc_html( strip_tags() )` is redundant, in a way.

What do you think?

kjohnson:
@JasonTheAdams I would agree that `esc_html( strip_tags() )` is redundant - if the HTML is stripped, then there is no HTML to escape. Rather, perhaps we should be using `esc_attr()` or is that moot with the `setAttribute()` call?

JasonTheAdams:
We're basically escaping on the front end, so I think we're good. You can try `esc_attr()` and see if it still solves the issue.

kjohnson:
@JasonTheAdams it looks like `esc_attr()` is already called in the context of "buttons", which is good because Sequoia pulls this into a data attribute, so it should be escaped as a (possible) attribute. For "radios" and "select" I just remove the `esc_html()` as redundant and did not replace with `esc_attr()` as these are never used as attributes.

JasonTheAdams (approved):
This is ready for release. Simple but important fix for folks using admin-defined recurring forms. Great job jumping on this so quickly, @kjohnson!

PR #5897, opened 2021-07-27
Fixes for checkbox template

flack:
## Description

the `placeholder` attribute does not have any effect on checkboxes. But if you happen to use HTML in your label, rendering it can lead to display errors, so remove it.

Also the `for` attribute of the label has to correspond to the `id` attribute of the checkbox, otherwise clicking the label will not affect the checkbox

## Affects

Labels on checkboxes are now clickable again and HTML labels don't break rendering

## Pre-review Checklist


knowler (approved):
Nice catch, @flack. Thanks for fixing that.

PR #5898, opened 2021-07-27
Fields API: add Html node type

knowler:
## Description

As a part of updating the Form Field Manager addon to use the new Fields API, we need to support some of its types which are not all desirable to add to core. In the last cycle, we created an `Html` type which was a sort of back door for us to allow creating some of these types. It was also this particular type which helped us realize the helpful distinction between fields and non-fields in a form.

This is an `Element` type of `Node`. This means that it is a non-`Field` which means that can be filtered out of a collection when appropriate (e.g. registering "required" fields — this wouldn’t make sense for a non-field).

## Affects

- Adds `Html` class to `Give\Framework\FieldsAPI`.
- Adds `HTML` constant to `Give\Framework\FieldsAPI\Types`.

## Testing Instructions

This is no manual testing required for this PR (other than making sure existing Fields API usage isn’t broken).

## Pre-review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+-   Add `Html` node type to Fields API (#5898)
+
 ### Fixed
 
 -   Add missing `TYPE` to Fields API `Group` node type (#5895)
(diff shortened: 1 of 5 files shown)


knowler:
@JasonTheAdams This is ready for review. I’m not actually 100% sold this is the direction we should go, however, this would be fairly flexible. The alternative is to actually think of new `Element` types, but I’m not sure if I want to do that for legacy stuff. The Fields API does seem like it would be mature enough to start creating node types like that.

JasonTheAdams (approved):
This is great! I'm really happy with the distinction of `Fields` and `Elements` — this fits perfectly into the latter group.

JasonTheAdams:
Oops! This needs to be added to the LegacyConsumer before we can merge this!

JasonTheAdams (approved):
Tested! Now I can put YouTube videos in my forms!
Image 2021-07-27 at 11 54 00 AM

PR #5868, opened 2021-07-08
Card declines on multi-step form now display an error message on first click

kjohnson:
Resolves #5867

Prevent a race condition between the error container and the relocated error node.

## Description

Adds a 1 millisecond timeout to prevent a race condition between the `.donation-errors` element being appended to the form and relocation of the specific error message node to that created element.

## Testing Instructions

- Enable Stripe Credit Card in **live** mode.
- Create a Donation Form with the Multi-Step Template (Sequoia)
- Attempt to make a donation with an _invalid_ card number, ie `4242 4242 4242 4244`
- Submitting the payment will should show a related error message.

## Pre-review Checklist

-   [x] Acceptance criteria satisfied and marked in related issue
-   [x] Relevant `@since` tags included in DocBlocks
-   [x] Changes logged to the `Unreleased` section of `CHANGELOG.md`
-   [ ] Includes unit and/or end-to-end tests
-   [x] [Self Review](https://give.gitbook.io/development-manual/devops/github/code-reviews#self-review) of code and UX completed

PR #5899, opened 2021-07-27
Update CHANGELOG.md

kjohnson:
Related: https://github.com/impress-org/givewp/pull/5868

## Description

It looks like this was removed from the CHANGELOG at some point prior to release.


JasonTheAdams (approved):
Good catch. Thanks!

PR #5900, opened 2021-07-27
New: make Fields macroable in Fields API

knowler:
## Description

This adds a trait called `Concerns\Macroable` to the Fields API which is used in the `Field` abstract class. This would ideally used by addons for adding functionality to core field types.

Example usage:

```php
use Give\Framework\FieldsAPI\Text;

Text::macro('repeatable', function ( $isRepeatable = true ) {
    $this->repeatable = $isRepeatable;
    
    return $this;
});

Text::macro('getRepeatable', function () {
    return $this->repeatable ?? false;
});
```

This implementation is a direct, reduced copy of `illuminate/macroable` without substantial change, so we probably need to include licensing for it or maybe use the package itself (though I think it adds more than we want and Composer deps in WP is a nightmare).

## Affects

Anything that extends `Give\Framework\FieldsAPI\Field`. It doesn’t change anything, just adds new functionality.

## Testing Instructions

Everything should still work.

## Pre-review Checklist


knowler:
@JasonTheAdams I'm leaving this as a draft since I don't think it's high priority. I was just interested in how this would be implemented, so I tried it out and thought I’d push it in a branch/draft PR.

JasonTheAdams (approved):
I'm always reluctant to use static properties as they break encapsulation. But this is a weird moment where we're changing the shape of the class itself for use at runtime. Adding a `MacroRegister` or something like that is doable, but feels potentially convoluted.

@kjohnson I'd appreciate if you could review this as well and give your thoughts.

kjohnson:
I would also have reached for the static property, as I think it does make sense here. Implementing `MacroRegister` would separate the concerns, but I agree that it sounds convoluted. I think the question here is if the static property is testable.

Also, it would be helpful to see some documentation and tests that demonstrate the usage examples.

knowler:
For reference, here are the tests for `illuminate/macroable` which this is a direct copy of:

https://github.com/laravel/framework/blob/277c2fbd0cebd2cb194807654d870f4040e288c0/tests/Support/SupportMacroableTest.php

JasonTheAdams:
Static properties aren't purely testable. That's why I'm not a fan. 🤷‍♂️ Still, that's not a dealbreaker for me in this _very_ specific situation.

As far as an example, the one that prompted this whole thing was giving add-ons a way of easily extending existing fields, such as Salesforce:
```php
Field::macro('mapInSalesforce', function(string $object, string $field) {
  $this->salesForceObject = $object.
  $this->salesForceField = $field;
});

// usage
Field::make('company')
  ->mapInSalesForce('account', 'name');
```

JasonTheAdams:
> For reference, here are the tests for `illuminate/macroable` which this is a direct copy of:
> 
> https://github.com/laravel/framework/blob/277c2fbd0cebd2cb194807654d870f4040e288c0/tests/Support/SupportMacroableTest.php

That's about as good a way of testing this as I've seen. Being in a trait really helps.

kjohnson:
I think the `MacroRegister` would be the correct way to do it, but the static property is practical. Although, if we have it well tested, maybe the complexity won't bee such a big deal. As long as the API is simple, the underlying system shouldn't require much maintenance.

(Thinking out loud).

JasonTheAdams:
I'm still leaning in the direction of the static property. If it really becomes a problem we can make a register (or something) without breaking backwards compatibility. Being inside of a trait makes it safely testable, and I'm happy with that. I can't think of any problems, otherwise.

JasonTheAdams (approved):

PR #5913, opened 2021-08-11
Payment errors no longer revert to the default donation amount

### LINKED ISSUE #5912: Payment errors should not revert to default donation amount
## Description

> If I select a custom amount for my donation and have my card declined by an incorrect card number, the donation amount should stay on the original custom amount entered and NOT revert to the default amount set for the form.

## Acceptance Criteria

- [x] Payment errors restore selected donation level.
- [x] Payment errors restore custom donation amounts.

kjohnson:
Resolves #5912

## Description

This PR adds a flag when the page is loaded for reference to prevent a race condition when adding an event listener.

## Testing Instructions

Intentionally triggering a payment error, ie card declined, should restore the user selected donation amount after the donation is attempted.

## Pre-review Checklist

### DIFF
--- a/includes/forms/functions.php
+++ b/includes/forms/functions.php
@@ -193,6 +193,11 @@ function give_send_back_to_checkout( $args = [] ) {
 	// Set the $level_id.
 	if ( isset( $_POST['give-price-id'] ) ) {
 		$defaults['level-id'] = sanitize_text_field( $_POST['give-price-id'] );
+
+		// If custom, set amount
+		if( 'custom' === $defaults[ 'level-id' ] ) {
+			$defaults['custom-amount'] = sanitize_text_field( $_POST['give-amount'] );
+		}
 	}
 
 	// Check for backward compatibility.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+-   Payment errors no longer revert to the default donation amount (#5913)
+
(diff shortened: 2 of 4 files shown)


JasonTheAdams (approved):
A little odd, but I think that's probably the best fix without a lot of refactoring. Since Zack has already tested this, you're good to merge!

kjohnson:
@JasonTheAdams is it odd? Yes. Does it work? Yes 😅

ravinderk:
@kjohnson I did not see any problem with event listener which restore previous donation form entries. Actually we do not have system in place to restore custom amount. I noticed that you updated logic to restore custom amount, so i thing there are not need to edit window load event listener logic.

But i find out that when restore donation form, everything restores expect donation level. it still remain set to default donation level. We need to fix this. Let me know if you have any question.

Support the Veterans Choir – give 2021-08-12 at 10 29 54 AM
image

includes/forms/functions.php
```diff
@@ -193,6 +193,11 @@ function give_send_back_to_checkout( $args = [] ) {
 	// Set the $level_id.
 	if ( isset( $_POST['give-price-id'] ) ) {
 		$defaults['level-id'] = sanitize_text_field( $_POST['give-price-id'] );
```
kjohnson, line 195: @ravinderk would this not select the `custom` level id? I remember that working during my testing (I may be misremembering).

ravinderk: @kjohnson this give me correct price id and redirect me to correct url.

kjohnson: Ok, I think I see now. This redirects correctly, but does not update the price id *after* the redirect.

PR #5917, opened 2021-08-12
Legacy Consumer: allow custom class name on root element

### LINKED ISSUE #5916: Legacy Consumer: support custom class names on root element
## Details

The Fields API does not allow class names to be set for nodes by default as it is an abstract representation of a form and does not attempt to describe the actual HTML structure. Adding class names would force consumers, like the legacy consumer, to render the HTML is a particular way and that would defeat the purpose of what we are trying achieve with the Fields API. Since each node in a Fields API collection must have a unique name, consumers can define their own API for styling. We need the legacy consumer to be able to support custom class names so that we can upgrade addons like Form Field Manager to use the Fields API.

I suggest that we use hooks to allow custom class names to be set on the root element of each Fields API node that is rendered by the legacy consumer. Ideally, we will want the hooks to be re-used beyond the legacy consumer, so that should be kept in mind of the naming.

## Acceptance Criteria

- [x] Hooks are created for setting class names on root elements rendered for Fields API node

knowler:
Resolves #5916

## Description

Since the Fields API does not support setting class names, we decided that consumers should use hooks for this instead. This pull request adds a hook/filter for setting a custom class name on the root element for a field.

The hook name is `give_form_{ID}_field_classes_{fieldName}`.

## Affects

Legacy consumer renderer

## Testing Instructions

Try the filter out when registering fields.

## Pre-review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### New
+
+- Add `give_form_{ID}_field_classes_{fieldName}` hook to legacy consumer for setting classes on field wrapper (#5917)
+
 ## 2.13.1 - 2021-08-20
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


knowler:
@JasonTheAdams Right now I have `give_form_field_class_name_{$field->getName()}` set which is not good since there needs to be something to identify the form. Using the form ID is probably the best solution. Also, I didn’t put a lot of thought into the hook name in general. Any suggestions would be appreciated.

JasonTheAdams:
We can pass form data to the template (its fine for field views to be aware of the form they're in in the Legacy Consumer). You can then pass the form id as the second filter parameter.

knowler:

JasonTheAdams (approved):
Looks good, @knowler. 👍

PR #5919, opened 2021-08-16
Fields API: support conditional visibility

### LINKED ISSUE #5903: Fields API: fields, elements, and groups should be allowed to be conditional based on field values
## Details

The `Field`, `Element`, and `Group` should be allowed to be conditionally displayed based on field values. Of course, the Fields API isn’t responsible for the logic for hiding and showing fields and that is up to the consumer, but this will allow users to specify the requirements for a node to be displayed.

I am being intentional with mentioning that `Field`, `Element`, and `Group`s should be allowed to be conditional since I don’t think that every node should be allowed to be conditional. This is because stuff like `Form`s shouldn’t be conditional and there might be other node types that emerge that do not need the conditionality. Without merging #5889 and this current proposal, anything that extends `Group` will be conditional, so `Form` will have the conditional functionality.

I suspect that as a part of implementing this we will create a new contract for the conditional functionality so that consumers can easily check if a node enforces it.

Some additional research will be required for the actual syntax and serialization of this feature.

## Acceptance Criteria

- [x] Implement reusable conditional visibility functionality
- [x] Fuctionality for `Field`, `Element`, and `Group` (probably via a trait)

knowler:
Resolves #5903 

## Description

Conditional visibility is something that we need the Fields API to allow if we want to allow users to have the long coveted feature of conditional fields in FFM. It is also useful for the entire GiveWP ecosystem since we will dogfood this API for building forms ourselves.

This PR introduces a trait `Give\Framework\FieldsAPI\Concerns\HasVisibilityConditions` which is used in all the major `Node` types: `Field`, `Element`, and `Group`. This trait adds a protected property `visibilityConditions` which is an array of `BasicCondition`s. In the future it will be an array of `Condition`s — the abstract class `BasicCondition` extends so that it can also include `NestedCondition`s. For now we chose to keep it simple. Further, this trait has four public methods:

- `HasVisibilityConditions::getVisibilityConditions` for... you guess it, getting the visibility conditions. 
- `HasVisibilityConditions::showIf` for setting a single visibility condition in a straightforward plain way.
- `HasVisibilityConditions::showWhen` for setting multiple complex conditions using `BasicCondition` or the corresponding array format (based on the former class’ constructor signature).
- `HasVisibilityConditions::jsonSerialize` to implement `JsonSerializable` for the class which lets us serialize the JSON.

It also introduces two concrete classes and one abstract class to the
`Give\Framework\FieldsAPI\Conditions` namespace:

- `Condition` is an abstract class used by the following condition type classes.
- `BasicCondition(string $field, string $operator, mixed $value, $boolean = 'and')` used to express a basic condition
- `NestedCondition(Condition[] $conditions, $boolean = 'and')` used to collect `Condition`s in order to express that they should be evaluated together (functions like parenthesis in maths).

The data structures used for conditions takes inspiration from Laravel’s internal query builder data structure. We changed to suit our purposes. Here’s a break down of a `BasicCondition`:

- `field`: this is the name of the field to compare.
- `operator`: the operator used in the condition. Right now, the API supports `=`, `!=`, `>`, `>=`, `<`, `<=`
- `value`: the value the field is being compared with.
- `boolean`: this is either `and` or `or`. This is used to epress how the condition should be evaluated with the preceding condition. This property has no effect for the first condition in an array since there is nothing that preceeds it.

When serialized, these objects also include a `type` property which is `basic` for `BasicCondition`.

## Affects

- `Field`, `Element`, and `Group` allow have the `HasVisibilityConditions` trait now.

## Testing Instructions

Run this code as a PHP file in the root of the plugin. Experiment with it.

```php
<?php

namespace Give\Framework\FieldsAPI; // Yes, I’m cheating.

require_once __DIR__ . '/vendor/autoload.php';

$form = Form::make('donations')->append(

    Select::make('fund')->options(
        Option::make('Save the dolphins'),
        Option::make('Protect the rainforests'),
        Option::make('Devin’s claw machine toy fund')
    ),

    Checkbox::make('inCannibalCorpse')
        ->label('Do you play in a band called Cannibal Corpse?')
        ->showIf('fund', '=', 'Devin’s claw machine toy fund'),

    Checkbox::make('areYouSure')
        ->label('Are you sure your name isn’t Corpsegrinder?')
        ->showWhen(
            ['fund', '=', 'Devin’s claw machine toy fund'],
            ['inCannibalCorpse', '=', false]
        )

);

// Look in the mirror, JSON.
var_export(json_encode($form, JSON_PRETTY_PRINT));
```

## Pre-review Checklist


knowler:

src/Framework/FieldsAPI/Conditions/BasicCondition.php
```diff
@@ -0,0 +1,96 @@
...
+	public $operator;
+
+	/** @var string */
+	public $boolean;
+
+	/** @var string[] */
+	const OPERATORS = [ '=', '!=', '>', '>=', '<', '<=' ];
+
+	/** @var string[] */
+	const BOOLEANS = [ 'and', 'or' ];
```
JasonTheAdams, line 31: Minor thing, but please group constants at the top.

JasonTheAdams (approved):
One minor improvement, but this looks great!

PR #5932, opened 2021-08-23
Allow Fields API Text fields to repeat

### LINKED ISSUE #5904: Fields API: allow fields to be repeatable
## Details

A user might want to have an array like field, where the field input is repeated and it collects multiple values. This is different than the multiple functionality the Fields API already provides, since in this case we are creating distinct fields for each input whereas the multiple functionality just indicates that the single input is able to collect multiple values.

As far as I can tell, this is functionality that will just be added to the `Field` abstract class and so all idiomatically defined fields (i.e. field type created by extending `Field`) will have this functionality. This means that there is no need for any extra contracts to aid consumers.

As far as how it is implemented, it probably will just be a trait with a setter method and a getter method. Consumers will be responsible for how this is implemented.

## Acceptance Criteria

- [x] Implement repeatable functionality for `Text`
- [x] Legacy consumer needs to support repeating fields

knowler:
Resolves #5904

## Description

We’ve determined that fields should be allowed to repeat in the Fields API. This adds a trait to describe that capability. Right now, we are limiting this feature to just `Text` field types as that is the only type we need it for.

Repeatability is really just a descriptor, so there’s no complex API that needs to be implemented in the Fields API. The complexity is in the legacy consumer. All this adds is a getter and a setter for a protected property.

## Affects

- `Text` Fields API type.

## Testing Instructions

```php
<?php

$text = Give\Framework\FieldsAPI\Text::make('middleName');

$text->isRepeatable(); // false

$text->repeatable();

$text->isRepeatable(); // true

$text->maxRepeatable( 4 );

$text->repeatable( false ); 

$text->isRepeatable(); // false

echo json_encode($text, JSON_PRETTY_PRINT); // See repeatable prop on object
```

## Pre-review Checklist

- [x] Includes unit and/or end-to-end tests

### DIFF
--- a/src/Form/LegacyConsumer/FieldView.php
+++ b/src/Form/LegacyConsumer/FieldView.php
@@ -52,6 +52,12 @@ class FieldView {
 				include static::getTemplatePath( 'label' );
 				include static::getTemplatePath( $type );
 				break;
+			case Types::TEXT:
+				/** @var \Give\Framework\FieldsAPI\Text $field */
+				$typeAttribute = 'text';
+				include static::getTemplatePath( 'label' );
+				include static::getTemplatePath( $field->isRepeatable() ? 'repeatable-text' : 'base' );
+				break;
 			// By default, include a template and use the base input template.
 			default:
 				// Used in the template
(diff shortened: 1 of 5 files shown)


knowler:
@JasonTheAdams Do you think I should add the legacy consumer stuff for this here? Or in a separate PR?

JasonTheAdams:
You can just do it here. That's fine.

PR #5961, opened 2021-09-14
Donor addressed is now spaced out on the Donor Dashboard receipt

### LINKED ISSUE #5874: Donor addressed should be spaced out on the Donor Dashboard receipt
## Details

In 2.11.0 we corrected address spacing when editing donors, unfortunately this affected how the address is displayed on the Donor Dashboard in the receipt (see screenshot). This needs some formatting to make it look nice again.

## Visuals

image

## Acceptance Criteria

-   [x] Donor addressed is spaced out on the Donor Dashboard receipt

alaca:
Resolves #5874

## Description
This PR resolves the issue where the Donor billing address was not spaced out. The reason for that is we used `<br>` tags to separate the address details instead of spaces. The issue has been resolved by simply replacing the `<br>` tags with spaces. 

## Affects

Donation receipt

## Visuals
| Before | After |
|---|---|
| image | image |

## Pre-review Checklist

### DIFF
--- a/src/DonorDashboards/Repositories/Donations.php
+++ b/src/DonorDashboards/Repositories/Donations.php
@@ -1,10 +1,11 @@
 <?php
 namespace Give\DonorDashboards\Repositories;
 
+use Give\Receipt\LineItem;
 use Give\ValueObjects\Money;
 use Give\Framework\Database\DB;
 use Give\Receipt\DonationReceipt;
-use InvalidArgumentException;
+use Give_Payment;
 
 /**
  * @since 2.10.0
--- a/src/DonorDashboards/Repositories/Donations.php
+++ b/src/DonorDashboards/Repositories/Donations.php
@@ -212,7 +212,6 @@ class Donations {
 		$receiptArr = [];
 
 		$sectionIndex = 0;
-		$lineBreaks = [ '<br />','<br>','<br/>' ];
 		foreach ( $receipt as $section ) {
 			// Continue if section does not have line items.
 			if ( ! $section->getLineItems() ) {
@@ -247,8 +246,7 @@ class Donations {
 				}
 
 				$label = html_entity_decode( wp_strip_all_tags( $lineItem->label ) );
-				$value = str_ireplace( $lineBreaks, "\r\n", $lineItem->value );
-				$value = $lineItem->id === 'paymentStatus' ? $this->getFormattedStatus( $payment->status ) : html_entity_decode( wp_strip_all_tags( $value ) );
+				$value = $lineItem->id === 'paymentStatus' ? $this->getFormattedStatus( $payment->status ) : html_entity_decode( wp_strip_all_tags( $lineItem->value ) );
 
 				$receiptArr[ $sectionIndex ]['lineItems'][] = [
 					'class' => $detailRowClass,
--- a/src/DonorDashboards/Repositories/Donations.php
+++ b/src/DonorDashboards/Repositories/Donations.php
@@ -212,6 +212,7 @@ class Donations {
 		$receiptArr = [];
 
 		$sectionIndex = 0;
+		$lineBreaks = [ '<br />','<br>','<br/>' ];
 		foreach ( $receipt as $section ) {
 			// Continue if section does not have line items.
 			if ( ! $section->getLineItems() ) {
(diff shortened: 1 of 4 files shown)


ravinderk:
@alaca donor address is not formatted as given in design https://www.figma.com/file/jaVXDA4wjQD9OqTi0l9KQe/Form-Templates?node-id=1882%3A44

Form Templates 2021-09-14 at 4 22 58 PM

alaca:
@ravinderk yeah, I was afraid of that. That's why I've requested your review.  
We could probably remove the line breaks and handle that with CSS, but I'm not sure that will always match the design. 
Anyways, I reverted the code and now I'm handling this with the only option available.

ravinderk (changes requested):
@alaca I am seeing the same issue. The address is not formatted.

Donor Dashboard Desing: https://www.figma.com/file/BKr2K0XdDyRRoccJbJKRD7/Epic-Donor-Profile?node-id=397%3A1743

Donor Dashboard – give 2021-09-14 at 6 30 45 PM

alaca:
@ravinderk yeah, I see. The problem here is that on this screen is React application which means if we want to match the design on this screen, each row will have to be a separate React element. That will require refactoring of display logic completely. This is what we use now  https://github.com/impress-org/givewp/blob/c02be08c6da34e3fc869c360144edcfbfec321ee/src/DonorDashboards/resources/js/app/components/donation-receipt/index.js#L24

I wonder is that worth the extra effort.

ravinderk:
> @ravinderk yeah, I see. The problem here is that on this screen is React application which means if we want to match the design on this screen, each row will have to be a separate React element. That will require refactoring of display logic completely. This is what we use now  https://github.com/impress-org/givewp/blob/c02be08c6da34e3fc869c360144edcfbfec321ee/src/DonorDashboards/resources/js/app/components/donation-receipt/index.js#L24
>
>I wonder is that worth the extra effort. 

@alaca I agree with you. this needs refactoring in React application. what do you think about this patch: https://github.com/impress-org/givewp/compare/ad2e17b67479...bebef06df34e

src/DonorDashboards/Repositories/Donations.php
```diff
@@ -211,6 +212,7 @@ protected function getReceiptInfo( $payment ) {
 		$receiptArr = [];
 
 		$sectionIndex = 0;
+		$lineBreaks = [ '<br />','<br>','<br/>' ];
```
alaca, line 215: Actually, since you did a really good job by handling the line breaks with CSS, this part of the code is not necessary. I'll explain below.

src/DonorDashboards/Repositories/Donations.php
```diff
@@ -245,7 +247,8 @@ protected function getReceiptInfo( $payment ) {
 				}
 
 				$label = html_entity_decode( wp_strip_all_tags( $lineItem->label ) );
-				$value = $lineItem->id === 'paymentStatus' ? $this->getFormattedStatus( $payment->status ) : html_entity_decode( wp_strip_all_tags( $lineItem->value ) );
+				$value = str_ireplace( $lineBreaks, "\r\n", $lineItem->value );
```
alaca, line 250: We don't have to replace anything if we fix the source of the problem.

src/Receipt/DonationReceipt.php
```diff
@@ -134,7 +134,7 @@ private function getDonorEmailLineItem() {
 	private function getDonorBillingAddressLineItem() {
 		$address = getDonationDonorAddress( $this->donationId );
 		$address = sprintf(
-			'%1$s<br>%2$s%3$s,%4$s%5$s<br>%6$s',
+			'%1$s<br>%2$s%3$s, %4$s%5$s<br>%6$s',
```
alaca, line 137: So basically, here is the source of the problem. If you just replacing the `<br>` tags with the newline character, the problem will be resolved. The CSS rule which will handle this properly is already there.

alaca (commented):
@ravinderk good idea to approach this problem from the CSS side, and it will be best to keep it on that side. That said, I left few comments explaining why we actually don't need to do any kind of reformatting if resolve this on the source of the problem.

src/DonorDashboards/Repositories/Donations.php
ravinderk: removed: https://github.com/impress-org/givewp/compare/959aaa5b55ed...2f9f4a276fad

src/Receipt/DonationReceipt.php
ravinderk: fixed: https://github.com/impress-org/givewp/compare/959aaa5b55ed...2f9f4a276fad

src/DonorDashboards/Repositories/Donations.php
ravinderk: removed: https://github.com/impress-org/givewp/compare/959aaa5b55ed...2f9f4a276fad

alaca (commented):
Thanks for your help on this, @ravinderk. It's working great.
Although this was a relatively simple issue, in the end, teamwork was what resolved it. :mechanical_arm: 

Merge it!!!

PR #5983, opened 2021-09-23
Prevent multiple donations on mobile devices

### LINKED ISSUE #5982: On mobile, multiple, fast clicks on the donate button will generate multiple donations
On rare occasions, if the donor is able to make multiple and really fast clicks on the donation button at the end of the form, multiple donations are be made.

https://givewp.canny.io/admin/board/bug-reports/p/on-mobile-multiple-fast-clicks-on-the-donate-button-will-generate-multiple-donat

alaca:
Resolves #5982

## Description

Although I was not able to reproduce this issue on multiple mobile devices and multiple browsers, I can understand, potentially, why this happens in the first place. The events attached to the Donate Now button are `touched` and `click`, so I suspect that some of the older mobile browsers handle those events incorrectly by firing both events at the same time. 

To be on the safe side and prevent potential double donations in the future, an additional JS check is added which will prevent this behavior. 

## Affects
Donate Now button

## Testing Instructions
I couldn't reproduce the issue I'm fixing in this PR, but the Support team says that on rare occasions, multiple and really fast clicks on Donate Now button trigger it. So, try to click multiple times very fast on the Donate Now button :sweat_smile: 

## Pre-review Checklist

### DIFF
--- a/assets/src/js/frontend/give-ajax.js
+++ b/assets/src/js/frontend/give-ajax.js
@@ -203,11 +203,22 @@ jQuery( document ).ready( function( $ ) {
 	} );
 
 	/**
+	 * Used to determine if a donation is already in progress
+	 * @type {boolean}
+	 */
+	let inProgress = false;
+
+	/**
 	 * Donation Form AJAX Submission
 	 *
 	 * @description: Process the donation submit
 	 */
 	$( 'body' ).on( 'click touchend', 'form.give-form input[name="give-purchase"].give-submit', function( e ) {
+		if ( inProgress ) {
+			return;
+		}
+
+		inProgress = true;
 		//this form object
 		const $this = $( this );
 		const this_form = $this.parents( 'form.give-form' );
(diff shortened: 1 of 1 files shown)


alaca:
@JasonTheAdams I would like you to be the second pair of eyes that will check the updated code of this important functionality.

JasonTheAdams (commented):
Honestly, @alaca, I think the simplest solution here is to drop the `touchend` event and leave `click`. I believe that combination is from the early days of touch devices, when browsers were still figuring themselves out. Having looked around, people do all sorts of things to prevent double-triggers for that event combination, but it looks like it became unnecessary around 2017 — which makes sense. Semi-modern browsers know how to handle the click event on touch devices.

alaca:
@JasonTheAdams make sense. I will remove the `touched` event.

alaca:
@JasonTheAdams resolved c669b4d5

PR #6002, opened 2021-09-30
Add additional scroll to top functionality when switching between steps 2 (donation amount) and 3 (payment screen)

### LINKED ISSUE #5969: Multistep form steps should start at top of viewport
Each step of the multistep donation form should scroll to the top of the viewport for improved UX on all devices.

Canny Feedback:

https://givewp.canny.io/admin/board/feature-requests/p/multistep-form-steps-should-start-at-top-of-viewport

DevinWalker:
Resolves #5969 

## Description

This work adds an additional scrolling feature to steps 2 and 3 of the donation process. When changing from the introduction to other slides you should have a better user experience, particularly on mobile.

I'd like to hear from @mathetos and @knowler on their opinions here. The Canny feedback post is here:  https://feedback.givewp.com/feature-requests/p/multistep-form-steps-should-start-at-top-of-viewport

_Questions:_

1. I'm leaning towards making this feature only enabled for mobile if not at all. Right now this is experimental because it's largely untested widely it touches a pretty important visual part of the multi-step donation form. 
2. Is there a better place to put the use of `scrolltoIframeTop()`?

## Affects

- Donation Form

## Testing Instructions

Checkout branch, npm run dev, checkout on mobile using devtools.

## Pre-review Checklist

### DIFF
--- a/src/Views/Form/Templates/Sequoia/assets/js/form.js
+++ b/src/Views/Form/Templates/Sequoia/assets/js/form.js
@@ -39,6 +39,7 @@
 			$( '.step-tracker' ).removeClass( 'current' );
 			$( '.step-tracker[data-step="' + step + '"]' ).addClass( 'current' );
 
+			// Handle introduction step (whether disabled or not).
 			if ( templateOptions.introduction.enabled === 'disabled' ) {
 				if ( $( '.step-tracker' ).length === 3 ) {
 					$( '.step-tracker' ).remove();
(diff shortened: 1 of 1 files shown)


knowler (approved):
This feature is good for smaller screens. It makes sense to return to the top of the form for each step. I think this helps offset the faux-scrolling effect caused by the height of the form changing between steps. It makes sense to direct the user’s focus to where the form starts rather than allowing the differing heights of steps place their focus on content that is not yet relevant.

mathetos:
I agree with @knowler -- I like this overall. I think this will be even more helpful for forms that have a lot of custom fields, or the Tributes add-on. The steps can get really really long. 

My only question is that it probably makes sense to have this applied to all the steps, specifically also for the Donation Confirmation page, since the payment slide can also get a bit long with multiple gateways and seeing the donation receipt from the top is really important.

mathetos (approved):
I commented on the main thread supporting this.

lcgitrdone:
Original change requester weighing in here...  

"My only question is that it probably makes sense to have this applied to all the steps, <snip>"

I agree with @mathetos and would remove the "probably."  What would be the use case where any other behavior would be desirable and not a headscratcher?

In any case, thank you all very much for fixing this!

PR #6023, opened 2021-10-07
Optimize Give icon SVG so that it doesn’t break Gutenberg

### LINKED ISSUE #6022: Gutenberg Blocks SVGs should not have identical IDs
So when selecting from the different Gutenberg blocks, Give uses an SVG to render the Give icon. However, it uses the following clip-path clipPath="url(#SVGID_2_)". Which would work like a charm if there was only one #SVGID_2_. Unfortunately, since GiveWP creates 5 different blocks, and each block defines it’s own set of #SVGID_1_ and #SVGID_2_‘s, we end up with 5 of each. This is against ID usage standards, which results in it breaking the clipping and not rendering the icon in some browsers (for example in Firefox):

https://givewp.canny.io/admin/board/bug-reports/p/gutenberg-blocks-svgs-have-identical-ids

- [x] The icons should display in Firefox

knowler:
Resolves #6022

## Description

The Give icon SVG is an export from Adobe Illustrator (circa 2015). This generated file does a few things that are outdated and unnecessary. I was able to simplify the SVG so that it doesn’t use `clip-path` at all. The use of the `id` attributes was causing issues in the Gutenberg editor in some browsers like Firefox. Overall, the size of the file has been greatly reduced. Further, there is now a single React component for the Give icon.

## Affects

Anywhere there is a Give icon that doesn’t use a gradient for its colour.

## Visuals

In Gutenberg (Firefox):

image

Settings pages:

image

## Testing Instructions

It would be good to go through most of the UI to confirm nothing is broken.

## Pre-review Checklist

### DIFF
--- a/includes/class-give-scripts.php
+++ b/includes/class-give-scripts.php
@@ -424,6 +424,7 @@ class Give_Scripts {
 				width: 18px;
 				height: 18px;
 				content: "\e800";
+				padding-top: 8px;
 			}
 		</style>
 		<?php


knowler:
@DevinWalker This PR affects the GiveWP icon everywhere that it is used in the plugin except for the admin menu (where it is a Dashicon via Icomoon). Would you be able to take a look and make sure it still looks correct?

PR #6049, opened 2021-10-20
Fix: conditions now work for the amount field and checkbox fields

### LINKED ISSUE #6047: Donation amount field should use appropriate numeric/decimal keypad on iOS
## User Story

As an iOS, I want the custom donation amount input to use the appropriate keypad so that I can use decimals in my donation amount.

## Details

Currently on iOS, the donation amount field uses the telephone keypad. This does not include decimals or commas (though commas can be added using the "pause" key — not ideal). The underlying input type `tel` is the issue and this can be fixed by using text with the `inputmode` attribute set to `decimal`.

## Expected Behavior

The custom donation amount input to use the appropriate keypad on iOS and Android. This means allowing for numbers as well as decimals and comma.

## Steps to Reproduce

1. Try setting the custom amount with a decimal value in the donation amount field on iOS.

## Visuals

https://user-images.githubusercontent.com/6908001/138127915-15cd7cdc-6ee3-4e73-a825-87282c7a01d7.mp4

givewp-amount-input-no-decimals-ios-keypad-numbers

givewp-amount-input-no-decimals-ios-keypad-symbols

## Additional Context

This field uses `type=tel` since that was a common hack to get mobile devices to use a numeric keypad.

About the hack:

- Since there is so much of a variety of telephone number formats, `type=tel` doesn’t have any built-in browser validation unlike `type=email` or `type=url`. It effectively is the same as `type=text` in this regard.
- On mobile, `type=tel` uses a numeric keypad on Android and the telephone keypad on iOS. This is obviously preferred over a standard keyboard — especially for amounts. The iOS keypad doesn’t allow for decimals which is a problem.
- From limited testing with VoiceOver, there seems to be no accessibility issues with using `type=tel` in this way. VoiceOver announces it as a regular text field. It is a better choice than `type=number` in this regard, since the `type=number` is announced as a "stepper" indicating to the user that the field works in steps and we do not want that.

About this proposed changed:

- Since the change from `type=text` to `type=tel` did not change how the input behaves as far as validation, there is no risk in reverting back to `type=text`.
- The `inputmode` attribute allows us to hint to mobile browsers that it should use a numeric/decimal keypad for input. This has widespread mobile support from iOS 12.2+ (March 2019), Chrome 66 (April 2018). It’s hard to tell exactly when some mobile browsers started supporting this since many of them are forks of their desktop counterparts, but it seems like it’s been at least 3.5 years (Firefox has potentially partially supported this since 2012).

## System Information

- iOS 14.5
- GiveWP 2.15.0

</details>

## Acceptance Criteria

- [ ] The custom donation amount input to use the appropriate keypad on iOS and Android.

JasonTheAdams:
Resolves #6047 

## Description

This wraps up the new conditions by making a couple special cases:

**Amount field**
A requested field to make conditions for the amount field. This field is special in that it can be set manually or have a button change the value. This is tricky because programmatically changing an input's value doesn't trigger a change event. JQuery, however, does have a system for this; and since JQuery is available, I made use of it. As a bonus, this also improves conditions for text input fields in general.

**Checkbox field**
Must fields check against the input's value, but for checkbox fields we really just want to know if the field is checked or not. Getting the value involved just mucks things up. So now checkboxes are just compared to see if the input is checked or not, which is compared against whether the comparison operator is equals or not equals. So having a field show up in either condition works.

**Bonus:**
Updated the amount input to show the optimal UI on mobile

## Affects

Conditional field checks

## Testing Instructions

Amount conditions:
1. Install FFM at this PR: https://github.com/impress-org/give-form-field-manager/pull/345
2. Create a field (I just use HTML fields)
3. Make a manual condition with the field name as `give-amount`
4. Play with the operator and value
5. Make sure the conditions work as expected

Checkbox conditions:
1. Create a checkbox field
2. Create two other fields
3. Set the condition for the first to be the checkbox, equals, and some (any) value
4. Set the condition for the latter to be the checkbox, NOT equals, and some (any ) value
5. View form — the first field should be visible when checked, and the latter should be visible when not

## Pre-review Checklist

### DIFF
--- a/src/Form/LegacyConsumer/resources/js/conditinal-fields.js
+++ b/src/Form/LegacyConsumer/resources/js/conditinal-fields.js
@@ -87,7 +87,7 @@ document.addEventListener('readystatechange', event => {
 					const comparisonResult = compareWithOperator(operator, inputValue, value);
 
 					if (fieldType === 'checkbox') {
-						if ((input.checked && operator === '=') || (!input.checked && operator === '!=')) {
+						if (comparisonResult && (input.checked && operator === '=') || (!input.checked && operator === '!=')) {
 							visible = true;
 						}
 					} else if (fieldType === 'radio') {


JasonTheAdams:
@knowler Can you please test this out when you have some time?

PR #6070, opened 2021-10-30
Setup the Classic template frontend

knowler:
## Description

The pull request contains substantial progress for the classic template form’s front-end. The current approach is as follows:

- The header’s markup is generated and added to the form server-side.
  - This uses a `<meter>` element for the progress bar. I’ve styled it in Chrome. Firefox doesn’t seem difficult. Safari might be a problem, so I could end up just using a `<div>` visually.
- `<section>` elements are used to create boundaries between markup added by hooks. We use an arbitrarily high/low priority numbers (i.e. 1000 and -1000) so that we can catch most of the markup, but still allow the user to add markup between the sections (not ideal, but this is WordPress, people like to do weird stuff).
- JS is used on the frontend to move around and add nodes that cannot be added server-side.
  - This doesn’t seem to cause any problems of stuff flashing on the screen as it moves.
  - Some of what I've done with JS might be able to be done server-side. We can always optimize later.
  - I’m using `vhtml` so I can write stuff with JSX (it makes strings — no vdom) since doing DOM stuff imperatively is slow to code and very hard to follow.
- I am dequeuing form styles from core and addons.
  - I figure there's no sense fighting against ourselves.
  - Also, this way all the styles live together rather than been fragmented.
  - This allows us to use less specific selectors which makes the template optimal for custom styles.
  - The CSS doesn’t affect the function of the forms.
- For "dynamic" styling, I am leveraging CSS variables where I can.

There will definitely be room for optimizing CSS selector naming and specificity later on. Right now I’m just letting the template tell me how it wants to be organized. Once we get there, I’ll go through and optimize it all.

PHP-wise I’m not doing stuff optimally. I figure I’ll leave that to Ante to figure out, while I just focus on the front-end.

## Affects

Adds an actual front-end for the Classic template.

## Pre-review Checklist


src/Views/Form/Templates/Classic/resources/js/not-jquery.js
```diff
@@ -0,0 +1,60 @@
...
+	temp.innerHTML = htmlString;
+	return temp.content.firstChild;
+}
+
+/**
+ * Execute a function when the DOM is fully loaded.
+ *
+ * @param {function} fn
+ */
+export const domIsReady = fn => document.readyState !== 'loading'
```
alaca, line 58: Is the `document.readyState` check really needed here? Why don't just add a listener to the `DOMContentLoaded` event?

alaca (approved):
Great work, @knowler. 
I only have one question.

src/Views/Form/Templates/Classic/resources/js/form.js
```diff
@@ -0,0 +1,99 @@
...
+	addPaymentDetailsDescription();
+	splitDonationLevelAmountsIntoParts();
+});
+
+/**
+ * Individual transformations
+ */
+
+function removeTestModeMessage() {
+	removeNode(document.querySelector('#give_error_test_mode')); // Get out of my way!
```
alaca, line 23: There is an error in the console pointing to this line 

image

knowler: Ah, I just need to make that conditional. That whole thing was temporary anyway.

src/Views/Form/Templates/Classic/resources/js/not-jquery.js
knowler: This function is ultimately derived from jQuery’s old `.ready()` function. jQuery does a little extra to support IE. Checking `document.readyState !== 'loading'` is for `async` scripts, since it is possible to miss the `DOMContentLoaded` event. `setTimeout(fn, 0)` makes the code execute asynchronously like how an event handler is called.

Now, I didn’t write this function with classic template in mind; I just copied and pasted this from an old PR to Sage that I had made to replace jQuery. Here we aren’t loading the JS asynchronously, but I would be tempted to just leave the extra check since this is WordPress and users can change how stuff is loaded.

I was going to bring this up at some point: WordPress has this function standalone as an include `wp-dom-ready` (I might have spelled something wrong). We can just enqueue that instead. Also, we could use jQuery just for this since it’s enqueued already. I’m personally for not using jQuery where it makes sense not to use it, so that there's less work later on when we drop it.

knowler: I’m actually going to switch to follow this approach here. It potentially fires before the DOM is interactive, which we want since we are swapping out content.

https://gomakethings.com/the-quickest-way-to-detect-when-the-dom-is-ready/

alaca: I don't mind if it's reliable.

PR #6076, opened 2021-11-03
Chore: set Babel to use classic runtime for React

knowler:
Resolves #

## Description

Laravel Mix doesn’t support file-relative Babel config files. Well, it actually likes to use a file-relative config as the project wide config which is a bit strange. It seems to be silently obstinate about fixing the matter. So this means we need to set the runtime for anything that’s using React. Since we just recently changed it to automatic as an effect of using LaraveL Mix, we know that we don’t actually need the automatic runtime quite yet. In order to do this, we need to stop using `.react()`. For simplicity sake, I just moved the entire Babel config into the Mix file. While I was at it I cleaned up the Babel config, since there were a few plugins that we didn’t really need.

I also moved the browserslist to the `package.json` file since that way other tools can use it (e.g. autoprefixer). We might want to actually tweak this to reflect what we actually support.

## Affects

Just the JS transpilation configuration.

## Testing Instructions

Build and test everything.

## Pre-review Checklist


knowler:
@JasonTheAdams This fixes the issues I was having with Babel. Can you review when you get a chance? Then I’ll merge this into the Classic Template epic branch.

JasonTheAdams (commented):
@knowler So does this mean that, since we're not using automatic, everywhere else we're using React we have to import it like in the olden days?

JasonTheAdams:
This isn't needed at this time, so we'll keep it simple.

PR #6093, opened 2021-11-10
Changed: Payment gateway selector renders HTML correctly

### LINKED ISSUE #6092: Payment gateway selector renders HTML correctly

alaca:
Resolves #6092

## Description

On each gateway change, an ajax request is sent to the server and then the response which contains HTML markup like form, submit button, etc. specific to the selected gateway is returned back. The problem with this approach is that returned markup is loaded into a single container, and it is impossible to match designs in Figma for the Classic Form template with this approach. 

This PR resolves this issue by adding logic that filters the ajax request which loads the gateway data before the request is sent to the server. And it splits the returned response markup so that each returned section can be injected into the proper section of the template. 

## Visuals
Screen Capture_select-area_20211110162903

## Pre-review Checklist


alaca:
@knowler you want to take care of the styling of the returned gateway HTML, or should I do it?

knowler:
@alaca Let’s leave that out of this PR and we can make that a separate task. Then we can work divide up that work. I’m happy with this PR just being the HTML stuff.

src/Views/Form/Templates/Classic/resources/js/form.js
```diff
@@ -101,11 +104,75 @@ function splitDonationLevelAmountsIntoParts() {
                 h('span', {className: "give-amount-decimal"}, decimalForAmount),
             ),
             // There’s an intentional leading space before the currency symbol.
-            currency.symbolPosition === 'after' && ` ${h(CurrencySymbol, {position: 'after'})}`,
+            //currency.symbolPosition === 'after' && ` ${h(CurrencySymbol, {position: 'after'})}`,
```
alaca, line 107: @knowler I forgot that I commented this out. There were some issues with displaying the currency. Can you uncomment this and see what is going on?

knowler: Resolved in a68dbd9

I think the `h` function just didn’t like the padded string. I realized that all that stuff is just visual and not going to be read by a screen reader so the space wasn’t important to keep.

knowler:
I’m interested if there are any existing scenarios where the the personal information from the response is actually different than what’s already in the form. I update the code to also re-add the title and description which are lost that section is replaced with what comes from the response. It just got me thinking whether we actually _need_ to be doing all this. I might take a look at the difference in the markup between gateways during cool down. I think it is fine for now and visually no one can tell what’s even happening. The DOM is pretty fast for stuff like this.

knowler:
Also, I need to update the CSS since I changed where the gateway payment details are placed. I put them inside of the list item rather than in another one.

knowler (approved):
This is good to go now. I might change the focus styles for the radio at some point. Ideally, they would get the entire label.

PR #6103, opened 2021-11-12
Epic: In-Plugin Upsells

knowler:
## Description

This introduces three upsell opportunities in GiveWP:
1. A cool new add-ons page
2. A slick and fresh recurring tab (when recurring is not installed)
3. A dope banner system for sharing sales in date ranges

Cowabunga

## Affects

The add-on link now goes to a page instead of going off-site.


knowler:
SInce this is more than just the add-ons page. We will need to re-organize the `InPluginUpsells` domain to reflect that.

JasonTheAdams (approved):
Thanks for all your excellent work, @alaca and @knowler! I think this may really help with our marketing and show users awesome features they didn't know about!

PR #6104, opened 2021-11-12
New: add Recurring Donations tab in Give settings

knowler:
Resolves #

## Description

This PR adds a tab for Recurring Donations in the Give settings. This is only shown when the plugin is not installed.

This PR is a work in progress.

## Pre-review Checklist

- [ ] [Self Review](https://give.gitbook.io/development-manual/devops/github/code-reviews#self-review) of code and UX completed


knowler:
@alaca I’ve begun updating the `InPluginUpsells` domain to reflect the multiple concerns (i.e. addons admin page, recurring donations tab, sale banner). My progress on the recurring donations tab is also here though I haven't been able to test it visually yet as I don't have the backend code set up for that settings tab to be shown in the Give settings.

alaca:
@knowler shouldn't recurring donations stuff be in recurring add-on and not in core? I mean, recurring is a separate add-on so it makes sense to have that logic there.

alaca:
Hmm... scratch that, I realized now that there is a page showing when the add-on is not installed. Okay, we should remove the settings page logic from the recurring add-on and add that to core.

knowler:
@alaca I see what you’re saying. That second view in Figma is functionality for when recurring _is_ installed. The part of me that likes categorizing and naming things correctly feels offended if we leave that in the upsells domain.

I wonder if we should just do the second view in the recurring addon. I’m not super worried about duplicating code. The frontend is just a single JS file and there are less commonalities between that view and any of the other views found in the upsells domain.

alaca:
Yeah, I'm not worried about duplicating the code, but I do think the second screen should be in the recurring add-on.

knowler:
@DevinWalker @alaca This is ready for review.

src/InPluginUpsells/resources/js/recurring-donations-settings-tab.js
```diff
@@ -0,0 +1,4 @@
+import {render} from 'react-dom';
+import {RecurringDonationsTab} from './components/RecurringDonationsTab';
+
+render(<RecurringDonationsTab />, document.getElementById('give-in-plugin-upsells'));
```
alaca, line 4: When the recurring add-on is activated, this container will be removed, but the URL will remain the same and this script will be loaded. So we have to check if element exists in DOM before rendering the app.

alaca (commented):
There is only one minor issue. Other than that, looks good 👍🏻

src/InPluginUpsells/resources/js/recurring-donations-settings-tab.js
knowler: > we could conditionally load these scripts based on whether the recurring add-on is activated or not.

This is what we should do.

alaca: Agreed. The easiest way to do this is to update the `RecurringDonationsTab::isShowing()` method by adding an additional check if the recurring add-on is installed or not. We can do something like this: `if ( defined('GIVE_RECURRING_VERSION') ) ...`

knowler:
@alaca Can you re-review this? I added the logic for checking if Recurring is active.

alaca (approved):
Great work, @knowler  👍🏻

PR #6112, opened 2021-11-17
New: Sale Banners

### LINKED ISSUE #6111: In-plugin sale banners

alaca:
Resolves #6111

## Description

We wanted a way to display sale banners on GiveWP admin pages. This pull request is a first slice for this feature. We’ve kept the system quite simple for now. We can build on this later on if we find it necessary.

### User Experience

https://user-images.githubusercontent.com/6908001/142496087-2597c1c3-9515-4343-a105-99db18459766.mp4

When the sale is active, the banner will display at the top of GiveWP admin pages underneath the page title. The user is presented with an action to visit the sale’s page in a new tab. 

https://user-images.githubusercontent.com/6908001/142496124-bffe4a66-1081-4981-8f79-88a777ead6cc.mp4

Each banner has a button to dismiss the banner. This option is remembered.

### Defining Banners

Banners can be defined in [the `Give\InPluginUpsells\SaleBanners::getBanners()` method](https://github.com/impress-org/givewp/blob/d2523429e565b08b9f23e93ec214c92f648dd8f1/src/InPluginUpsells/SaleBanners.php#L28-L50). Each banner must have a unique identifier (`id`) as these are stored in the `give_hidden_sale_banners` WordPress option when dismissed. Other required fields are:

- `iconURL`: This is the URL for the icon that displays at the start of the banner.
- `accessibleLabel`: This is announced to screen readers. Keep it concise.
    - For example, this will be announced “Black Friday/Giving Tuesday Sale complementary” in VoiceOver on macOS (“complementary” being the `<aside>` element that wraps the banner).
- `leadText`: This is the strong text at the start of the banner’s content.
- `contentText`: This is the regular text that follows the strong text.
- `actionText`: What should the sale link say? This appears after the `contentText`.
- `actionURL`: Where does the sale link go to?
- `startDate`: The exact time when the sale starts. Timezone should _not_ be included.
- `endDate`: The exact time when the sale ends. Timezone should _not_ be included.

Once banners are past their end date `iconURL`, `accessibleLabel`, `leadText`, `contentText`, `actionText`, `actionURL` will never be used provided the user has their timezone set correctly. For the future of this feature, it becomes safe to change the API of how banners are defined so long as the `id`, `startDate`, and `endDate` are kept. It might be useful to leave the definitions of past sales in place since then they can be referenced so that previous `id`s can be referenced when creating new ones (which, again, need to be unique). Honestly, we don’t need to worry too much about this right now — just thought I’d share that for posterity's sake.

## Testing Instructions

- `npm run dev`
- Try modifying the dates in the `Give\InPluginUpsells\SaleBanners::getBanners`
  method (link to code location above). Remember these are using the
  `America/Los_Angeles` timezone.
  - Set the plan’s start date to be future, then check if the banner displays. It should not.
  - Set the plan’s start date to be past and end date to be future, then check if the banner displays. It should display.
    - While the banner displays, try dismissing it, then refresh to ensure the preference is respected. It should not display.
      - Make sure to reset the banner hidden preferences by deleting the `give_hidden_sale_banners` WordPress option: `wp option delete give_hidden_sale_banners`.
  - Set the plan’s end date to be past, then check if the banner displays. It should not.

## Pre-review Checklist


knowler:
I’m done my work here (at least for today). It works. It’s styled. Looks decent on all pages. The correct banner content will need to be added. Otherwise, we are very close to finished.

src/InPluginUpsells/SaleBanners.php
```diff
@@ -0,0 +1,136 @@
...
+	 */
+	public function getVisibleBanners()
+	{
+		$currentDateTime = current_datetime()->getTimestamp();
+		$currentUserId = get_current_user_id();
+
+		return array_filter(
+			$this->getBanners(),
+			function ($banner) use ($currentDateTime, $currentUserId) {
+				$isHidden = in_array($banner['id'] . "_{$currentUserId}", $this->hiddenBanners);
```
alaca, line 61: Underscore between `$banner['id']` and `$currentUserId` should be removed. Otherwise, the hide banner action won't work.

src/InPluginUpsells/resources/js/sale-banner.js
```diff
@@ -0,0 +1,36 @@
+const bodyIsReady = onReady => window.requestAnimationFrame( document.body ? onReady : bodyIsReady );
```
alaca, line 1: Actually, we don't even need this because this script is loaded in page footer 😅

src/InPluginUpsells/SaleBanners.php
```diff
@@ -0,0 +1,136 @@
...
+
+	/**
+	 * Render admin page
+	 */
+	public function render()
+	{
+		$banners = $this->getVisibleBanners();
+
+		if (!empty($banners)) {
+			$saleIconURL = GIVE_PLUGIN_URL . 'assets/dist/images/admin/sale-icon.png';
```
alaca, line 92: I'm not sure this icon will be always reusable, so it might be better to add it as a banner definition prop.

alaca:
@knowler thank you for stepping in, great work!

src/InPluginUpsells/resources/js/sale-banner.js
knowler: Resolved in eb5de4e

src/InPluginUpsells/SaleBanners.php
knowler: Resolved in 2f377f3

src/InPluginUpsells/SaleBanners.php
knowler: Resolved in ab09675

knowler:
I’ve implemented all the feedback from @alaca and tested this all again. We just need the final content for the banner, then this PR will be ready.

knowler:

JasonTheAdams (approved):
Great work, @alaca and @knowler!

@DevinWalker Do you want to review the design before I merge this? Or are you good from the screenshots?

DevinWalker:
Looks good to me 👍🏻

JasonTheAdams:
Then let's make a like a tree... and merge this sucker. 😎

PR #6118, opened 2021-11-19
New: Donation Goal stats works

### LINKED ISSUE #6116: Donation Goal works

alaca:
Resolves #6116

## Description

This PR adds support for Donation Goal stats.

## Affects

Donation Goal stats

## Visuals
image

## Testing Instructions

1. Create a new form and enable Donation goal
2. View form and make a few donations
3. Check the goal stats

## Pre-review Checklist

### DIFF
--- a/src/Views/Form/Templates/Classic/resources/views/header.php
+++ b/src/Views/Form/Templates/Classic/resources/views/header.php
@@ -2,42 +2,58 @@
 /**
  * @var string $title
  * @var string $description
- * @var bool   $isSecureBadgeEnabled
- * @var bool   $secureBadgeContent
+ * @var bool $isSecureBadgeEnabled
+ * @var bool $secureBadgeContent
+ * @var bool $hasGoal
+ * @var array $goalStats
  */
+
 ?>
 <div class="give-form-header">
     <div class="give-form-header-top-wrap">
         <h1 class="give-form-title"><?= $title ?></h1>
         <p class="give-form-description"><?= $description ?></p>
-        <?php if ($isSecureBadgeEnabled) : ?>
-        <aside class="give-form-secure-badge">
-            <svg class="give-form-secure-icon">
-                <use href="#give-icon-lock" />
-            </svg>
-            <?= $secureBadgeContent ?>
-        </aside>
-        <?php endif; ?>
+        <?php
+        if ($isSecureBadgeEnabled) : ?>
+            <aside class="give-form-secure-badge">
+                <svg class="give-form-secure-icon">
+                    <use href="#give-icon-lock"/>
+                </svg>
+                <?= $secureBadgeContent ?>
+            </aside>
+        <?php
+        endif; ?>
     </div>
-    <aside class="give-form-stats-panel">
-        <ul class="give-form-stats-panel-list">
-            <li class="give-form-stats-panel-stat">
-                <span class="give-form-stats-panel-stat-number">$0</span> raised
-            </li>
-            <li class="give-form-stats-panel-stat">
-                <span class="give-form-stats-panel-stat-number">50</span> donations
-            </li>
-            <li class="give-form-stats-panel-stat">
-                <span class="give-form-stats-panel-stat-number">$10,000</span> goal
-            </li>
-            <li class="give-form-goal-progress">
-                <div class="give-form-goal-progress-meter" role="meter" aria-labelledby="meter-label" aria-valuemin="0" aria-valuemax="1000000" aria-valuenow="200000" aria-valuetext="20%">
-                    <div id="meter-label" hidden>
-                        $20,000 of $1,000,000 goal
+    <?php
+    if ($hasGoal): ?>
+        <aside class="give-form-stats-panel">
+            <ul class="give-form-stats-panel-list">
+                <li class="give-form-stats-panel-stat">
+                    <span class="give-form-stats-panel-stat-number">
+                        <?= $goalStats[ 'raised' ]; ?>
+                    </span> <?= __('raised', 'give'); ?>
+                </li>
+                <li class="give-form-stats-panel-stat">
+                    <span class="give-form-stats-panel-stat-number">
+                         <?= $goalStats[ 'count' ]; ?>
+                    </span> <?= $goalStats[ 'countLabel' ]; ?>
+                </li>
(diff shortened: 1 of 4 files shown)


src/Views/Form/Templates/Classic/resources/views/header.php
```diff
@@ -2,42 +2,60 @@
...
-        <?php if ($isSecureBadgeEnabled) : ?>
-        <aside class="give-form-secure-badge">
-            <svg class="give-form-secure-icon">
-                <use href="#give-icon-lock" />
-            </svg>
-            <?= $secureBadgeContent ?>
-        </aside>
-        <?php endif; ?>
+        <?php
+        if ($isSecureBadgeEnabled) : ?>
```
knowler, line 17: I know PSR-12 is doing this and this comment is unrelated to this PR, but I think we should not apply PSR-12 to files that are substantially HTML. You can just leave this though.

knowler (changes requested):
Nice work @alaca. I have a few minor requests.

src/Views/Form/Templates/Classic/resources/views/header.php
knowler: @JasonTheAdams Do you think we should use `.html.php` and then exclude that extension from being formatted to PSR-12, so stuff like this doesn’t happen in HTML?

JasonTheAdams: Would we use prettier instead, then?

src/Views/Form/Templates/Classic/resources/views/header.php
```diff
@@ -2,42 +2,60 @@
...
+                </li>
+                <li class="give-form-goal-progress">
+                    <div class="give-form-goal-progress-meter" role="meter" aria-labelledby="meter-label" aria-valuemin="0" aria-valuemax="1000000"
+                         aria-valuenow="<?= $goalStats[ 'progress' ]; ?>%" aria-valuetext="<?= $goalStats[ 'progress' ]; ?>%">
+                        <div id="meter-label" hidden>
+                            <?= sprintf(__('%s of %s goal', 'give'), $goalStats[ 'raised' ], $goalStats[ 'goal' ]); ?>
+                        </div>
+                        <div
+                            class="give-form-goal-progress-meter-bar"
+                            style="width: <?= $goalStats[ 'progress' ]; ?>%; background: linear-gradient(180deg, <?= $goalStats[ 'color' ]; ?> 0%, <?= $goalStats[ 'color' ]; ?> 100%), linear-gradient(180deg, #fff 0%, #ccc 100%); background-blend-mode: multiply;"></div>
```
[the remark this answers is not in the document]
alaca: Resolved 235e03d8

src/Views/Form/Templates/Classic/resources/views/header.php
knowler: I guess, ya. I haven’t really seen how Prettier formats HTML.

JasonTheAdams: This may kill you to hear, but I'm less interested in visual perfection and more interested in consistency. So I don't want to exclude something without a working and reliable alternative.

knowler: I just tried Prettier with HTML and I think it's good.

My thing is that the formatter should serve the primary language that it's formatting and when most of the file is HTML it just makes sense to me to format it like HTML.

alaca: Oh, man. Code formatting is hard.

knowler: Cause code formatting is really to serve the reader, right?

knowler: I'm also fine with just leaving this. We'll probably switch to React in GiveWP 3.

JasonTheAdams: Yeah, let's just move along for now. You're right, this will likely become React.

Formatting, to me, falls into a 90/10 trap. That is, most standards will take code 90% of the way to making it entirely readable. In fact, it's probably more. The last 10% (should curly brackets be on the same line!?!?) is just personal preference and highly subjective. So long as code is sensibly readable (90%), I don't care. At that point I just want consistency so our commits aren't ridiculous and more difficult to review.

knowler: You know what I don’t know if this is actually formatted with PSR-12 though. I just ran the `phpcbf --standard=PSR12` on it and it doesn’t seem to have any issues with keeping the `if` and `endif` on the same lines.

knowler: Resolved by https://github.com/impress-org/givewp/commit/f9a6954e544f6a04f2bf0197e950c6cc829cfb1e

PSR-12 actually doesn’t create issues formatting files with lots of HTML. We can carry on.

src/Views/Form/Templates/Classic/resources/views/header.php
```diff
@@ -2,42 +2,62 @@
...
+                </li>
+                <li class="give-form-goal-progress">
+                    <div
+                        role="meter"
+                        class="give-form-goal-progress-meter"
+                        style="--progress: <?= $goalStats[ 'progress' ]; ?>%"
+                        aria-labelledby="meter-label"
+                        aria-valuemin="0"
+                        aria-valuemax="<?= $goalStats[ 'goal' ]; ?>"
+                        aria-valuenow="<?= $goalStats[ 'raised' ]; ?>"
```
knowler, line 51: I realized after my first review that these two values should actually both be unformatted values. I think we will need to add new entries to the `$goalStats` array for both of these.

knowler (changes requested):
@alaca I have just one more change for you here. (Also, if your editor changes the formatting from what I changed it to you can just commit it. I realized that I was using PSR-12 while the shared PhpStorm config does PSR-12 and a little extra.)

src/Views/Form/Templates/Classic/resources/views/header.php
alaca: Resolved e8876edd

knowler (approved):
Nice work @alaca!

PR #6123, opened 2021-11-23
New: add offsite and routeMethods to gateway api

jonwaldstein:
## Description

This PR introduces a new _offsite_ concept for the gateway api.  When gateways need to go offsite to complete the transaction, they can now extend the `OffSitePaymentGateway` class that extends `PaymentGateway`, implements the new `OffsiteGatewayInterface`, and uses the new trait `OffsiteGateway`

Screen Shot 2021-11-29 at 11 15 17 AM

Screen Shot 2021-11-29 at 11 15 25 AM

..and inside `createPayment` they will return the new command `RedirectOffsite` which accepts a redirect url.  There is also a helper in the trait to generate the return url back to GiveWP: `$this->generateReturnUrlFromRedirectOffsite($paymentId, $args)`;

Screen Shot 2021-11-30 at 9 14 19 AM

Finally, gateways will have an access point `returnFromOffsiteRedirect` to complete the process and return GiveWP a status command.  At this time we will allow super global to access data the gateways need to complete themselves. Eventually there will be a more elegant solution for this part.

Screen Shot 2021-11-30 at 9 14 52 AM

The api for creating custom gateway routes that resolved to custom route methods will live within the gateway class using the new `routeMethods` constant.  

Route methods are used to extend the gateway api. By adding a custom route method, a gateway is effectively registering a new route url that will resolve itself and call the specified method.

```
/**
 * @inheritDoc
 */
public $routeMethods = [
    'testGatewayMethod'
];

 /**
 * @inheritDoc
 */
public function createPayment(GatewayPaymentData $paymentData)
{
    $redirectUrl = $this->generateGatewayRouteUrl('testGatewayMethod', $donationId);

    return new RedirectOffsite($redirectUrl);
}

 /**
 * @return PaymentComplete
 */
public function testGatewayMethod()
{
    $transactionId = "test-gateway-transaction-id";

    return new PaymentComplete($transactionId);
}
```

## Affects

- new command `RedirectOffsite` and handler `RedirectOffsiteHandler`
- new interface `OffsiteGatewayInterface` and trait `OffsiteGateway`

## Testing Instructions

In `RegisterPaymentGateways` uncomment out `TestGatewayOffsite` and use "Test Gateway Offsite" to donate.

## Pre-review Checklist


JasonTheAdams:
@jonwaldstein I don't think the transaction ID would be generated prior to the redirect, as the transaction occurs during the offsite flow and would be provided upon return.

src/PaymentGateways/Actions/RegisterPaymentGatewayRoutes.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Give\PaymentGateways\Actions;
+
+use Give\Framework\PaymentGateways\Routes\ReturnFromOffsiteRedirectRoute;
+use Give\Helpers\Hooks;
+
+class RegisterPaymentGatewayRoutes
```
JasonTheAdams, line 8: Rather than setting up one action per route (n+1), use a single action which matches the query string with the listener.

src/PaymentGateways/Gateways/TestGateway/TestGatewayOffsite.php
```diff
@@ -0,0 +1,93 @@
...
+
+        return $legacyFormFieldMarkup();
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function createPayment(GatewayPaymentData $paymentData)
+    {
+        $transactionId = "test-gateway-transaction-id-{$paymentData->paymentId}";
```
JasonTheAdams, line 74: I believe the transaction is generated offsite, so it's available in `returnFromOffsiteRedirect`, not here.

src/Framework/PaymentGateways/DataTransferObjects/GatewayOffsiteReturnData.php
```diff
@@ -0,0 +1,41 @@
...
+     *
+     * @return self
+     */
+    public static function fromRequest(array $request)
+    {
+        $self = new static();
+
+        $self->gatewayId = $request['give-gateway-id'];
+        $self->gatewayMethod = $request['give-gateway-method'];
+        $self->paymentId = $request['give-payment-id'];
```
JasonTheAdams, line 37: All query args are strings, so if this is an integer you'll need to cast it.

JasonTheAdams (changes requested):
Great work! I'm continuing to love how simple the resulting gateways are. I left some notes and have one additional idea I'm open to your feedback on:

What do you think about adding an `OffSiteGateway` abstract class that extends the base class, implements the interface, and uses the trait. The thinking here is that most gateways will be on-site or off-site, and having a single class to extend simplifies the API for many folks.

src/PaymentGateways/Gateways/TestGateway/TestGatewayOffsite.php
jonwaldstein: Yes, the `transactionId` would likely come after returning but I was just trying to show both an example of adding args to the return url and how to receive data within `returnFromOffsiteRedirect`.  I'm open to using another example but needed something tangible like the `transactionId` to use for this flow example.

src/PaymentGateways/Actions/RegisterPaymentGatewayRoutes.php
jonwaldstein: Fixed this up to be more clear / fluent 8ad87141

jonwaldstein:
> Great work! I'm continuing to love how simple the resulting gateways are. I left some notes and have one additional idea I'm open to your feedback on:
> 
> What do you think about adding an `OffSiteGateway` abstract class that extends the base class, implements the interface, and uses the trait. The thinking here is that most gateways will be on-site or off-site, and having a single class to extend simplifies the API for many folks.

Updated with a new abstract class 435a1f0a

src/Framework/PaymentGateways/DataTransferObjects/GatewayOffsiteReturnData.php
jonwaldstein: resolved 74ea9eb0

src/PaymentGateways/Gateways/TestGateway/TestGatewayOffsite.php
JasonTheAdams: That's fine I guess, but the `paymentId` wouldn't be available in the returning DTO.

jonwaldstein: I'm sending the `paymentId` in the query params for the return url so it's available when the route resolves.

src/Framework/PaymentGateways/Routes/GatewayRouteFacade.php
```diff
@@ -0,0 +1,112 @@
...
+     */
+    private $paymentGatewayRegister;
+
+    /**
+     * @param  PaymentGatewayRegister  $paymentGatewayRegister
+     */
+    public function __construct(PaymentGatewayRegister $paymentGatewayRegister)
+    {
+        $this->paymentGatewayRegister = $paymentGatewayRegister;
+        $this->paymentGateways = $this->paymentGatewayRegister->getPaymentGateways();
```
JasonTheAdams, line 30: I don't recommend doing this as it's splitting the source of truth. The gateways are locked in at this point, here. But if gateways are modified before being used within this class, it won't be reflected.

JasonTheAdams (changes requested):
On comment, but I also don't think we're connecting on the product id stuff. We shouldn't be checking for the product ID from the returning request as we don't know what shape it will take. The only thing we know, and should be included in the DTO, are our own parameters. Right now we're adding the product ID to the query parameters, but we really shouldn't be and it's only proving to confuse this flow.

Happy to hop on a call if it's still not making sense.

src/Framework/PaymentGateways/Routes/GatewayRouteFacade.php
jonwaldstein: Are you talking about the paymentGateways property?

jonwaldstein:
> On comment, but I also don't think we're connecting on the product id stuff. We shouldn't be checking for the product ID from the returning request as we don't know what shape it will take. The only thing we know, and should be included in the DTO, are our own parameters. Right now we're adding the product ID to the query parameters, but we really shouldn't be and it's only proving to confuse this flow.
> 
> Happy to hop on a call if it's still not making sense.

I think you mean `paymentId` right? It's a little tricky because we're leaving the server and don't have the `GatewayData` with `paymentId` anymore.  Once we return we need a way to get reference `paymentId` to finish things up. Unless we're able to store it in session somehow, the only way I was able to do this was adding it to the return url params.

src/Framework/PaymentGateways/Routes/GatewayRoute.php
```diff
@@ -0,0 +1,86 @@
...
+
+            /** @var PaymentGateway $gateway */
+            $gateway = give($paymentGateways[$data->gatewayId]);
+
+            $gatewayMethods = array_filter($gateway::routeMethods, static function ($method) use ($data) {
+                return $method === $data->gatewayMethod;
+            });
+
+            if (is_a($gateway, OffsiteGatewayInterface::class)) {
+                $gatewayMethods = array_merge($gatewayMethods, OffSitePaymentGateway::routeMethods);
```
JasonTheAdams, line 43: @jonwaldstein Remember that the class is for convenience. Under the hood the gateways or compositional. As such, let's add `routeMethods` to the interface instead of the class.

jonwaldstein: @JasonTheAdams I tried that first but was getting an error _Cannot inherit previously-inherited or override constant from interface_

JasonTheAdams: Interesting. We could try either static properties (my only issue with this is that the property becomes externally writeable), or just give the interface unique constant names. I'm fine with this as they're really only intended for internal use.

src/Framework/PaymentGateways/Routes/GatewayRoute.php
```diff
@@ -0,0 +1,88 @@
...
+                    $allowedGatewayMethods,
+                    OffsiteGatewayInterface::defaultRouteMethods
+                );
+            }
+
+            foreach ($allowedGatewayMethods as $gatewayMethod) {
+                if (!method_exists($gateway, $gatewayMethod)) {
+                    throw new PaymentGatewayException('The gateway method does not exist.');
+                }
+            }
```
JasonTheAdams, line 52: This is a bit over-complicated:
```php
            $allowedGatewayMethods = $gateway::routeMethods;

            if (is_a($gateway, OffsiteGatewayInterface::class)) {
                $allowedGatewayMethods = array_merge(
                    $allowedGatewayMethods,
                    OffsiteGatewayInterface::defaultRouteMethods
                );
            }

            if (!in_array($data->gatewayMethod, $allowedGatewayMethods, true) || !method_exists($gateway, $data->gatewayMethod) {
                throw new PaymentGatewayException('The gateway method does not exist.');
            }
```

jonwaldstein: nice suggestion 👍  208ef1d4

JasonTheAdams (approved):
Great work, @jonwaldstein! We added some features even more fundamental than just offsite in this. I'm excited for the continued progress.

PR #6133, opened 2021-11-30
New: Custom Google Font

### LINKED ISSUE #6132: Template can use custom Google font

alaca:
Resolves #6132

## Description
This PR adds support for custom Google fonts. 
Inspiration for the font list came from https://kinsta.com/blog/best-google-fonts/ 

## Affects

Classic template font.

## Visuals

|Backend|Frontend|
|---|---|
|image|image|

## Testing Instructions

1. Set Primary Font to be Custom Google Font
2. Choose font from the Google Font dropdown, save and preview the form
3. Repeat step 2 for all fonts :) 

## Pre-review Checklist


knowler:
I know it’s not in the design, but I wonder if we can simplify this and just have **Google Font** and **System Font** as the options? Then put **Montserrat** as the default choice. We could even use a single select control with the list of Google fonts, then have **System Font** as an option (at the top with **Montserrat**). @DevinWalker what do you think?

alaca:
@knowler good suggestion, I think it would be better to have 2 distinct options instead of everything in one dropdown, but let's wait for @DevinWalker

DevinWalker:
@alaca I like @knowler's suggestion. It just seems more clean and less confusing to have fewer options that do the same functionality.

knowler:
If the select field in the admin supported `<optgroup>` elements, we could put all the Google Fonts including Montserrat under there to make it clear that they are different.

alaca:
@knowler there is no [support](https://github.com/impress-org/givewp/blob/0b020fee3aebf1c9cdfc3adb36c334c0cc206ef3/includes/admin/give-metabox-functions.php#L854) to have `optgroup` elements.

knowler (approved):
Nice work, @alaca!

PR #6141, opened 2021-12-03
New: Classic Donation Form Template

knowler:
Resolves #

## Description

This pull request implement a new donation form template: classic. The template’s design is based on the multi-step template, however, it is presented in a single flow similar to the legacy template.

## Pre-review Checklist


knowler:
@DevinWalker I’ve pushed all of my changes for your requests. I am working on the new template selector stuff now. Can you review the form as it is while I work on that?

knowler:
@DevinWalker the template selector has also been updated. Let me know what you think.

JasonTheAdams:
@knowler I just pulled this down and made a new form, and I'm seeing the same notice at two spots:

Screen Shot 2021-12-23 at 4 31 24 PM

JasonTheAdams:
I'm also noticing a "null" value on a fresh form.

image

JasonTheAdams:
Just moved the feedback above to the Canny board: https://feedback.givewp.com/classic-template

knowler:
I think that I will need to set the `decimals_enabled` option with some default value for compatibility with other templates, but maybe hide it since we aren’t supporting it right now in the template. Not too sure how to do this though.

knowler:
So looks like the admin form API doesn’t support a `hidden` type so that doesn’t work. The other option is to just check if the `decimals_enable` option exists in `Give\Helpers\Form\Template::handleOptionsBackwardCompatibility` and set it as null or something if it doesn’t.

JasonTheAdams:
I noticed that the box shadow at the bottom of the form is clipped after making a donation. Alas, this is actually a result of the iframe-resizer library that we're using. When it resizes the iframe it doesn't take into consideration the height of the box shadow when calculating the new height.

It's not a big deal, but I did make an Issue on that repo to hopefully fix this in the future: https://github.com/davidjbradshaw/iframe-resizer/issues/1030

knowler:
@JasonTheAdams Ya, at one point I was going to apply the box shadow to the iframe and then remove all of the extra whitespace inside of it. I decided that would be too much work and too complicated.

JasonTheAdams:
Yeah, I gave myself 5 minutes to figure it out. That was my conclusion. 😁

PR #6143, opened 2021-12-08
Condition operators are now decoded in case of values stored encoded

kjohnson:
Resolves https://github.com/impress-org/give-form-field-manager/issues/376

## Description

This PR adds a line to decode the operator before validation, in case the operator value is coming from the database and encoded as an HTML entity.

## Testing Instructions

- Using GiveWP 2.17.1
- Using FFM 2.0.1
- Set an FFM field's visibility to Less Than or Less Than Or Equals
- Form **should** load and **should not** throw the following error:
`Fatal error: Uncaught Give\Framework\Exceptions\Primitives\InvalidArgumentException: Invalid operator: &lt;=. Must be one of: =, !=, >, >=, <, <= in give/src/Framework/FieldsAPI/Conditions/BasicCondition.php on line 45`

## Pre-review Checklist

### DIFF
--- a/src/Framework/FieldsAPI/Conditions/BasicCondition.php
+++ b/src/Framework/FieldsAPI/Conditions/BasicCondition.php
@@ -41,6 +41,13 @@ class BasicCondition extends Condition
      */
     public function __construct($field, $operator, $value, $boolean = 'and')
     {
+        /**
+         * Decode (maybe) the operator in case the operator is stored as an encoded setting value.
+         * @link https://github.com/impress-org/give-form-field-manager/issues/376
+         * @unreleased
+         */
+        $operator = html_entity_decode( $operator );
+
         if ($this->invalidOperator($operator)) {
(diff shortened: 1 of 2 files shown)


JasonTheAdams (changes requested):
@kjohnson Right solution but wrong place. Yes, we should be encoding it when placing it in the HTML attribute, but no it should not be encoded in the Field API. It only needs to be encoded at the exact point that we're applying it to the HTML.

kjohnson:
@JasonTheAdams I would agree, but there is also the case that these values are already stored encoded in the database, so this is a defensive positioning for that reason.

JasonTheAdams:
If we need to encode/decode in multiple places, that's fine. But I don't want the experience within the Field API to be comparing against HTML entities. That's a completely different context.

kjohnson:
Closing in favor of https://github.com/impress-org/give-form-field-manager/pull/377.

PR #6156, opened 2021-12-23
Payment Commands now extend base class for shared functionality

kjohnson:
## Description

This PR adds a new abstract base class for payment handlers, which handles the shared functionality of updating a payment status, setting a transaction ID, and optionally adding payment notes.

The concrete payment handlers need only to specify the related payment status value:

```php
class PaymentCompleteHandler extends PaymentHandler {
    protected function getPaymentStatus() { return 'complete'; }
}
```

```php
abstract class PaymentHandler
{
    abstract protected function getPaymentStatus();

    public function __invoke(PaymentCommand $paymentCommand, $donationId)
    {
        give_update_payment_status($donationId, $this->getPaymentStatus());

        if( $paymentCommand->gatewayTransactionId ) {
            give_set_payment_transaction_id($donationId, $paymentCommand->gatewayTransactionId);
        }

        foreach( $paymentCommand->paymentNotes as $paymentNote ) {
            give_insert_payment_note( $donationId, $paymentNote );
        }
    }
}
```

Likewise, an abstract payment command class adds the shared functionality to include payment notes

```php
abstract class PaymentCommand implements GatewayCommand {
    public function setPaymentNotes( ...$paymentNotes ) {
        $this->paymentNotes = $paymentNotes;
        return $this;
    }
}
```
The fluent `setPaymentNotes()` method can be used in the webhook handler like so:

```php
return PaymentProcessing::make($this->data->id)->setPaymentNotes(
    sprintf(esc_html__('BitPay invoice ID: %s', 'give-bitpay'), $this->data->id)
);
```

## Pre-review Checklist

### DIFF
--- a/src/Framework/PaymentGateways/CommandHandlers/PaymentHandler.php
+++ b/src/Framework/PaymentGateways/CommandHandlers/PaymentHandler.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Give\Framework\PaymentGateways\CommandHandlers;
+
+use Give\Framework\PaymentGateways\Commands\PaymentCommand;
+
+abstract class PaymentHandler
+{
+    /**
+     * @unreleased
+     * @return string
+     */
+    abstract protected function getPaymentStatus();
+
+    /**
+     * @unreleased
+     *
+     * @param  PaymentCommand  $paymentCommand
+     * @param  int  $donationId
+     * @return void
+     */
+    public function __invoke(PaymentCommand $paymentCommand, $donationId)
+    {
+        give_update_payment_status($donationId, $this->getPaymentStatus());
+
+        if( $paymentCommand->gatewayTransactionId ) {
+            give_set_payment_transaction_id($donationId, $paymentCommand->gatewayTransactionId);
+        }
+
+        foreach( $paymentCommand->paymentNotes as $paymentNote ) {
+            give_insert_payment_note( $donationId, $paymentNote );
+        }
+    }
+}
--- a/src/Framework/PaymentGateways/CommandHandlers/PaymentHandler.php
+++ b/src/Framework/PaymentGateways/CommandHandlers/PaymentHandler.php
@@ -7,27 +7,48 @@ use Give\Framework\PaymentGateways\Commands\PaymentCommand;
 abstract class PaymentHandler
 {
     /**
+     * @var PaymentCommand
+     */
+    protected $paymentCommand;
+
+    /**
      * @unreleased
      * @return string
      */
     abstract protected function getPaymentStatus();
 
     /**
+     * @param PaymentCommand $paymentCommand
+     */
+    public function __construct(PaymentCommand $paymentCommand)
+    {
+        $this->paymentCommand = $paymentCommand;
+    }
+
+    /**
+     * @param PaymentCommand $paymentCommand
+     * @return static
+     */
+    public static function make(PaymentCommand $paymentCommand)
+    {
+        return new static($paymentCommand);
+    }
+
+    /**
      * @unreleased
      *
-     * @param  PaymentCommand  $paymentCommand
      * @param  int  $donationId
      * @return void
      */
-    public function __invoke(PaymentCommand $paymentCommand, $donationId)
+    public function handle($donationId)
     {
         give_update_payment_status($donationId, $this->getPaymentStatus());
 
-        if( $paymentCommand->gatewayTransactionId ) {
-            give_set_payment_transaction_id($donationId, $paymentCommand->gatewayTransactionId);
+        if( $this->paymentCommand->gatewayTransactionId ) {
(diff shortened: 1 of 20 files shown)


jonwaldstein:
@kjohnson I like this direction, since each payment command will be so similar with different statuses - this could clean up the redundancy.  I do have a couple concerns that i'll point out here.

src/Framework/PaymentGateways/Commands/PaymentComplete.php
```diff
@@ -5,21 +5,4 @@
...
-     *
-     * @param  string  $gatewayTransactionId
-     */
-    public function __construct($gatewayTransactionId)
-    {
-        $this->gatewayTransactionId = $gatewayTransactionId;
-    }
-}
\ No newline at end of file
+class PaymentComplete extends PaymentCommand {}
```
jonwaldstein, line 8: We'll need to make it clear that `$gatewayTransactionId` is _not_ optional for this command.  So let's overload the parent construct here to make it required.

src/Framework/PaymentGateways/CommandHandlers/PaymentHandler.php
```diff
@@ -0,0 +1,34 @@
...
+    abstract protected function getPaymentStatus();
+
+    /**
+     * @unreleased
+     *
+     * @param  PaymentCommand  $paymentCommand
+     * @param  int  $donationId
+     * @return void
+     */
+    public function __invoke(PaymentCommand $paymentCommand, $donationId)
```
jonwaldstein, line 22: One thing that makes this abstraction different, is we are generalizing the `$paymentCommand` in our handler instead of being specific about the command the handler consumes.  I realize that might be a trade-off for the benefits of this abstraction though so open for discussion.

kjohnson: What if me moved the `$paymentCommand` to an overloaded constructor?

src/Framework/PaymentGateways/Commands/PaymentComplete.php
kjohnson: The transaction ID might be optional, though. In the case of BitPay we actually have to set the transaction ID up front, since we are dealing with invoices, which are then paid.

kjohnson: Like the `withNotes()` method, we could add a `withTransactionID()` method so that setting the transaction ID is declarative.

src/Framework/PaymentGateways/CommandHandlers/PaymentHandler.php
jonwaldstein: Yeah that will work.  if we're going to enforce a constructor then we should probably turn `__invoke` into `handle`

src/Framework/PaymentGateways/Commands/PaymentComplete.php
jonwaldstein: okay let's just keep it optional for now.

src/Framework/PaymentGateways/CommandHandlers/PaymentHandler.php
kjohnson: That makes sense.

kjohnson: Resolved by bee2ae01a.

src/Framework/PaymentGateways/Commands/PaymentComplete.php
kjohnson: See 11a2835.

jonwaldstein (approved):
This cleans up the redundant payment handlers, Nice job 👍

src/Framework/PaymentGateways/Commands/PaymentCommand.php
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * @param  string|string[]  ...$paymentNotes
+     * @return $this
+     */
+    public function withNote(...$paymentNotes)
+    {
+        $this->paymentNotes = $paymentNotes;
+        return $this;
+    }
```
JasonTheAdams, line 51: @kjohnson This can simply be `notes`, same as below. The `withX` convention, in my experience, is intended for immutable object changes. That is, `withNote` would clone the current object, apply the notes to the new object, and then return the new object.

JasonTheAdams (approved):
Nice! I just had one minor suggestion. 👍

src/Framework/PaymentGateways/Commands/PaymentCommand.php
kjohnson: Yeah, I hadn't thought about that 👍🏻 Resolved by d4ea6da89.

PR #6162, opened 2021-12-29
Gateway API Route Methods now use the Response API

kjohnson:
## Description

This PR replaces the use of commands in the Route Method callback, in favor of responses. This allows the Route Method callbacks to act as controllers, either for client/server communication or webhook handling. Either way, the controller is expected to handle itself and return a response type.

Each Route Method callback is now passed the `$donationId`, since each callback is now responsible for updating the payment. This is as simple as using the existing API methods, such as `give_update_payment_status()`, etc.

## Testing Instructions

For a given gateway, add a new `$routeMethod`, with a matching method definition that returns a response (ie JsonReponse).

Below is an example route method URL for BitPay:

```
https://wordpress.test/?give-listener=give-gateway&give-gateway-id=bitpay&give-gateway-method=handleFoo&give-donation-id=275
```

The corresponding route method:
```php
public function handleFoo( $donationId )
{
    give_update_payment_status( $donationId, 'complete' );

    return response()->json([
        'payment_id' => $donationId,
    ]);
}
```

... and the resulting response in the browser

```json
{"data":{"payment_id":275}}
```

## Pre-review Checklist


jonwaldstein (changes requested):
@kjohnson this is looking great.  Couple things we'll need to do before merging:

1. Let's update all the instances of `OffsiteGateway` and pass in the `$donation` param into `returnFromOffsiteRedirect`.
2. Then let's go ahead and keep the `TestGateway` updated by implementing the new way of returning from `returnFromOffsiteRedirect` in `TestGatewayOffsite`.

P.S. i've been using the `TestGateway` to test the gateway api and also keep it as an example for implementing new gateways.

ravinderk (changes requested):
@kjohnson I think we should not use the `Give_Payment` class in gateway API. We can use it internally in any function. I do not recommend it because this class does not scale well and causes a lot of queries and also it is a legacy class.

kjohnson:
Here is an updated example implementation for `TestGatewayOffsite`.

```php
/**
 * An example gateway method for extending the Gateway API for a given gateway.
 *
 * @param int $donationId
 * @return RedirectResponse|JsonResponse
 */
public function testGatewayMethod( $donationId )
{
    give_insert_payment_note( $donationId, 'NOTE GOES HERE' );
    give_update_payment_status( $donationId, 'complete' );
    give_set_payment_transaction_id( $donationId, "test-gateway-transaction-id" );

    return response()->json();
}
```

jonwaldstein (approved):
@kjohnson I made some updates to the offsite interface and related to match the handling of the gateway route methods.  Everything else looks good to me 👍

PR #6167, opened 2022-01-07
New: Migrate legacy give-recurring subscriptions into core

### LINKED ISSUE #6166: Subscriptions should be a core concept
## Details

Right now the concept of Subscriptions (including renewals) is only found in GiveWP Recurring.  This is problematic because subscriptions is actually a core concept of donations as a whole.  Having core models like this living outside our main codebase makes it difficult to manage and harder to integrate with.

## Additional Context

This is a big job and will be an iterative process. The goal is to start small and just move files over to core, then eventually we can optimize the functionality for the future.

## Acceptance Criteria

-   [x] Subscriptions and related functionality lives in give core

jonwaldstein:
Resolves #6166
🛑  Depends on [Give Recurring PR](https://github.com/impress-org/give-recurring/pull/1111)

## Description

This is the first step into identifying _Subscriptions_ as a core concept. We are relocating all the key subscription classes into give core.   All that should be left in give-recurring is logic that uses these concepts and/or extends give core - not WordPress.

## Affects

- A new `LegacySubscriptions` domain & service provider with the legacy files for Subscriptions and related database functionality.
- added `@property-read Give_Recurring_DB_Subscription_Meta $subscription_meta` to main Give class

The classes below have been relocated to Give Core from give-recurring
- `Give_Recurring_Cache`
- `Give_Recurring_Cron`
- `Give_Recurring_DB_Subscription_Meta`
- `Give_Recurring_Subscriber`
- `Give_Subscription`
- `Give_Subscriptions_API`
- `Give_Subscriptions_DB`

## Testing Instructions

- Make sure you're recurring branch is aligned with this PR https://github.com/impress-org/give-recurring/pull/1111
- Create a donation form with recurring
- Donate with recurring
- Uninstall recurring and donate with the same form

## Pre-review Checklist


jonwaldstein:
@JasonTheAdams this is ready for review

JasonTheAdams:
@jonwaldstein What happens if someone uses this while being on an old version of Recurring? This is a fairly likely scenario as folks will probably update GiveWP core first, then update Recurring. Does it throw a fatal error with classes being re-defined? If so, we may need to add a check to the Service Provider to not load the classes if they're already loaded by Recurring.

jonwaldstein:
@JasonTheAdams yup, I found this was the safest way to check 8f2b9fb0 since those legacy files are just loaded with autoload what seems like _after_ give core.

JasonTheAdams:
That works! Let's keep this in mind as we move more things to core; we have to be mindful of the update flow and making sure it doesn't brick sites, preventing the admin from even updating the plugins. It's _ok_ if the site mildly breaks, wherein the admin can and should update the corresponding add-on.

src/LegacySubscriptions/ServiceProvider.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        $recurringIsInstalled = defined('GIVE_RECURRING_VERSION') && GIVE_RECURRING_VERSION;
+        $recurringMeetsRequirements = $recurringIsInstalled && GIVE_RECURRING_VERSION > '1.14.1';
+
+        if ($recurringMeetsRequirements || !$recurringIsInstalled) {
+            $this->includeLegacyFiles();
+            $this->bindClasses();
+        }
+
+        $this->includeLegacyHelpers();
```
JasonTheAdams, line 32: @jonwaldstein Why isn't this included in the above condition?

src/LegacySubscriptions/ServiceProvider.php
```diff
@@ -0,0 +1,105 @@
...
+ */
+class ServiceProvider implements ServiceProviderInterface
+{
+    /**
+     * @inheritDoc
+     */
+    public function register()
+    {
+        $recurringIsInstalled = defined('GIVE_RECURRING_VERSION') && GIVE_RECURRING_VERSION;
+        $recurringMeetsRequirements = $recurringIsInstalled && GIVE_RECURRING_VERSION > '1.14.1';
```
JasonTheAdams, line 25: I recommend using [version_compare](https://www.php.net/manual/en/function.version-compare.php)

JasonTheAdams (changes requested):
I was able to make a recurring donation with Recurring on the newest version using the test gateway, but I noticed that the Stripe Credit Card gateway is not loading and has a console error.

src/LegacySubscriptions/ServiceProvider.php
jonwaldstein: 👍 770cd0f1

src/LegacySubscriptions/ServiceProvider.php
jonwaldstein: So this file basically takes `give_recurring()->is_recurring`, `give_recurring()->is_donation_recurring`, and `give_recurring()->get_period` and pulls them in as functions that don't rely on `give_recurring()`.  This way for the time being, we don't need to use the main give_recurring class to figure out if a donation is recurring.

jonwaldstein:
@JasonTheAdams interesting, i'll take a look into the Stripe gateway to see what's going on.

jonwaldstein:
@JasonTheAdams regarding Stripe, I was not able to reproduce an error. I've tried with and without recurring installed as well as develop branch vs draft branch - all successful.  Can you tell me more about your environment/config so I can try to reproduce?

JasonTheAdams (approved):
Figured out the Stripe thing (it was from testing another PR). This is good to go!

JasonTheAdams:
Actually, bump the version in this to 2.18.0, or else folks will have problems when updating Recurring `develop`

PR #6176, opened 2022-01-20
New: Introducing donation, donor, and subscription models 🎉

### LINKED ISSUE #6246: GiveWP core should have models
## Details

Currently, core concepts like donations, donors, and subscriptions are all being treated like WordPress items.  Typical scenarios are hooks and filters for everything, inconsistency around data persistence, and _hard to remember_ data structures.  This was fine to start as GiveWP _is_ a WordPress plugin - but as the codebase expands, and the development team continues to grow, we need to structure our key architectural concepts in a way that improves the developer experience and allows us to build meaningful features more efficiently.

Enter the world of Models...

Models on a high level, serve as a data access layer where data is fetched and returned in formats usable throughout your application.  Modern frameworks like Laravel take this concept to the next level by infusing models with patterns like [Object Relational Mapper](http://en.wikipedia.org/wiki/Object_relational_mapper) and [Active Record Pattern](http://en.wikipedia.org/wiki/Active_record_pattern).  This makes interacting with your data easier and more consistent by giving you specific access points for all your application needs.

For the case of GiveWP, our initial models will be for donations, donors, and subscriptions.  Each model will abstract the database layer through repositories and allow you to just think about common business logic like retrieving, creating, updating, and deleting.  They will also have clear paths for relationships so you won't have to think about _how_ to find the related record through the database.  In fact, one of the main goals is _not_ having to think about the database structure.  This will allow us to focus our time and energy on developing the product instead of trying to figure out how to use our own data concepts.

## Additional Context

Inspiration for this direction comes from much experience with Laravel's Eloquent ORM https://laravel.com/docs/9.x/eloquent

## Acceptance Criteria

-   [x] GiveWP has a concept of models
-   [x] There is a model for Donations, Donors, and Subscriptions

jonwaldstein:
Resolves #6246

### Models

Models are the future of GiveWP.  They will give us a consistent way of representing our data concepts, along with specific ways of interacting with them.  No more trying to figure out how to retrieve a donation and update a value.  The models will abstract all that away so the developer can focus on actually building things.

This PR introduces 3 fully functional models: `Donor`, `Donation`, and `Subscription`.  They will all come with a defined data structure and CRUD functionality.  All of the underlying logic for their database transactions are located inside corresponding repositories.

### Repositories

Each new model has a corresponding repository in their domain - but can be referenced from the give container.  
```
give()->donorRepository->getById();
give()->donations->getById();
give()->subscriptions->getById();
```

This is where _all_ the heavy lifting of data abstraction & database queries/transactions are located.  This is intentional so models are always referencing a method inside a repository that will always return the same thing - no matter the change in data structure over time (think a new donation table vs posts table).

#### Model Basics

Each model will have a `$properties` array which identifies each model property key & type.  You can think of this array as a constructor of object properties, because you will be retrieving them the same way you would if they _were_ properties.
```
class MockModel extends Model {
    protected $properties = [
        'id' => 'int',
        'firstName' => 'string',
        'lastName' => 'string'
    ];
}
```

When creating an instance of the model you will pass in an array of attributes.  Each of those values passed will validate their type identified from the `$properties` array and throw an exception if not assigned correctly.  This forces us to assign our properties the correct value type for consistency.

```
$mockModel = new MockModel(['firstName' => 'Bill', 'lastName' => 'Murray']);

$mockModel->firstName; // 'Bill'
$mockModel->lastName; // 'Murray'
```

**Model Dirty State / Changes**

We can also check the internal state of the model properties.

```
$mockModel = new MockModel(['firstName' => 'Bill', 'lastName' => 'Murray']);

$mockModel->firstName = "William"

$mockModel->getDirty() // ['firstName' => 'William']
$mockModel->isDirty('firstName') // true
```

#### Model CRUD functionality

At the time of writing our CRUD enabled Models are `Donation`, `Donor`, and `Subscription`.  Each of these implements the `ModelCrud` contract and has full CRUD functionality.  All of the underlying logic for these database transactions are located inside corresponding repositories.  Let's see what they can do:

Find a Donation using its ID.

```
```

Now you have the model representation for that Donation.  Here's what all those values look like:

```
$donation->toArray() //
[
     "amount" => 50,
     "currency" => "USD",
     "donorId" => 1,
     "firstName" => "Bill",
     "lastName" => "Murray",
     "email" => "billmurray@givewp.com",
     "id" => 370,
     "status" => Give\Donations\ValueObjects\DonationStatus {#2777},
     "parentId" => 0,
     "subscriptionId" => 16,
     "updatedAt" => DateTime @1645820358 {#2779
       date: 2022-02-25 15:19:18.0 America/New_York (-05:00),
     },
     "createdAt" => DateTime @1669784460 {#2778
       date: 2022-11-30 00:01:00.0 America/New_York (-05:00),
     },
     "gateway" => "test-gateway",
     "mode" => Give\Donations\ValueObjects\DonationMode {#2780},
     "formId" => 1,
     "billingAddress" => Give\Donations\Properties\BillingAddress {#2781
       +country: null,
       +address1: null,
       +address2: null,
       +city: null,
       +state: null,
       +zip: null,
     },
     "formTitle" => "Donation Form",
     "purchaseKey" => "55b47b12938981a257372f05c71235a6",
     "donorIp" => "127.0.0.1",
     "anonymous" => false,
     "levelId" => 0,
     "gatewayTransactionId" => "",
   ];
```

Let's update a property value and save it.

```
$donation->gatewayTransactionId = "gateway-transaction-id"
$donation->save();
```

Let's get the donor associated to the donation

```
$donation->donor()->get();
```

Let's get the subscription associated to the donation

```
$donation->subscription()->get();
```

How about creating a new donation?

```
Donation::create([
    'status' => DonationStatus::PENDING(),
    'gateway' => TestGateway::id(),
    'mode' => DonationMode::TEST(),
    'amount' => 50,
    'currency' => 'USD',
    'donorId' => 1,
    'firstName' => 'Bill',
    'lastName' => 'Murray',
    'email' => 'billmurray@givewp.com',
    'formId' => 1,
    'formTitle' => 'Form Title',
]);
```

Don't worry, each create method will validate all the required information it needs to be successful.  So an exception will be thrown if you forgot to include something required like amount. On top of that, a Donation will make sure a real donor exists with the donorId you provide before proceeding.

Here's the list of CRUD commands that come with each of our current models:

```
Model::find($id); // returns the model instance from the database
Model::create($attributes); // inserts the model into the database
Model::query() // returns the query builder representation of the model
$model->save(); // saves the current state of the model
$model->delete(); // deletes the model & all it's information (including meta under the hood)
```

#### Model Factories
Another benefit of models is the ease of unit testing.  Being able to spin up an environment with specific data is powerful in the world of testing.  To make this even easier model factories are included.  This means you can do this in your tests:
```
$donor = Donor::factory()->create();
$donation = Donation::factory()->create(['donorId' => $donor->id]);

$this-assertEquals($donation->donor()->get()->id, $donor->id);
```

Under the hood, those static factory methods are calling a Model Factory instance with defined attributes that can be overwritten.

## Affects

- New Donations, Donors, and Subscriptions domains with matching models and repositories.
- Backwards compatibility with the following actions:
```
give_insert_payment
give_manual_insert_payment
give_update_payment_status
give_subscription_post_create
give_recurring_add_subscription_payment
give_recurring_record_payment
```

## Testing Instructions

There are unit tests in place to test.  Feel free to open up Tinkerwell and play around with the models.

## Pre-review Checklist


src/Donations/Repositories/DonationRepository.php
```diff
@@ -0,0 +1,519 @@
...
+            foreach ($this->getCoreDonationMetaForDatabase($donation) as $metaKey => $metaValue) {
+                DB::table('give_donationmeta')
+                    ->where('donation_id', $donation->id)
+                    ->where('meta_key', $metaKey)
+                    ->delete();
+            }
+        } catch (Exception $exception) {
+            DB::query('ROLLBACK');
+
+            Log::error('Failed deleting a donation');
```
JasonTheAdams, line 254: Seems like we could add the donation and exception for context, here? This doesn't seem very helpful as is. See other error logs in this repo as well, please.

src/Donations/Repositories/DonationRepository.php
```diff
@@ -0,0 +1,519 @@
...
+    }
+
+    /**
+     * @unreleased
+     *
+     * @param  Donation  $donation
+     *
+     * @return array
+     */
+    public function getCoreDonationMetaForDatabase(Donation $donation)
```
JasonTheAdams, line 273: Looks like this can be protected or private.

src/Donors/Models/Donor.php
```diff
@@ -0,0 +1,227 @@
...
+        return give()->donorRepository->getByEmail($email);
+    }
+
+    /**
+     * @unreleased
+     *
+     * @param  string  $email
+     * @return bool
+     */
+    public function hasEmail($email)
```
JasonTheAdams, line 76: This can be static since it's not using `$this`. I believe the point is to see if an existing donor is already using an email. Not sure if this belongs in the model, anyway. I think `give()->donors->isUsingEmail($email)` is more readable.

src/Donors/Models/Donor.php
```diff
@@ -0,0 +1,227 @@
...
+     * @unreleased
+     *
+     * @param  string  $email
+     * @return bool
+     * @throws Exception
+     */
+    public function addAdditionalEmail($email)
+    {
+        return give()->donorRepository->insertAdditionalEmail($this->id, $email);
+    }
```
JasonTheAdams, line 104: This is breaking the model flow. This immediately persists, whereas models are supposed to enter hold persistence until `save()` is called.

src/Donors/Models/Donor.php
```diff
@@ -0,0 +1,227 @@
...
+
+    /**
+     * @unreleased
+     *
+     * @return int
+     */
+    public function totalDonations()
+    {
+        return (int)give()->donations->getTotalDonationCountByDonorId($this->id);
```
JasonTheAdams, line 185: Doesn't seem like this needs to be cast again?

JasonTheAdams (changes requested):
@jonwaldstein I didn't quite finish this (25/61 files), but here's what I've seen so far.

JasonTheAdams:
@jonwaldstein Grr, stupid Slack notifications don't work when a PR is in draft mode. Please note that I left a review. 👆

src/Donations/LegacyListeners/DispatchGiveInsertPayment.php
```diff
@@ -0,0 +1,47 @@
...
+                'lastName' => $donor->lastName,
+                'title' => $donor->prefix,
+                'email' => $donor->email
+            ],
+        ]);
+
+         /**
+         * @deprecated
+         */
+        Hooks::dispatch('give_insert_payment', $donation->id, $giveInsertPaymentData->toArray());
```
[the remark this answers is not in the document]
jonwaldstein: I'm okay with that, the main reason I made this was to log all the do_actions after they are called for sanity sakes.  I don't think that's necessary for production but I could see it being helpful during development in the future. So maybe we can wrap that logger to check for `WP_DEBUG_LOG`. What do you think?

src/Donations/ServiceProvider.php
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            /**
+             * @notice
+             * Anytime we call give_update_payment_status the donor purchase_value and purchase_count get affected.
+             * We are doing this in the gateway api and in many other places.
+             * The listener below matches the functionality but the count seems to be overwritten elsewhere.
+             * Leaving this commented out until resolved or needed.
+             */
+            //Call::invoke(UpdateDonorPurchaseValueAndCount::class, $donation);
```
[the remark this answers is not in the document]
jonwaldstein: Nope, let's leave this uncommented for now.

The problem is all the gateways are using `give_update_payment_status` which triggers these side effects.  So I think resolving this will need to happen in the gateway api / implementation phase but I want to make sure we don't forget about it.

src/Donations/LegacyListeners/UpdateDonorPaymentIds.php
```diff
@@ -0,0 +1,25 @@
...
+     * @return void
+     * @throws Exception
+     */
+    public function __invoke(Donation $donation)
+    {
+        $ids = give()->donations->getAllDonationIdsByDonorId($donation->donorId);
+
+        $paymentIds = implode( ',', array_unique( array_values(  $ids ) ) );
+
+        give()->donorRepository->updateLegacyColumns($donation->donorId, ['payment_ids' => $paymentIds]);
```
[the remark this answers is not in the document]
jonwaldstein: hmm i'm not sure about that.

src/Donations/Models/Donation.php
```diff
@@ -0,0 +1,218 @@
...
+use Give\Donations\Factories\DonationFactory;
+use Give\Donations\Properties\BillingAddress;
+use Give\Donations\ValueObjects\DonationMode;
+use Give\Donations\ValueObjects\DonationStatus;
+use Give\Framework\Exceptions\Primitives\InvalidArgumentException;
+use Give\Framework\Models\Contracts\ModelCrud;
+use Give\Framework\Models\Contracts\ModelHasFactory;
+use Give\Framework\Models\Model;
+use Give\Framework\QueryBuilder\QueryBuilder;
+use Give\ValueObjects\Money;
```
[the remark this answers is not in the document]
jonwaldstein: Agreed, I would rather use that

src/Framework/Models/Model.php
```diff
@@ -0,0 +1,299 @@
...
+     * Set a given attribute on the model.
+     *
+     * @unreleased
+     *
+     * @param  string  $key
+     * @param  mixed  $value
+     */
+    public function setAttribute($key, $value)
+    {
+        $this->attributes[$key] = $value;
```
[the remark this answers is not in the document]
jonwaldstein: validation is done within `__set` before this is called.

src/Framework/Models/Model.php
```diff
@@ -0,0 +1,299 @@
...
+     * @throws RuntimeException
+     */
+    public function getAttribute($key)
+    {
+        if (!$key) {
+            return null;
+        }
+
+        return array_key_exists($key, $this->attributes) ? $this->attributes[$key] : null;
+    }
```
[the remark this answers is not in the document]
jonwaldstein: I had that setup initially but was running into a conflict with the repository validators.  I'll revisit this and see if I can fix it.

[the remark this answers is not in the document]
jonwaldstein: Resolved c3f9ecb7

src/Framework/Models/Model.php
```diff
@@ -0,0 +1,299 @@
...
+     * Determine if an attribute exists on the model.
+     *
+     * @unreleased
+     *
+     * @param  string  $key
+     * @return bool
+     */
+    public function __isset($key)
+    {
+        return !is_null($this->getAttribute($key));
```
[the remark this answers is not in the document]
jonwaldstein: Resolved c3f9ecb7

src/Framework/Models/Model.php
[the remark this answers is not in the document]
jonwaldstein: I just moved the validator into this function so it's all together.

src/Donations/Repositories/DonationRepository.php
jonwaldstein: resolved dfe29ac4

src/Donations/Repositories/DonationRepository.php
jonwaldstein: resolved 17b71194

src/Donors/Models/Donor.php
jonwaldstein: I moved the cast into the repository 297d04c5

JasonTheAdams: We should update it within the QB, then. Wouldn't it make sense that count always returns an integer?

src/Donors/Models/Donor.php
jonwaldstein: Yeah that's obviously not right, it should be referencing an ID.  Removing for now since it's only used during implementation 55a634b3.  Once we have a proper `additionalEmails` property this method won't be be necessary.

src/Donations/Models/Donation.php
```diff
@@ -0,0 +1,218 @@
...
+    {
+        return give()->donations->delete($this);
+    }
+
+    /**
+     * @unreleased
+     *
+     * @return QueryBuilder
+     */
+    public function donor()
```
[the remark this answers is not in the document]
jonwaldstein: I thought we settled on doing that in an additional PR?  I can do it here though if we want.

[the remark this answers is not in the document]
JasonTheAdams: Ah, yeah, another PR is fine. I just know people will start using this quickly. So we'll just need to hop on that.

src/Donations/LegacyListeners/UpdateDonorPaymentIds.php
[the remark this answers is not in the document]
JasonTheAdams: Just having `give()->donorsRepository` breaks the naming convention, and since open to 3rd party use, I'd like to get it right.

src/Donations/LegacyListeners/DispatchGiveInsertPayment.php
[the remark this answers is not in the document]
JasonTheAdams: So long as we're using `Log::info()` under the hood it's already only logging those with debug mode enabled. Honestly, it may be worth adding a `Log::debug()` method that checks for something like `GIVE_DEBUG`. What do you think?

src/Donations/Factories/DonationFactory.php
```diff
@@ -0,0 +1,33 @@
...
+            'gateway' => TestGateway::id(),
+            'mode' => DonationMode::TEST(),
+            'amount' => $this->faker->numberBetween(50, 50000),
+            'currency' => 'USD',
+            'donorId' => 1,
+            'firstName' => $this->faker->firstName,
+            'lastName' => $this->faker->lastName,
+            'email' => $this->faker->email,
+            'formId' => 1,
+            'formTitle' => 'Form Title',
```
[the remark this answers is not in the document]
JasonTheAdams: Huh. Ok. 🤷‍♂️

src/Donations/Repositories/DonationRepository.php
```diff
@@ -0,0 +1,519 @@
...
+                    ->insert([
+                        'donation_id' => $donationId,
+                        'meta_key' => $metaKey,
+                        'meta_value' => $metaValue,
+                    ]);
+            }
+        } catch (Exception $exception) {
+            DB::query('ROLLBACK');
+
+            Log::error('Failed creating a donation', compact('donation'));
```
JasonTheAdams, line 161: @jonwaldstein Is that right? I thought the second parameter was an associative array with expected keys (such as `'context'`)? I could be wrong.

src/Donations/LegacyListeners/DispatchGiveInsertPayment.php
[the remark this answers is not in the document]
JasonTheAdams: It's true. I just think there's some logging that's truly informative to everyone (such as a webhook being received), and then there's stuff like logging _every_ action that can easily flood the log.

src/Donations/Repositories/DonationRepository.php
jonwaldstein: Yes this is right, context is just any array key that you want to add for additional information. 
Screen Shot 2022-03-15 at 1 10 43 PM

src/Donations/LegacyListeners/UpdateDonorPaymentIds.php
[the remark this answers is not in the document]
jonwaldstein: How would you know if someone is trying to use the legacy give()->donors vs the new one though?  the former is currently used a lot.

src/Donations/LegacyListeners/DispatchGiveInsertPayment.php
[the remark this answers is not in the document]
jonwaldstein: Yeah I agree

src/Donations/Models/Donation.php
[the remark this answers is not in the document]
jonwaldstein: I think i'm only using our `Money` helper for getting the amount from the database and then within `$donation->getMinorAmount()` for use in legacy hooks.  I don't completely understand how our amount formatting works so I would want to discuss that before adding MoneyPHP.

src/Donations/LegacyListeners/DispatchGiveInsertPayment.php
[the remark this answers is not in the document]
jonwaldstein: Do you want me to want to this in this PR, or a seperate one?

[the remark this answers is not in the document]
JasonTheAdams: It's so tiny, just go ahead and do it here.

src/Donations/LegacyListeners/UpdateDonorPaymentIds.php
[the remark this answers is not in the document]
JasonTheAdams: I don't think we have any methods with the same name, presently. You could just add a `__call()` magic method which falls back to the legacy methods.

src/Donations/Models/Donation.php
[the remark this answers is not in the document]
JasonTheAdams: That's fair. Let's handle this in a separate PR.

src/Donors/Models/Donor.php
```diff
@@ -0,0 +1,215 @@
...
+ * @property string $firstName
+ * @property string $lastName
+ * @property string $email
+ * @property Subscription[] $subscriptions
+ * @property Donation[] $donations
+ */
+class Donor extends Model implements ModelCrud, ModelHasFactory
+{
+    /**
+     * @var string[]
```
JasonTheAdams, line 39: This should probably be `@inheritDoc`

src/Framework/Models/Model.php
```diff
@@ -0,0 +1,337 @@
...
+     *
+     * @unreleased
+     *
+     * @param  string  $attribute
+     * @return bool
+     */
+    public function isDirty($attribute)
+    {
+        return array_key_exists($attribute, $this->getDirty());
+    }
```
JasonTheAdams, line 99: Let's make `$attribute` optional; if it's null then it checks if _any_ value is dirty.

Also, let's add an inverted method: `isClean` — which can just call this one and reverse the boolean.

src/Framework/Models/Model.php
```diff
@@ -0,0 +1,337 @@
...
+     * @param $key
+     *
+     * @return Model|Model[]
+     *
+     * @throws InvalidArgumentException
+     */
+    protected function getRelationship($key)
+    {
+        if (!is_callable([$this, $key])) {
+            throw new InvalidArgumentException("$key() does not exist in.");
```
JasonTheAdams, line 322: "does not exist in..." — is this missing the class or something?

src/Framework/QueryBuilder/Concerns/CRUD.php
```diff
@@ -2,13 +2,24 @@
...
 /**
  * @unreleased
  */
 trait CRUD
 {
+    /**
+     * @var string
+     */
+    private $model;
+
```
JasonTheAdams, line 22: I'm not sure I'm a fan of coupling the QB to the models. Where is this Trait used?

Also, if we do go down this path, we can make use of generics: https://blog.jetbrains.com/phpstorm/2021/12/phpstorm-2021-3-release/#infer_generic_types_from_constructor

src/PaymentGateways/Actions/RegisterPaymentGateways.php
```diff
@@ -17,7 +18,7 @@ class RegisterPaymentGateways
      */
     public $gateways = [
         // When complete, the Test Gateway will eventually replace The legacy Manual Gateway.
-        //TestGateway::class,
+        TestGateway::class,
```
JasonTheAdams, line 21: Do we need to comment this again before merging?

src/Subscriptions/LegacyListeners/DispatchGiveSubscriptionPreCreate.php
```diff
@@ -0,0 +1,36 @@
...
+			'recurring_fee_amount' => $subscription->feeAmount,
+			'bill_times'           => $subscription->installments,
+			'parent_payment_id'    => 0,
+			'form_id'              => $subscription->donationFormId,
+			'created'              => null,
+			'expiration'           => $subscription->expiration(),
+			'status'               => $subscription->status->getValue(),
+			'profile_id'           => $subscription->gatewaySubscriptionId,
+		];
+
```
JasonTheAdams, line 33: Tabs?

JasonTheAdams (changes requested):
Alright! I made it through everything! A few more notes, but for 5000 lines of code, I'm surprised there's not more. I'm stoked to get this in core, @jonwaldstein!

src/Donations/LegacyListeners/UpdateDonorPaymentIds.php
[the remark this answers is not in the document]
jonwaldstein: `update()` and `insert()` exist in both places.  I guess I could just suffix the latter ones with the model name.

src/Subscriptions/LegacyListeners/DispatchGiveSubscriptionPreCreate.php
jonwaldstein: reformatted file, not sure what happened there e9d2a7aa

src/PaymentGateways/Actions/RegisterPaymentGateways.php
jonwaldstein: Yup good catch 9ebc74f4

src/Framework/QueryBuilder/Concerns/CRUD.php
jonwaldstein: This whole setModel part within the QB was inspired by Laravel, although not the same thing. Basically we needed a way to set the model on the query builder so we can return instances of the Model instead of generic objects that gets runs through each model's implementation of `fromQueryBuilderObject()`.  This cleaned up sooo much redundant code.  You can find these being set in each repository within `prepareQuery()`.

src/Framework/Models/Model.php
jonwaldstein: woops no just does not exist is good 73aa2c68

src/Framework/Models/Model.php
jonwaldstein: all set 88c4d73f

src/Donors/Models/Donor.php
jonwaldstein: resolved ffe9bdcb

src/Framework/QueryBuilder/Concerns/CRUD.php
alaca: Having QB coupled with models would be nice if we could query models by using model properties. The problem with supporting that is our current DB structure. Sure, it is possible to do such thing by adding a significant amount of complexity, but I'm not sure what would be the benefit of doing that. Having these nice and organized models it's easy to forget that most of our model properties are actually representations of **two** table rows in the meta table. The alternative is to move all the _extra_ things we need from our models into the repositories. 

I like what @jonwaldstein is doing here, but I would suggest doing that by using a slightly different approach - extend the QB class and override the `get` and `getAll` methods, or add more specific methods like `getModelName`, `getAllModelName`. QB is meant to be an all-purpose QB used in Give core, our add-ons, 3rd party add-ons, etc. so it might be a good idea to separate these things.

jonwaldstein: @alaca in the CRUD trait i created 2 methods `getAsModel()` and `getAllAsModel()` which does what you are explaining.  They are currently being used inside `get()` & `getAll()` when the model is set to the QB. I went back and forth with using those explicitly and decided the first example seemed redundant. I do see your point about model properties & the database structure.  I prefer how it works now because the fluent api stays within context of the query - but if we all agree it's better to just have separate explicit methods for returning the query as model I can be persuaded.

`Donation::query()->getAllAsModel()`
`$donation->subscriptions()->getAllAsModel()`

vs

`Donation::query()->getAll()`
`$donation->subscriptions()->getAll()`

alaca: What I suggested wouldn't affect the current API. I'm just suggesting moving the model logic outside the QB into a subclass.

jonwaldstein: Ahh I see what you're saying now, sorry I misunderstood.  Yes, that's another good option.

alaca: No worries. I probably have to finish one, or two more coffees before writing anything in the morning. Anyway, what you have here is fine, and it will help us to write less code. I'm just thinking how we could, maybe, have this better organized.

jonwaldstein: Okay, let me see what I can do.

jonwaldstein: All set b6f9247c

src/Donations/LegacyListeners/UpdateDonorPaymentIds.php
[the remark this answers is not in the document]
jonwaldstein: resolved with the DonorRepositoryProxy 00817495 9a57bd24

src/Framework/Models/ValueObjects/Relationship.php
```diff
@@ -0,0 +1,14 @@
...
+
+use Give\Framework\Support\ValueObjects\Enum;
+
+/**
+ * @method static ONE_TO_ONE();
+ * @method static ONE_TO_MANY();
+ */
+class Relationship extends Enum {
+    const ONE_TO_ONE = 'one-to-one';
+    const ONE_TO_MANY = 'one-to-many';
```
JasonTheAdams, line 13: Let's add:
- MANY_TO_MANY
- BELONGS_TO
- BELONGS_TO_MANY

Make sure to update the models as makes sense.

JasonTheAdams (changes requested):
Alright, @jonwaldstein! We're getting down to it. I have two more notes, and then I think we're done. 💪

src/Framework/Models/ValueObjects/Relationship.php
jonwaldstein: Okay updated 77ada265

Do we want to stick with `ONE_TO_ONE`, `ONE_TO_MANY` or `HAS_ONE`, `HAS_MANY`?

jonwaldstein: I personally think the `HAS` syntax is easier to comprehend so added 04d0efcb.  can use either or.

JasonTheAdams: Yeah, that's a good idea.

JasonTheAdams: Let's just remove the generic ones. They're not clear.

jonwaldstein: all set db24f3a2

JasonTheAdams (approved):
Chefs kiss

Issue #6175, opened 2022-01-18
“All Forms” admin UI should be refreshed


knowler:
The donation forms list page should be updated to use the new GiveWP Admin UI. The page should be a single page application that doesn’t need to refresh for new data or different views. We will be starting simple so some existing feature will not be implemented until further scopes related to other list pages.

## Acceptance Criteria

- [x] Visual design is updated to match new Admin styles
- [x] Admin API endpoint is created for donation forms
- [x] Data is paginated

LMNTL:
Resolved by #6241.

PR #6177, opened 2022-01-20
enhancement: style the pagination links and clean up their markup

LMNTL:
Part of #6175

## Description

This PR styles the pagination component for the All Forms admin page (with the current CSS reset solution) and cleans up some of its markup (since it no longer needs to conform to the standard WP table markup).

## Visuals

image

## Testing Instructions

1. Create at least one Donation Form
2. Install GiveWP from this PR
3. Navigate to Donations -> All Forms in the admin
4. Verify that the pagination component works as expected and is properly styled

## Pre-review Checklist

### DIFF
--- a/src/DonationForms/resources/components/Pagination.js
+++ b/src/DonationForms/resources/components/Pagination.js
@@ -20,88 +20,69 @@ const Pagination = ({currentPage, totalPages, disabled, setPage}) => {
     const previousPage = parseInt(currentPage) - 1;
 
     return (
-        <nav className={styles.container}>
-            {previousPage > 0 ? (
-                <>
-                    <a
-                        href="#"
-                        className={cx(styles.navDirection, styles.navElement)}
-                        onClick={(e) => {
-                            e.preventDefault();
-                            if (!disabled) {
-                                setPage(1);
-                            }
-                        }}
-                    >
-                        «
-                    </a>{' '}
-                    <a
-                        href="#"
-                        className={cx(styles.navDirection, styles.navElement)}
-                        onClick={(e) => {
-                            e.preventDefault();
-                            if (!disabled) {
-                                setPage(parseInt(currentPage) - 1);
-                            }
-                        }}
-                    >
-                        ‹
-                    </a>
-                </>
-            ) : (
-                <>
-                    <span className={cx(styles.navDirection, styles.navElement, styles.disabled)}>«</span>
-                    <span className={cx(styles.navDirection, styles.navElement, styles.disabled)}>‹</span>
-                </>
-            )}
-
-            <span id="table-paging" className="paging-input">
-                <label htmlFor="current-page-selector" className={styles.visuallyHidden}>Current Page</label>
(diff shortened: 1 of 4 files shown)


src/DonationForms/resources/components/Pagination.js
```diff
@@ -19,86 +20,89 @@ const Pagination = ({currentPage, totalPages, disabled, setPage}) => {
...
+                        className={cx(styles.navDirection, styles.navElement)}
+                        onClick={(e) => {
+                            e.preventDefault();
+                            if (!disabled) {
+                                setPage(parseInt(currentPage) - 1);
+                            }
+                        }}
+                    >
+                        ‹
+                    </a>
```
knowler, line 49: Let’s make these `<button type="button">` elements since we are preventing the browser default and we aren’t doing anything meaningful to the URL anyway. If we were setting a URL search param, I would keep these as `<a>` elements, but we can cross that bridge when/if we get to it.

src/DonationForms/resources/components/Pagination.js
```diff
@@ -19,86 +20,89 @@ const Pagination = ({currentPage, totalPages, disabled, setPage}) => {
...
-                                    }
-                                }}
-                            >
-                                ‹
-                            </a>
-                        </>
-                    ) : (
-                        <span className="tablenav-pages-navspan button disabled">‹</span>
-                    )}
+        <nav className={styles.container}>
```
knowler, line 23: Since there is already an element with `role=navigation` on the page (a part of the WordPress admin), we should add an `aria-label` attribute with a meaningful name for this.

```suggestion
        <nav aria-label={__('donation forms table', 'give')} className={styles.container}>
```

Screen readers announce “navigation” after the label, so this would read “donation forms table navigation.”

src/DonationForms/resources/components/Pagination.js
```diff
@@ -19,86 +20,89 @@ const Pagination = ({currentPage, totalPages, disabled, setPage}) => {
...
+                        className={cx(styles.navDirection, styles.navElement)}
+                        onClick={(e) => {
+                            e.preventDefault();
+                            if (!disabled) {
+                                setPage(totalPages);
+                            }
+                        }}
+                    >
+                        »
+                    </a>
```
knowler, line 97: Make these `<button>` elements too.

src/DonationForms/resources/components/Pagination.module.scss
```diff
@@ -0,0 +1,74 @@
+:global(#give-admin-donation-forms-root) {
```
knowler, line 1: I’m not going to look too much in the styles, because I will need to change a few things when I merge this into my branch. Looks good otherwise.

knowler (changes requested):
Nice work @LMNTL. I just have a few suggestions, otherwise, once those are addressed. Ping me and I’ll approve so you can merge.

knowler (approved):
:shipit:

PR #6183, opened 2022-01-25
New: Query Builder

alaca:
## Description

This PR adds P2P QueryBuilder to the Give core. Also, the QueryBuilder is improved significantly and many new features were added. 

## Test Instructions

Check the README.md file for more information and example usages. 

## Pre-review Checklist


jonwaldstein (changes requested):
@alaca this is looking & working great! Just a heads up, we're going to set this QueryBuilder PR to be merged into develop because it's an isolated deliverable.  As a result, we'll need to remove the model branch code from this PR so it's not depending on it.

kjohnson (commented):
@alaca I'm not seeing any unit tests included in this PR. Framework code (and non-legacy code) like this should be easily testable. With the Query Builder in Peer-to-peer I focused on unit testing the resulting SQL code.

src/Framework/Database/DB.php
```diff
@@ -99,6 +100,44 @@ public static function last_insert_id()
...
+
+                $builder->from($tableName, $alias);
+            } else {
+                $builder->from($table);
+            }
+        });
+
+        return $builder;
+    }
+
```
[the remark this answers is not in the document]
JasonTheAdams: This seems a bit over-complicated. Why not just have `$table, $alias = null` parameters like we do in `$builder->from`? Then the whole `array_walk` would become `$builder->from($table, $alias)`.

Or is this able to set multiple `FROM` tables? Is that a thing?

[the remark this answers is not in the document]
JasonTheAdams: I guess it's possible using `GROUP BY`. Still this is intended to just be a convenience method, so let's just keep it simple and assume it's setting a single table and returning the query builder. Folks can always add more `from` calls after getting it back.

src/Donations/Models/Donation.php
```diff
@@ -0,0 +1,138 @@
...
+     */
+    public $updatedAt;
+    /**
+     * @var string
+     */
+    public $status;
+    /**
+     * @var int
+     */
+    public $amount;
```
ravinderk, line 40: @alaca Is it the total amount?

src/Donations/Models/Donation.php
```diff
@@ -0,0 +1,138 @@
...
+     */
+    public $lastName;
+    /**
+     * @var string
+     */
+    public $email;
+    /**
+     * @var int
+     */
+    public $parentId;
```
ravinderk, line 68: @alaca what is this?

src/Donations/Repositories/DonationRepository.php
```diff
@@ -0,0 +1,256 @@
...
+                DB::insert($this->donationMetaTable, [
+                    'donation_id' => $donationId,
+                    'meta_key' => $metaKey,
+                    'meta_value' => $metaValue,
+                ], null);
+            }
+        } catch (Exception $exception) {
+            DB::query('ROLLBACK');
+
+            Log::error('Failed creating a donation');
```
ravinderk, line 186: @alaca can we store additional information about exception in the log?

src/Donations/Repositories/DonationRepository.php
```diff
@@ -0,0 +1,256 @@
...
+                    'meta_value' => $metaValue,
+                ], [
+                    'donation_id' => $donation->id,
+                    'meta_key' => $metaKey
+                ]);
+            }
+        } catch (Exception $exception) {
+            DB::query('ROLLBACK');
+
+            Log::error('Failed creating a donation');
```
ravinderk, line 227: @alaca  is the message incorrect, I think you want to say `updating` instead of creating?

src/Framework/Database/DB.php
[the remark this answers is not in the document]
alaca: Yes, multiple `FROM` is a thing, and you don't have to use `GROUP BY` with it. TBH I didn't see someone use this method in a while, but using multiple `FROM` is similar if not exact same thing as `JOIN`. I guess everyone is using `JOIN` nowadays - which makes more sense and queries are more legible.  

I agree, let's change the method signature to `DB:::table(tableName, alias);`

src/Framework/QueryBuilder/Where.php
```diff
@@ -0,0 +1,112 @@
...
+            return $comparisonOperator;
+        }
+
+        throw new InvalidArgumentException(
+            sprintf(
+                'Unsupported comparison operator %s. Please use one of the supported operators (%s)',
+                $comparisonOperator,
+                implode(',', $operators)
+            )
+        );
```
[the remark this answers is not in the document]
alaca: Good catch, @jonwaldstein. And you are right, I will update the code with the suggested change.

src/Subscriptions/Repositories/SubscriptionRepository.php
```diff
@@ -16,12 +16,12 @@ class SubscriptionRepository
...
-        $db = give(Give_Subscriptions_DB::class);
+        $query = DB::table(DB::prefix('give_subscriptions'))
+            ->select('*')
+            ->where('id', $subscriptionId)
+            ->get();
 
-        $subscriptionObject = $db->get($subscriptionId);
-
-        return SubscriptionObjectData::fromObject($subscriptionObject)->toSubscription();
+        return SubscriptionQueryData::fromObject($query)->toSubscription();
```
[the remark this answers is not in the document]
alaca: Looks great! Love it. One note, If you want to select all columns, calling `select('*')`[ is redundant.](https://github.com/impress-org/givewp/blob/ca8e7d7e510f8381dba40f07be865cf271a15d5d/src/Framework/QueryBuilder/Concerns/SelectStatement.php#L45)

In other words, this is all you need.

```php
DB::table(DB::prefix('give_subscriptions'))->where('id', $subscriptionId)->get();
```

alaca:
@kjohnson thanks for bringing that up. I will include the tests.

src/Donations/Models/Donation.php
alaca: Yes, @ravinderk. This is the total amount.

src/Donations/Models/Donation.php
alaca: Of course.

src/Donations/Repositories/DonationRepository.php
alaca: What do you have in mind?

src/Donations/Repositories/DonationRepository.php
alaca: Good catch!

alaca:
What do you guys think about prefixing the table name automatically when calling` DB::table(tableName)`?

I would like to avoid doing things like this
`DB::table(DB::prefix('give_subscriptions'))`

Because this looks much nicer
`DB::table('give_subscriptions')`

But if we go this route, all the other methods like `from`, `join`, and `attachMeta` should prefix the table names as well 🤔

alaca:
@JasonTheAdams good thinking. I like both ideas, and we should probably have something like `DB::raw` which can be used everywhere, in `select`, `where`, etc.

About prefixing, the only tables that are shared in a multisite environment are `users` and `usersmeta`, and those two should be in the list of known tables. We can include new tables created by multisite install into the list as well. The only thing that worries me is that I'm not sure if GiveWP and all the add-ons can work in a multisite environment? If yes, then we have nothing to worry about. 

Correct me if I'm wrong, but in my understanding, `$wpdb->prefix` will already have the correct prefix in a multisite environment. For example, `"{$wpdb->prefix}posts"` will return `wp_posts`, `wp_1_posts`, `wp_2_posts`, etc. depending on the current site? Also, there is a [wpdb::get_blog_prefix](https://developer.wordpress.org/reference/classes/wpdb/get_blog_prefix/) method which we can use if `$wpdb->prefix` is not what we are looking for. 

**Workflow**

1. If the table name is an instance of `RawSQL` - return the _raw_ table name
2. check if provided table name has a prefix,  and remove it if it does.
3. check if the table name is in the list of known tables - if yes, return the table name from the list 
4. prefix the table name using `$wpdb->prefix` or `wpdb::get_blog_prefix` and return the prefixed table name

JasonTheAdams:
@alaca Cool! And yes, `$wpdb` has two prefix properties:
- `$wpdb->base_prefix` — the exact prefix as defined in wp-config.php, does not include the multisite id prefix
- `$wpdb->prefix` — the prefix which will include the multisite prefix

GiveWP is, as far as I know, MS compatible. So we will want to make sure our QB does not break that.

The rest sounds good. I'm on the fence about detecting a pre-existing prefix —  as opposed to simply saying passing a prefixed table is "wrong".

alaca:
Closing this because it is completely messed up. 
QueryBuilder work is extracted and it is now in a separate PR(#6184 ). 
Donation Model work will be extracted and pushed to the `feature/donation-model` branch.

PR #6196, opened 2022-01-31
New: Update gateway api with secure route methods

jonwaldstein:
## Description

The gateway api has been updated to support `secureRouteMethods`, which works the same as `routeMethods` but will create and verify a signature upon url resolution.

The route methods have also become more flexible by replacing the `$donationId` with $queryParms`  which will contain  any additional $args passed to create the route url.

Also, to reduce confusion with our un-biased gateway api - this PR removes all mention of Offsite Gateways including the interface, abstract class, and trait.

## Affects

- Replace `$donationId` with `$queryParams` param in gateway's custom routeMethods.
- Remove OffsiteGateway interface, class, and trait
- Add new `generateSecureGatewayRouteUrl()` helper to PaymentGateway

## Visuals

Screen Shot 2022-01-31 at 6 06 41 PM

Screen Shot 2022-01-31 at 6 06 50 PM

## Testing Instructions

- Uncomment out TestGatewayOffsite and use to donate.
- Try passing in a dummy string as the nonce in `generateSecureGatewayRouteUrl` to make the validation fail

## Pre-review Checklist


jonwaldstein:
@JasonTheAdams this is ready for review.  I plan on writing tests for this just don't have the time right now.  Also i'm not exactly sure what the best way to throw a 403 from the gateway route is - if you wouldn't mind taking a look where I left off on that.

JasonTheAdams (commented):
@kjohnson Would removing the `OffsiteGatewayInterface` interfaces break BitPay?

src/Framework/PaymentGateways/PaymentGateway.php
```diff
@@ -128,13 +139,43 @@ public function createSubscription(GatewayPaymentData $paymentData, GatewaySubsc
...
+     * @param  string  $method
+     * @param  array  $queryParams
      *
+     * @return void
      * @since 2.18.0
      *
+     * @unreleased - replace $donationId with $queryParams array
+     *
+     */
+    public function handleGatewayRouteMethod($method, $queryParams)
```
JasonTheAdams, line 152: I'm conflicted about having this within the gateway itself. Why doesn't this live within the framework?

src/Framework/PaymentGateways/Routes/GatewayRoute.php
```diff
@@ -2,74 +2,90 @@
...
 class GatewayRoute
 {
     /**
-     * @since 2.18.0
+     * This is our entry point into the Gateway Routing system.
      *
      * @return void
      * @throws PaymentGatewayException
+     * @since 2.18.0
+     * @unreleased - validate secureRouteMethods
```
JasonTheAdams, line 23: Should be
- Description
- `@since`
- `@param`
- `@return`
- `@throws`

src/Framework/PaymentGateways/Routes/GatewayRoute.php
```diff
@@ -2,74 +2,90 @@
...
             }
 
-            $gateway->handleGatewayRouteMethod($data->donationId, $data->gatewayMethod);
+            // If method is in secureRouteMethods then, validate signature
+            if (in_array($data->gatewayMethod, $gateway->secureRouteMethods, true)) {
+                $this->validateSignature($data->routeSignature, $data);
+            }
+
+            // Navigate to our payment gateway api to handle calling the gateway's method
+            $gateway->handleGatewayRouteMethod($data->gatewayMethod, $data->queryParams);
```
JasonTheAdams, line 70: Yeah, why not just handle the logic here and pull it out of the gateway? This also secures protected and private gateway methods.

JasonTheAdams (changes requested):
Good work, @jonwaldstein! I left a few notes. My biggest concern with what we're removing is if it will break BitPay upon updating GiveWP core. We may need to leave some of the Interfaces/Traits and mark them as deprecated.

src/Framework/PaymentGateways/PaymentGateway.php
ravinderk: +1

src/Framework/PaymentGateways/Routes/GatewayRoute.php
ravinderk: @jonwaldstein This is a list of PHPDoc tags in PHPStorm.

Preferences 2022-02-01 at 8 16 45 AM

src/Framework/PaymentGateways/Routes/GatewayRoute.php
```diff
@@ -86,4 +102,22 @@ private function isValidListener()
...
+     * @param  string  $routeSignature
+     * @param  GatewayRouteData  $data
+     * @return void
+     */
+    private function validateSignature($routeSignature, GatewayRouteData $data)
+    {
+        $action = RouteSignature::make($data->gatewayId, $data->gatewayMethod, $data->queryParams);
+
+        if (!wp_verify_nonce($routeSignature, $action->toString())) {
+            wp_redirect(response('Forbidden', 403));
```
ravinderk, line 120: @jonwaldstein 

- I think this will cause an error because [`wp_redirect`](https://developer.wordpress.org/reference/functions/wp_redirect/) expects string param instead of `Response|ResponseFactory`.
- Always use `exit` after `wp_redirect`

src/Framework/PaymentGateways/PaymentGateway.php
jonwaldstein: Honestly, I think these handle methods used to pull things in from the gateway in the earlier iterations - but now they're much more agnostic.  I'll abstract these out of the gateway class.

src/Framework/PaymentGateways/Routes/GatewayRoute.php
jonwaldstein: My editor shuffled these around. Is this something that's in the coding standards repo?  Could be a setting that didn't make it after getting the new mac.

src/Framework/PaymentGateways/Routes/GatewayRoute.php
jonwaldstein: Figured it out, I needed to use wp_die for this

src/Framework/PaymentGateways/Routes/GatewayRoute.php
```diff
@@ -2,74 +2,94 @@
...
+
+            // get all registered gateway ids
             $gatewayIds = array_keys($paymentGateways);
 
+            // make sure required params are valid
             if (!$this->isValidRequest($gatewayIds)) {
                 throw new PaymentGatewayException('This route is not valid.');
             }
 
+            // create DTO from GET request
```
ravinderk, line 47: @jonwaldstein We are sanitizing query param data. We can use `give_clean($_GET)` to pass sanitized data to `GatewayRouteData::fromRequest`

src/Framework/PaymentGateways/PaymentGateway.php
```diff
@@ -214,63 +226,45 @@ public function handleGatewaySubscriptionCommand(
...
-        if ($type instanceof JsonResponse) {
-            wp_send_json(['data' => $type->getData()]);
-        }
+        $nonce = RouteSignature::make($this->getId(), $gatewayMethod, $args);
+
+        return Call::invoke(
+            GenerateGatewayRouteUrl::class,
+            $this->getId(),
+            $gatewayMethod,
+            array_merge($args, [
```
ravinderk, line 265: @jonwaldstein this will generate a PHP warning because default value of `$args` is `null` but function expect `array`

src/Framework/PaymentGateways/Routes/GatewayRoute.php
jonwaldstein: 👍  77038462

jonwaldstein:
@JasonTheAdams the BitPay PR for using the gateway api is not merged yet.  So, I can still remove the offsite stuff there.

src/Framework/PaymentGateways/Routes/GatewayRoute.php
```diff
@@ -2,74 +2,94 @@
...
             }
 
-            $gateway->handleGatewayRouteMethod($data->donationId, $data->gatewayMethod);
+            // If method is in secureRouteMethods then, validate signature
+            if (in_array($data->gatewayMethod, $gateway->secureRouteMethods, true)) {
+                $this->validateSignature($data->routeSignature, $data);
+            }
+
+            // Navigate to our payment gateway api to handle calling the gateway's method
+            $this->handleGatewayRouteMethod($gateway, $data->gatewayMethod, $data->queryParams);
```
ravinderk, line 74: @jonwaldstein this will last point of the script after that we do not want anything to run. Can will add `exit` after this?

JasonTheAdams:
> @JasonTheAdams the BitPay PR for using the gateway api is not merged yet. So, I can still remove the offsite stuff there.

@jonwaldstein Ah! I didn't realize that! Clear sailing, then! ⛵

src/Framework/PaymentGateways/Routes/GatewayRoute.php
JasonTheAdams: We need to generate an `.editorConfig` that contains everything. For now, you have to set those manually in PhpStorm.

src/Framework/PaymentGateways/Routes/GatewayRoute.php
jonwaldstein: @ravinderk Is that necessary? I would think the try...catch inside the function compensates for finishing the request.

src/Framework/PaymentGateways/Traits/ResponseHelpers.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Give\Framework\PaymentGateways\Traits;
+
+use Give\Framework\Http\Response\Types\JsonResponse;
+use Give\Framework\Http\Response\Types\RedirectResponse;
+
+trait ResponseHelpers {
```
JasonTheAdams, line 8: @jonwaldstein Let's change this to `HandleHttpResponses` — the "helpers" term is easily overused and not really descriptive. It hangs out playing buzzed darts with its cousin Utility. 😆

src/Framework/PaymentGateways/Routes/GatewayRoute.php
jonwaldstein: That would be convenient

src/Framework/PaymentGateways/Routes/GatewayRoute.php
ravinderk: @jonwaldstein Yes, I think so.
Process a dummy gateway route then you will see that home page HTML will load if the payment gateway does not use exit in their gateway method.
With my suggestion, we can prevent code execution after this point.

src/Framework/PaymentGateways/Traits/ResponseHelpers.php
jonwaldstein: Fair enough lol dd9ea2cd

src/Framework/PaymentGateways/Routes/GatewayRoute.php
jonwaldstein: @ravinderk would you mind sharing a screenshot? I'm not able to reproduce

ravinderk: image

jonwaldstein: @ravinderk exit added 3c53f012

jonwaldstein:
@JasonTheAdams @ravinderk comments resolved.  Ready for re-review.  cc @kjohnson

ravinderk (approved):

JasonTheAdams (approved):
Great work, @jonwaldstein! 👍

PR #6203, opened 2022-02-01
Hot Fix: Classic template displays payment fields when there is only one active gateway

### LINKED ISSUE #6202: Classic Template should display the gateway when there is only one active gateway
## Details

When only one gateway is active — either from Payment Gateway settings or Per Form Gateways add-on — the Classic Template does not properly display the payment fields. Instead, it's just an empty white area. Like the north pole.

## Expected Behavior

Payment gateway field should show up. Always. Forever. Amen.

## Steps to Reproduce

1. Go to the Gateway settings page
2. Deactivate all gateways but one
3. Create a new form that uses the Classic template
4. View the form
5. Stare in shock at the lack of payment fields

## Visuals

image

## Acceptance Criteria

-   [x] Payment fields display, without the header, description, and radio buttons, when there is one active gateway
-   [x] Payment gateways fields display and function as usual when multiple gateways or selected

JasonTheAdams:
Resolves #6202 

## Description

The Classic template, up to this point, moved the default payment fields into the payment gateway selection area when the form renders. The problem with this is that the gateway selection area has a `display: none` applied to it when rendered on the server. An option was to simply remove this for Classic, but it looks odd to be prompted as to how to pay when there's only one payment option.

This patch makes moving the fields around conditional, and moves the test notice into view as well. It also brings over the CSS necessary to style things properly when in the new area.

## Affects

Classic template

### Before
image

### After
image

## Testing Instructions

See the Issue for instructions on reproducing the issue. It should look _and work_ fine with this PR in place. Be sure to also test multiple gateways, to avoid regression issues.

## Pre-review Checklist

### DIFF
--- a/give.php
+++ b/give.php
@@ -5,7 +5,7 @@
  * Description: The most robust, flexible, and intuitive way to accept donations on WordPress.
  * Author: GiveWP
  * Author URI: https://givewp.com/
- * Version: 2.18.0
+ * Version: 2.18.1
  * Requires at least: 4.9
  * Requires PHP: 5.6
  * Text Domain: give
@@ -282,7 +282,7 @@ final class Give
     {
         // Plugin version.
         if ( ! defined('GIVE_VERSION')) {
-            define('GIVE_VERSION', '2.18.0');
+            define('GIVE_VERSION', '2.18.1');
         }
 
         // Plugin Root File.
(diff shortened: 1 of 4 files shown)


jonwaldstein (approved):
@JasonTheAdams confirming this resolved the issue. 👍

PR #6227, opened 2022-02-11
Tributes display styles are updated

kjohnson:
Related to impress-org/give-tributes#303

## Description

This PR updates the style in GiveWP that relate to the Tributes add-on.

Specifically, updating the selected state when using buttons (instead of radios) caused some layout/overlap issues.

### Multi-Step Form

| Before | After |
| --- | --- |
| image | Screenshot 2022-02-11 at 11-20-06 Donation Form with Tributes – WordPress |

## Testing Instructions

With the Tributes add-on installed and active:

- Create a donation form
- Add multiple tribute options, ie "In honor of"
- Update the tributes display style to "buttons"

## Pre-review Checklist

### DIFF
--- a/src/Views/Form/Templates/Sequoia/assets/css/tributes.scss
+++ b/src/Views/Form/Templates/Sequoia/assets/css/tributes.scss
@@ -82,11 +82,6 @@
         }
     }
 
-    // Match style of payment level buttons.
-    input[type='radio']:checked + label.give-tribute-type-button {
-        color: #28C77B !important;
-    }
-
     // Adjust layout to avoid overlapping text.
     .give-tributes-type-button-list {
         margin-top: 20px;
--- a/src/Views/Form/Templates/Sequoia/assets/css/tributes.scss
+++ b/src/Views/Form/Templates/Sequoia/assets/css/tributes.scss
@@ -43,7 +43,7 @@
         position: absolute !important;
     }
 
-    input[type='radio'] + label {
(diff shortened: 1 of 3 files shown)


src/Views/Form/Templates/Sequoia/assets/css/tributes.scss
```diff
@@ -82,7 +82,32 @@
         }
     }
 
-    input[type='radio']:checked + label::after {
+    // Match style of payment level buttons.
+    input[type='radio']:checked + label.give-tribute-type-button {
+        color: #28C77B !important;
```
LMNTL, line 87: This looks good! The only thing I would change is to remove the `!important` here and use the form's primary color for text. The primary color is added to the rest of the form in the `$dynamicCss` string here:

https://github.com/impress-org/givewp/blob/develop/src/Views/Form/Templates/Sequoia/Sequoia.php#L128

So you could add a little CSS there to style tribute buttons, then set the text color using `$primaryColor`.

LMNTL (changes requested):
Setting  $dynamicCss

src/Views/Form/Templates/Sequoia/assets/css/tributes.scss
kjohnson: Ah, primary color! Good catch 😅 

I pushed an update to use the Dynamic CSS string so that the primary color is incorporated.

image

LMNTL (approved):

Issue #385, opened 2015-11-06
Modal window not in mobile viewport


konus1:
I created a form with the setting that payment fields opens in a modal window.
If I try to use the form on my mobile phone (Android with Chrome Browser), then the modal window is below my current viewport. I have to scroll down to see the window.

The same happens, if  I this as widget into my sidebar (Theme: http://themegrill.com/themes/esteem/)

I am not sure, if this is a bug with my theme only. On my desktop, everything works OK.

mathetos:
Hi Konrad,

Our Github repo is really intended just for development of the plugin, pull-requests, and bugs. Your request is more a support request which we handle at the WordPress Plugin Directory here:

https://wordpress.org/support/plugin/give

Nevertheless, that is by design. Any given Give form might potentially be very LONG specifically on narrow devices, so it's important that the modal be capable of stretching outside the view in order to contain the whole form yet still be scrollable. But if that functionality isn't desired then it might be better to use a link in your sidebar to point your user to the full Give single page instead and use the "Show On Page" display method instead.

konus1:
I am Sorry. My problem was not that the Give form is stretched outside of the view because it is to long. The problem is that the modal window is not visible at all after opening on a mobile device. The user sees the darkend screen only and has no visual hint what he should do. 
Thats why I had the impression that this must be a bug.

mathetos:
I see... do you have a link to the site so I can see the problem live? I have a feeling that's a theme conflict.

konus1:
I created a test site at
http://beta.dd4kids.de/donations/modal-test/
you can also look at it wiht
http://www.responsinator.com/?url=http%3A%2F%2Fbeta.dd4kids.de%2Fdonations%2Fmodal-test%2F


Am 09.11.2015 um 19:27 schrieb Matt Cromwell:
>
> I see... do you have a link to the site so I can see the problem live? 
> I have a feeling that's a theme conflict.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/WordImpress/Give/issues/385#issuecomment-155147429>.
>

-- 
mit freundlichen Grüßen
Konrad Bauckmeier

Steirische Str.55, 01279 Dresden
Tel. 0351/2137976
http://www.dd4kids.de (Tipps für Eltern in Dresden)

mathetos:
OK, I'm pretty sure this is related to the styles your theme has set on lines 427-436:

`
/* Clearing */
.clear:before,
.clear:after,
[...]

PR #6268, opened 2022-02-24
feature: Check visibility before validating required.

kjohnson:
Resolves impress-org/give-form-field-manager#389

## Description

This PR implements a visibility check on fields before validating each field as required. This is in addition to the client-side validation that was previously solely responsible for considering visibility for required validation.

The issue in impress-org/give-form-field-manager#389 manifested as an issue with checkboxes because an unchecked checkbox does not submit with the form.

Additionally, this PR updates the required check to check for empty values. Previously it just checked that the value was seet in the `$_POST` array, but that would allow empty values to pass validation on the server.

## Acceptance Criteria

- [x] Required checkbox **fails** validation if no value is selected and the field is visible.
- [x] Required checkbox **passes** validation if no value is selected and the field is **not** visible.
- [x] Required radio **fails** validation if no value is selected and the field is visible.
- [x] Required radio **passes** validation if no value is selected and the field is **not** visible.
- [x] Required textbox **fails** validation<sup>1</sup> if no value is provided and the field is visible.
- [x] Required textbox **passes** validation if no value is provided and the field is **not** visible.

<sup>1</sup> Previously, required validation for textboxes was only checked by the browser (HTML5 validation), but the server did not validate correctly because an empty value was passed, so the field was not checked (an empty string passes an isset check). This wasn't noticeable before because the HTML5 validation didn't allow the form to submit, but the server side validation is required because HTML5 validation can be bypassed.

## Pre-review Checklist

### DIFF
--- a/src/Form/LegacyConsumer/Actions/DetermineVisibilityForRequest.php
+++ b/src/Form/LegacyConsumer/Actions/DetermineVisibilityForRequest.php
@@ -1,0 +1,115 @@
+<?php
+
+namespace Give\Form\LegacyConsumer\Actions;
+
+use ArrayObject;
+use Give\Framework\FieldsAPI\Conditions\BasicCondition;
+use Give\Framework\FieldsAPI\Conditions\Condition;
+use Give\Framework\FieldsAPI\Field;
+use Give\ValueObjects\Money;
+
+/**
+ * @unreleased
+ */
+class DetermineVisibilityForRequest
+{
+    /** @var bool */
+    const IS_VISIBLE = true;
+
+    /** @var Field */
+    private $field;
+
+    /** @var ArrayObject */
+    protected $postData;
+
+    /**
+     * @unreleased
+     * @param Field $field
+     * @param array $postData
+     */
+    public function __construct(Field $field, array $postData )
+    {
+        $this->field = $field;
+        $this->postData = new ArrayObject( $postData );
+    }
+
+    /**
+     * @unreleased
+     */
+    public function __invoke()
+    {
+        if( ! $this->fieldHasVisibilityConditions() ) {
+            return self::IS_VISIBLE;
+        }
+
+        $conditions = $this->field->getVisibilityConditions();
+        return array_reduce( $conditions, [$this, 'reduceVisibility'], self::IS_VISIBLE );
+    }
+
+    protected function fieldHasVisibilityConditions()
+    {
+        return method_exists( $this->field, 'hasVisibilityConditions' )
+            && $this->field->hasVisibilityConditions();
+    }
+
+    protected function reduceVisibility( $visibility, Condition $condition )
+    {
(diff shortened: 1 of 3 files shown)


src/Form/LegacyConsumer/Actions/DetermineVisibilityForRequest.php
```diff
@@ -0,0 +1,115 @@
...
+            : $visibility || $result;
+    }
+
+    protected function compareConditionWithOperator( Condition $condition )
+    {
+        if( is_a( $condition, BasicCondition::class ) ) {
+            return $this->compareBasicConditionWithOperator( $condition );
+        }
+
+        // @TODO Implement nested conditions.
```
kjohnson, line 70: I don't think the Fields API supports nested conditions yet, although the groundwork is in place in the API structure.

alaca: Yeah, nested conditions are not supported yet so we don't have to worry about that now.

src/Form/LegacyConsumer/Commands/SetupFieldValidation.php
```diff
@@ -60,6 +61,10 @@ public function __invoke($hook)
      */
     public function validate(Field $field)
     {
+        if( ! $this->isFieldVisible( $field ) ) {
+            return;
+        }
+
```
LMNTL, line 67: Short-circuiting the validation process if a field isn't visible might cause some issues. For example, a required File Upload field that only accepts image files is set to display when then donation amount is $25. The user can set the donation amount to $25, upload a .exe file, and then change the donation amount. Now the file upload field is hidden, and the donation continues even though the upload is invalid. Since hidden fields are still committed to the database, the file is uploaded successfully. (I'm not sure if hidden fields **should** get saved, but that's outside of the scope of the original issue.)

Note: This is related to #385, but that only applies when the field isn't required.

LMNTL (changes requested):
I think this looks great, there's just one edge case that might need addressing.

src/Form/LegacyConsumer/Commands/SetupFieldValidation.php
kjohnson: I would say that hidden fields *should not* be persisted - they should be treated as empty. This needs to be addressed as well, but in the Form Field Manager add-on.

kjohnson: @LMNTL I have created a separate issue in FFM to address this, see https://github.com/impress-org/give-form-field-manager/issues/391.

LMNTL: @kjohnson That makes sense to me! Approving this.

LMNTL (approved):

PR #6287, opened 2022-03-07
Fix: don't strip table prefixes in Query Builder

### LINKED ISSUE #6282: Donation Forms table should not break when the database prefix is give_
## User Story

As an admin, I want to view a list of my Donation Forms.

## Details

When Give's database is prefixed with `give_`, the Donation Forms page doesn't load, infinitely displaying the loading message.

## Expected Behavior

Donation Forms page should work regardless of DB prefix

## Steps to Reproduce

1. Install GiveWP
2.  Change your database prefix for your WP install to `give_`.
Note: This plugin allows you to do that very easily from the WP admin: https://wordpress.org/plugins/brozzme-db-prefix-change/
As always, BACK UP! before doing any procedure that deals with your database. (Thanks, @rickalday!)
3. In the admin area, navigate to the Donation Forms page (Donations -> All Forms) and notice that the page does not display

## System Information

</details>

## Acceptance Criteria

-   [ ] Donation Forms page displays correctly with a table prefix of `give_` or any other valid prefix.

LMNTL:
Resolves #6282 

## Description

When Give's database is prefixed with give_, the Donation Forms page doesn't load, infinitely displaying the loading message.

After some digging, it appears this was happening because of a Query Builder issue. The call to `attach_meta()` in the Donation Forms repository was stripping all of the instances of the prefix from the table name before prefixing it, so `give_give_formmeta` became `give_formmeta`. After some discussion, it was determined the best solution was to keep Query Builder from stripping the prefix altogether.
 
Before, Query Builder methods would either accept the prefixed or unprefixed table name. After this PR, Query Builder methods will only work with unprefixed table names. The raw SQL methods (i.e. `joinRaw', `selectRaw`) will still use the full prefixed table names.

## Affects

This will affect any other places where the Give Core version of Query Builder is currently being used.

Since the Donation Forms page was showing the loading spinner/message even after the initial API request failed, I've also tightened up error handling on that page. The `There was a problem retrieving the donation forms` error message now displays if the initial load fails, and will not wait until SWR is done retrying the request to display.

## Testing Instructions

1. Install GiveWP
2. Change your database prefix for your WP install to `give_`.
**Note**: This plugin allows you to do that very easily from the WP admin: https://wordpress.org/plugins/brozzme-db-prefix-change/
As always, BACK UP! before doing any procedure that deals with your database. (Thanks, @rickalday!)
3. In the admin area, navigate to the Donation Forms page (Donations -> All Forms) and make sure the page displays correctly
4. Change your database prefix to something other than `give_`.
5. Repeat step 3

## Pre-review Checklist

### DIFF
--- a/src/DonationForms/resources/components/DonationFormsTable.tsx
+++ b/src/DonationForms/resources/components/DonationFormsTable.tsx
@@ -113,7 +113,7 @@ export default function DonationFormsTable({statusFilter: status, search}: Donat
                     setPage={setPage}
                 />
             </div>
-            {initialLoad ? (
+            {( initialLoad && !error ) ? (
                 <div className={styles.initialLoad}>
                     <div className={cx(styles.tableGroup)}>
                         <Spinner size={'large'} />
@@ -253,7 +253,7 @@ export default function DonationFormsTable({statusFilter: status, search}: Donat
                         {isEmpty && (
                             <div className={styles.statusMessage}>{__('No donation forms found.', 'give')}</div>
                         )}
-                        {error && !isValidating && (
+                        {error && (
                             <>
                                 <div className={styles.statusMessage}>
                                     {__('There was a problem retrieving the donation forms.', 'give')}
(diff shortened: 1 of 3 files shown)


alaca (approved):
Looks good!

PR #6289, opened 2022-03-09
New: Query Builder now supports having the % wild card in whereLike* functions

### LINKED ISSUE #6288: Query Builder should support setting the % wild card in whereLike* functions
## User Story

As a developer, I want to be able to set the `%` wild card when using `whereLike*` functions so that I can better optimize my SQL queries.

## Details

By default, Query Builder will add a `%` wild card on both sides of the condition value. That works, but it is not optimized and can lead to performance issues. 

For example, if we want to select all meta keys from the meta table that are starting with `_give_donor_address_`,  using `whereLike('meta_key', '_give_donor_address_')` function, the current QB implementation will generate SQL like this.
```sql 
... WHERE meta_key LIKE '%_give_donor_address_%'
```
Since the meta_key column is indexed, this query is not optimized because MySQL has to do a full table scan in order to get all keys that are matching were clause because there is no access predicate, and MySQL is doing a full-text search on each row. This can be easily avoided by allowing developers to set the `%` wild card on the side they want.
So in the example above, the most efficient query would be:
```sql 
... WHERE meta_key LIKE '_give_donor_address_%'
```
Now MySQL uses `_give_donor_address_` as an access predicate on the indexed columns which will narrow down the scanned indexes rows significantly. 

## Acceptance Criteria

-   [ ] Developer is able to set `%` wild card on both sides of condition value  
Example: `whereLike('column', 'value%')`

alaca:
Resolves #6288

## Description

This PR adds support for having an `%` wild card in the Query Builder's `whereLike*` functions on both sides of the `WHERE` condition value which will help us to write more optimized SQL queries. 

Note: This is not a breaking change and all the previous usages of the `whereLike* `functions are working just fine.  

## Affects

Query Builder's `whereLike*` functions

## Testing Instructions

To have a sense of performance impact with this _small_ change:

1. Generate donors by using WP CLI  
`wp give test-donors --count=1000`
3. Now login to MySQL console  
`mysql -u yourusername -p yourpassword` 
4. Run explain command for unoptimized query  
`explain select meta_key from wp_give_donormeta where meta_key like '%donor%'`
5. Now repeat that for the optimized version  
 `explain select meta_key from wp_give_donormeta where meta_key like 'donor%'`

Screen Capture_select-area_20220309181757

[Clarification of the `rows` column in the screenshot. ](https://dev.mysql.com/doc/refman/8.0/en/explain-output.html#explain_rows)
_The rows column indicates the number of rows MySQL believes it must examine to execute the query._ 

The _unoptimized_ query is doing the full-table scan.
The _optimized_ query has to do much less than the _unoptimized_ query, and it is much more performant than the _unoptimized_ query.

**Edit**: it looks like the test data will set only `_give_donor_first_name` and `_give_donor_last_name` columns for donor. 
To have realistic results, we should update the database to look more _organic_.  In order to do that, we can simply run 
```sql
update wp_give_donormeta set meta_key = 'donor_first_name' where meta_key = '_give_donor_first_name' limit 500;
``` 
and then repeat the steps in testing instructions.

## Pre-review Checklist

### DIFF
--- a/src/Framework/QueryBuilder/Concerns/WhereClause.php
+++ b/src/Framework/QueryBuilder/Concerns/WhereClause.php
@@ -481,11 +481,13 @@ trait WhereClause
             case Operator::LIKE:
             case Operator::NOTLIKE:
                 return DB::prepare(
-                    "%1s %2s %3s '%%%s%%'",
+                    "%1s %2s %3s %s",
                     $where->logicalOperator,
                     $where->column,
                     $where->comparisonOperator,
-                    DB::esc_like($where->value)
+                    strpos($where->value, '%') !== false
+                        ? $where->value
+                        : sprintf('%%%s%%', $where->value )
                 );
 
             // Handle NULL conditions
--- a/tests/unit/tests/Framework/QueryBuilder/WhereTest.php
+++ b/tests/unit/tests/Framework/QueryBuilder/WhereTest.php
@@ -256,6 +256,37 @@ final class WhereTest extends TestCase
         );
     }
 
+    public function testWhereLikeWithWildCardPositionLeft()
+    {
+        $builder = new QueryBuilder();
+
+        $builder
+            ->select('*')
+            ->from(DB::raw('posts'))
+            ->whereLike('post_title', '%Donation');
+
+        $this->assertContains(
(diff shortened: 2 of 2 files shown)


JasonTheAdams (approved):
Great addition! 👍

PR #6300, opened 2022-03-14
New: Preload Donation Forms admin page data

LMNTL:
## Description

Link preloading can be used to speed up page loads by making requests for resources early. Since the Donation Forms page JS first fully loads and then makes a fetch request, I anticipated there could be a significant page load speed increase by preloading its data.

First, I tried preloading the fetch using `<link rel="preload">`, but that doesn't allow you to set necessary headers and didn't work with WP REST due to caching. Next, I tried using [@wordpress/api-fetch](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-api-fetch/) along with `usePreloadingMiddleware`, but that ended up being a poorly documented path that led to a brittle solution.

So I settled on this solution: make a REST request manually before enqueuing scripts, then include the resulting data in the global along with the WP REST API details. This can be backported to the Logs admin page or any other page that makes a predictable WP REST API request on load. With a piece of added middleware, this works cleanly with SWR, regardless of fetcher function used. On pages that don't use SWR, it's as simple as not making a fetch to the WP REST API on page load and using the preloaded data as the initial state. 

This PR speeds up the load process of the Donation Forms page, making it feel more snappy. The results are most pronounced when on faster connections, because the biggest savings come from skipping the fixed overhead of a single HTTP/S request.

## Affects

N/A

## Visuals

https://user-images.githubusercontent.com/7819986/158235079-0ee0c3ee-e4e5-435e-8489-f00a7f205fdd.mp4
Left: After this PR
Right: Before this PR

## Testing Instructions

Install GiveWP
Navigate to Donations -> All Forms
Make sure you have at least 11 donation forms to test pagination; if not, you can duplicate forms until you have 11
Verify that the page loads correctly
Verify that changing pages works correctly, loading new data
Verify that searching forms by title/ID works correctly

## Pre-review Checklist


src/DonationForms/DonationFormsAdminPage.php
```diff
@@ -49,6 +60,30 @@ public function loadScripts()
         );
     }
 
+    /**
+     * Make REST request to Donation Forms endpoint before page load
+     * @unreleased
+     */
+    public function preloadForms()
```
alaca, line 67: I'm not sure I'm a fan of this approach - running additional HTTP requests from the backend. 

What do you think about we pull this data from the repository? We can also create a `ViewModel` class for donation forms and use it here and in the `ListForms` endpoint. I know this would be extra work to do almost the same thing, but in the end, it will run much faster. In addition to `useFallbackAsInitial` middleware, this would be a win-win situation.

alaca: I will finish backend tasks for Scope 3 soon, so I can give you a hand here.

src/DonationForms/resources/api.ts
```diff
@@ -38,16 +39,21 @@ const Fetcher = (params) => {
...
-            if(error.name == 'AbortError') return;
-            if (retryCount >= 5) return
-            const retryAfter = (retryCount + 1) * 500;
-            setTimeout(() => revalidate({ retryCount }), retryAfter);
+    const {data, error, isValidating} = useSWR(
+        {page, perPage, status, search},
+        Fetcher,
+        {
+            use: [useFallbackAsInitial, lagData],
+            fallbackData: window.GiveDonationForms.preload,
```
alaca, line 47: Nice 👍🏻

src/DonationForms/resources/hooks/useFallbackAsInitial.ts
```diff
@@ -0,0 +1,24 @@
...
+    return (key, fetcher, config) => {
+
+        const hasMounted = useRef(false);
+
+        useEffect(() => {
+            hasMounted.current = true;
+        }, []);
+
+        // Actual SWR hook.
+        const swr = useSWRNext(key, fetcher, {
```
alaca, line 16: Magic 🪄

alaca (changes requested):
Looks good. I have a couple of suggestions, really minor things. Overall, great work. Preloading donation forms is an excellent idea 💯

src/DonationForms/DonationFormsAdminPage.php
LMNTL: @alaca As far as I understand, the `preloadForms` method currently doesn't do an HTTP request, either on the client or the server. It creates a mock `WP_REST_Request` object, then passes it to `rest_do_request`. The REST server matches the request to an endpoint, and then calls the endpoint's handler function using `call_user_func` and returns the response object. So we're just passing around objects and calling functions on them, without getting HTTP involved. Like this: https://wpscholar.com/blog/internal-wp-rest-api-calls/

alaca: Great, I didn't know that. As long as we don't make HTTP requests, I'm fine with that.

LMNTL:
Closed - this will be implemented (with some tweaking) as part of [#6328](https://github.com/impress-org/givewp/issues/6328).

PR #6307, opened 2022-03-18
Refactor: use generics for the ModelQueryBuilder

JasonTheAdams:
## Description

The existing `ModelQueryBuilder` required a return union type with every single type of Model, which then meant the IDE would autocomplete _every_ method and property for a given model, as it didn't know which one it is.

Using generics, we can inform the instance which Model class it represents. PHP doesn't actually have generics, but PhpStorm recently added support for inferring generics from docblocks. As you can see, this cleans things up quite a bit.

As a note, the `prepareQuery` return type is needed, for example `@return ModelQueryBuilder<Donation>`. The reason is PhpStorm doesn't _yet_ support [inferring the type from the `class-string<>` parameter](https://youtrack.jetbrains.com/issue/WI-65811).

## Affects

Existing model repositories

## Testing Instructions

Just make the repositories still work and that autocomplete works better in PhpStorm.

## Pre-review Checklist

### DIFF
--- a/src/Donations/Models/Donation.php
+++ b/src/Donations/Models/Donation.php
@@ -149,7 +149,7 @@ class Donation extends Model implements ModelCrud, ModelHasFactory
     /**
      * @unreleased
      *
-     * @return ModelQueryBuilder
+     * @return ModelQueryBuilder<Donor>
      */
     public function donor()
     {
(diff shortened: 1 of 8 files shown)


JasonTheAdams:
@jonwaldstein Can you give this a quick review?

jonwaldstein (approved):
@JasonTheAdams this is awesome, thanks!

I updated the models as well.

PR #6335, opened 2022-03-29
Fixed Block encountered error while adding class name

mhimon:
After inserting the block if you try to add any class name its broke.

Fix: Added some attributes to register_block_type

```
'className' => array(
'type' => 'string',
'default' => '',
),
'align' => array(
'type' => 'string',
'default' => '',
),
'extUtilities' => array(
'type' => 'object',
'default' => array(),
),
```
ServerSideRender requires the same attributes to render the block. By default, the class name field is empty ( Advanced Inspectors ) so it doesn't display any error but if you try to add any class name you will see the error.

### DIFF
--- a/blocks/donation-form/class-give-donation-form-block.php
+++ b/blocks/donation-form/class-give-donation-form-block.php
@@ -14,6 +14,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
+use Give\Helpers\Html;
+
 /**
  * Give_Donation_Form_Block Class.
  *
@@ -117,6 +119,18 @@ class Give_Donation_Form_Block {
 					'id'                  => array(
 						'type' => 'number',
 					),
+					'className' => array(
+						'type' => 'string',
+						'default' => '',
+					),
+					'align' => array(
+						'type' => 'string',
+						'default' => '',
+					),
+					'extUtilities' => array(
+						'type' => 'object',
+						'default' => array(),
+					),
 					'prevId'              => array(
 						'type' => 'number',
 					),
@@ -172,7 +186,17 @@ class Give_Donation_Form_Block {
 		$parameters['display_style']         = $attributes['displayStyle'];
 		$parameters['continue_button_title'] = trim( $attributes['continueButtonTitle'] );
 
-		return give_form_shortcode( $parameters );
+		$class_names = Html::classNames(
+			isset( $attributes['extUtilities'] ) ? $attributes['extUtilities'] : null,
+			! empty( $attributes['align'] ) ? 'align' . $attributes['align'] : null,
+			isset( $attributes['className'] ) ? $attributes['className'] : null
+		);
+
+		$content = '<div class="' . $class_names . '">';
+		$content .= give_form_shortcode( $parameters );
+		$content .= '</div>';
+
+		return $content;
 	}
 
 	/**
--- a/blocks/donation-form/class-give-donation-form-block.php
+++ b/blocks/donation-form/class-give-donation-form-block.php
@@ -123,10 +123,6 @@ class Give_Donation_Form_Block {
 						'type' => 'string',
 						'default' => '',
 					),
-					'align' => array(
-						'type' => 'string',
-						'default' => '',
-					),
 					'extUtilities' => array(
 						'type' => 'object',
 						'default' => array(),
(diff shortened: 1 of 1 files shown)


ravinderk:
@mhimon thank you for your contribution. I was able to reproduce this issue.

mhimon:
Hi,
will you merge later?

blocks/donation-form/class-give-donation-form-block.php
```diff
@@ -172,7 +186,17 @@ public function render_donation_form( $attributes ) {
...
 		$parameters['continue_button_title'] = trim( $attributes['continueButtonTitle'] );
 
-		return give_form_shortcode( $parameters );
+		$class_names = Html::classNames(
+			isset( $attributes['extUtilities'] ) ? $attributes['extUtilities'] : null,
+			! empty( $attributes['align'] ) ? 'align' . $attributes['align'] : null,
+			isset( $attributes['className'] ) ? $attributes['className'] : null
+		);
+
+		$content = '<div class="' . $class_names . '">';
```
ravinderk, line 195: @mhimon Is there any reason to add this extra `div`?

Changes in DOM structure can cause arbitrary errors on the front. I would like to suggest you use [`give_form_wrap_classes`](https://github.com/impress-org/givewp/blob/78f7634b6da410d722196cb22922d79cd3c1961e/includes/class-give-donate-form.php#L876) filter to add custom classes to the donation form wrapper.

blocks/donation-form/class-give-donation-form-block.php
```diff
@@ -117,6 +119,18 @@ public function register_block() {
...
 					),
+					'className' => array(
+						'type' => 'string',
+						'default' => '',
+					),
+					'align' => array(
+						'type' => 'string',
+						'default' => '',
+					),
+					'extUtilities' => array(
```
ravinderk, line 130: @mhimon Where I can find the `align` setting and the `extUtilities` setting because the `Advanced` tab only contains setting for custom classes.

ravinderk (changes requested):
@mhimon I added review to this pull request. Let me know if you have any questions.

blocks/donation-form/class-give-donation-form-block.php
mhimon: ok, I will update the code. Thanks

blocks/donation-form/class-give-donation-form-block.php
mhimon: We can remove align attribute as that block doesn't have support for that. 

And extUtilities is in Advanced Inspector Controls.

Screen Shot 2022-05-05 at 6 13 09 PM

mhimon:
Check new commit

blocks/donation-form/class-give-donation-form-block.php
ravinderk: @mhimon it seems that `Exntendify Utilities` is not the default setting field. Does it appear when you install plugin https://wordpress.org/plugins/extendify/

mhimon: Sorry! yes, I think so. Thanks man

mhimon (commented):
Removed extUtilities

blocks/donation-form/class-give-donation-form-block.php
```diff
@@ -172,6 +178,14 @@ public function render_donation_form( $attributes ) {
 		$parameters['display_style']         = $attributes['displayStyle'];
 		$parameters['continue_button_title'] = trim( $attributes['continueButtonTitle'] );
 
+		$class_names = Html::classNames(
+			isset( $attributes['className'] ) ? $attributes['className'] : null
+		);
+
+		add_filter( 'give_form_wrap_classes', function() use ( $class_names ) {
```
ravinderk, line 185: `give_form_wrap_classes ` filter hook returns list of classes in `array` format. 

The current implementation will override other classes and only return a list of custom classes which can cause of unexpected output on the frontend. Can you update your logic to merge custom classes with the list of classes.

ravinderk (changes requested):
@mhimon We are almost ready. I added a suggestion. Let me know if you have any questions.

mhimon (commented):
Updated code to merge custom classes

blocks/donation-form/class-give-donation-form-block.php
mhimon: Thanks! I forgot to merger.

blocks/donation-form/class-give-donation-form-block.php
```diff
@@ -172,6 +178,14 @@ public function render_donation_form( $attributes ) {
 		$parameters['display_style']         = $attributes['displayStyle'];
 		$parameters['continue_button_title'] = trim( $attributes['continueButtonTitle'] );
 
+		$class_names = Html::classNames(
+			isset( $attributes['className'] ) ? $attributes['className'] : null
+		);
+
+		add_filter( 'give_form_wrap_classes', function( $classes ) use ( $class_names ) {
+			return Html::classNames( $classes, $class_names );
```
ravinderk, line 186: @mhimon this function should return an array.
I tested this with [Multi-step form template](https://givewp.com/documentation/core/give-forms/form-templates/) and classes are not added to the form wrapper.

mhimon: Maybe problem with that filter hook

ravinderk: @mhimon this is not an issue with the filter. `Html::classNames` returns a string of class names instead of an array. 

https://github.com/impress-org/givewp/blob/452aad332e35f65767a021314d9630bf6ce2fa5d/includes/class-give-donate-form.php#L876

kjohnson:
Closing this for now. We can re-evaluate using the new [feedback](https://feedback.givewp.com/) workflow is this is still an issue.

PR #6341, opened 2022-03-30
Change: Add subscription table migrations to core

JasonTheAdams:
## Description

Now that more subscription code has been moved to core, more actions are taking place in various circumstances. For example, when deleting a form, the `Give_Recurring_Cache` attempts to clear out the subscription tables. This throws an error when attempting to delete forms elsewhere. That's no good.

We've also recently introduced Subscription models, and using those would also throw errors without the table in place. As such, I think it makes sense to move the table into core. It will most likely remain empty if the customer doesn't have the Recurring add-on installed, but honestly it won't hurt anything and is less likely to cause issues by existing.

I also cheated and added a line to the new Form Admin UI which deletes form meta upon deletion. Presently, the form post was being deleted, but its meta stayed.

## Affects

1. When the subscription and subscription meta tables are introduced
2. Deleting a form from the new form archive

## Testing Instructions

**Deleting a form**
1. Delete a form from the forms table
2. Check the DB to make sure the meta is removed
3. Make sure no errors occurred

**New Install**
1. Make a fresh install
2. Install GiveWP
3. Check for the tables
4. Make sure deleting a form doesn't throw errors
5. Activate Recurring add-on and make sure nothing breaks
6. Make a new Subscription donation to confirm all is well

**Upgrading**
1. Install an old version of GiveWP
2. Activate the Recurring add-on (to add the tables)
3. Update GiveWP (to trigger the new migration)
4. Make sure everything still works!

## Pre-review Checklist

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -52,9 +52,7 @@ jobs:
                     coverage: none
 
             -   name: Install composer dependencies
-                uses: php-actions/composer@v5
-                with:
-                    php_version: ${{ matrix.php }}
+                run: composer install --no-progress --no-interaction
 
             -   name: Store MySQL details in environment
                 run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,9 +51,7 @@ jobs:
                     coverage: none
 
             -   name: Install composer dependencies
-                uses: php-actions/composer@v5
-                with:
-                    php_version: ${{ matrix.php }}
+                run: composer install --no-progress --no-interaction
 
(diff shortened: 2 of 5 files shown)


LMNTL (approved):
This works in my testing.

jonwaldstein (approved):
Works for me 👍

rickalday (approved):
Tested on an old site
- DB migrations runs without any issues
- Form metadata is deleted when a form is deleted
- New subscriptions work without any errors

New install:
- No errors on activation
- No errors after activating Recurring Donations
- New subscriptions work without any issues.

PR #6348, opened 2022-04-04
Enhancement: Add excerpt length option to Form Grid Block

### LINKED ISSUE #6323: Excerpt length for donation form grid block
Currently it's not possible to control the excerpt length, an option for in the block settings would be great.

https://givewp.canny.io/admin/board/feature-requests/p/excerpt-length-for-donation-form-grid-block

LMNTL:
Resolves #6323

## Description

The Form Grid block lacked the ability to set the displayed length of the form excerpt, a feature that the Form Grid shortcode has. It would always show the first 15 words of the excerpt. This PR fixes that by adding an "Excerpt Length" setting in the Form Grid Block options. The "Excerpt Length" option is only visible when "Show Excerpt" is set to true, otherwise it's hidden using `display: none;`.

## Affects

N/A

## Visuals

image

## Testing Instructions

1. Install GiveWP
2. Create a new form or edit an existing form
3. Scroll to the bottom of the Edit Donation Form screen and add an excerpt, then click "Publish"
4. Create a new post or page and add a Form Grid block (click the '+' and search for Donation Form Grid)
5. Click on the Form Grid block and toggle "Show Excerpt" on
6. Input various values for the "Excerpt Length" option and verify that the excerpt shows that number of words
7. Publish the post/page
8. View the form grid from the frontend and verify that the excerpt shows the correct number of words

## Pre-review Checklist


blocks/donation-form-grid/edit/inspector.js
```diff
@@ -98,7 +98,20 @@ const Inspector = ( { attributes, setAttributes } ) => {
 					label={ __( 'Show Excerpt', 'give' ) }
 					checked={ !! showExcerpt }
 					onChange={ ( value ) => saveSetting( 'showExcerpt', value ) } />
-				<ToggleControl
+                <TextControl
+                    name="excerptLength"
+                    label={ __( 'Excerpt Length', 'give' ) }
+                    value={ excerptLength ? excerptLength : '' }
+                    className={ !showExcerpt && 'hidden' }
+                    onChange={ ( value ) => {
```
alaca, line 106: We are setting the `excerptLength` value only if `value === ''` and `intValue == value`?
Users should be able to set `excerptLength` to some arbitrary value, but they can't do that now.

alaca (changes requested):

blocks/donation-form-grid/edit/inspector.js
alaca: Resolved 68ecac6d

alaca (approved):
Looking good now 😃

blocks/donation-form-grid/edit/inspector.js
LMNTL: Users probably shouldn't be able to set `excerptLength` to any arbitrary value, only an empty string or a positive integer. With the way it's just been refactored, the user can set a negative `excerptLength`. Not the biggest deal, and I appreciate the simplification of the code, just something to consider.

PR #6351, opened 2022-04-08
Refactor: update models repositories to mutate model instances

jonwaldstein:
## Description

When count was added to the model factories during https://github.com/impress-org/givewp/commit/ecd9878275174ada39397f9a5513ae170a06d15c the created instances of model factories were not being returned - only their initial instantiated instances.  This resulted in model factories not returning important properties like their ID and created_at which caused all the model tests to fail.

This brought us to discuss how model repositories _should_ handle the given instance.  The conclusion was to let the repositories mutate the model instance and not return anything.

This means the api is changing slightly, so instead of returning a new instance every time, the current instance will be updated.

Here is a basic example

```
$model = new Model();
$model->foo = 'bar';
$model->save()
$model->id // 1
$model->foo // 'bar'
``` 

During this PR a few other updates were made:

1. The Subscription model was missing an important conditional in its save method.
2. Change action `give_subscription_deleted` to `give_subscription_model_deleted`. There was already an action with this name so needed to avoid conflict.
3. added `afterMaking` method to model factory.
4. Updated default model factory properties that require ids with factory creations (relationships)
5. Updated all Model `create` methods to return the model instance
6. Updated all Model `save` methods to return void
7. Updated all Model `insert` methods to return void
8. Updated all model repositories to update their instances on insert.
9. Updated all model & repository tests to reflect the change.

## Affects

- Donor, Donation, and Subscription models & how they use create & save
- Donor, Donation, and Subscription repositories & how they use insert & update
- All Models and Model Repository tests

## Testing Instructions

1. Make sure when running Model::factory()->create() it actually returns the created instance - not the instantiated instance.
2. Play around with the Model create & save methods.
3. Run the model & repository unit tests. 

## Pre-review Checklist


JasonTheAdams (commented):
@jonwaldstein I'm a bit confused by something, here. When a model is saved what happens to the instance? Does it get destroyed? Why would we need to specifically store the instance returned from the save method?

jonwaldstein:
@JasonTheAdams please review this.  I don't know how our WordPress Tests action did not alert us of failed unit tests when this was committed. Maybe because it wasn't a PR?

jonwaldstein:
@JasonTheAdams saving inserts or updates depending on if an ID is present.  So the instance needs to be returned in its new state.

JasonTheAdams:
Objects properties are shared even when object variables are passed by value. So if the repository (or whatever) sets the `id`, then all references to that object will immediately reflect this. This is only not true if an entirely new instance is instantiated.

JasonTheAdams:
This is because objects are a group of pointers to positions in memory. So even if you pass it by value and all the pointer values are copied, the addresses remain the same, thus the "new" object instance still points to the same property addresses as the previous.

JasonTheAdams:
As an example:
```php
class Repo {
  public function insert(Model $model) {
    $model->id = 1;
  }
}

class Model {
  public $id;
  
  public function save() {
    (new Repo)->insert($this);
  }
}

$model = new Model;
$model->id; // null

$model->save();
$model->id; // 1
```

jonwaldstein:
@JasonTheAdams right now we're inserting and updating the database and returning the new instance, so I think what your'e suggesting is that we need to update the model properties after those transactions as well?

JasonTheAdams:
Yeah, I think it's counterintuitive otherwise. I think "oh, I'm saving *this* model", but then the model I'm referencing actually changes and is no longer the same. We could do `$this = $repo->insert()` if we think it's important to return an instance.

jonwaldstein:
@JasonTheAdams okay per our conversation, this PR is ready for review.  I'll probably be on PTO by the time this is reviewed so feel free to make any updates you see fit.

src/Donations/Repositories/DonationRepository.php
```diff
@@ -166,19 +168,64 @@ public function insert(Donation $donation)
...
+            $donation->levelId = 0;
+        }
+
+        if (!isset($donation->gatewayTransactionId)) {
+            $donation->gatewayTransactionId = null;
+        }
+
+        if (!isset($donation->parentId)) {
+            $donation->parentId = 0;
+        }
```
JasonTheAdams, line 217: @jonwaldstein Do you think any of these should simply be model defaults?

JasonTheAdams (commented):
Great work, @jonwaldstein. My biggest thing is that I'm wondering if some of these insertion property checks should simply be model property defaults — that is, the default of the given property when the model is instantiated. Thoughts?

jonwaldstein:
@JasonTheAdams yeah, honestly default properties would probably clean this up a lot. We technically don't have a first class way of defining defaults outside of just doing an array_merge within create.

src/Donations/Repositories/DonationRepository.php
JasonTheAdams: I added defaults: a82748c

JasonTheAdams (approved):

JasonTheAdams:
@kjohnson I added property defaults as part of this and Jon is out for the next couple weeks. I've reviewed Jon's work. Would you mind glancing over the default property stuff I added and making sure it's all sensible?

In short, all properties are stored upon instantiation with `null` as the default. But the following syntax can be used to provide a default:
```php
protected $properties = [
  'id' => 'int', // null by default
  'fristName' => ['string', 'Kyle'], // Kyle by default
];
```

You'll notice this is limited to non-expression defaults. So in the `Donation` model you'll see I overrode the `getPropertyDefaults()` parent method to add more defaults that required method calls and such.

src/Framework/Models/Model.php
```diff
@@ -245,11 +249,47 @@ protected function validatePropertyType($key, $value)
...
+     *
+     * @unreleased
+     *
+     * @param $key
+     *
+     * @return mixed|null
+     */
+    protected function getPropertyDefault($key)
+    {
+        return is_array(
```
kjohnson, line 274: This combination of a multi-line, multiple logic check combined with an inline ternary can be a bit confusing. This could be resolved with changing to linebreaks.

```php
return is_array(...) && isset(...)
    ? $this->properties[$key][$i]
    " null;
```

kjohnson (approved):

src/Framework/Models/Model.php
JasonTheAdams: Fixed in f419cfccc

jonwaldstein:
@JasonTheAdams thanks for adding the model property defaults, this should simplify a lot of tediousness. 👍

PR #6393, opened 2022-04-29
New: Delete donations, donors, and donation forms CLI command

### LINKED ISSUE #6391: CLI command to delete data
As an admin, I want to be able to delete donations using the WP CLI.
A scenario I could use is when I have thousands of spam donations that I want to delete. Right now, I can bulk delete max 30 donations at a time, and using WP CLI I could delete much more.

https://givewp.canny.io/admin/board/feature-requests/p/cli-command-to-delete-data

alaca:
Resolves #6391

## Description

This PR adds a new WP CLI command that allows users to delete donors, donations, and donation forms. 

The new command is `wp give delete` which is used with required additional parameters:

`--donations` - delete donations
`--donors` - delete donors
`--donors-and-donations` - delete donors and their donations
`--donation-forms` - delete donation forms

Note: it is possible to use multiple parameters at the same time:

`wp give delete --donations --donors --donation-forms`

## Affects

Give CLI commands.

## Testing Instructions

Commands:
`wp give delete --donations`
`wp give delete --donors`
`wp give delete --donors-and-donations`
`wp give delete --donation-forms`

## Pre-review Checklist

### DIFF
--- a/includes/class-give-cli-commands.php
+++ b/includes/class-give-cli-commands.php
@@ -779,13 +779,18 @@ class GIVE_CLI_COMMAND {
      * [--donors]
      * : Delete all donors
      *
+     * [--donors-and-donations]
+     * : Delete all donors and their donations
+     *
      * [--forms]
      * : Delete all donation forms
      *
      * ## EXAMPLES
      *     wp give delete --donations
      *     wp give delete --donors
+     *     wp give delete --donors-and-donations
      *     wp give delete --forms
+     *     wp give delete --donations --donors --forms
      *
      * @unreleased
      */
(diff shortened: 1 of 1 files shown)


kjohnson:
@alaca I think I understand now from where you derived the command names - I had not considered the cascading nature of the relationships. Forms have donations, have donors, etc.

kjohnson:
Also, the Canny feedback doesn't give us much direction. I see how this change fits into the current CLI conventions.

That said, I think we should consider deprecating the existing commands and refactor to a resource based convention in the Next Gen plugin, using models.

kjohnson (commented):
I'm considering on pushing back on this change, not because of the solution, but because I find the original feedback lacking. 

Specifically, we need to consider deleting single items for a given resource, in addition to deleting all items for a given resource. Also, we need to consider how we handle cascading relationship deletions.

Further, I might suggest re-implementing this as a first step towards a more resource based CLI toolset, with an interface such as `wp give donations delete`, as opposed to extending the current set of CLI Commands.

PR #6399, opened 2022-04-29
Feature: cache model relations

JasonTheAdams:
## Description

Consider the following:
```php
$donation->donor->firstName = 'Jason';
$donation->donor->lastName = 'Adams';
```

Presently, this will actually trigger two implicit queries. When the `donor` property is accessed it queries the donor under the hood. This means that in both cases when the donor is accessed it's querying the model. This is bad for a few reasons:
1. It's really counterintuitive
2. It's an easy performance sink
3. The object returned changes every time, so the second line is actually done to a different model

This PR adds a very light caching layer that stores relations after the first query. This allows for things like the example to happen and perform as one would intuit.

## Affects

Model relationships

## Testing Instructions

Grab a donation and then retrieve the donor twice. The two objects should be the literal (`===`) same.

## Pre-review Checklist

### DIFF
--- a/src/Framework/Models/Model.php
+++ b/src/Framework/Models/Model.php
@@ -42,6 +42,13 @@ abstract class Model implements Arrayable
     protected $relationships = [];
 
     /**
+     * Relationships that have already been loaded and don't need to be loaded again.
+     *
+     * @var Model[]
+     */
+    private $cachedRelations = [];
+
+    /**
      * Create a new model instance.
      *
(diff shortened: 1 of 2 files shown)


JasonTheAdams:
@jonwaldstein This small but mighty PR is ready for review. 💪

jonwaldstein (approved):
@JasonTheAdams very nice! 🚀

PR #6401, opened 2022-05-02
Enhancement: Add redaction and better context handling to logs

JasonTheAdams:
## Description

We're running into an issue with sensitive information being stored in logs. We've handled this in a couple specific places, but we really should be trying to prevent this information from being stored on a Log level, otherwise we'll keep patching this sort of stuff as time goes on. Obviously this doesn't perfectly catch everything, but it would've caught all known issues thus far.

This PR adds checking for sensitive information and also makes a couple other improvements:
1. It makes sure the details modal fits within the available vertical space and provides scrolling in case of overflow.
2. It stores the contextual data as actual JSON, rather than a `print_r()` string. This is important as it helps us to better format and display the data on the front-end.

As a note, in cleaning up when objects are stored, I opted to add a "class" property when an object is being serialized. This way we have all of the properties stored, but also the name of the class the properties belong to.

Related: #6389 

## Visuals

image

## Testing Instructions

You can use the following to reproduce the visual:
```php
use Give\Log\Log;

Log::info('Here\'s some stuff', [
  'name' => 'Jason Adams',
  'password' => 'password1!',
  'confirm_password' => 'mypassword',
  'credit_card' => '4242',
  'deeper_array' => [
    'business' => 'GiveWP',
    'business_card' => '4242 4242',
    'foo' => new Foo()
  ]
]);

class Foo {
  public $bar = 'bar';
  public $secret = 'so secret';
}
```

Also view some old logs where the context was stored using `print_r` to make sure there's no regressions.

## Pre-review Checklist

### DIFF
--- a/src/Log/Log.php
+++ b/src/Log/Log.php
@@ -65,22 +65,14 @@ use Give\Log\ValueObjects\LogType;
  */
 class Log
 {
+    const REDACTION_LIST = ['card', 'password', 'secret', 'token'];
+
     public function __call($name, $arguments)
     {
         list ($message, $context) = array_pad($arguments, 2, null);
 
         if (is_array($context)) {
-            // Convert context values to string
-            $context = array_map(
-                function ($item) {
-                    if (is_array($item) || is_object($item)) {
-                        $item = print_r($item, true);
-                    }
-
-                    return $item;
-                },
-                $context
-            );
+            $context = $this->serializeAndRedactContext($context);
 
             // Default fields
             $data = array_filter(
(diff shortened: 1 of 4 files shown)


JasonTheAdams:
@jonwaldstein @alaca I'd appreciate your thoughts on this. I _think_ it's a pretty clear improvement. 😄

JasonTheAdams:
Test are failing... says there's an error in the `Model` class on line 418... there are only 400 lines. 😳

ravinderk:
> Test is failing... says there's an error in the `Model` class on line 418... there are only 400 lines. 😳

@JasonTheAdams tests are failing because We are still running tests on PHP `5.6` but in code, we are writing `7.0` compatible code.

ravinderk:
@JasonTheAdams I am facing the following issues:

1. Object is not formatted corrected instead of that I see a long string.
Snip - GiveWP Tools ‹ GiveWP Playground — WordPress

2. I am getting a javascript error on the log page.
Snip - GiveWP Tools ‹ GiveWP Playground — WordPress

JasonTheAdams:
@ravinderk Please pull again and let me know if that resolves your issues. Remember to run `npm run dev`. 😉

ravinderk:
> @ravinderk Please pull again and let me know if that resolves your issues. Remember to run `npm run dev`. 😉

I still facing issues:

Screenshot

src/Log/Log.php
```diff
@@ -117,15 +109,49 @@ function ($key) {
...
 
     /**
-     * Static helper for calling the logger methods
+     * Takes the context array, serializes it, and redacts sensitive data.
      *
-     * @param  string  $name
-     * @param  array  $arguments
+     * @unreleased
+     */
+    private function serializeAndRedactContext(array $context): array
```
alaca, line 116: Interesting approach, I like it. Little extra work is needed on larger nested arrays, but the benefit is obvious. The only thing I would suggest here is to add more keywords to the redaction list. For example, CVV (card security code) comes in multiple variations. More about that [here](https://en.wikipedia.org/wiki/Card_security_code#Naming).

alaca (commented):
Looks good. I have only one suggestion for the redaction list.

alaca:
@ravinderk JS warning resolved here 76ea42e1

src/Log/Log.php
ravinderk: @alaca I think @JasonTheAdams added only `card` because in GiveWP all credit card fields start with this prefix.

JasonTheAdams: Yup, that's exactly right. I'm trying to ride the line and not get into other languages and such. I did have the idea, however, to make a `getRedactionList()` method with a filter so customers can add more if they need to.

JasonTheAdams: Added a filterable redaction list in 6a04ca0

JasonTheAdams:
I believe this should be ready to rock if someone doesn't mind providing another review. I've added another migration which removes all logs containing any of words from the redaction list.

ravinderk (approved):

PR #6520, opened 2022-07-22
Fix: add additional escaping to the legacy form rendering

JasonTheAdams:
## Description

To further proactively improve security in GiveWP, this PR adds a bunch more escaping to the form rendering template. We haven't had any reported exploits, but the additional escaping will help make sure no injection is possible within forms — arguably the most critical part of GiveWP.

As escaping goes, this shouldn't break any existing proper functionality, but doing things like setting the currency symbol to contain html, or setting the legacy template modal button caption to include markup, and things like that, should now just show up as text in the front-end — it should not allow for new DOM elements or, especially, JavaScript.

## Affects

How the amount field render, and also the modal display button in the legacy template.

## Testing Instructions

Primarily we just need to test that everything still work as expected. Also good to try injecting some markup into form settings to see if anything is rendering on the front-end.

## Pre-review Checklist

-   [ ] Includes unit tests


JasonTheAdams:
@ravinderk Would appreciate your eyes on this when you have some time. You don't need to drop what you're doing. Let's just can get this out next week. I think it should be pretty quick.

ravinderk (approved):

PR #6526, opened 2022-08-09
Fix: prevent MAX_JOIN_SIZE limit from occurring and blocking queries

### LINKED ISSUE #6527: Large queries should not prevent donations
## Details

We are seeing a handful of reports of donations failing with a cryptic error (front end) about contacting site administrators. On the back end in the GiveWP logs the error registers an error about a "Database Query"

The problem seems to be related to the new Models introduced in recent GiveWP versions, and certain servers not being able to handle JOINS over a certain size. We're looking into a fix, but for now the workaround is to get your host to increase the MAX_JOIN_SIZE to something large like 20000. We're treating this bug with urgency, but need to be able to reliably reproduce it first. Please comment here with any ways to reproduce it.

https://givewp.canny.io/admin/board/bug-reports/p/large-queries-should-not-prevent-donations

## Acceptance Criteria

-   [ ] GiveWP queries are able to run successfully on lower max_join_size settings

jonwaldstein:
Resolves #6527

## Description

This adds an action `givewp_db_pre_query` to our DB class for optimizing mysql before running data retrieval queries.

There is also an easy way to set mysql big selects using the DB class `DB::enableBigSelects()`.

With this action people can do things like this:

```
add_action('givewp_db_pre_query', function() {
    static $bigSelects = false;

    if (!$bigSelects) {
        global $wpdb;

        $wpdb->query('SET SESSION SQL_BIG_SELECTS=1;');

        $bigSelects = true;
    }
});
```

or with our helper
```

add_action('givewp_db_pre_query', function() {
    \Give\Framework\Database\DB::enableBigSelects();
});

```

or target more specific queries

```
use Give\Framework\Database\DB;
use Give\Framework\Support\Facades\Str;

add_action('givewp_db_pre_query', function($sql) {
    if (Str::contains($sql, 'LEFT JOIN')){
        DB::enableBigSelects();
    }
});
```

## Affects

Our database queries using the query builder.

## Testing Instructions

- Set your mysql to max_join_size = 1500000000
- Try to view the donations list table (should not load)
- add snippet below and view donations list table (should load now)

```
add_action('givewp_db_pre_query', function() {
    \Give\Framework\Database\DB::enableBigSelects();
});
```

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/give.php
+++ b/give.php
@@ -57,6 +57,7 @@ use Give\Form\LegacyConsumer\ServiceProvider as FormLegacyConsumerServiceProvide
 use Give\Form\Templates;
 use Give\Framework\Exceptions\UncaughtExceptionLogger;
 use Give\Framework\Migrations\MigrationsServiceProvider;
+use Give\Framework\Database\ServiceProvider as DatabaseServiceProvider;
 use Give\Framework\PaymentGateways\PaymentGatewayRegister;
 use Give\Framework\WordPressShims\ServiceProvider as WordPressShimsServiceProvider;
 use Give\LegacySubscriptions\ServiceProvider as LegacySubscriptionsServiceProvider;
@@ -197,6 +198,7 @@ final class Give
         PromotionsServiceProvider::class,
         LegacySubscriptionsServiceProvider::class,
         WordPressShimsServiceProvider::class,
+        DatabaseServiceProvider::class,
     ];
 
     /**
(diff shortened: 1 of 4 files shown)


jonwaldstein:
@JasonTheAdams this is ready for review.  As we discussed, i'd like to put this to the test in different real-world situations before introducing a more permanent GiveWP setting.

rickalday (approved):
I tested on 2 customer sites and a local by updating the DB.php file and applying the snippet.
I also made several Stripe test donations to make sure no errors came up.

PR #6538, opened 2022-08-16
Nodes in the Fields API can now be tapped

kjohnson:
## Description

This PR introduces a tap method to `Element`, `Field`, and `Group` nodes. Tapping a node allows for that node to be updated without breaking the fluent syntax of the Fields API.

```php
$field = Field::make( 'my-field' )
    ->label( 'My Field' )
    ->tap(function( $node ) {
        // Custom code here
    })
    ->isRequired();

```

For group, child nodes can be tapped using the fluent syntax of the Fields API.

```php
// Tap a nested node by name.
$group->getNodeByName( 'child' )->tap( ... );
```

## Testing Instructions

Create a node and call the tap function. Changes to that node should be reflected and the fluent syntax should not be broken.

## Pre-review Checklist

-   [x]  Includes unit tests

### DIFF
--- a/src/Framework/FieldsAPI/Concerns/TapNode.php
+++ b/src/Framework/FieldsAPI/Concerns/TapNode.php
@@ -2,9 +2,14 @@
 
 namespace Give\Framework\FieldsAPI\Concerns;
 
+/**
+ * @unreleased
+ */
 trait TapNode
 {
     /**
+     * @unreleased
+     *
      * @param callable $callback
      *
      * @return $this
--- a/src/Framework/FieldsAPI/Concerns/TapNode.php
+++ b/src/Framework/FieldsAPI/Concerns/TapNode.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Give\Framework\FieldsAPI\Concerns;
+
+trait TapNode
+{
+    /**
+     * @param callable $callback
+     *
+     * @return $this
+     */
+    public function tap(callable $callback)
+    {
+        call_user_func($callback, $this);
+
+        return $this;
+    }
+}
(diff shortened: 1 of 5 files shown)


src/Framework/FieldsAPI/Concerns/TapNode.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Give\Framework\FieldsAPI\Concerns;
+
+/**
+ * @unreleased
+ */
+trait TapNode
```
JasonTheAdams, line 8: @kjohnson Why not just add this method to the `Node` class and let inheritance do its thing? Why apply the trait to each child class separately?

kjohnson: `Node` is an interface. `Element`, `Field`, and `Group` are the base concrete implementations.

```php
interface Node extends JsonSerializable {
     ...
}
```

JasonTheAdams: Oh. That would explain that. 😁

kjohnson: Yeah, caught me off guard. "What do I do now???"

JasonTheAdams (approved):
Looks great! Are we adding the group tapNode function in this PR? If another, to ahead and merge and we'll bust out the other one.

kjohnson:
@JasonTheAdams Since `Node` is an interface and I had to add `tap()` as a trait, I don't have a place for `tapNode()` that feels right. Adding a second `TapNodes` trait seemed odd, because it is so similarly named, and adding it directly to the `Group` class seemed odd, because it assumes that underlying nodes support a method that can't be added to the interface directly.

If you think it is important enough, then I can get over the above and find a good enough place for `tapNode()`.

JasonTheAdams:
Nah, just add it to the Group class. That's fine. If you want to add it to the Collection interface that's fine.

PR #6542, opened 2022-08-17
Nodes in a Group can now be tapped

kjohnson:
## Description

This PR introduces a `tapNode()` method to the `Group` node type. Similar to the `tap()` method, except that `tapNode()` will return the group node.

```php
$group = Group::make('my-group')->append(
    $node = Text::make('my-text')
)->tapNode('my-text', function( $tappedNode ) {
    //
}); // returns $group
``` 

## Testing Instructions

Create a group with a nested node and call the `tapNode()` method on the group, specifying the nested node by name. Changes to that node should be reflected and the fluent syntax should not be broken.

## Pre-review Checklist

-   [x] Includes unit tests

### DIFF
--- a/src/Framework/FieldsAPI/Group.php
+++ b/src/Framework/FieldsAPI/Group.php
@@ -59,9 +59,17 @@ class Group implements Node, Collection
         return new static($name);
     }
 
-    public function tapNode($name, callable $callback)
+    /**
+     * Gives th ability to fluently "tap" a specific node within the group. This is useful when fluently calling methods
+     * on the group, and making a change to a specific node without breaking the fluency.
+     *
+     * @unreleased
+     *
+     * @return $this
+     */
+    public function tapNode(string $name, callable $callback)
     {
-        call_user_func($callback, $this->getNodeByName($name));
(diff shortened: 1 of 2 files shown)


src/Framework/FieldsAPI/Group.php
```diff
@@ -58,4 +58,11 @@ public static function make($name)
     {
         return new static($name);
     }
+
+    public function tapNode($name, callable $callback)
```
JasonTheAdams, line 62: Please add a doc block

JasonTheAdams (approved):
Love this, @kjohnson! Just needs a little documentation!

src/Framework/FieldsAPI/Group.php
JasonTheAdams: Resolved in 912315bef

JasonTheAdams:
In case you didn't know, @kjohnson, in PHP 7+ you can call a variable like a function, so you don't need to use `call_user_func`:
```php
call_user_func($callback, $this->getNodeByName($name)); // a bit confusing to look at

$callback($this->getNodeByName($name)); // much clearer
```

PR #6551, opened 2022-08-24
Refactor: make use of Stripe's PaymentMethod Api

jonwaldstein:
…od object and add GetStripeGatewayData action

Resolves #

## Pre-review Checklist

-   [ ] Includes unit tests


jonwaldstein:

src/PaymentGateways/Gateways/Stripe/Actions/GetStripeGatewayData.php
```diff
@@ -0,0 +1,44 @@
...
+            /**
+             * This supplies the actual Stripe Api PaymentMethod
+             */
+            $gatewayData['paymentMethod'] = $paymentMethod;
+            /**
+             * This internal ValueObject is currently being used throughout the Stripe gateways.
+             * Eventually this array key will be @deprecated in favor of using Stripe's PaymentMethod
+             * Object directly via $gatewayData['paymentMethod']
+             */
+            $gatewayData['stripePaymentMethod'] = new PaymentMethod($paymentMethod->id);
```
kjohnson, line 31: Should we go ahead and mark `stripePaymentMethod` as deprecated?

src/PaymentGateways/Gateways/Stripe/Actions/GetStripeGatewayData.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Give\PaymentGateways\Gateways\Stripe\Actions;
+
+use Give\Donations\Models\Donation;
+use Give\Framework\Exceptions\Primitives\Exception;
+use Give\Framework\PaymentGateways\Log\PaymentGatewayLog;
+use Give\PaymentGateways\Gateways\Stripe\ValueObjects\PaymentMethod;
```
kjohnson, line 8: Should we add an alias that further communicates that this is the Give Value Object? Something like `LegacyPaymentMethodValueObject`?

I think it would be helpful to avoid confusion when this is re-visited.

src/PaymentGateways/Gateways/Stripe/Actions/GetStripeGatewayData.php
jonwaldstein: Is it possible to mark an array item as deprecated?  I wasn't sure about that.

src/PaymentGateways/Gateways/Stripe/Actions/GetStripeGatewayData.php
jonwaldstein: Good idea, i'll update that

src/PaymentGateways/Gateways/Stripe/Actions/GetStripeGatewayData.php
kjohnson: I was assuming that you could and questioning the statement "Eventually this array key will be @deprecated".

That said, it doesn't look like PHPStan supports the `@deprecated` tag for array key items - unfortunately.

kjohnson: The (perhaps harebrained) idea that comes to mind is that we could replace the array with an `ArrayAccess` object and then throw deprecation notices in a magic `__get` method.

Probably overkill in this situation.

jonwaldstein:
@kjohnson let's revisit this next cooldown.  We need to make sure this will work downstream with all stripe gateways.

PR #6562, opened 2022-09-20
Enhancement: Improvements to recurrence in the Donation and Subscription models

JasonTheAdams:
## Description

Related: https://github.com/impress-org/give-recurring/pull/1133

This PR aims to improve the experience of dealing with recurrence in the Donation and Subscription models. There were two primary issues to be tackled:
1. Donations didn't have a clear "type" to indicate whether it was singular, the start of a subscription, or a renewal. This is a regular question within business logic and had to be determined via a mix of confusing meta.
2. Subscriptions had an "expiration" method that was terribly confusing. It had nothing to do with when the subscription actually ended, but rather when the next renewal would be. Super confusing.
3. Subscriptions and Initial Donations self-referenced. The proper order for creating a subscription should be to make the subscription and then make the donation (which references the subscription). But in our current database structure the subscription and initial donation both reference each other... This creates a chicken and egg situation.
4. Testing Subscriptions was weird, as the factory for the Subscription was odd (due to the self-referencing).

It was a bit of a tangled mess, which is why this is a single PR instead of 4. The hope though, is that the strangeness is isolated to the repository and the models are easy to use, pointing us towards how we _want_ them to work.

```php
$donation->type->isRenewal(); // checks whether its a renewal or not

$subscription = Subscription::create(['frequency' => 1, 'period' => SubscriptionPeriod::MONTH());
$subscription->bumpRenewalDate(); // sets the next renewal date; just a helper
$subscription->renewsAt; // gets/sets the next renewal date
```

## Affects

The Donation and Subscription models and repositories.

These are used in the latest Google Analytics release, so we'll need to make a PR for that as well.

## Testing Instructions

We'll want to test anywhere these models are used, which I know is a bit spotty. For the subscription renewals, I wasn't sure what we might need to test/update to make sure the renewal dates are properly set. This should work with legacy code as the underlying data structure wasn't changed (still uses the expiration column).

## Pre-review Checklist

-   [x] Includes unit tests


JasonTheAdams:
@jonwaldstein This should be ready for review (I'll fix any failing tests). As a note, I'm considering removing the `initialDonationId` property from the Donation model entirely. I don't really want folks to get/set that. I'd probably make a method somewhere for retrieving the id for the legacy places it's used. I'm open to your thoughts.

jonwaldstein:
@JasonTheAdams I haven't tested anything yet but this is looking great, nice job!  

Re: `initialDonationId` yes let's get rid of that from the donation model.  It's very much an underlying architecture term that would be better suited for behind the scenes.  In the future we'll want the renewals to reference the subscription record as their _parent_ not the initial donation.

JasonTheAdams:
I went ahead and removed the `initialDonationId` property, @jonwaldstein. It was actually simpler than I thought. Since the legacy data structure depends on it in the database, I'm retrieving it during insert and update for donations — but only for renewals. Also, the method being used to retrieve the subscription donation doesn't rely on that column, so the repositories effectively work without it.

JasonTheAdams:
@jonwaldstein With Glauber getting to Recurring soon it's more important that we get this done so he's ready to rock over there.

src/Donations/Repositories/DonationRepository.php
```diff
@@ -183,9 +184,9 @@ public function insert(Donation $donation)
                     'post_date_gmt' => get_gmt_from_date($dateCreatedFormatted),
                     'post_modified' => $dateCreatedFormatted,
                     'post_modified_gmt' => get_gmt_from_date($dateCreatedFormatted),
-                    'post_status' => $donation->status->getValue(),
+                    'post_status' => $this->getPersistedDonationStatus($donation)->getValue(),
                     'post_type' => 'give_payment',
-                    'post_parent' => $donation->parentId ?? 0
+                    'post_parent' => $donation->type->isRenewal() ? give()->subscriptions->getInitialDonationId($donation->subscriptionId) : 0,
```
jonwaldstein, line 189: Let's add a method to cleanup this duplicated logic for inserting and updating.

src/Donations/ValueObjects/DonationStatus.php
```diff
@@ -11,22 +11,22 @@
...
  * @method static DonationStatus REFUNDED()
  * @method static DonationStatus FAILED()
- * @method static DonationStatus CANCELED()
+ * @method static DonationStatus CANCELLED()
  * @method static DonationStatus ABANDONED()
  * @method static DonationStatus PREAPPROVAL()
  * @method static DonationStatus PROCESSING()
  * @method static DonationStatus REVOKED()
- * @method static DonationStatus RENEWAL()
+ * @method static DonationStatus RENEWAL() @deprecated
```
jonwaldstein, line 19: I'm not sure if this actually marks it deprecated but we should add a description as well.

src/Subscriptions/Repositories/SubscriptionRepository.php
```diff
@@ -256,6 +258,21 @@ public function delete(Subscription $subscription): bool
...
+     * @return void
+     */
+    public function updateLegacyParentPaymentId(int $subscriptionId, int $donationId)
+    {
+        DB::table('give_subscriptions')
+            ->where('id', $subscriptionId)
+            ->update([
+                'parent_payment_id' => $donationId,
+            ]);
+    }
```
jonwaldstein, line 274: Since this is a public method on the repository, we should add some validation. I'm thinking at the very least make sure the donation exists since the where clause would fail if the subscriptionId does not exist. Also wrapping in a transaction wouldn't hurt.

jonwaldstein (changes requested):
Very nice! Just a few suggestions.

src/Donations/Repositories/DonationRepository.php
```diff
@@ -468,11 +408,36 @@ private function validateDonation(Donation $donation)
...
         }
 
+        if ( $donation->subscriptionId && $donation->type->isSingle()) {
+            throw new InvalidArgumentException('Subscription ID can only be set for recurring donations.');
+        }
+
+        if ( !$donation->subscriptionId && ( $donation->type->isRenewal() || $donation->type->isSubscription() ) ) {
+            throw new InvalidArgumentException('Subscription ID is required for recurring donations.');
+        }
+
```
jonwaldstein, line 418: @JasonTheAdams I just ran into an issue submitting a recurring donation using stripe and this seems to be the culprit.  Remember, the donation is created _before_ the subscription right now so the `subscriptionId` is going to be null even if the type is subscription.

jonwaldstein: In `LegacyPaymentGatewayAdapter->handleBeforeGateway` we're trying to save the status as subscription before the subscription exists so there is no subscriptionId

JasonTheAdams: Resolved in 07edc009f

src/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayAdapter.php
```diff
@@ -56,11 +58,10 @@ public function handleBeforeGateway(array $legacyDonationData, PaymentGatewayInt
         );
 
         $donation = $formData->toDonation($donor->id);
+        $donation->type = $isRecurring ? DonationType::SUBSCRIPTION() : DonationType::SINGLE();
         $donation->save();
```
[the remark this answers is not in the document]
JasonTheAdams: Resolved in 07edc009f

src/Donations/ValueObjects/DonationStatus.php
JasonTheAdams: Fun fact, I learned there's no way to mark magic methods as deprecated. I even reported this to PhpStorm: https://youtrack.jetbrains.com/issue/WI-68849/Marking-a-magic-method-method-as-deprecated-does-nothing

JasonTheAdams: Further details included in 7c1ac47e6

src/Donations/Repositories/DonationRepository.php
JasonTheAdams: Resolved in 81c61aff1

src/Subscriptions/Repositories/SubscriptionRepository.php
JasonTheAdams: I added further documentation in 515d11cbd.

I'm not really sure any further validation is necessary. It's not like we check to see if the subscription exists whenever we save a donation with a `subscriptionId`. The where clause won't fail if the subscription doesn't exist, it will simply do nothing.

Also, a transaction wouldn't do anything here. If it fails there's nothing to rollback.

JasonTheAdams:
Thanks, @jonwaldstein! Good catch on the subscription/donation save order. 🤦‍♂️ 

I addressed all the comments so back to you!

src/Subscriptions/Repositories/SubscriptionRepository.php
jonwaldstein: Okay thanks, it might be beneficial to actually add that validation when saving a donation (if a subscription exists) - we _are_ validating that the donor exists.  Can discuss another time.

src/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayAdapter.php
```diff
@@ -70,22 +68,23 @@ public function handleBeforeGateway(array $legacyDonationData, PaymentGatewayInt
...
-            give()->donations->updateLegacyDonationMetaAsInitialSubscriptionDonation($donation->id);
-            give()->subscriptions->updateLegacyColumns(
-                $subscription->id,
-                [
-                    'parent_payment_id' => $donation->id,
-                    'expiration' => $subscription->expiration()
-                ]
-            );
+            $donation->type = DonationType::SUBSCRIPTION();
+            $donation->save();
```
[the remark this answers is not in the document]
JasonTheAdams: Resolved in 3f908b3

src/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayAdapter.php
```diff
@@ -70,22 +68,23 @@ public function handleBeforeGateway(array $legacyDonationData, PaymentGatewayInt
                 'donorId' => $donor->id,
                 'installments' => (int)$subscriptionData->times,
                 'status' => SubscriptionStatus::PENDING(),
-                'donationFormId' => $formData->formId
+                'donationFormId' => $formData->formId,
             ]);
```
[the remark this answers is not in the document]
JasonTheAdams: I took a stab at resolving this in 398671c, but I'd like your input.

We're chasing our tail a bit here. It is required — in the database — so there's no getting around that. And I made the bump function to make it easy to set the `renewsAt` property. Normally, I'd say we just need to do that when building a Subscription, but this wouldn't work for `Subscription::create()`, as that immediately builds the model and saves it.

I thought about simply bumping it in the `create` function, but thought perhaps it's just useful to have it auto-generated when saving. I then did it in the repository since the properties required for that function are validated. This meant I had to remove it from the required list, so I added a manual check to `SubscriptionRepository::update()`. I _think_ this is fine, practically speaking, it just feels a little weird.

Open to your thoughts!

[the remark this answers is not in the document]
jonwaldstein: This solution is fine.  Having to define the `renewsAt` property when you create a subscription would be tedious because the default relies on the frequency and period which we're already defining.  Since we're validating those properties _before_ dynamically setting the property, there should be no issues.  This is pretty much how it was working with the previous `$subscription->expiration()` method which was set under-the-hood _after_ creating the subscription.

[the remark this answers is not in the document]
JasonTheAdams: I wouldn't mind requiring folks to define the `renewsAt`, as they could do something like:
```php
$sub->frequency = 1;
$sub->period = SubscriptionPeriod::MONTH();
$sub->bumpRenewalDate();
```
It's the specific complication this creates with `Subscription::create()` that mostly bothers me. So the question is whether `renewsAt` is generated at that point specifically, or as a side-effect whenever saving.

JasonTheAdams:
@jonwaldstein Back to you for reivew!

src/PaymentGateways/Actions/RegisterPaymentGateways.php
```diff
@@ -36,8 +38,8 @@ class RegisterPaymentGateways
      */
     public $gateways = [
         // When complete, the Test Gateway will eventually replace The legacy Manual Gateway.
-        //TestGateway::class,
-        //TestGatewayOffsite::class,
+        TestGateway::class,
+        TestGatewayOffsite::class,
```
jonwaldstein, line 42: I'm assuming leaving these open was not intentional but I want to point out that I don't think we're quite ready to fully replace the legacy "Test Donation".  We're really close though and I'm thinking we can open a PR for that specific purpose very soon - just don't want customers to be confused about multiple test gateways.

JasonTheAdams: Oops! Yes, that was an oops. Resolved in f480a9a29

src/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayAdapter.php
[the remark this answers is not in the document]
jonwaldstein: I think in this case it makes the most sense to keep what you have (as a side-effect during save) - in fact our other time properties like `createdAt` and `updatedAt` work the same way.  Unless defined, they are generated during persistence. In order to set the property dynamically during create, we would have to add additional validation for the properties (like frequency & period) it requires to be generated which is not common to any other models.

jonwaldstein (approved):
@JasonTheAdams great updates!  let's send this off to QA 🚀

Benunc (approved):
This passes all automated tests.

PR #6564, opened 2022-09-22
Hotfix: `give_goal` and `give_form' display the goal progress correctly

### LINKED ISSUE #6563: `give_goal` and `give_form` with a legacy template should display the form goal
## Details

Currently if using the shortcodes `give_goal` or `give_form` with a legacy template. The form goals are not displaying.

## Expected Behavior

While using either shortcode with a legacy template the goals display accordingly.

## Steps to Reproduce

1. Create a form with the legacy template.
2. Create a page and use the shortcodes [give_goal] and [give_form].
3. View page.

## Visuals

Uploading Screen Recording 2022-09-22 at 1.25.05 PM.mov…

## Acceptance Criteria

-   [ ] The shortcode [give_goal] works correctly by displaying the goal of the form id given.
-   [ ] The shortcode [give_form] works correctly with all templates by displaying the goal of the form id given.
-   [ ] While changing the goal options for a form, the correct options display with both [give_goal] and [give_form]

JoshuaHungDinh:
Resolves #6563

## Description

The PR removes a get_id() Wordpress function in the `give_goal` shortcode. The function was not being used and causing the shortcode to fail when rendering. It was an oversight from PR https://github.com/impress-org/givewp/pull/6541.  Because the Legacy template relies on `give_goal` to display the goal progress  - fixing the issue in `give_goal` has corrected the issue with `give_form` while using a legacy template.

## Affects

`give_goal` shortcode

## Visuals

https://user-images.githubusercontent.com/75056371/191848931-c33b6e21-afe5-44d6-9131-03fc89b94dee.mov

## Testing Instructions

- Create a form with the Legacy template.
- Enable & set donation goal options to amount raised.
- Process a donation.
- Create a page using `give_goal` and `give_form`
-  View Page to verify the goal section displays.
- Change the donation goal to each different option and view each option from the shortcodes to verify the different goal options continue to work.
- Repeat steps with all templates.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/templates/shortcode-goal.php
+++ b/templates/shortcode-goal.php
@@ -3,7 +3,6 @@
  * This template is used to display the goal with [give_goal]
  */
 
-$form_id          = get_the_ID(); // Form ID.
 $form        = new Give_Donate_Form( $form_id );
 
 $goal_option = give_get_meta( $form->ID, '_give_goal_option', true );
--- a/templates/shortcode-goal.php
+++ b/templates/shortcode-goal.php
@@ -209,7 +209,6 @@ $progress = apply_filters( 'give_goal_amount_funded_percentage_output', $progres
         </div>
     <?php endif; ?>
 
-
     <?php if ( ! empty( $show_bar ) ) :
         $style = "width:{$progress_bar_value}%";
 


templates/shortcode-goal.php
```diff
@@ -3,7 +3,6 @@
  * This template is used to display the goal with [give_goal]
  */
 
-$form_id          = get_the_ID(); // Form ID.
 $form        = new Give_Donate_Form( $form_id );
```
glaubersilva, line 6: I know that the **$form_id** value is passed through the **give_show_goal_progress()** from the **includes/forms/template.php** file, but at first, the look seems stranger to using a variable without initializing that. So, I would suggest initializing it using something like this: `$form_id = isset($form_id) ? $form_id : 0;`

JasonTheAdams: I wouldn't advise this. If the `$form_id` comes from a parent context it just needs to be documented. If we check and set the value to 0 then we're just setting ourselves up for a potentially unexpected outcome. I would first document it via:
```php
/**
 * @var int $form_id form id passed from the give_show_goal_progress() context
 */
```

Then, if we feel there's a chance this could be misused, I would do something like:
```php
if ( empty($form_id) ) {
    Log::error(...);
    return false;
}
```
I really don't like returning false, as that rarely means what we think it means, but since it's what's already happening that's fine.

glaubersilva (approved):
Seems good to me, thanks! =)

templates/shortcode-goal.php
```diff
@@ -1,10 +1,21 @@
...
  */
 
-$form_id          = get_the_ID(); // Form ID.
-$form        = new Give_Donate_Form( $form_id );
+/**
+ * @var int $form_id form id passed from the give_show_goal_progress() context
+ */
+
+if ( empty($form_id) ) {
+    Log::error('Error: Form ID can not be empty.');
```
JasonTheAdams, line 14: @JoshuaHungDinh Few notes on good logging:
1. You don't need to prefix it with "error" or anything, as you are already marking it as an error: `Log:error()`
2. Ask yourself if the description would help you to find the issue. Where is this happening? Would we know that?
3. If you have data to provide to the 'context' array then please do.
4. Make the log happen somehow so you can see what's actually recorded and make sure its enough and would really help. This is a favor to our future selves.

JoshuaHungDinh: Resolved in : https://github.com/impress-org/givewp/pull/6564/commits/9cf4b9c7
Screen Shot 2022-09-27 at 11 05 19 AM

Benunc (approved):
This fix passes all automated tests, and I visually confirmed that the progress bar shows up now.

PR #6573, opened 2022-10-12
Tests: Introducing api for add-on test suite bootstrapping

jonwaldstein:
## Description

Introducing a new way of bootstrapping our test suite in GiveWP add-ons: `GiveTests\Framework\Addons\Bootstrap`. 

**Imaginary Give Addon**

```
give-addon
  - give-addon.php
  - tests
    - bootstrap.php
```

**bootstrap.php file**

```
<?php

use GiveTests\Framework\Addons\Bootstrap;

require __DIR__ . '/../../give/vendor/autoload.php';

(new Bootstrap(__DIR__ . '/../give-addon.php'))
    ->load();
```

Notice all we need to do is
1. Pull in the main give autoload file
2. Instantiate the `Bootstrap` class with the path to our main add-on file and call `->load()`

Under the hood, we are using the path to load our add-on using the WP test filter `'muplugins_loaded'` - then loading the GiveWP bootstrap file.

## Affects

This does not directly affect anything until used.  

There's one minor update in our main GiveWP bootstrap file to use our new `TestHooks::addFilter`.

## Testing Instructions

This would need to be tested within a real add-on which will accompany this PR in the near future.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -20,9 +20,6 @@ $currentTestEnvironment = $testEnvironment->current();
 // define for use in WP bootstrap file
 define('WP_TESTS_CONFIG_FILE_PATH', $currentTestEnvironment->config());
 
-// pull in wp test functions like tests_add_filter
-require_once $currentTestEnvironment->functions();
-
 // load GiveWP
 TestHooks::addFilter('muplugins_loaded', static function () {
     require_once __DIR__ . '/../give.php';
@@ -33,6 +30,9 @@ TestHooks::addFilter('setup_theme', static function () {
     echo 'Installing GiveWP.....' . PHP_EOL;
     give()->install();
 });
+
+// pull in wp test functions like tests_add_filter
+require_once $currentTestEnvironment->functions();
 
 // pull in WP bootstrap file which looks for WP_TESTS_CONFIG_FILE_PATH defined above
 require_once $currentTestEnvironment->bootstrap();
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -31,9 +31,6 @@ TestHooks::addFilter('setup_theme', static function () {
     give()->install();
 });
 
-// pull in wp test functions like tests_add_filter
-require_once $currentTestEnvironment->functions();
-
 // pull in WP bootstrap file which looks for WP_TESTS_CONFIG_FILE_PATH defined above
 require_once $currentTestEnvironment->bootstrap();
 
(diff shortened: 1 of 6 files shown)


tests/bootstrap.php
```diff
@@ -19,20 +20,20 @@
 // define for use in WP bootstrap file
 define('WP_TESTS_CONFIG_FILE_PATH', $currentTestEnvironment->config());
 
-// pull in wp test functions like tests_add_filter
-require_once $currentTestEnvironment->functions();
-
 // load GiveWP
-tests_add_filter('muplugins_loaded', static function () {
+TestHooks::addFilter('muplugins_loaded', static function () {
```
JasonTheAdams, line 24: What's the difference in using the new `TestHooks` over `tests_add_filter`?

JasonTheAdams (commented):
Looking at this, are we just making the assumption that to run tests in CI we'll just need to make sure that GiveWP core is adjacent to the add-on directory?

tests/bootstrap.php
jonwaldstein: They are the same thing functionally.  We were pulling in _all_ the WP test suite functions in this file just to use that `tests_add_filter` function.  It was also tricky because that functions file is a dynamic path depending on what environment you are in (local vs workflow).  Also, the main WP bootstrap file being included at the end _already_ includes that functions file.  So `TestHooks` pulls out the things we need like addFilter and can be used in add-ons without double loading WP's functions file.

jonwaldstein:
> Looking at this, are we just making the assumption that to run tests in CI we'll just need to make sure that GiveWP core is adjacent to the add-on directory?

Yes, unless you have a better idea - I think mimicking the structure of WP plugins (adjacent to each other) makes the most sense since that's how they are loaded in WordPress.  cc: @glaubersilva - will be working on the CI portion of add-on unit tests.

glaubersilva (approved):
This is great and worked really well! I did some tests using P2P as a sample and was able to run all tests from the add-on even after replacing the [P2P bootstrap file](https://github.com/impress-org/give-peer-to-peer/blob/ebf69e13be472406a3f2ddc1112227fed8de9079/tests/unit/bootstrap.php) content with this:

```
<?php

use GiveTests\Framework\Addons\Bootstrap;

require __DIR__ . '/../../../give/vendor/autoload.php';

(new Bootstrap(__DIR__ . '/../../give-peer-to-peer.php'))->load();
```

PR #6585, opened 2022-10-18
Fix: Display error messages on Classic template

### LINKED ISSUE #6584: Classic template should display error messages
## Details

Currently, the Classic template does not display error messages. When you submit a donation form and there's an error while processing the form, nothing is returned.

## Expected Behavior

Whenever you have an error, it must be printed on the form so the user is aware something happened with the transaction.

## Steps to Reproduce
_The following steps are only one example of how to trigger an error._

1. Create a form that accepts recurring payments
2. Select the Classic Template
3. Go to the page with that form
4. Select the option to pay recurrently
5. Select Offline Payment as the method (it does not support recurring payments)
6. Submit the form and see that nothing happens

## Visuals
CleanShot 2022-10-18 at 14 44 39
*Multi-step form presenting the error message*

CleanShot 2022-10-18 at 14 45 34
*Classic template form with no error message*

CleanShot 2022-10-18 at 14 46 20
*Return of the ajax call*

## Acceptance Criteria

-   [ ] When the form submission responds with an error, it's printed out.

pauloiankoski:
Resolves #6584

## Description

Errors messages are not being displayed on forms using the Classic template. This PR fixes the selector being used to print the ajax response data when on a Classic template form and adds margins when multiple errors are displayed at the same time.

## Affects

Classic template

## Visuals

CleanShot 2022-10-18 at 15 28 28

## Testing Instructions

1. Create a form using the Classic template
2. Go to the page with that form
3. Enter "123" as the first name
4. An error must be displayed

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/assets/src/js/frontend/give-ajax.js
+++ b/assets/src/js/frontend/give-ajax.js
@@ -252,7 +252,12 @@ jQuery( document ).ready( function( $ ) {
 				$this.val( complete_purchase_val );
 				loading_animation.fadeOut();
 				this_form.find( '.give_errors' ).remove();
-				this_form.find( '#give_purchase_submit input[type="submit"].give-submit' ).before( data );
+
+                if ( this_form.has( '.give-payment-details-section .give-payment-mode-label' ) ) {
+                    this_form.find( '.give-payment-details-section .give-payment-mode-label' ).after( data );
+                } else {
+                    this_form.find( '#give_purchase_submit input[type="submit"].give-submit' ).before( data );
+                }
 
 				// Enable the form donation button.
 				Give.form.fn.disable( this_form, false );
--- a/assets/src/js/frontend/give-ajax.js
+++ b/assets/src/js/frontend/give-ajax.js
@@ -248,13 +248,15 @@ jQuery( document ).ready( function( $ ) {
 
 				this_form.trigger( 'give_form_validation_passed' );
 			} else {
+                var $giveFormHeader = this_form.parent().find( '.give-form-header' );
+
 				//There was an error / remove old errors and prepend new ones
 				$this.val( complete_purchase_val );
 				loading_animation.fadeOut();
-				this_form.find( '.give_errors' ).remove();
+				this_form.parent().find( '.give_errors' ).remove();
 
-                if ( this_form.has( '.give-payment-details-section .give-payment-mode-label' ) ) {
-                    this_form.find( '.give-payment-details-section .give-payment-mode-label' ).after( data );
+                if ( $giveFormHeader.length > 0 ) {
+                    $giveFormHeader.after( data );
                 } else {
                     this_form.find( '#give_purchase_submit input[type="submit"].give-submit' ).before( data );
                 }
--- a/assets/src/js/frontend/give-ajax.js
+++ b/assets/src/js/frontend/give-ajax.js
@@ -248,7 +248,7 @@ jQuery( document ).ready( function( $ ) {
 
 				this_form.trigger( 'give_form_validation_passed' );
 			} else {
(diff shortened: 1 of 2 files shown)


src/Views/Form/Templates/Classic/resources/css/_errors-notices.scss
```diff
@@ -26,6 +33,10 @@
     padding-inline: 0.75rem;
     position: relative;
 
+    &:not(:last-child) {
+        margin-bottom: 0.9rem;
```
kjohnson, line 37: Do we need both uses of the `margin-bottom` rule, or can one of the two uses be dropped?

assets/src/js/frontend/give-ajax.js
```diff
@@ -252,7 +252,12 @@ jQuery( document ).ready( function( $ ) {
 				$this.val( complete_purchase_val );
 				loading_animation.fadeOut();
 				this_form.find( '.give_errors' ).remove();
-				this_form.find( '#give_purchase_submit input[type="submit"].give-submit' ).before( data );
+
+                if ( this_form.has( '.give-payment-details-section .give-payment-mode-label' ) ) {
+                    this_form.find( '.give-payment-details-section .give-payment-mode-label' ).after( data );
```
kjohnson, line 257: Should these errors be added at the top specifically to the payment details section? I see that the test mode notice is there, but that is specific to payments.

Should more generic notices be closer to the top (like multi-step) or closer to the submit button?

/cc @jdghinson

pauloiankoski: CleanShot 2022-10-18 at 16 36 22
*Errors moved to top*

kjohnson: Ok, yeah, I think that makes more sense here. Thanks for relocating those.

kjohnson (approved):
Looks good to me! Thanks for the quick resolution on the discussions 👍🏻

kjohnson:
@pauloiankoski can you send a build to QA?

pauloiankoski:
> @pauloiankoski can you send a build to QA?

Sure. I'm generating the zip file.

pauloiankoski:
@kjohnson I just added one more commit based on a suggestion from Rick to scroll the page up after printing errors. Could you review it?

CleanShot 2022-10-19 at 14 13 22

assets/src/js/frontend/give-ajax.js
```diff
@@ -261,6 +261,8 @@ jQuery( document ).ready( function( $ ) {
                     this_form.find( '#give_purchase_submit input[type="submit"].give-submit' ).before( data );
                 }
 
+                this_form.closest( 'body' )[0].scrollIntoView({ behavior: 'smooth' });
```
kjohnson, line 264: I'm assuming that the closest `body` would be the top of the iframe. but not all templates use the iframe - for example the Legacy Template is just HTML on the page.

Instead, maybe we could scroll to the Form Header, if in use? Or maybe there is another scroll to top example that you could reference.

kjohnson (changes requested):

assets/src/js/frontend/give-ajax.js
pauloiankoski: CleanShot 2022-10-19 at 15 03 04
*Legacy Template*

CleanShot 2022-10-19 at 15 03 54
*Classic Template*

kjohnson:
@pauloiankoski if this is otherwise approved by QA then go ahead and do a squash and merge.

kjohnson:
Ah, yeah, I forgot to re-approve after the changes - I'll do that now.

kjohnson (approved):

rickalday (approved):
Passed manual tests

PR #6591, opened 2022-10-20
Form Grid donate button text color now defaults to black

### LINKED ISSUE #6590: Form Grid donate button text is not visible on a white background
Design updates to the Form Grid changed the presentation of the Donate button to be styled as a link. This is a design decision for the Form Grid moving forward.

As for the Shortcode it seems that the default text color for the Donate Button is set to white, presumably a carry over from text color styled as a button, but now the button text displayed as a link is not readable.

> When the Form Grid is being used, there are a some display issues.
>
> Block:
> There should be a Donate button (if the option is selected), but it is a link.
>
> Shortcode:
> no link or button is showing.
>
> A CSS workaround for the Block grid might look like this:

```
.give-form-grid button {
    border: 2px solid #28C77B !important;
    background: #28C77B !important;
    color: white;
    text-decoration: none;
    padding: 1rem!important;
    border-radius: 3px;
}    

.give-form-grid button span {
    text-decoration: none!important;
    color: white!important;
    
}
```

https://givewp.canny.io/admin/board/bug-reports/p/on-the-form-grid-the-donate-button-is-no-longer-a-button

kjohnson:
Resolves #6590

## Description

This PR updates the default value for the `donate_button_text_color` property of the `give_form_grid` shortcode to inherit the text color to match the surrounding elements.

## Affects

The `give_form_grid` shortcode used directly uses the `donate_button_text_color`, where as the newer Block wrapper for the shortcode always specifies a text color, thus the default is not used.

## Visuals

| Shortcode used directly |
| :---: |
| `[give_form_grid show_donate_button=true]` |
| Screen Shot 2022-10-20 at 15 57 26 |

| Block wrapping the shortcode |
| --- |
| Screen Shot 2022-10-20 at 15 57 56 |

## Testing Instructions

Add the Form Grid shortcode to a page/post with the `show_donate_button` attribute enabled.

Example: `[give_form_grid show_donate_button=true]`

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1203209111237825

### DIFF
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -791,7 +791,7 @@ add_shortcode( 'give_totals', 'give_totals_shortcode' );
  * @type string $display_style How the form is displayed, either in new page or modal popup.
  *                                       Default 'redirect'. Accepts 'redirect', 'modal'.
  *
- * @unreleased Updates the default text color for the donate button, see #6591.
+ * @unreleased Updated the default text color for the donate button, see #6591.
  * @since 2.21.2 change tag_background_color, progress_bar_color to official green color #69b868.
  *             change tag_text_color color to #333333.
  * @since 2.20.0 $show_donate_button Option to show donate button
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -791,6 +791,7 @@ add_shortcode( 'give_totals', 'give_totals_shortcode' );
  * @type string $display_style How the form is displayed, either in new page or modal popup.
  *                                       Default 'redirect'. Accepts 'redirect', 'modal'.
  *
+ * @unreleased Updates the default text color for the donate button, see #6591.
  * @since 2.21.2 change tag_background_color, progress_bar_color to official green color #69b868.
  *             change tag_text_color color to #333333.
  * @since 2.20.0 $show_donate_button Option to show donate button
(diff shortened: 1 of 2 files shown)


kjohnson:
@jdghinson I updated the text color for the Donate Button to inherit the color, instead of setting white text as the default (which is unreadable in a lot of themes).

JoshuaHungDinh (changes requested):
This works best for defined contrasted colors. I think we might want to consider a  dark grey or black default. If the parent container of the shortcode happens to be white or any really light color it inherits it's characteristics which results in something like this : 
Screen Shot 2022-10-20 at 3 20 35 PM
or in the case of white wont display at all.

On one hand any defined color will work amazingly and fit their theme very well. On the other hand it can become visibly challenging.

kjohnson:
@JoshuaHungDinh fair point. I went ahead and updated the default color to black, instead of trying to be clever with `inherit`.

JoshuaHungDinh (approved):

kjohnson:
Added to the QA Pipeline, see https://app.asana.com/0/1203193833018151/1203209111237825.

alexwcoleman (approved):

Issue #3907, opened 2018-12-18
fix: prevent error when importing JSON settings file


samsmith89:
# Bug Report

## User Story

As a site admin, I want to import the JSON settings file from another installation so that I can quickly get Give set up in another site.

## Current Behavior

When importing the JSON file from another site, I receive an error stating that the file is invalid.

## Expected Behavior

I expect to import the file without receiving an error message.

## Bug Type
- [ ] I am not sure whether this functionality ever worked as expected.

## Steps to Reproduce

1. Export a JSON file from another Give installation
2. In an empty site attempt to import the JSON settings file
3. See that an error is thrown stating the file is invalid, and the import doesn't take place

## Possible Solution

![screen shot 2018-12-18 at 10 00 39 am](https://user-images.githubusercontent.com/20672471/50162675-eae4bb80-02ac-11e9-93c6-c316847d05f7.png)

## Related

https://secure.helpscout.net/conversation/734568449/31321

## Acceptance Criteria

- [x] I can import a settings JSON file without receiving an error 
- [x] The import goes through without issue

## Environment

<details>
    <summary>Operating System</summary>
    <ul>
        <li>Platform: Mac OS X</li>
    </ul>
</details>

<details>
    <summary>Browser</summary>
    <ul>
        <li>Name: Chrome</li>
    </ul>
</details>

<details>
    <summary>WordPress System Info</summary>
### WordPress Environment ###

Home URL: http://destroy.local
Site URL: http://destroy.local
WP Version: 4.9.9
WP Multisite: –
WP Memory Limit: 256 MB
WP Debug Mode: –
WP Cron: ✔
Language: en
Permalink Structure: /%postname%/
Show on Front: page
Page on Front: Home – Movement For A Peoples Party (#9)
Page for Posts: Blog (#84)
Table Prefix Length: wp_
Table Prefix Length: 3
Table Prefix Status: Acceptable
Admin AJAX: Accessible
Registered Post Statuses: publish, future, draft, pending, private, trash, auto-draft, inherit, request-pending, request-confirmed, request-failed, request-completed, refunded, failed, revoked, cancelled, abandoned, processing, preapproval

### Server Environment ###

Hosting Provider: DBH: localhost, SRV: destroy.local
TLS Connection: Connection uses TLS 1.2
TLS Connection: Probably Okay
Server Info: Apache/2.4.10 (Debian)
PHP Version: 7.0.3
PHP Post Max Size: 1,000 MB
PHP Time Limit: 900
[...]

PR #6596, opened 2022-10-27
Fix Import GiveWP settings errors

Genevieve-K:
There is often an error while importing GiveWP settings:  [error] => 'Sorry, you are not allowed to upload this file type'. I sent an email in January 2019 and Sam suggested using the plugin called Disable Real Mime Check. But this plugin can be dangerous as it allows the upload of any file type. So I investigated. I saw there is 2 passes in the wp_check_filetype_and_ext() function, the first with mime type 'application/json', the second with 'text/plain'. So I suggest you to add 'text/plain' at the authorized mime types.

### DIFF
--- a/includes/admin/tools/import/class-give-import-core-settings.php
+++ b/includes/admin/tools/import/class-give-import-core-settings.php
@@ -485,6 +485,7 @@ if ( ! class_exists( 'Give_Import_Core_Settings' ) ) {
 		public static function json_upload_mimes( $existing_mimes = array() ) {
 
 			$existing_mimes['json'] = 'application/json';
+            $existing_mimes['text'] = 'text/plain';
 
 			return $existing_mimes;
 		}


JoshuaHungDinh:
> There is often an error while importing GiveWP settings: [error] => 'Sorry, you are not allowed to upload this file type'. I sent an email in January 2019 and Sam suggested using the plugin called Disable Real Mime Check. But this plugin can be dangerous as it allows the upload of any file type. So I investigated. I saw there is 2 passes in the wp_check_filetype_and_ext() function, the first with mime type 'application/json', the second with 'text/plain'. So I suggest you to add 'text/plain' at the authorized mime types.

Hi @Genevieve-K, I just wanted to reach out and let you know that we have received this pull request as well. I've included it with your previous PR information for the team. Thanks!

JoshuaHungDinh (changes requested):
Hi @Genevieve-K, looks good! Thank you for updating your PR. While user testing I was unable to select .text files for upload. Due to this I might suggest including .text files in the input fields "accept" attribute.

https://github.com/impress-org/givewp/blob/9697a7864851a7655d05be1a83332c1e8ca7abf3/includes/admin/tools/import/class-give-import-core-settings.php#L366

JoshuaHungDinh:
Hi @Genevieve-K , I had spoken with some of our team members about your PR. Its been brought to my attention that the code snippet being modified was introduced as a workaround for a MIME type bug in WP core 4.9.9/5.0.1. Please refer to https://github.com/impress-org/givewp/issues/3907. Considering the functionality is expecting JSON there is some reservation in loosening those restrictions even further. This will require additional time as we work through the underlying issue.

PR #6605, opened 2022-11-11
Enhancement: New Validation framework

JasonTheAdams:
### Overview
As we have built Next Gen, we've run into pain points with how validation works, namely:
- Tracking validation settings is tightly coupled to the Field API and tedious to add new validations
- Validations are tracked as traits, so they must be tied directly to the Field class, which is inflexible
- The actual act of validating the data isn't handled and is left up to the request handler (as such it's not portable)

On top of that, validation happens _all the time!_ Literally every form in every part of GiveWP (settings, endpoints, etc.) should be validated. So having a proper system for easily defining validation would be a big time saver.

When I first started making this it was directly coupled to the Field API, but @jonwaldstein pointed out that it could be easily made agnostic. As such, it _can_ be seamlessly used in the Field API, however it can just as easily be used elsewhere.

#### Rules
First stop are the rules themselves. There are two types of rules:
- ValidationRule classes
- Rule closures

The validation rule classes must implement the `ValidationRule` interface, which is a self-guided tour if you read the docs. The meat though is the `ValidationRule::__invoke()` method, which will has the same signature as the closures. So here's what that might look like:

```php
public function __invoke($value, Closure $fail, string $key, array $values) {
   if ( $value !== 'awesome') {
    $fail('{field} must be awesome');
  }
}
```

Parameters:
- `$value`: The value to be validated; can be anything
- `$fail`: A closure to call if the validation fails. Pass a string as the message to the end-user and use {field} as the placeholder for the field label
- `$key`: The key of the value in the values array, usually also the field name
- `$values`: All of the values. This is useful for rules like required, which checks to see if the value exists, or "confirm password" which compares the value with another value in the array.

**Front-End Rules**
Some rules can't be used on the front end. If, for example, a rule checked to see if an email already exists in the user table, the front-end obviously can't do that. But the "Required" rule obviously can. If a Rule implements the `ValidatesOnFrontEnd` interface then it will be serialized and included in the rules sent to the front-end. The front-end validation adapter will have to support the rule, but that's the job of another framework.

**Sanitizing Rules**
It's a bit of a gray area, but validation rules can actually perform sanitization. It's recommended to keep it simple at this time, but it sure is useful. The "integer" validation rule, for example, validates that the given value is in fact an integer, and then casts the value to an int before moving it along. To give a Validation Rule the ability to sanitize, implement the `Sanitizer` interface.

#### ValidationRuleSet

This is the class which stores all of the rules. It has "array" in the name as it can be iterated over like an array using something like a normal `foreach`. It has a bunch of functions for adding, removing, and retrieving rules. It's also smart enough to validate rules and take rules in multiple forms. Here's an example of rules being passed as either a string, validation class, or closure:

```php
$validations->rules('required', 'integer', 'min:1', new Max(10000), function($value, $fail) {
  if ( $value === 9999 ) {
    $fail('Come on, just give {field} one more');
  }
});
```

Notice that the Closure is only using two of the parameters, which is fine. You can ignore the last two parameters if you don't need them in Closure Rules.

#### HasValidationRules

This handy trait just makes it easy to get up and running with validation rules. Use this trait and a `$validationRules` property will be added to the class, along with methods for safely interacting with the rules array. As an example, this trait is being used on the `Field` class in the Field API.

#### Validator

Now we come to the point. We obviously want to actually run all these validation rules. For that we use the Validator class. Here's a framework agnostic way of doing it:

```php
$values = ['name' => 'Jason', 'age' => 36];
$labels = ['name' => 'Name', 'age' => 'Age'];
$rules = [
  'name' => give(ValidationRuleSet)->rules('required'),
  'age' => give(ValidationRuleSet)->rules('required', 'min:18', 'max:150'),
  'height' => give(ValidationRuleSet)->rules('required'),
];

$validator = new Validator($rules, $values, $labels);

// will fail because the height rule is required but the value is missing
if ( $validator->fails() ) {
  return $validator->errors();
} else {
  return $validator->validated(); // returns the validated, sanitized values
}
```

As you can see it's pretty darn easy to use all by itself! But it's also really easy to use it with the Field API:
```php
$form = new Form('profile');

$form->append(
  Section::make('personal')->append(
    Text::make('name')->rules('required'),
    Text::make('age')->rules('required', 'min:18', 'max:150'),
  )
);

$validator = Validator::fromForm($form, $values);
```

It's also possible to just pass an array of rules (or combination thereof) to the Validator and it will convert the arrays for you:
```php
$validator = new Validator($values, [
  'name' => ['required'],
  'age' => ['required', 'min:18', 'max:150']
]);
```

Finally, the last parameter of the `Validator` constructor is a `$labels` array. When creating an error, the validator will use the corresponding label if one exists. If there is no label for a value, then the key is fallen back on.

## Affects

The existing Field API Field and validations were updated to use the new system

## Testing Instructions

This is largely unit tested. This does affect fields being added by Form Field Manager, so we should mostly test that to make sure we didn't that!

## Pre-review Checklist

-   [x] Includes unit tests


JasonTheAdams:
Ok, @jonwaldstein, this bad boy is officially ready for review. Just a cool 3000 lines of code. So... have fun with that. 😁

assets/src/css/admin/sale-banners.scss
```diff
@@ -11,12 +11,6 @@
     row-gap: 1.875rem;
     font-family: 'Open Sans', system-ui, sans-serif;
 
-    /* Settings page specific styles */
-    .give-settings-page & {
-        margin-top: 2rem;
-        margin-bottom: 3.25rem;
-    }
-
     #give-in-plugin-upsells & {
```
jonwaldstein, line 14: Was this css change intentional?

src/Framework/FieldsAPI/Concerns/HasMaxLength.php
```diff
@@ -2,39 +2,43 @@
...
-     *
-     * @return $this
      */
-    public function maxLength($maxLength)
+    public function maxLength(int $maxLength): self
     {
-        $this->validationRules->rule('maxLength', $maxLength);
+        if ( $this->validationRules->hasRule('max') ) {
+            $this->validationRules->removeRuleWithId('max');
+        }
```
jonwaldstein, line 23: Interesting 🤔

src/Framework/Validation/Rules/Currency.php
```diff
@@ -0,0 +1,235 @@
...
+     * @unreleased
+     *
+     * @return string[]
+     */
+    public static function currencyCodes(): array
+    {
+        static $codes = null;
+
+        if ($codes === null) {
+            $codes = [
```
jonwaldstein, line 43: Out of curiosity where did you get this list from?  The first thing I noticed is it's not alphabetical 😅

src/Framework/Validation/Rules/Max.php
```diff
@@ -0,0 +1,98 @@
...
+     * @unreleased
+     */
+    public function __construct(int $size)
+    {
+        if ($size <= 0) {
+            throw new InvalidArgumentException('Max validation rule requires a non-negative value');
+        }
+
+        $this->size = $size;
+    }
```
jonwaldstein, line 57: I noticed _some_ rules are using a `__construct` and it appears because they store some values internally.  Should we add some kind of contract to enforce the construct for these types of rules?

src/Framework/Validation/Rules/Currency.php
JasonTheAdams: Internet. 😂 😂

src/Framework/FieldsAPI/Concerns/HasMaxLength.php
JasonTheAdams: Hmm. I guess rules with options, such as Size, Max, and Min, should probably have a way of being retrieved and modified. If someone wants to grab a field and modify a rule, it should be possible without having to replace the rule entirely (and changing the rule order). Would you agree, @jonwaldstein?

src/Framework/Validation/Rules/Max.php
JasonTheAdams: I thought about this and didn't see any benefit in constraining the constructors. They're never statically instantiated. I like the flexibility until we otherwise have a solid reason to reduce it.

assets/src/css/admin/sale-banners.scss
JasonTheAdams: No. I meant to put it on `develop`, which I did cherry-pick later. So just ignore this.

src/Framework/FieldsAPI/Field.php
```diff
@@ -24,26 +24,24 @@ abstract class Field implements Node
     use Concerns\Macroable;
     use Concerns\SerializeAsJson;
     use Concerns\TapNode;
-
-    /** @var ValidationRules */
-    protected $validationRules;
+    use HasValidationRules {
+        HasValidationRules::__construct as private __validationRulesConstruct;
+    }
```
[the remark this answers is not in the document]
JasonTheAdams: The `HasValidationRules` trait has a constructor to instantiate the `ValidationRuleSet` when the class is instantiated. In the case of the `Field` class, the constructor is already being used and has a different signature. So the trait constructor is being renamed and called manually in the `Field` constructor. It's a nifty trait trick.

src/Framework/FieldsAPI/Concerns/HasMaxLength.php
jonwaldstein: Yeah I would agree with that.

src/Framework/Validation/ValidationRulesRegister.php
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Give\Framework\Validation;
+
+use Give\Framework\Exceptions\Primitives\InvalidArgumentException;
+use Give\Framework\Validation\Contracts\ValidationRule;
+
+/**
+ * @unreleased
+ */
+class ValidationRulesRegister
```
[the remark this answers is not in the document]
JasonTheAdams: Resolved in [f82ce0a](https://github.com/impress-org/givewp/pull/6605/commits/f82ce0a0e4ab98a61c05e9bad26186afaf7a4b5a)

src/Framework/FieldsAPI/Concerns/HasMaxLength.php
JasonTheAdams: Resolved in 19cb33527

src/Framework/Validation/Rules/Currency.php
JasonTheAdams: Better? 😁 bac55aea9

JasonTheAdams:
@jonwaldstein I just made the changes. I also removed the `Validator::fromForm()` method and instead made it a function. After some thought, I realized both packages were being coupled to each other, which is unnecessary. The Field API depends on the Validation library, but it really doesn't need to be the other way around (especially since I plan on pulling this out).

jonwaldstein (approved):
It's beautiful 😍  🌈

PR #6614, opened 2022-11-22
Epic/form UI column sorting

pauloiankoski:
## Description
The current PR mainly adds the ability to sort columns of the admin list tables.

Under the hood, it does the following:
- implements dynamic table lists that read the available columns from a global list and renders the content for each column using Interweave JS;
- implements ListTable classes containing ModelColumn instances that are filled once receiving a model;
- changes the API endpoints to use the ModelQueryBuilder instance that will be passed to the ListTable class in order to return formatted HTML for each column instead of RAW data so that we can easily extend the responded columns without changing the frontend;
- adds support to order by some of the columns that are pre-defined.

## Affects
Admin List Tables

## Visuals

https://user-images.githubusercontent.com/75056371/204906931-14ca921c-9cd1-4732-b5b6-194352d4603d.mov

https://user-images.githubusercontent.com/75056371/204907006-7abcb3d8-6596-43b7-bc45-9e1124933957.mov

## Pre-review Checklist

-   [x] Includes unit tests

---
  - https://app.asana.com/0/0/1203468388185982


Benunc:
I've left feedback on the Asana task for this that needs to be addressed, and I'm happy to continue the discussion there or in Slack.

rickalday:
The only thing that's not working on my tests is sorting by donations in the Donors page.  It throws some errors in the console.
This video explains what's going on: https://somup.com/c0VnYpwNmj

JasonTheAdams:
@pauloiankoski Any updates on this?

pauloiankoski:
@JasonTheAdams finally had the test issues solved. There were some tests needing updates after some changes we did to the DB attachMeta and having clauses earlier on this epic.

In case you wanna see what has been changed since the last code review, this is the commits range:
https://github.com/impress-org/givewp/pull/6614/files/beb6261f5460cc7548cadfcb380428316ac2e2ec..d93436b097634f2c09e8795d9195272ba84c2343

@rickalday [here is the updated zip](https://github.com/impress-org/givewp/actions/runs/3843456237) with a fix for that 500 error when sorting by donations count. Regarding the Donations Form column being sorted by id, that's how it works on the legacy table list. Due to the current data structure, that's what we can do for now.

kjohnson:
@pauloiankoski I ran into an issue where a form did not have the `_give_price_option` meta defined, which caused a fatal error. This is probably not common, but is an unrecoverable error.

Specifically, `\Give\DonationForms\DataTransferObjects\DonationFormQueryData::getDonationFormLevels` assumes a non-multi price option if the value is not set, in which case a fatal error is thrown because `DonationFormLevel::fromPrice` expects a `string`.

@JasonTheAdams should we establish a fallback, or is this considered breaking intentionally?

JasonTheAdams:
@kjohnson Can you elaborate on the circumstances that led to the meta missing?

rickalday (approved):
Passed QA tests

kjohnson:
> `\Give\DonationForms\DataTransferObjects\DonationFormQueryData::getDonationFormLevels `assumes a non-`multi` price option if the value is not set, in which case a fatal error is thrown because `DonationFormLevel::fromPrice()` expects a string.

@JasonTheAdams I came across this assumption when testing backwards compatibility of Next Gen Forms with the updated Forms Table UI. The Next Gen forms did not have the  `_give_price_option` which caused a fatal error.

My concern is that a single corrupted donation form (missing meta key/value) brings down the Forms Table in such a way that is not recoverable by the user.

JasonTheAdams:
@kjohnson I think I'm fine to proceed, then. Unless there's some specific scenario I'm not understanding, the same argument could be made for any other missing data that's assumed to exist. Next Gen is a special circumstance, wherein the forms themselves are being rebuilt, so it makes sense to add some legacy meta — as opposed to a legacy form that, under some legit circumstance, may not have this meta.

Make sense?

kjohnson:
@JasonTheAdams yeah - that makes sense. I don't think it is a hard stop, but did want to mention it since I ran across it.

I do not have any examples of this being an issue that occurs "naturally" during normal use of the legacy forms.

kjohnson (approved):
I was able to test this with the backwards compatibility updates in Next Gen and everything works as expected.

PR #6631, opened 2022-12-08
Refactor/speed up endpoint performance

pauloiankoski:
## Description

Optimize list table endpoints performance replacing expensive queries with simple ones.

## Affects

- On `DonationsAdminPage.php`, `DonorsAdminPage.php` and `SubscriptionsAdminPage.php` there is a method `getForms` that was triggering a request to the ListDonationForms endpoint what needed to load the entire model. We are replacing that request with a simple DB query returning only the necessary columns.
- When calling the method `count()` in a query it used to just append `COUNT(*) as count` to the query then get that value from the results. We are now removing any other column from the `SELECT` statement so that the query only results the count column.
- On every list table endpoint we have a counter method that returns the total number of items based on the current arguments so that we calculate paginations. It used to make use of the `prepareQuery()` method but that attaches many extra meta data not necessary for the counting method. We are replacing the `prepareQuery()` call with a custom `ModelQueryBuilder` where only the necessary meta data for that endpoint is now being attached.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/DonationForms/Endpoints/ListDonationForms.php
+++ b/src/DonationForms/Endpoints/ListDonationForms.php
@@ -3,6 +3,7 @@
 namespace Give\DonationForms\Endpoints;
 
 use Give\DonationForms\ListTable\DonationFormsListTable;
+use Give\DonationForms\Models\DonationForm;
 use Give\Framework\Models\ModelQueryBuilder;
 use WP_REST_Request;
 use WP_REST_Response;
@@ -176,7 +177,11 @@ class ListDonationForms extends Endpoint
      */
     public function getTotalFormsCount(): int
     {
-        $query = give()->donationForms->prepareQuery();
+        $builder = new ModelQueryBuilder(DonationForm::class);
+
+        $query = $builder->from('posts')
+            ->where('post_type', 'give_forms');
+
         $query = $this->getWhereConditions($query);
 
         return $query->count();
--- a/src/Donations/Endpoints/ListDonations.php
+++ b/src/Donations/Endpoints/ListDonations.php
@@ -3,6 +3,8 @@
 namespace Give\Donations\Endpoints;
 
 use Give\Donations\ListTable\DonationsListTable;
+use Give\Donations\Models\Donation;
+use Give\Donations\ValueObjects\DonationMetaKeys;
 use Give\Donations\ValueObjects\DonationMode;
 use Give\Framework\ListTable\Exceptions\ColumnIdCollisionException;
 use Give\Framework\Models\ModelQueryBuilder;
@@ -189,7 +191,26 @@ class ListDonations extends Endpoint
      */
     public function getTotalDonationsCount(): int
     {
-        $query = give()->donations->prepareQuery();
+        $builder = new ModelQueryBuilder(Donation::class);
+
+        $columnsForAttachMetaQuery = [
+            DonationMetaKeys::EMAIL(),
+            DonationMetaKeys::FIRST_NAME(),
+            DonationMetaKeys::LAST_NAME(),
+            DonationMetaKeys::DONOR_ID(),
+            DonationMetaKeys::FORM_ID(),
+            DonationMetaKeys::MODE(),
+        ];
+
+        $query = $builder->from('posts')
+            ->attachMeta(
+                'give_donationmeta',
+                'ID',
+                'donation_id',
+                ...DonationMetaKeys::getColumnsForAttachMetaQueryFromArray($columnsForAttachMetaQuery)
+            )
+            ->where('post_type', 'give_payment');
+
         $query = $this->getWhereConditions($query);
 
         return $query->count();
--- a/src/Donors/Endpoints/ListDonors.php
+++ b/src/Donors/Endpoints/ListDonors.php
@@ -3,6 +3,7 @@
 namespace Give\Donors\Endpoints;
 
 use Give\Donors\ListTable\DonorsListTable;
+use Give\Donors\Models\Donor;
 use Give\Framework\Models\ModelQueryBuilder;
 use Give\Framework\QueryBuilder\QueryBuilder;
 use WP_REST_Request;
@@ -181,7 +182,9 @@ class ListDonors extends Endpoint
      */
     public function getTotalDonorsCount(): int
     {
-        $query = give()->donors->prepareQuery();
+        $builder = new ModelQueryBuilder(Donor::class);
+
+        $query = $builder->from('give_donors');
         $query = $this->getWhereConditions($query);
 
         return $query->count();
--- a/src/Subscriptions/Endpoints/ListSubscriptions.php
+++ b/src/Subscriptions/Endpoints/ListSubscriptions.php
@@ -2,6 +2,8 @@
 
 namespace Give\Subscriptions\Endpoints;
 
+use Give\Donations\Models\Donation;
+use Give\Donations\ValueObjects\DonationMetaKeys;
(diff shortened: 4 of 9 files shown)


src/DonationForms/Endpoints/ListDonationForms.php
```diff
@@ -176,7 +177,11 @@ public function getForms(): array
      */
     public function getTotalFormsCount(): int
     {
-        $query = give()->donationForms->prepareQuery();
+        $builder = new ModelQueryBuilder(DonationForm::class);
```
jonwaldstein, line 180: If you're not using the models for these, then you should not be using the `ModelQueryBuilder`.  Instead you should just use the regular `QueryBuilder` like this:

`DB::table('posts')->where()`

src/Donations/Endpoints/ListDonations.php
```diff
@@ -189,7 +191,26 @@ public function getDonations(): array
      */
     public function getTotalDonationsCount(): int
     {
-        $query = give()->donations->prepareQuery();
+        $builder = new ModelQueryBuilder(Donation::class);
```
jonwaldstein, line 194: Same comment as before about not using the `ModelQueryBuilder`

src/Donors/Endpoints/ListDonors.php
```diff
@@ -181,7 +182,9 @@ public function getDonors(): array
      */
     public function getTotalDonorsCount(): int
     {
-        $query = give()->donors->prepareQuery();
+        $builder = new ModelQueryBuilder(Donor::class);
```
jonwaldstein, line 185: Same comment as before about not using the `ModelQueryBuilder`

src/Subscriptions/Endpoints/ListSubscriptions.php
```diff
@@ -184,7 +186,9 @@ public function getSubscriptions(): array
      */
     public function getTotalSubscriptionsCount(): int
     {
-        $query = give()->subscriptions->prepareQuery();
+        $builder = new ModelQueryBuilder(Donation::class);
```
jonwaldstein, line 189: Same comment here about not using the `ModelQueryBuilder`

jonwaldstein (changes requested):
@pauloiankoski nice optimization, I just pointed out places that you don't need to use the `ModelQueryBuilder` and instead should just use `QueryBuilder` or `DB::table()`.  The `ModelQueryBuilder`'s main purpose is to return a model instance when you call `get()` or `getAll()`.  Since your doing manual queries and calling `count()` in the end, there would be no reason to use the `ModelQueryBuilder`.

src/DonationForms/Endpoints/ListDonationForms.php
pauloiankoski: Resolved: e13dd57ce3cc1184463bf52711077f0c3b212542

src/Donations/Endpoints/ListDonations.php
pauloiankoski: Resolved: e13dd57ce3cc1184463bf52711077f0c3b212542

src/Donors/Endpoints/ListDonors.php
pauloiankoski: Resolved: e13dd57ce3cc1184463bf52711077f0c3b212542

jonwaldstein (approved):
Great job man!

JasonTheAdams:
@jonwaldstein @pauloiankoski I'll be honest, this PR actually makes me a little nervous. The point of making the Model-oriented list tables was to further decouple (albeit not 100%) the list tables from the database structure. The bit in here for clearing the selects on `count()` is fine, but in other places we're coupling ourselves right back to the database.

jonwaldstein:
@JasonTheAdams from what I understand, the main thing is not using the models for getting total counts specifically.  I'm fine with that optimization as we don't need to join a bunch of tables just to get the total records number.  Maybe some of these simple queries can just be moved into the repository itself so they're not called directly in the list tables.

JasonTheAdams:
@jonwaldstein @pauloiankoski Did you explore using [SQL_CALC_FOUND_ROWS and FOUND_ROWS()](https://dev.mysql.com/doc/refman/5.7/en/information-functions.html#function_found-rows) so we don't have to reconstruct the query for the count? It seems like the count is just being used to calculate the total rows for the query?

pauloiankoski:
@JasonTheAdams According to the [most recent docs](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_found-rows), I believe we are in line with the most recent recommendations:

> The SQL_CALC_FOUND_ROWS query modifier and accompanying [FOUND_ROWS()](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_found-rows) function are deprecated as of MySQL 8.0.17; expect them to be removed in a future version of MySQL. As a replacement, considering executing your query with LIMIT, and then a second query with [COUNT(*)](https://dev.mysql.com/doc/refman/8.0/en/aggregate-functions.html#function_count) and without LIMIT to determine whether there are additional rows.

@jonwaldstein At first, I had this idea of adding a parameter to the `prepareQuery` method, something like `$columnsForAttachedMetaQuery` where we could modify the metadata being attached. The selects would be removed on the `count()` method as they already are now.

jonwaldstein:
@pauloiankoski thanks, we wouldn't want to do that though because those `prepareQuery` methods on the repository are meant to prepare the query to be used specifically as a model and not be overwritten for any other reason.  

However I was thinking we could just have some simple methods on the repositories that do counts like:
`give()->donationForms->getTotalCount($query)` or even another option I just thought of, is we could have a modified `count()` method in the `ModelQueryBuilder` that knows how to strip out everything it doesn't need to run a simple count statement.

pauloiankoski:
@jonwaldstein As I have to check some conditions related to the filters we have on the tables, I'd need to have an instance of `QueryBuilder` back in that method we are using to count.

Detach the meta from `prepareQuery` would be an option but IHMO, it's a bigger effort undo things than to create a new query. Then, the other solution that came up to me was to have a `initQuery` method that does just the basic: start the `queryBuilder` object and set the minimum requirements for that domain, like `table` and `post_type` where necessary. The `prepareQuery` would use that method as the starter point to add the other clauses like `select` and `attachMeta`.

I'd be able to use that function for the counting. It's an empty query builder where I can easily run the count method. PS: As a good developer, naming functions is not one of my strengths.

I would love to also hear your thoughts @JasonTheAdams.

CleanShot 2022-12-08 at 19 35 06

JasonTheAdams:
@pauloiankoski Out of curiosity, for the count query, if all we do is remove all other selects upon `count()` and just stick with the exact same query as it otherwise being used, where does that put us in terms of time? I know the joins do slow things down, but it seemed like the major time consumer was the what-having stuff we found yesterday.

pauloiankoski:
@JasonTheAdams After a long battle against the system over the last couple of days, I was finally able to run the tests.

I ran the queries 3 times for each amount and tabulated the median value (Y axis in log scale):

CleanShot 2022-12-11 at 16 37 49


This is not science but I do believe the results are meaningful. What I can read from the results are:
- Running on the local environment is much faster than on any server so exact numbers should not be considered but what they represent to the whole;
- Using `having` on the main query reduces the query time by 60% on average;
- The main query as limiting results to 30 took a stable amount of time for all queries;
- The counting query consistently doubles as the number of rows in the database also doubles. I can see a 1:1 ratio there;
- Restricting the amount of `left join` on the counting query reduces the query time by 75% on average;
- Using `having` on the counting query slightly increases the query time;

PR #6632, opened 2022-12-08
Available font weights for Open Sans is normalized

kjohnson:
## Description

This PR normalizes the available font weights for Open Sans (via fonts.google.com) to include a range of `400` -` 700`.

Previously, font weights were included piecemeal. This means that the full range of font weights were not available to designers, depending on the admin screen.

**Usage**: `wp_enqueue_style('givewp-admin-fonts');`

## Testing Instructions

Nothing should be visibly different.


JasonTheAdams:
@kjohnson If we're cleaning this up and want it to be uniform, then we could use `wp_register_style` in once place, then reference that tag in all the places its being enqueued. Then we wouldn't have the redundancy, and if someone adds a weight in the future its done everywhere.

kjohnson:
@JasonTheAdams good idea.

kjohnson:
@JasonTheAdams I've updated this PR to register a single `givewp-admin-fonts` styles that can be enqueued where needed.

give.php
```diff
@@ -234,6 +234,8 @@ public function boot()
 
         add_action('plugins_loaded', [$this, 'init'], 0);
 
+        add_action('admin_enqueue_scripts', [$this, 'enqueueAdminScripts'], -1);
+
```
jonwaldstein, line 238: @kjohnson can we move this into a service provider?  I'd like to avoid adding to the main give file since we have the service providers for all our entry points.

kjohnson: @jonwaldstein resolved, see https://github.com/impress-org/givewp/pull/6632/commits/a2a76fbe10ded6b01f0515f210dcf2e8125b8afb.

jonwaldstein (approved):
Great job!

PR #6639, opened 2022-12-14
Refactor/list table columns for addons

pauloiankoski:
## Description

In order to add columns to add-ons, these changes have been necessary:
- renamed the method that return columns from a List Table as it was conflicting with the method from the Columns concern. That means once we were trying to get all columns added to the list table instance, we were actually getting the columns registered on the specific list table;
- added self return to the methods that we use to add/remove columns from the list table making them chainable;
- added a new parameter to the `setColumnVisibility` method so we can invoke it, passing a value that is different from its default;

## Pre-review Checklist

-   [x] Includes unit tests


jonwaldstein (approved):
@pauloiankoski looks good to me 👍

PR #6645, opened 2022-12-16
Hotfix: makes the Square credit card fields be rendered in Safari

### LINKED ISSUE #6644: The gateway selector list should prevent a Safari bug that does not automatically set the selected item as the active element
## User Story

When a user clicks on an item of the gateway list on the payment form, the item selected gets automatically set as active in most browsers. But on Safari this behavior does not happen and as a result, when we try to use the `event.currentTarget.activeElement` to get the active item, we receive the `body` element as the active item instead of the item selected in the list. 

## Details

To check more details about this problem, check this link: https://stackoverflow.com/q/41035743

## Additional Context

As a sample, [check this Square addon method](https://github.com/impress-org/give-square/blob/develop/assets/src/js/frontend/give-square-frontend.js#L83-L94) that is trying to use `event.currentTarget.activeElement` and because of this problem instead of loading the gateway data, it shows an error in the console and does not render the credit card fields when the user is using Safari.

image

## Acceptance Criteria

-   [ ] When a user clicks on an item in the gateway list, the item selected gets automatically set as the active element.

glaubersilva:
Resolves #6644

## Description

This PR forces the focus on the selected item on the gateways list from the payment form and ensures that this selected item is set as the active element in the DOM.

## Affects

Makes possible to render Square credit card fields on the Safari browser.

## Visuals

https://user-images.githubusercontent.com/16308864/208195749-1e38d546-a03f-4fc8-a415-00093d4b8c96.mp4

## Testing Instructions

1. Enable Square and any other gateway.
2. Open the form in Safari and switch to Square. If Square is the default gateway, switch to any other gateway and back to Square.

**Important:** If necessary, use a service like [Lambda Test](https://www.lambdatest.com/) to make the test.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1203579167893822

### DIFF
--- a/assets/src/js/frontend/give-donations.js
+++ b/assets/src/js/frontend/give-donations.js
@@ -242,6 +242,7 @@ jQuery( function( $ ) {
 			// Select payment gateway.
 			$( this ).prop( 'checked', true );
 			$( this ).parent().addClass( 'give-gateway-option-selected' );
+			$( this ).focus();
 
 			// Get new payment gateway.
 			new_payment_gateways = Give.form.fn.getGateway( $form );


JoshuaHungDinh (approved):
Interesting safari bug.. Thanks for linking the post in the issue. This worked in my tests.

glaubersilva:
Thanks @JoshuaHungDinh ! I already sent it to QA. =)

rickalday (approved):
Passed manual tests

JasonTheAdams:
Released in 2.25.0. It didn't show up here as I merged it via squash locally.

PR #6651, opened 2022-12-22
Fix: Classic form's payment amount summary reflects custom input amounts

### LINKED ISSUE #6650: Classic form's summary should reflect users custom input amount
## Details

When using the Classic Template and a donor manually enters a custom donation amount, the Payment Amount section in the donation summary does not update to reflect the donation amount.  

## Expected Behavior

Considering, the Donation Total section in the summary updates to reflect the total amount the payment amount should also.

## Steps to Reproduce

1. Create a classic form.
2. Select the any donation amount - view the summary at the bottom and notice the payment amount section updates.
3. Select "Custom amount" and input a custom amount.
4. View the summary at the bottom and you will notice that the payment amount section does not update.

## Visuals

https://user-images.githubusercontent.com/75056371/209043378-ebc7266c-ce79-4db9-a32b-f77a4f756773.mov

## Acceptance Criteria

-   [x] Payment Amount section in the donation summary reflects users custom input amount.

JoshuaHungDinh:
Resolves #6650

## Description

When changing the donation payment via the custom amount field (a user typed input), the data-amount on the element does not update until another event on the form occurs. Because the mutation observer is tracking changes to the elements attribute and the attribute does not immediately change the payment amount field is not updating until the following interaction.  I included an event listener for the input to update the payment amount field when changes are committed to the input.

## Affects
Donation Summary Payment Amount

## Visuals

https://user-images.githubusercontent.com/75056371/209046902-9b03b5f8-ee55-4e44-894e-b426990be44b.mov

## Testing Instructions

- View each form type and submit custom amounts and a donation set level amount for each form.
- Verify payment amounts in the donation summary update accordingly.
- See "Visual" for reference.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1203615080998862


kjohnson (changes requested):
@JoshuaHungDinh it looks like `Object.observe()` has since been deprecated, so we may want to go ahead and refactor out the `Object.observe()` and just use the `addEventListener`.

kjohnson (changes requested):
@JoshuaHungDinh I may have spoken too soon. Looking closer this uses a `MutationObserver` under the hood, which should be fine. I'm curious as to why it doesn't already catch the change event, though.

In any case, let's move this (redundant?) event listener outside of the `observe` callback and just add it inside of `initAmount()`.

JoshuaHungDinh:
> @JoshuaHungDinh I may have spoken too soon. Looking closer this uses a `MutationObserver` (not `Object.observe`) under the hood, which should be fine. I'm curious as to why it doesn't already catch the change event, though.
> 
> In any case, let's move this (redundant?) event listener outside of the `observe` callback and just add it inside of `initAmount()`.

It is odd, here are some similar issues that pointed me toward a solution:
 https://stackoverflow.com/questions/12048645/how-do-you-get-a-mutation-observer-to-detect-a-value-change-of-a-textarea
https://stackoverflow.com/questions/32383349/detect-input-value-change-with-mutationobserver

I've updated the PR. Thanks!

kjohnson (approved):
Thanks for the change. Let's get this to QA.

rickalday (approved):
Passed manual tests

PR #6655, opened 2022-12-23
Fix: donation outdate post cache

### LINKED ISSUE #6654: The donation repository should clear the post cache after having updated it
## User Story

When the [update method](https://github.com/impress-org/givewp/blob/develop/src/Donations/Repositories/DonationRepository.php#L242-L280) from the donation repository is used, the cache for the post updated is not cleared and it can generate inconsistent results when the WP native functions that rely on this post cache are used.

## Expected Behavior

After updating a donation, the post cache related to it should be cleared.

## Steps to Reproduce

1. Update a donation status from `'pending'` to `'publish'` using model functions like `$donation->save()`;
2. Retrieve the donation status using the `get_post_status()` function native from WP and you'll see that the status still will be `'pending'` instead of `'publish'`.
## Additional Context

This problem is preventing the [give_is_payment_complete()](https://github.com/impress-org/givewp/blob/develop/includes/payments/functions.php#L721-L739) function from the Give core to return the correct value and this is directly affecting the [is_receipt_link_allowed()](https://github.com/impress-org/give-pdf-receipts/blob/develop/includes/class-give-pdf-receipts-engine.php#L377-L386) function from the PDF recipes addon that relies on the return of the first function mentioned. As result, this bug is preventing the "Download Receipt" link to be available in the e-mail sent to the donors.

## Visuals

pdf-receipts-side-effects

pdf-receipts-side-effects-2

## Acceptance Criteria

-   [ ] The update method from the donation repository cleans the post cache after updating it

glaubersilva:
Resolves #6654

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Donations/Repositories/DonationRepository.php
+++ b/src/Donations/Repositories/DonationRepository.php
@@ -276,6 +276,8 @@ class DonationRepository
 
         DB::query('COMMIT');
 
+        clean_post_cache($donation->id);
+
         Hooks::doAction('givewp_donation_updated', $donation);
     }
 
--- a/src/Donations/Repositories/DonationRepository.php
+++ b/src/Donations/Repositories/DonationRepository.php
@@ -230,11 +230,12 @@ class DonationRepository
     }
 
     /**
-     * @since 2.23.1 Use give_update_meta() method to update entries on give_donationmeta table
-     * @since 2.23.0 retrieve the post_parent instead of relying on parentId property
-     * @since 2.21.0 replace actions with givewp_donation_updating and givewp_donation_updated
-     * @since 2.20.0 return void
-     * @since 2.19.6
+     * @unreleased Cleans the post cache after updating it
+     * @since      2.23.1 Use give_update_meta() method to update entries on give_donationmeta table
+     * @since      2.23.0 retrieve the post_parent instead of relying on parentId property
+     * @since      2.21.0 replace actions with givewp_donation_updating and givewp_donation_updated
+     * @since      2.20.0 return void
+     * @since      2.19.6
      *
      * @return void
      * @throws Exception|InvalidArgumentException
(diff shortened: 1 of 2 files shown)


src/Donations/Repositories/DonationRepository.php
```diff
@@ -276,6 +277,8 @@ public function update(Donation $donation)
 
         DB::query('COMMIT');
 
+        clean_post_cache($donation->id);
+
```
jonwaldstein, line 281: @glaubersilva instead of adding this directly here, can you please move this into a new legacy listener called `ClearDonationPostCache`.  In `src/Donations/ServiceProvider.php` you'll find where all the legacy listeners are being called from our new givewp hooks.  The idea is that we can use actions like `givewp_donation_updated` to support all backwards compatibility in an isolated way instead of adding directly to the repository methods.

tests/Unit/Donations/Repositories/TestDonationRepository.php
```diff
@@ -236,4 +237,31 @@ public function testFirstDonationLatestDonationSortOrder()
...
+    /**
+     * @unreleased
+     *
+     * @see https://github.com/impress-org/givewp/issues/6654
+     *
+     * @return void
+     *
+     * @throws Exception
+     */
+    public function testUpdateDonationShouldClearPostCache()
```
jonwaldstein, line 250: Love this matching unit test!!!

jonwaldstein (changes requested):

src/Donations/Repositories/DonationRepository.php
glaubersilva: I did some commits to implementing it. ;-)

jonwaldstein (approved):
Great job!

Benunc (approved):
This passes all automated tests, and appears to work on a few manual tests to download the PDF.

PR #6658, opened 2022-12-27
Feature: add refund checkbox to the framework

### LINKED ISSUE #6657: Gateway API should add a refund checkbox automatically on the donation admin screen
## User Story

Currently, when changing a donation status on the admin screen to "refund", the checkbox to refund it on the gateway side is not automatically displayed by the Framework. In fact, is necessary that each gateway implementation add the checkbox in the admin screen, which is not ideal - the ideal is to have this refund checkbox added automatically.

## Acceptance Criteria

-   [ ] The refund checkbox is displayed automatically on new gateway implementations

glaubersilva:
Resolves #6657

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayAdapter.php
+++ b/src/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayAdapter.php
@@ -170,4 +170,73 @@ class LegacyPaymentGatewayAdapter
 
         return $donor;
     }
+
+    /**
+     * @unreleased
+     */
+    public function addOptRefundCheckbox(int $donationId, PaymentGatewayInterface $registeredGateway)
+    {
+        $donation = Donation::find($donationId);
+        if ($donation->gatewayId === $registeredGateway::id()) {
+            ?>
+            <div id="give-gateway-opt-refund-wrap"
+                 class="give-gateway-opt-refund give-admin-box-inside give-hidden">
+                <p>
+                    <input type="checkbox" id="give-gateway-opt-refund" name="give_gateway_opt_refund" value="1" />
+                    <label for="give-gateway-opt-refund">
+                        <?php
+                        esc_html_e(sprintf('Refund the donation at %s?', $registeredGateway::id()), 'give');
+                        ?>
+                    </label>
+                </p>
+            </div>
+            <script>
+                const donationStatus = document.getElementById('give-payment-status');
+
+                if (!!donationStatus) {
+                    donationStatus.addEventListener('change', function (event) {
+                        const refundCheckbox = document.getElementById('give-gateway-opt-refund');
+
+                        if (null === refundCheckbox) {
+                            return;
+                        }
+
+                        refundCheckbox.checked = false;
+
+                        if ('refunded' === event.target.value) {
+                            document.getElementById('give-gateway-opt-refund-wrap').style.display = 'block';
+                        } else {
+                            document.getElementById('give-gateway-opt-refund-wrap').style.display = 'none';
+                        }
+                    });
+                }
+            </script>
+            <?php
+        }
+    }
+
+    /**
+     * @unreleased
+     */
+    public function maybeRefundOnGateway(
+        int $donationId,
+        string $newStatus,
+        string $oldStatus,
+        PaymentGatewayInterface $registeredGateway
+    ) {
+        $gatewayOptRefund = ! empty($_POST['give_gateway_opt_refund']) ? give_clean($_POST['give_gateway_opt_refund']) : '';
+        $canProcessRefund = ! empty($gatewayOptRefund) ? $gatewayOptRefund : false;
+
+        // Only move forward if refund requested.
+        if ( ! $canProcessRefund) {
+            return;
+        }
+
+        $donation = Donation::find($donationId);
+        if ($donation->gatewayId === $registeredGateway::id() &&
+            'refunded' === $newStatus &&
+            'refunded' !== $oldStatus) {
+            $registeredGateway->refundDonation($donation);
+        }
+    }
 }
--- a/src/Framework/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayRegisterAdapter.php
+++ b/src/Framework/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayRegisterAdapter.php
@@ -39,6 +39,30 @@ class LegacyPaymentGatewayRegisterAdapter
                 $legacyPaymentGatewayAdapter->handleBeforeGateway(give_clean($legacyPaymentData), $registeredGateway);
             }
         );
+
+        /**
+         * Remove all refund checkboxes added by the gateways with priority 11 and add a new refund checkbox with priority 12
+         */
+        remove_all_actions("give_view_donation_details_totals_after", 11);
+        add_action(
+            "give_view_donation_details_totals_after",
+            static function (int $donationId) use ($registeredGateway, $legacyPaymentGatewayAdapter) {
+                $legacyPaymentGatewayAdapter->addOptRefundCheckbox($donationId, $registeredGateway);
+            },
+            12
+        );
+        add_action(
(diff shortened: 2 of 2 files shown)


glaubersilva:
@jonwaldstein I liked the button idea, seems to be more intuitive from the UX perspective. About the `handleRefundDonation` I suppose that it would be used like this: `$registeredGateway->handleRefundDonation($donation);` instead of calling the `refundDonation` method directly [as I did here](https://github.com/impress-org/givewp/pull/6658/files#diff-f04cd0d24b8efa092be5aef9c1f4dabb3f58a25084f84ce8289b850fc30228c6R239), right?

jonwaldstein:
@glaubersilva yup exactly!  I think for now we can leave the ux the way it is since this page is going to be redesigned in the near future but we'll definitely revisit when that time comes.  Here's what i'm thinking to wrap this one up:

1. Let's add a method `handleRefundDonation` to `PaymentGateway` (similar to createPayment) with a try / catch.  The catch will handle any Exceptions.  Basically, If a refund did not work the gateway can just throw a `PaymentGatewayException` and our framework will deliver the message.
2. This method will handle the `PaymentRefunded` command (I just realized this exists) using the `PaymentRefundedHandler` (also exists lol).  This really just ensures the framework is setting the refunded status on the donation and redirecting to the appropriate location.

glaubersilva:
@jonwaldstein I did implement your suggestions and change the PR description to better explain how it works after the changes.

src/Framework/PaymentGateways/PaymentGateway.php
```diff
@@ -497,4 +547,17 @@ private function isFunctionImplementedInGatewayClass(string $methodName): bool
...
+     * Checks to see if the provided method is returning a specific command type as result.
+     * It always returns 'false' if the provided method doesn't explicitly declare the return type.
+     *
+     * @unreleased
+     */
+    private function isFunctionReturningCommand(string $methodName, PaymentCommand $command): bool
+    {
+        $reflector = new ReflectionMethod($this, $methodName);
+
+        return $reflector->getReturnType() ? $reflector->getReturnType()->getName() : '' === get_class($command);
```
jonwaldstein, line 561: @glaubersilva looks like we can't use `$reflector->getReturnType()->getName()` until php 7.1 (we support 7.0 right now).  My other comment might end up removing this anyway.

Screenshot 2023-02-23 at 4 26 06 PM

src/Framework/PaymentGateways/PaymentGateway.php
```diff
@@ -76,6 +78,21 @@ public function __construct(SubscriptionModule $subscriptionModule = null)
...
+     *             by default, it returns 'false' to prevent:
+     *             1) Duplication of refund checkboxes on gateways that already implement it;
+     *             2) Adding the refund checkbox on the admin screen in gateways with an empty refundDonation() method;
+     *             3) Overwriting legacy refund methods with empty refundDonation() methods.
+     *             4) Break gateways that implement the refundDonation() method but don't return the PaymentRefunded command as result
+     */
+    public function supportsRefund(): bool
+    {
+        return $this->isFunctionReturningCommand('refundDonation', (new PaymentRefunded()));
+    }
```
jonwaldstein, line 94: @glaubersilva thanks for really thinking through this! I love how you outlined all the scenarios - it's really helpful for me to understand what's going on.

I do wonder if we're being a little too strict on the implementation of a return type though, as some folks might be building gateways and not be used to doing that sort of thing in php. 

For that reason, (if possible) i'm thinking we should just add a `refundDonation()` to the main PaymentGateway class that just returns false and let `supportsRefund()` use `$this->isFunctionImplementedInGatewayClass('refundDonation')`.  I like the simplicity of saying that a gateway supports refunds if it's using the refundDonation method.

To prevent duplication maybe we can use some kind of combination of `doing_filter` and/or `doing_action` before proceeding in the legacy adapter.  Another options is in our legacy adapter in the js script we just check for the checkbox in the dom to make sure it doesnt' exist yet.

Open to your thoughts here.

jonwaldstein (changes requested):

JasonTheAdams:
@jonwaldstein @glaubersilva I would like to prioritize getting this resolved during cooldown, please.

src/Framework/PaymentGateways/PaymentGateway.php
glaubersilva: Thanks for the insights! I tried an approach with the `doing_action` but it seems not to be the most suitable for this situation, so I implemented a JS approach here: 1e486288916efdb506bee0d57655c257a44ad4d2

Basically what the changes implemented in the commit mentioned above are doing is: 

1. Ensure that the framework is the last one to register a callback in the `give_view_donation_details_totals_after` hook.
2. Before rendering the checkbox on the hook from the previous step, verify if there is just one hidden checkbox in the DOM's proper place, if there is more than one, means that the gateway already added it, so the framework will do nothing.

glaubersilva:
@jonwaldstein Thanks for the insights! I did implement your suggestions and left a comment explaining some details. I also changed the PR description again. Let me know if there is anything more to improve. =)

src/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayAdapter.php
```diff
@@ -178,4 +179,73 @@ private function getOrCreateDonor(
...
+            </div>
+            <script>
+                if (!!document.getElementById('give-payment-status') &&
+                    1 === document.querySelectorAll('div.give-admin-box > div.give-hidden input[type="checkbox"]').length
+                ) {
+                    document.getElementById('give-payment-status').addEventListener('change', function (event) {
+                        const refundCheckbox = document.getElementById('give-gateway-opt-refund');
+
+                        if (null === refundCheckbox) {
+                            return;
```
jonwaldstein, line 210: @glaubersilva wouldn't we need to hide this checkbox if one already exists?  We might want to be more specific about targeting the checkbox also like checking if the id attribute has the string `opt-refund` in it to make sure it's the right checkbox.

glaubersilva: Resolved in 1c9da987e4d771d2006e0489352deffc221ed8fb

About your question:

> wouldn't we need to hide this checkbox if one already exists?

That's not necessary because each checkbox adds its own script to handle the visibility and the framework default checkbox already starts hidden because it uses the give-hidden class.

src/Framework/PaymentGateways/PaymentGateway.php
```diff
@@ -123,6 +134,32 @@ public function handleCreatePayment(Donation $donation)
...
     }
 
+    /**
+     * @unreleased
+     */
+    public function handleRefundDonation(Donation $donation)
+    {
+        try {
+            $command = $this->refundDonation($donation);
+            $this->handleGatewayPaymentCommand($command, $donation);
```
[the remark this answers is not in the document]
glaubersilva: Resolved in c01e3140d80abece9fc8e34eb9f7d11ca963a086

glaubersilva:
@jonwaldstein I finally moved the existing code to the controller class, so now there is no more merge conflict. Let me know if I can merge this one or if there is anything more to adjust. Thanks! =)

jonwaldstein (approved):
@glaubersilva great job man, this looks good to me.  please send this to QA to test on a few different gateways.

jonwaldstein:
@glaubersilva one more thing, would you mind adding a test to `GatewayPaymentControllerTest` for the new refund method? You should be able to duplicate what's already there for the create method and swap a few things out to return the refunded command.

JasonTheAdams:
Ignore. Just testing Slack.

JasonTheAdams:
Testing again.

Benunc:
This has been tested and is good to merge. Will need to run all automated tests still. But it passes manual tests both with existing gateways and with the example gateway repo.

Benunc (approved):

PR #6660, opened 2023-01-02
Feature: add donation note type to donation note model

jonwaldstein:
## Description

I recently discovered we have two types of donation notes, one for admins and another for donors.  Currently, when you add a donation note for a donor - there is an action that hooks into an email notification sent to the donor you are adding a note for.  This PR adds the new `$donationNote->type` attribute with a legacy listener for the donor type.

## Affects

Donation notes

## Visuals

Using the Donation Note model programmatically using the Test Gateway:

Screen Shot 2023-01-02 at 4 40 25 PM

Screen Shot 2023-01-02 at 4 39 23 PM

## Testing Instructions

This would have to be done manually by using the model like this:

```
$donation = Donation::factory()->create();

DonationNote::create([
'type' => DonationNoteType::DONOR(),
'donationId' => $donation->id,
'content' => 'message to donor'
]);
```

## Pre-review Checklist

-   [x] Includes unit tests


jonwaldstein:
@JasonTheAdams can you review this when you get a chance?  Not urgent since the parity functionality is only used in the admin right now (as far as I can tell), but something I realized our `DonationNote` model was missing.

JasonTheAdams:
Good catch, @jonwaldstein! Just thinking through this, here. If the donation notes are persisted and used in entirely different contexts, what's the benefit of keeping them as a combined concept, as opposed to two separate types of notes? I'm thinking of things like `$donation->notes` — does that return donor notes, admin notes, or both? And if both, how are they sorted?

jonwaldstein:
@JasonTheAdams the only place i'm aware of these notes being listed anywhere is inside the donation details page.  It's just a list of notes in chronological order, the _type_ is only used for creating the note since a donor note will trigger a notification.  A donor note is not something that lives with a donor record, these are for donation records.  In this regard, I think `$donation->notes` returns all notes and can choose to display the _type_ of note.

Screen Shot 2023-01-02 at 4 58 05 PM

JasonTheAdams:
> A donor note is not something that lives with a donor record, these are for donation records. In this regard, I think `$donation->notes` returns all notes and can choose to display the type of note.

Cool, I'm tracking. Am I correct in looking at the code that the note is stored in two different places based on its type? That's what's weirding me out.

jonwaldstein:
> > A donor note is not something that lives with a donor record, these are for donation records. In this regard, I think `$donation->notes` returns all notes and can choose to display the type of note.
> 
> Cool, I'm tracking. Am I correct in looking at the code that the note is stored in two different places based on its type? That's what's weirding me out.

Nope the notes are all stored in the `give_comments` table and use `give_commentmeta` to store the `note_type` value.  As far as I can tell, that's only necessary to determine a donation note is specifically created for a donor.

  
Screen Shot 2023-01-03 at 9 35 23 AM

Screen Shot 2023-01-03 at 9 35 29 AM

JasonTheAdams (approved):
My apologies for dropping this one, @jonwaldstein. Thank you for answering my questions and clearing up my misunderstanding. Let's get this merged.

PR #6664, opened 2023-01-05
Feature: Replace Validation Framework with StellarWP Library

JasonTheAdams:
## Description

Per the interest in the Validation framework, it's been extracted into its own [StellarWP/Validation](https://github.com/stellarwp/validation) package. Pretty cool!

This PR removes the framework code and adds the library. It also adds [Strauss](https://github.com/BrianHenryIE/strauss) to prefix the validation classes as to not conflict with other libraries. This is also cool since it will set us up for doing this with additional packages in the future, avoiding conflicts with other plugins and themes.

~One concern I have in this is that I had to bump the platform PHP version to 7.4 since it's [not currently possible](https://github.com/composer/composer/issues/11253) to specify platform requirements for dev. This doesn't necessarily mean PHP 7.0 compatibility is broken, but it does mean we'll need to be careful and check that this PR didn't break compatibility. If it did, we can try specifying exact versions of packages to bring it back down.~

I ended up adding the PHAR version of Strauss as a binary in `bin/strauss`. The benefit of this is that it doesn't impact the project's composer dependencies — since Strauss requires PHP 7.4 to run. It's still a bit complicated to get the automated tests and such to run, as Strauss has to be run on PHP 7.4 or later, but the final build and tests themselves all work on PHP 7.0+.

## Affects

This affects how validations work in GiveWP, which is currently only the Fields API. I believe this does also affect Form Field Manager — which I will provide a PR for.

## Testing Instructions

Reduce your local version of PHP to 7.0 and make sure everything still works.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1203760890899574


JasonTheAdams:
After some pains, @jonwaldstein, I got this working. Please give this a review when you get a chance. I'll send it off to QA if it looks good to you.

give.php
```diff
@@ -524,5 +525,13 @@ function give($abstract = null)
...
 
 require __DIR__ . '/vendor/autoload.php';
+require __DIR__ . '/vendor/vendor-prefixed/autoload.php';
+
+// Setup and initialize the Validation library
+ValidationConfig::setServiceContainer(give());
+ValidationConfig::setHookPrefix('givewp_');
+ValidationConfig::setInvalidArgumentExceptionClass(InvalidArgumentException::class);
+ValidationConfig::setValidationExceptionClass(ValidationException::class);
+ValidationConfig::initialize();
```
jonwaldstein, line 535: Can we please move this into its own service provider?

jonwaldstein (changes requested):

give.php
JasonTheAdams: Hah! Yeah, I kept debating that. Thanks for pushing me in that direction. Resolved in eef378d9a

JasonTheAdams:
@jonwaldstein I've addressed your issue and all tests are passing. Back to you!

jonwaldstein:
@JasonTheAdams thanks, I try to avoid adding anything directly to the main give file.

I did notice this which makes me wonder if we're trying to enforce the container's shape:

Screen Shot 2023-01-10 at 2 24 34 PM

Screen Shot 2023-01-10 at 2 24 38 PM

JasonTheAdams:
@jonwaldstein I updated the Container to implement the interface, I updated its typing as well since the interface required it. Back to you!

JasonTheAdams:
@jonwaldstein Just as a heads up, I ended up pull the `strauss.phar` file out and am instead just downloading it during `composer install` if it doesn't already exist. I plan on doing something similar with PHPStan in another PR, later, so we can run PHPStan locally.

jonwaldstein (approved):
🦾

rickalday (approved):

PR #6667, opened 2023-01-11
chore: Integrate next gen form button

kjohnson:
Related https://github.com/impress-org/givewp-next-gen/pull/112

## Description

This PR checks for a registered Next Gen Form URL and conditionally renders a button.

I tried to keep this simple. The Next Gen feature plugin will register the URL on the window.

## Visuals

image

## Testing Instructions

Install and enable the Next Gen plugin, using https://github.com/impress-org/givewp-next-gen/pull/112, and the "Add Next Gen Form" button should render.

### DIFF
--- a/src/DonationForms/resources/components/DonationFormsListTable.tsx
+++ b/src/DonationForms/resources/components/DonationFormsListTable.tsx
@@ -13,6 +13,10 @@ declare global {
             authors: Array<{id: string | number; name: string}>;
             table: {columns: Array<object>};
         };
+
+        GiveNextGen?: {
+            newFormUrl: string;
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


jonwaldstein:
@kjohnson you think it's possible to add a filter for the buttons that we can just hook into? If so, I would consider that for all table button cc @pauloiankoski .

I'm thinking we can pass the array of buttons using applyFilters to the front-end.  There's also the [`wp.hooks.applyFilters`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-hooks/#api-usage) js version but that would be less convenient.

jonwaldstein (approved):
As we discussed, this is good for now.

PR #6675, opened 2023-01-19
Fix: Prevent fatal error for null price option in the Forms Table

### LINKED ISSUE #6674: Forms with missing price option meta cause fatal error
Related https://github.com/impress-org/givewp/pull/6614#issuecomment-1372471668

## Details

With the new Form Table column sorting feature, a fatal error can be thrown if a form is missing a value for the Price Option meta (`_give_price_option`).

## Steps to Reproduce

1. Checkout `develop`.
2. Delete the `_give_price_option` meta value for a donation form.
3. Navigate to `/wp-admin/admin.php?page=give-forms`

## Visuals

image

kjohnson:
Resolves #6674

## Description

This PR prevents a fatal error caused by donation forms missing the Price Option meta (`_give_price_option`).

Previously, `\Give\DonationForms\DataTransferObjects\DonationFormQueryData::getDonationFormLevels()` would throw a fatal error when the Price Option was `null` because it assumed a non-`mulit` value for the Price Option was the value of `simple` which results in a `TypeError` in `Give\DonationForms\Properties\DonationFormLevel::fromPrice()`.

## Testing Instructions

1. Checkout `develop`.
2. Delete the `_give_price_option` meta value for a donation form.
3. Navigate to `/wp-admin/admin.php?page=give-forms`

---
  - https://app.asana.com/0/0/1203776617039851

### DIFF
--- a/src/DonationForms/DataTransferObjects/DonationFormQueryData.php
+++ b/src/DonationForms/DataTransferObjects/DonationFormQueryData.php
@@ -102,19 +102,17 @@ final class DonationFormQueryData
      */
     public function getDonationFormLevels($object): array
     {
-        if ('multi' === $object->{DonationFormMetaKeys::PRICE_OPTION()->getKeyAsCamelCase()}) {
-            $levels = [];
-            $array = maybe_unserialize($object->{DonationFormMetaKeys::DONATION_LEVELS()->getKeyAsCamelCase()});
-
-            foreach ($array as $level) {
-                $levels[] = DonationFormLevel::fromArray($level);
-            }
-
-            return $levels;
+        switch( $object->{DonationFormMetaKeys::PRICE_OPTION()->getKeyAsCamelCase()} ) {
+            case 'multi':
+                return array_map(function($level) {
+                    return DonationFormLevel::fromArray($level);
+                }, maybe_unserialize($object->{DonationFormMetaKeys::DONATION_LEVELS()->getKeyAsCamelCase()}));
+            case 'simple':
+                return [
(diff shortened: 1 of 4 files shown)


matheuswd (approved):
It worked on my site - I was facing this problem today.

pauloiankoski:
@jonwaldstein I'm working on it!

pauloiankoski:
@jonwaldstein We have all tests in green now!

Benunc (approved):
This passes automated QA tests.

jonwaldstein:
A few more scenarios we should think about:

1. A Form's `Donation Option` is set to `Multi-level Donation` but for some reason does not have any levels.
2. A Form's `Donation Option` is set to 'Set Donation' but does not have a `Set Donation` value

If any of those scenarios were to occur we would face a problem. I have manually tried to see if those scenarios are possible by creating and editing a form and so far they are not.  Can someone else confirm as well?

JasonTheAdams:
@kjohnson @jonwaldstein Another approach, here, would be to add a `try...catch` to the controller to simply display something if an error occurs. If we're not confident in the integrity of this data, that may be a good approach.

JasonTheAdams:
@kjohnson did a great job of addressing the immediate issue. He, myself, and @Benunc also decided it would be very helpful if a failing cell simply caused that one cell to fail, instead of breaking the whole table. I've gone ahead and implemented this and, I'm pleased to say, since every List Table inherits from a base class, this applies to all cells on all tables quite easily.

image

The error is now logged with lots of context and a fallback message is displayed. This feels _way_ better!

JasonTheAdams (approved):

PR #6678, opened 2023-01-19
Fix: Prevent donation summary with donor from filtering twice

Genevieve-K:
Both functions getSummary() and getSummaryWithDonor() apply the same filter. As getSummaryWithDonor() calls getSummary(), the filter is applied twice. As a result, additional data is added twice.
Here I remove the filter for getSummaryWithDonor() , another solution is to use a different filter or add another parameter to the filter.

Resolves #

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Framework/PaymentGateways/DonationSummary.php
+++ b/src/Framework/PaymentGateways/DonationSummary.php
@@ -114,13 +114,14 @@ class DonationSummary
     protected function trimAndFilter(string $text): string
     {
         /**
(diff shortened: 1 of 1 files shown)


kjohnson (approved):
@Genevieve-K thanks for the pull request! I added some unreleased tags along with a small refactor using the `trim` method. I've also opened an additional PR with a test for this change. I'll merge this into my new branch and then into `develop`.

PR #6681, opened 2023-01-20
Feature: adds request methods to the ConnectClient class and a way to debug a local connect server

### LINKED ISSUE #6680: Calls to the server gateway URL should be handled in a centralized place
## User Story

In each place where we need to do a call to the give gateway server, we are using the `https://connect.givewp.com` string directly in the core and also on the addons.

So a good improvement would be to manage it in a centralized place/class that allows us to make calls to the gateway server without writing the URL each time. 

## Details

This way, instead of using this directly:

```
$response =  wp_remote_post('https://connect.givewp.com/square/token-status)', [
  'foo' => 'bar'
]);
```
We would be able to use it similarly to this:

```
$response = $client->post('square/token-status', [
  'foo' => 'bar'
]);
```

## Additional Context

Another advantage is that with this approach we can change the server URL for a local server easily (through the use of a const in the wp-config.php file) and be able to debug changes before deploying it to the live gateway server.

Example: `define( 'GIVE_GATEWAY_SERVER_URL', 'http://127.0.0.1:8000' );`

## Acceptance Criteria

glaubersilva:
Resolves #6680

## Description

This PR adds the `get`, `post`, and `request` methods to the `ConnectClient` class that should be used as a centralized place to retrieve the connect server URL and also to send requests to the connect server. 

It also allows you to debug your local connect server by setting the `GIVE_CONNECT_URL` const in your `wp-config.php` file.

## Affects

Now you can do requests to the server using the following methods:

```
$response = give(ConnectClient::class)->get('your/endpoint');
$response = give(ConnectClient::class)->post('your/endpoint');
$response = give(ConnectClient::class)->request('your/endpoint', ['method' => 'PUT']);
```

If you need to debug your local connect server, just set a const in your `wp-config.php` file like this:

`define( 'GIVE_CONNECT_URL', 'http://127.0.0.1:8000' );`

Now, when you use this:

`$url = give(ConnectClient::class)->getApiUrl('your/endpoint'); `

Instead of getting the default live server URL, you'll get this:

`http://127.0.0.1:8000/your/endpoint`

## Visuals

image

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1203928770361433


src/ConnectClient/ConnectClient.php
```diff
@@ -19,17 +21,108 @@ class ConnectClient
...
      *
-     * @return string
+     * @throws Exception
      */
-    public function getApiUrl($endpoint)
+    private function validateWpErrorInResponse($response)
     {
-        return "{$this->apiUrl}/{$endpoint}";
+        if (is_wp_error($response)) {
+            throw new Exception(
```
JasonTheAdams, line 117: @glaubersilva Let's make a custom `RequestException` that extends `RuntimeException` so folks can specifically catch this.

src/ConnectClient/ConnectClient.php
```diff
@@ -2,11 +2,13 @@
 
 namespace Give\ConnectClient;
```
JasonTheAdams, line 3: I'd rather not have this as a domain in and of itself. @jonwaldstein where do you suggest this goes?

src/Onboarding/Setup/PageView.php
```diff
@@ -139,7 +141,7 @@ public function stripeConnectURL()
                 'website_url' => get_bloginfo('url'),
                 'give_stripe_connected' => '0',
             ],
-            esc_url_raw('https://connect.givewp.com/stripe/connect.php')
+            esc_url_raw(give(ConnectClient::class)->getApiUrl('stripe/connect.php'))
```
JasonTheAdams, line 144: This class is already using dependency injection in the constructor. Use that instead of retrieving it this way.

src/PaymentGateways/Stripe/Admin/AccountManagerSettingField.php
```diff
@@ -608,7 +610,7 @@ public function getStripeConnectButtonMarkup()
                 'website_url' => get_bloginfo('url'),
                 'give_stripe_connected' => '0',
             ],
-            esc_url_raw('https://connect.givewp.com/stripe/connect.php')
+            esc_url_raw(give(ConnectClient::class)->getApiUrl('stripe/connect.php'))
```
JasonTheAdams, line 613: Same here, use dependency injection.

src/ConnectClient/ConnectClient.php
```diff
@@ -19,17 +21,108 @@ class ConnectClient
...
      *
-     * @since 2.8.0
+     * @since      2.8.0
+     *
+     * @param string $endpoint The route on the server. E.g: 'gateway-name/connect'
+     */
+    public function getApiUrl(string $endpoint = ''): string
+    {
+        if ( ! empty($endpoint)) {
+            $this->apiUrl .= '/' . ltrim($endpoint, '/');
```
JasonTheAdams, line 44: I don't see the value in modifying the `apiUrl` property to reflect the endpoint. It's only written and never referenced. In fact, this would have strange behavior:
```php
$client = new ConnectClient();

$client->get('/foo'); // gateway.givewp.com/foo
$client->get('/bar'); // gateawy.givewp.com/foo/bar <-- ???
```

It seems like making this stateless would be more intuitive. It would also mean the class can be stored in the Service Container as a singleton. You'd also be able to pass the connect URL as a constructor, which is passed in via the Container:

```php
give()->singleton(ConnectClient::class, static function() {
    return new ConnectClient(GIVE_CONNECT_URL);
});
```

This would be useful for unit testing.

JasonTheAdams (changes requested):
Great work, @glaubersilva! Left some thoughts!

src/ConnectClient/ConnectClient.php
glaubersilva: Resolve in eac07fc85e7a099ef0c74b9ba1f5aca9db00ed79 and 7539b3c51978a1131ab745e7eca30325f176dc3a

src/Onboarding/Setup/PageView.php
glaubersilva: Resolved in b870f753d05dfb509090e716c9810e4640404030

src/PaymentGateways/Stripe/Admin/AccountManagerSettingField.php
glaubersilva: Resolved in b870f753d05dfb509090e716c9810e4640404030

src/ConnectClient/ConnectClient.php
glaubersilva: @JasonTheAdams  @jonwaldstein Just a note, this class already existed before and is already used in other places with this domain, I just added the new methods and the local server support. However, I believe that makes more sense that it lives in the `PaymentGateways` folder.

JasonTheAdams: @glaubersilva Is this being used in any add-ons? Or is it safe to move?

glaubersilva: @JasonTheAdams In the addons that I have installed, I can't see any using it, just the core. But I don't have sure if any other addon is using it.

src/ConnectClient/ConnectClient.php
```diff
@@ -19,17 +21,110 @@ class ConnectClient
...
+    /**
+     * @unreleased
+     */
+    public function __construct($giveConnectUrl = '')
+    {
+        if ( ! empty($giveConnectUrl)) {
+            $this->apiUrl = $giveConnectUrl;
+        } elseif (defined('GIVE_CONNECT_URL') && GIVE_CONNECT_URL) {
+            $this->apiUrl = GIVE_CONNECT_URL;
+        }
```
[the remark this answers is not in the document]
glaubersilva: @JasonTheAdams In fact, in most cases, the constant will not be defined. Currently, this is not on the `give.php` file, the initial idea was to use it only if it is defined on the `wp-config.php` file.

src/ConnectClient/ConnectClient.php
glaubersilva: @JasonTheAdams According to [this search](https://github.com/search?q=org%3Aimpress-org+ConnectClient&type=code), just a few places in the core are using it. Based on it, seems to be safe to change it. So, let me know if I can move forward with the change to this folder that I mentioned previously.

src/ConnectClient/ConnectClient.php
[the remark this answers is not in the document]
JasonTheAdams: So then... unless a customer adds that constant to their `wp-config.php` file the gateway server requests will stop working? 😅

src/ConnectClient/ConnectClient.php
JasonTheAdams: Yeah, if we can move it then great. Still hoping for @jonwaldstein to weigh in on the location. I think `PaymentGateways\ConnectServer\ConnectClient` could be good. My only reservation on "PaymentGateway" is that this is also used for currency switcher. It's really anytime we need GiveWP to interact with external systems.

src/ConnectClient/ConnectClient.php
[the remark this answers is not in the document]
glaubersilva: Nope! They'll keep using the `$apiUrl` default value.

https://github.com/impress-org/givewp/blob/d2561dc542daf9f1ae819101bcdbf7ab0b83ce28/src/ConnectClient/ConnectClient.php#L22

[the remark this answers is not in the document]
JasonTheAdams: Ohhhh. Right. 🤦‍♂️

src/ConnectClient/ConnectClient.php
jonwaldstein: Although its mainly used for gateways, the connect server is actually not specific to gateways.  It's sort of a general purpose api server.  That makes me think we should have a client in our framework domain. We can add to existing `Http` -> `Framework\Http\ConnectServer\Client`.  Then later we could even extract a new client for gateways specifically that uses the framework one _under-the-hood_.

JasonTheAdams: Yeah, I like that. Let's do that. 🙂

glaubersilva: Resolved in 5072d4b0b96b1ba151177e3a44f22a5bf15e9d4e

src/Framework/Http/ConnectServer/Client/ConnectClient.php
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @unreleased Set default endpoint value as empty
+     * @since      2.8.0
+     *
+     * @param string $endpoint The route on the server. E.g: 'gateway-name/connect'
+     */
+    public function getApiUrl(string $endpoint = ''): string
+    {
+        if ( ! empty($endpoint)) {
+            return $this->apiUrl . '/' . ltrim($endpoint, '/');
```
kjohnson, line 47: This should use [trailingslashit()](https://developer.wordpress.org/reference/functions/trailingslashit/) to ensure that a trailing slash is added and only one trailing slash is added. This protects against passing in a value that already contains a trailing slash.

src/Framework/Http/ConnectServer/Client/Exceptions/RequestException.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Give\Framework\Http\ConnectServer\Client\Exceptions;
+
+use Give\Framework\Exceptions\Primitives\RuntimeException;
+
+/**
+ * @unreleased
+ */
+class RequestException extends RuntimeException
```
kjohnson, line 10: Should this extend our Loggable exception base class?

src/Framework/Http/ConnectServer/Client/ConnectClient.php
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * @unreleased
+     */
+    public function __construct($giveConnectUrl = '')
+    {
+        if ( ! empty($giveConnectUrl)) {
+            $this->apiUrl = $giveConnectUrl;
+        } elseif (defined('GIVE_CONNECT_URL') && GIVE_CONNECT_URL) {
+            $this->apiUrl = GIVE_CONNECT_URL;
```
kjohnson, line 32: We should probably setup a singleton in the container that injects the value from the constant. This would decouple the class from the existence of the constant and improve test-ability.

kjohnson (changes requested):
A couple of suggestions.

src/Framework/Http/ConnectServer/Client/ConnectClient.php
glaubersilva: Resolved in e828c2f884f27896b1d2dee242519110a0cb164b

src/Framework/Http/ConnectServer/Client/Exceptions/RequestException.php
glaubersilva: Resolved in 9c9d78e710cace0a740a6669d8927cc4347a633f

src/Framework/Http/ConnectServer/Client/ConnectClient.php
glaubersilva: @kjohnson Thanks for the review, but I didn't understand why this is necessary, by default this const should not exist and it should use the [**$apiUrl default value**](https://github.com/impress-org/givewp/blob/feature/gateway-server-client-6680/src/Framework/Http/ConnectServer/Client/ConnectClient.php#L22). Basically, this const should be used just for local development, for test-ability with different URLs we can pass it as a param, like this: `$client = new ConnectClient('my-custom-url');`

Maybe we should rename it to `GIVE_DEV_CONNECT_URL` to let it more clear?

kjohnson: Generally, my point is that the configuration should be injected into the class - so that the class is not concerned with determining which URL to use.

The use of a constant is configuring application state, which is specific to the WordPress environment. This can make it harder to test the class in isolation. Instead, if the URL is passed into the class then it can be tested more easily.

This isn't strictly necessary and is a recommendation. I don't want to hold up the PR over what is simply a refactor.

kjohnson (approved):
@glaubersilva thanks for addressing my feedback. There is one open [comment](https://github.com/impress-org/givewp/pull/6681#discussion_r1104651144), but I'll otherwise approve this PR for merge at your discretion.

src/Framework/Http/ConnectServer/Client/ConnectClient.php
kjohnson: I realize now how my original suggestion would be unclear. My apologies for rushing through that.

Your refactor moving the `ConnectClient` to a singleton looks good. The only change that I would make would be to make the `$giveConnectUrl` constructor argument required and pass the default in from the service container.

This creates a single source of truth for where the URL is being set, which would make debugging much easier.



```php
        give()->singleton(ConnectClient::class, static function () {
            $giveConnectUrl = (defined('GIVE_CONNECT_URL') && GIVE_CONNECT_URL)
                ? GIVE_CONNECT_URL
                : 'https://connect.givewp.com';

            return new ConnectClient($giveConnectUrl);
        });
```

```php
    public $apiUrl;

    public function __construct($giveConnectUrl)
    {
        $this->apiUrl = $giveConnectUrl;
    }
```

glaubersilva: @kjohnson No problem! Thanks for the extra improvement suggestions. Resolved in fe0ec8b1548249b520d1a5d89e7eca9bb959138f

kjohnson:
@glaubersilva there is a failing test pointing to an instance of `ConnectClient` that is being instantiated manually - and not resolving from the service container.

Resolving it from the container is a best practice here anyway, so a good catch by the automation.

```
ArgumentCountError: Too few arguments to function Give\Framework\Http\ConnectServer\Client\ConnectClient::__construct(), 0 passed in /home/runner/work/givewp/givewp/tests/Unit/Onboarding/Setup/PageViewTest.php on line 19 and exactly 1 expected
```

glaubersilva:
@kjohnson Thanks for the note, I fixed this last issue and send it to QA.

rickalday (approved):
Passed manual tests

PR #6682, opened 2023-01-20
feature: add setting for automatic formatting

JasonTheAdams:
## Description

Since 2.24.0 we're having issues with customers whose servers are running PHP without the intl extension installed. This is currently used for formatting currency amounts in the list tables. While we'd like to improve internationalization and make auto-formatting of currencies happen, we need to give folks time to upgrade their systems. Truth is, most folks won't really even notice the difference (especially admin-side).

This PR does the following:
1. Adds a new Currency setting to allow customers to opt in or out
2. Adds a migration which sets the initial value based on whether they have the intl extension
3. Provides a failsafe so, no matter what, the intl classes won't be used if not available

**Note:** This intentionally does not affect auto-formatting when it occurs in the browser — which I think only happens in Next Gen. Simply put, I can't think of a downside to auto-formatting, so I'd like it to be the future of GiveWP. The Intl object in JavaScript is simply reliable since it's in [all modern browsers](https://caniuse.com/?search=intl).

Here's the expected flow when formatting a currency:

```mermaid
graph TD;
	start(Begin formatting currency)
	hasIntl{Intl extension installed}
	enabled{Auto-formatting enabled}
    giveFormat(Format using GiveWP)
	intlFormat(Format using intl)
	throwWarning([Throw a warning])

	start-->enabled
	enabled-- Yes -->hasIntl
    enabled-- No -->giveFormat
    hasIntl-- Yes -->intlFormat
	hasIntl-- No -->throwWarning
	throwWarning-->giveFormat
```

## Testing Instructions

If using LocalWP, then use the 7.3.5 PHP version to test **without** the intl extension. Use the 8.0.22 version to test **with** the intl extension.

- [ ] It should not be possible to enable auto-formatting if intl is not present
- [ ] Disabling auto-formatting should use the give formatting options regardless of whether intl is enabled
- [ ] Under no circumstances should having the intl extension missing break any of the list tables

## Pre-review Checklist

---
  - https://app.asana.com/0/0/1203855868473757

### DIFF
--- a/includes/admin/settings/class-settings-general.php
+++ b/includes/admin/settings/class-settings-general.php
@@ -56,7 +56,9 @@ if (!class_exists('Give_Settings_General')) :
         /**
          * Get settings array.
          *
+         * @unreleased add auto_format_currency setting
          * @since  1.8
+         *
          * @return array
          */
         public function get_settings()
--- a/includes/admin/settings/class-settings-general.php
+++ b/includes/admin/settings/class-settings-general.php
@@ -215,7 +215,7 @@ if (!class_exists('Give_Settings_General')) :
                                     'If enabled, amounts are automatically formatted based on the selected currency. This option requires the intl extension for PHP. Read more about <a href="%s" target="_blank">this option and how GiveWP values internationalization in web applications.</a>',
                                     'give'
                                 ),
-                                'https://givewp.com'
+                                'https://docs.givewp.com/currency-format'
                             ),
                             'id' => 'auto_format_currency',
                             'type' => 'checkbox',
--- a/includes/admin/settings/class-settings-general.php
+++ b/includes/admin/settings/class-settings-general.php
@@ -212,7 +212,7 @@ if (!class_exists('Give_Settings_General')) :
                             'desc' => sprintf(
                             /* translators: %s: Link to the GiveWP docs. */
                                 __(
-                                    'Currencies will be automatically formatted based on the user\'s location. Requires the PHP intl-extension. Read more about this <a href="%s" target="_blank">in our docs.</a>',
+                                    'If enabled, amounts are automatically formatted based on the selected currency. This option requires the intl extension for PHP. Read more about <a href="%s" target="_blank">this option and how GiveWP values internationalization in web applications.</a>',
                                     'give'
                                 ),
                                 'https://givewp.com'
--- a/includes/admin/settings/class-settings-general.php
+++ b/includes/admin/settings/class-settings-general.php
@@ -3,592 +3,712 @@
  * Give Settings Page/Tab
  *
  * @package     Give
- * @subpackage  Classes/Give_Settings_General
+ * @since       1.8
  * @copyright   Copyright (c) 2016, GiveWP
  * @license     http://opensource.org/licenses/gpl-2.0.php GNU Public License
- * @since       1.8
+ * @subpackage  Classes/Give_Settings_General
  */
 
-if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Exit if accessed directly.
+if (!defined('ABSPATH')) {
+    exit; // Exit if accessed directly.
 }
 
-if ( ! class_exists( 'Give_Settings_General' ) ) :
+if (!class_exists('Give_Settings_General')) :
 
-	/**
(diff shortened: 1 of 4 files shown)


JasonTheAdams:
@Benunc and @jonwaldstein I'd appreciate your eyes and thoughts on this. I feel like this could be a could solution for our customers having problems.

JasonTheAdams:
Also, I found out that if the intl extension is not installed then it is _very subtly_ surfaced as a recommendation in the Site Health tool:

Screenshot 2023-01-20 at 3 28 48 PM

includes/admin/settings/class-settings-general.php
```diff
@@ -3,592 +3,714 @@
...
+                            ],
+                        ],
+                        [
+                            'name' => __('Auto-format server currency amounts', 'give'),
+                            'desc' => sprintf(
+                            /* translators: %s: Link to the GiveWP docs. */
+                                __(
+                                    'Currencies will be automatically formatted based on the user\'s location. Requires the PHP intl-extension. Read more about this <a href="%s" target="_blank">in our docs.</a>',
+                                    'give'
+                                ),
```
JasonTheAdams, line 217: @Benunc Please give this some messaging love.

src/Framework/Support/Facades/CurrencyFacade.php
```diff
@@ -76,6 +78,20 @@ public function formatToDecimal(Money $amount): string
      */
     public function formatToLocale(Money $amount, $locale = null): string
     {
+        $useAutoFormatting = give_get_option('auto_format_currency');
+        if (!class_exists(NumberFormatter::class) || !$useAutoFormatting) {
+            if ($useAutoFormatting) {
+                Log::warning(
+                    'Auto-formatting enabled but the INTL extension is not available. Please install the INTL extension to enable auto-formatting.'
```
JasonTheAdams, line 85: @Benunc Feel free to adjust this message, too.

includes/admin/settings/class-settings-general.php
Benunc: I believe this is correct: we are not formatting based on user or donor location, but based on currency, right? An individual donor or admin user experience will not differ based on location. It'll format it for everybody based on the currency. 

```suggestion
                                    'If enabled, amounts are automatically formatted based on the selected currency. This option requires the intl extension for PHP. Read more about <a href="%s" target="_blank">this option and how GiveWP values internationalization in web applications.</a>',
                                    'give'
                                ),
```

Benunc: Also, we got rid of lots of checkboxes once upon a time in favor of radio "enabled/disabled" options. Not sure if we should continue that here, and make this enabled/disabled. My suggested message above leans that direction.

src/Framework/Support/Facades/CurrencyFacade.php
Benunc:
```suggestion
                    'Auto-formatting is enabled  at Donations > Settings > General > Currency but the INTL extension for PHP is not available. Please install the INTL extension to enable auto-formatting, or disable the Auto-formatting setting to prevent this error message. Most web hosts can help with installing and activating INTL. GiveWP is falling back to formatting based on the legacy settings.'
```

Benunc (changes requested):
I've added my messaging help.

JasonTheAdams:
Your suggestions have been merged, @Benunc.

Any further feedback, @jonwaldstein, before I send this to QA?

I see this continuing to impact users, so I'd like to move this along.

JasonTheAdams:
@jonwaldstein Yeah, it's just a placeholder until we approve this and write some docs. I wanted to make sure we all agree this is a good idea before we get Stephanie (and me) writing docs for it.

jonwaldstein:
@JasonTheAdams okay other than that looks good to me.

jonwaldstein (approved):
👍

JasonTheAdams:
Great! Then I will work on getting a write-up together today and over to Stephanie, @Benunc.

Benunc (approved):
This passes automated tests, and I installed it on a server where intl is not enabled, and it appropriately fell back with no fatal errors. :shipit:

PR #6684, opened 2023-01-26
Feature: update gateway api registration order

jonwaldstein:
## Description

- Moved the place where we connect the gateways to the legacy adapter.  I found this would be called twice during an unregister which means it was being called too early.  In this scenario the gateway's fields were being displayed twice.  Moving this part to _after_ all the registration actions have been called, instead of the register method specifically has proved to be more reliable - and makes more sense since this is specifically for legacy adaption. 
- Remove unused `ResponseHelpers` trait

## Affects

- The gateway registration

## Visuals

**Before: example of PayPal being registered by core, unregistered, then registered again**

This is because the register() method was being called twice and thus connecting to the adapter twice.

Screenshot 2023-01-27 at 11 02 58 AM

**After: example of PayPal being registered by core, unregistered, then registered again**

Now its only being connected to the adapter once, after all gateways have been registered.

Screenshot 2023-01-27 at 11 03 42 AM

## Testing Instructions

Just make sure a donation form with a few gateways work properly.  

## Pre-review Checklist

-   [ ] Includes unit tests


jonwaldstein:
@JasonTheAdams please review this when you have a moment.

JasonTheAdams (approved):
Ah yes! This definitely belongs in the gateway registration and not in the framework. Good catch!

PR #6695, opened 2023-02-08
Refactor: update gateway api `$gatewayData` param

jonwaldstein:
## Description

Since our `PaymentGateway` is an abstract class, assigning the `$gatewayData` param to null in createSubscription() is forcing gateways to assign the param to null by default in order to be compatible. Gateways extending the PaymentGateway class can still assign the `$gatewayData` param as null for backwards compatibility but should not be forced to. This PR removes the assignment from the abstract class.

## Affects

Any gateways overloading the createSubscription class directly in their gateway class.  

## Visuals

N/A

## Testing Instructions

We will need to test recurring gateways using the gateway api.  In this PR I have updated the test gateways which can be uncommented in `RegisterPaymentGateways` and tested manually using a recurring form.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1204449389516441


jonwaldstein:
@JasonTheAdams please review this one.
cc: @ravinderk

JasonTheAdams:
@jonwaldstein So what's the benefit of preventing this from being null? Is the idea that it's reliably an array, albeit an empty one if no data is passed?

jonwaldstein:
@JasonTheAdams that's the idea but also I don't want gateways to be forced to set the `$gatewayData` property as null.  If they don't set it to null right now, it will cause a fatal error because this is an abstract class:

Screenshot 2023-02-17 at 9 09 12 AM


Since we introduced the `$gatewayData` param _after_ some gateways were built, @ravinderk had to set that param to null for backwards compatibility, but as far as I can tell this would not be breaking change anymore.

JasonTheAdams:
@jonwaldstein Ah, I see. The reason it doesn't throw an error is because the parameter itself isn't typed as an array and is therefore `mixed`. We would only need the optional null assignment if we had it cast as an array: `array $gatewayData = null`.

jonwaldstein:
@JasonTheAdams we decided on continuing with this update right?

Benunc (approved):
This passes automated tests.

PR #6696, opened 2023-02-09
Fix: Prevent donation summary with donor from filtering twice

kjohnson:
Related #6678.

## Description

Adds a test for #6678, which should fail on `develop`, but pass when #6678 is merged into this branch.

## Testing Instructions

After the test fails, merge #6678 to re-trigger the tests.

---
  - https://app.asana.com/0/0/1203939212131551


kjohnson:
I approved and merged #6678 into this branch. Building and submitting for QA now.

JasonTheAdams (approved):
We can go ahead and merge this. We don't need QA as the entire problem is caught in your excellent unit test. 💪

PR #6697, opened 2023-02-09
Feature/new synced command

glaubersilva:
Resolves #

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Framework/PaymentGateways/CommandHandlers/SubscriptionSyncedHandler.php
+++ b/src/Framework/PaymentGateways/CommandHandlers/SubscriptionSyncedHandler.php
@@ -17,8 +17,19 @@ class SubscriptionSyncedHandler
      */
     public function __invoke(SubscriptionSynced $subscriptionSynced): JsonResponse
     {
+        $subscriptionAttributesChanged = $subscriptionSynced->subscription->getDirty();
+        $subscriptionOldStatus = $subscriptionSynced->subscription->getOriginal('status');
+        $subscriptionNewStatus = $subscriptionSynced->subscription->status->getValue();
+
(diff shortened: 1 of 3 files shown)


jonwaldstein:
@JasonTheAdams can you please review this one

JasonTheAdams:
Nice work, @glaubersilva! This makes sense! The only thing I'm wondering about is if synchronization is currently done in one go of it, or if it's batched. There are potentially a lot of API calls going on, which risks a timeout if there are a lot of donations to sync.

I'm curious how GiveWP currently does this? Is it all one go or batched?

glaubersilva:
@JasonTheAdams It gets all transactions for a subscription in the last 6 months. So, a monthly subscription should return `6` transactions and in the worse scenario a daily subscription should return `~180` transactions. 

I remember a call we had with @Benunc (last year I believe) and he said that this sync feature is like a last resource when something goes wrong. I think we also talked about this worst scenario and we conclude that it would not be something too frequent as using a daily subscription is not so common.

So, I believe that we should not have timeout problems if we keep this 6-month time frame in future implementations.

JasonTheAdams (approved):
That is _very well_ reasoned out, @glaubersilva. Thank you for laying that all out for me!

In that case, this is sufficient for what it needs to be at this point in time. 👍

glaubersilva:
@JasonTheAdams Thanks! Do you able to also review [this other related PR](https://github.com/impress-org/give-recurring/pull/1164) on the Recurring addon?

rickalday (approved):
Passed manual QA tests

rickalday (approved):
Passed manual QA tests

PR #6708, opened 2023-02-22
Hotfix: make donations get finished when Authorize.Net is the default gateway in sites with more than one gateway enabled.

### LINKED ISSUE #6707: The classic form template should work with Authorize.Net when this is the default gateway
## User Story

Currently, Authorize.Net donations made using the classic form template, don't get finished when this is the default gateway.

## Details

It's necessary to have more than one gateway enabled to reproduce the bug.

## Expected Behavior

It's expected that the donations work when Authorize.Net is the default gateway in a site with more than one gateway enabled.

## Steps to Reproduce

1. Create a form using the classic template
3. Enable the Authorize.Net gateway and set it as the default
4. Enable the test gateway
5. Load the form page, try to do a donation and nothing will happen - the browser console should display the following error message: `E_WC_01: Please include Accept.js library from cdn.`

## Visuals

image

## Acceptance Criteria

-   [ ] The donations get finished when Authorize.Net is the default gateway in sites with more than one gateway enabled.

glaubersilva:
Resolves #6707

## Description

This PR fixes the related issue by changing the way how we clone the children nodes from the `#give_purchase_form_wrap` in the `moveDefaultGatewayDataIntoActiveGatewaySection()` method. Now instead of selecting just `fieldset` nodes, we select all nodes, except the `.give-donation-submit` element.

But why does this fix the problem? Because now, we also include the `script` nodes that are necessary to load the Accept.js library required by the Authorize.Net gateway. 

As a note, this problem wasn't happening in situations where Authorize.Net isn't the default gateway because in these cases, the details are loaded through an AJAX call that already was properly loading everything.

## Affects

Classic Template. 

## Visuals

image

## Testing Instructions

1. Create a form using the classic template
2. Enable the Authorize.Net gateway and set it as the default
3. Enable the test gateway
4. Load the form page, make a donation using Authorize.Net and it should work as expected

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1204033725468285

### DIFF
--- a/src/Views/Form/Templates/Classic/resources/js/form.js
+++ b/src/Views/Form/Templates/Classic/resources/js/form.js
@@ -246,10 +246,11 @@ function moveDefaultGatewayDataIntoActiveGatewaySection() {
 
     addSelectedGatewayDetails(createGatewayDetails());
 
-    document.querySelector('#give_purchase_form_wrap').removeChild(document.querySelector('.give-donation-submit'));
-    document.querySelector('.give-gateway-details').append(...document.querySelector('#give_purchase_form_wrap').children);
+    const purchaseFormWrap = document.querySelector('#give_purchase_form_wrap');
+    purchaseFormWrap.removeChild(purchaseFormWrap.querySelector('.give-donation-submit'));
+    document.querySelector('.give-gateway-details').append(...purchaseFormWrap.children);
 
-    removeNode(document.querySelector('#give_purchase_form_wrap'));
+    removeNode(purchaseFormWrap);
 }
 
 function attachRecurringDonationEvents() {
--- a/src/Views/Form/Templates/Classic/resources/js/form.js
+++ b/src/Views/Form/Templates/Classic/resources/js/form.js
@@ -246,9 +246,8 @@ function moveDefaultGatewayDataIntoActiveGatewaySection() {
 
     addSelectedGatewayDetails(createGatewayDetails());
 
-    document
-        .querySelector('.give-gateway-details')
-        .append(...jQuery("#give_purchase_form_wrap").children().not('.give-donation-submit').clone(true, true));
+    document.querySelector('#give_purchase_form_wrap').removeChild(document.querySelector('.give-donation-submit'));
+    document.querySelector('.give-gateway-details').append(...document.querySelector('#give_purchase_form_wrap').children);
 
     removeNode(document.querySelector('#give_purchase_form_wrap'));
 }
--- a/src/Views/Form/Templates/Classic/resources/js/form.js
+++ b/src/Views/Form/Templates/Classic/resources/js/form.js
@@ -248,7 +248,7 @@ function moveDefaultGatewayDataIntoActiveGatewaySection() {
 
     document
         .querySelector('.give-gateway-details')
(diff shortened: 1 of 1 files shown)


src/Views/Form/Templates/Classic/resources/js/form.js
```diff
@@ -248,7 +248,7 @@ function moveDefaultGatewayDataIntoActiveGatewaySection() {
 
     document
         .querySelector('.give-gateway-details')
-        .append(...document.querySelectorAll('#give_purchase_form_wrap fieldset:not(.give-donation-submit)'));
+        .append(...jQuery("#give_purchase_form_wrap").children().not('.give-donation-submit').clone(true, true));
```
pauloiankoski, line 251: I try to keep vanilla JS functions free of jQuery and also have control over which elements are copied as it can be beneficial to reduce unnecessary code and avoid introducing unexpected bugs. So a simple modification of the original selector seems to achieve the same result. Can you confirm it?

```suggestion
        .append(...document.querySelectorAll('#give_purchase_form_wrap :is(fieldset:not(.give-donation-submit), script)'));
```

pauloiankoski (changes requested):

src/Views/Form/Templates/Classic/resources/js/form.js
JasonTheAdams: Good note, Paulo. We're really trying to move away from jQuery. Honestly, you can do pretty much anything in vanilla JS these days that used to be better done in jQuery, and it's faster.

src/Views/Form/Templates/Classic/resources/js/form.js
```diff
@@ -246,9 +246,8 @@ function moveDefaultGatewayDataIntoActiveGatewaySection() {
 
     addSelectedGatewayDetails(createGatewayDetails());
 
-    document
-        .querySelector('.give-gateway-details')
-        .append(...document.querySelectorAll('#give_purchase_form_wrap fieldset:not(.give-donation-submit)'));
+    document.querySelector('#give_purchase_form_wrap').removeChild(document.querySelector('.give-donation-submit'));
+    document.querySelector('.give-gateway-details').append(...document.querySelector('#give_purchase_form_wrap').children);
 
     removeNode(document.querySelector('#give_purchase_form_wrap'));
```
pauloiankoski, line 252: That's a good approach. In case you think it's relevant, we can also optimize the code a bit avoiding unnecessary use of `document.querySelector()` by storing the reference to the `#give_purchase_form_wrap` element in a variable:

```suggestion
    const purchaseFormWrap = document.querySelector('#give_purchase_form_wrap');
    purchaseFormWrap.removeChild(purchaseFormWrap.querySelector('.give-donation-submit'));
    document.querySelector('.give-gateway-details').append(...purchaseFormWrap.children);

    removeNode(purchaseFormWrap);
```

glaubersilva: Resolved in dd3165a8122493ce0b6d7d082e34f268e48e3ed7

pauloiankoski (approved):
Looks good! Ready for QA!

rickalday (approved):
Passed manual QA tests

PR #6712, opened 2023-02-27
Feature: Add color parameter to give_goal shortcode

rickalday:
Resolves # https://feedback.givewp.com/documentation/p/website-feedback-givewp-goal-shortcode

## Description

Adds **color** parameter to the Give Goal Shortcode.

## Affects

None

## Visuals

Screen Shot 2023-02-27 at 11 27 40 a m

## Pre-review Checklist

-   [ ] Includes unit tests


templates/shortcode-goal.php
```diff
@@ -32,7 +32,7 @@
 
 $goal_progress_stats = give_goal_progress_stats( $form );
 $goal_format         = $goal_progress_stats['format'];
-$color               = give_get_meta( $form_id, '_give_goal_color', true );
+$color               = sanitize_hex_color( $args['color'] ) ?? give_get_meta( $form_id, '_give_goal_color', true );
```
JasonTheAdams, line 35: This is a little fragile. The order of operation here is that the `sanitize_hex_color()` function will _always_ be called, even if `$args['color']` is empty. Given the default of an empty string, then an empty string will be returned, and that would actually pass in a null coalescing operator comparison. Consider:
```php
echo null ?? 'hi'; // 'hi'
echo $array['not_set'] ?? 'hi'; // 'hi'
echo '' ?? 'hi': // ''
```
I recommend changing this to:
```php
$color = empty($args['color']) ? give_get_meta( $form_id, '_give_goal_color', true ) : sanitize_hex_color( $args['color'] );
```
This way the meta value is used if the parameter isn't set or is an [empty](https://www.php.net/manual/en/function.empty.php) value, and the sanitization method is only called if we're confident a value is actually set.

JasonTheAdams (changes requested):
Thanks for contributing, @rickalday! I provided some feedback. Let me know if you have any questions. 😄

templates/shortcode-goal.php
JasonTheAdams: @rickalday This is still waiting for you. 😄

rickalday: @JasonTheAdams I've committed the changes.

JoshuaHungDinh: @rickalday Looks good. It's been tested and working.

JoshuaHungDinh (approved):

PR #6719, opened 2023-03-06
Fix: Patch SSRF in add-on changelog

kjohnson:
Re: https://app.asana.com/0/0/1204069668576428/f

## Description

This PR patches a potential SSRF vulnerability in which a malicious user with elevated permissions (GiveWP Manager) can provide arbitrary URLs to be fetched by the server.

---
  - https://app.asana.com/0/0/1204129686534554

### DIFF
--- a/includes/ajax-functions.php
+++ b/includes/ajax-functions.php
@@ -812,6 +812,16 @@ function give_get_content_by_ajax_handler() {
 		die();
 	}
 
+    /**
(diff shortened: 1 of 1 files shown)


jonwaldstein (approved):
This looks like a solid solution, thanks!.  We just need to make sure it doesn't affect reading our add-on changelogs. I'll get this into QA to confirm.

Benunc (approved):
This passes manual tests. I can still read all the changelogs. Also passes automated tests.

PR #6723, opened 2023-03-07
Add support for PayPal Order ID via gateway data (Next Gen)

kjohnson:
## Description

This PR adds support for passing the PayPal Order ID via the Gateway Data.

Previously, the Order ID was pulled directly from the `$_POST` global, which resulted in a `PayPalOrderIdException` for a missing Order ID in the Next Gen PayPal Commerce gateway implementation.

## Testing Instructions

Make a donation using the Next Gen updated PayPal Commerce gateway.

---
  - https://app.asana.com/0/0/1204267265032333


src/PaymentGateways/Actions/RegisterPaymentGateways.php
```diff
@@ -140,7 +140,10 @@ private function addGatewayDataToPayPalCommerce()
                 PayPalCommerce::id()
             ),
             function ($gatewayData) {
-                $gatewayData['paypalOrder'] = (new GetPayPalOrderFromRequest())();
+                if(!isset($gatewayData['payPalOrderId'])) {
+                    $gatewayData['payPalOrderId'] = give_clean($_POST['payPalOrderId']);
+                }
+                $gatewayData['paypalOrder'] = (new GetPayPalOrderFromRequest())($gatewayData['payPalOrderId']);
```
jonwaldstein, line 146: @kjohnson can we clean this up a bit while we're here?

`GetPayPalOrderFromRequest`(by it's name) should either have all the request logic for retrieving the `payPalOrderId` or we can do the request logic here and just call `give(PayPalOrder::class)->getOrder($paypalOrderId)` directly without the action.

Both ways should be using a try/catch as well.

jonwaldstein (changes requested):

jonwaldstein (approved):
@kjohnson thanks for the update, looks much better.  Just to be safe, we should probably send this to QA to make sure it doesn't affect existing PayPal commerce/donations.

kjohnson:
Added to the QA pipeline, see https://app.asana.com/0/1203193833018151/1204267265032333/f.

Benunc:
@kjohnson I ran tests on this ZIP and it worked, too: https://github.com/impress-org/givewp/actions/runs/4568552280

:shipit:

PR #6726, opened 2023-03-08
Feature: Admin FormNavigation with actionable buttons

JoshuaHungDinh:
## Description

This PR includes  the Admins FormPage Navigational component, a reusable Button & the MoreActionsMenu component. The FormNavigation is rendered inside the FormPage. Its props are statically being passed by the Donation Details app but will eventually derive this data from the window.

## Updated - on 03/16

Includes: 
- An Icons folder holding all the svgs that will be used throughout the app.
- Separated type files into a per component basis. This is how I began handling it in the PaymentBanner branch as types became more robust.
- Removed Nav middle donation selector. New design is in preparation and we will add this to the UI when we include the final actions for the action menu.
- Removed any un-setup type files

I have also added a test input for testing form submission, button states & the JOI schema. ..This will be replaced.

### pageId, pageTitle, pageDescription: 
These are presentational details that will display in header according to the selected donation.

### navigationalOptions:
 This is being passed an array of donation numbers to filter through. The design on the select element used to render these options needs to be addressed.

### onSubmit:
This simply handles formSubmission with the SaveChanges button.

### isDirty: 
a boolean value that determines the form state and adjusts if the SaveChanges button is disabled

### actionConfig:
a config of titles and actions for the MoreActions button to list. Since the actions will very depending on the detail page the actions will be defined in the detail app component. A separate MoreActionsMenu component will then receive this config through FormNavigation. It will then map through the config and display their properties.

## Additional Context
- Inline svgs are currently being used. I imagine this will be replaced by loading the svgs as images and placing their files in the assets directory. But I am also open to other suggestions? 

## Visuals & Testing Instructions

https://user-images.githubusercontent.com/75056371/223852853-9090bb18-9ec9-41ea-8f9a-68c289cc345d.mov

Form submissions and actions are setup to throw an alert on the screen when called. Try changing the input, submitting the form and selecting different actions from the MoreActionsList.

## Pre-review Checklist

-   [ ] Includes unit tests


src/Views/Components/AdminUI/Button/style.module.scss
```diff
@@ -0,0 +1,76 @@
...
+    border-color: var(--givewp-red-500);
+
+    &:hover {
+        background-color: var(--givewp-red-400);
+    }
+
+    &:active {
+        background-color: var(--givewp-red-600);
+    }
+}
```
pauloiankoski, line 61: You have many things being repeated here. A reduced version of it could be:
```scss
.primary,
.secondary,
.danger {
    &:hover {
        background-color: var(--givewp-blue-400);
    }

    &:active {
        background-color: var(--givewp-blue-600);
    }
}

.primary {
    color: var(--givewp-shades-white);
    background: var(--givewp-blue-500);
}

.secondary {
    background: var(--givewp-shades-white);
    color: var(--givewp-blue-500);
    border-color: currentColor;
}

.danger {
    background-color: var(--givewp-red-500);
    color: var(--givewp-shades-white);
    border-color: var(--givewp-red-500);
}

File: src/Views/Components/AdminUI/Button/style.module.scss [hunk 2]
```diff
@@ -0,0 +1,76 @@
...
+
+.large {
+    padding: var(--givewp-spacing-4) var(--givewp-spacing-20);
+}
+
+.button:disabled {
+    background-color: var(--givewp-grey-300);
+    color: var(--givewp-shades-white);
+    border-color: var(--givewp-grey-300);
+}
```
pauloiankoski, line 75: As this is a state of the button element, it would make more sense if staying closer to it:
```scss
.button {
    ...

    &:disabled {
        background-color: var(--givewp-grey-300);
        color: var(--givewp-shades-white);
        border-color: var(--givewp-grey-300);
    }
}

File: src/Views/Components/AdminUI/FormNavigation/index.tsx [hunk 3]
```diff
@@ -0,0 +1,62 @@
...
+    const [toggleActions, setToggleActions] = useState(false);
+
+    const toggleMoreActions = () => {
+        setToggleActions(!toggleActions);
+    };
+
+    return (
+        <header className={styles.formPageNavigation}>
+            <div className={styles.wrapper}>
+                <button className={styles.container} onClick={() => window.history.back()}>
```
pauloiankoski, line 34: I wonder if wouldn't be better if we had the URL for the respective List Table here instead. For instance, if you share the donation details URL with someone in your team. Once that person opens the page, its history will be empty and the button won't do anything.

pauloiankoski (changes requested):

src/Views/Components/AdminUI/Button/style.module.scss
JoshuaHungDinh: Resolved: https://github.com/impress-org/givewp/pull/6726/commits/8786d71f09125c0225e79bff0ac95dc922ef68b3

src/Views/Components/AdminUI/Button/style.module.scss
JoshuaHungDinh: Hey @pauloiankoski, Each of the these css properties & values are just slightly different from each other. The first two button css values are swapped with each property. Im not sure the proposed solution will have the same outcome

src/Views/Components/AdminUI/Button/style.module.scss
pauloiankoski: I think my suggestion is slightly different from your implementation. I meant to keep it close as you have the `.button` class on line 1, so it makes sense just to move the `:disabled` statement to be appended to the class on the top.

JoshuaHungDinh: resolved: https://github.com/impress-org/givewp/pull/6726/commits/3ae7d14428acfabba13d64c6f2f3638ec2a21d9e

src/Views/Components/AdminUI/FormNavigation/index.tsx
JoshuaHungDinh: Yea if you want to pass it. Otherwise I think react-router-dom has a history.back that preserves the origin page.

pauloiankoski: It's okay for now but let's change this in the future.

pauloiankoski (approved):

PR #6736, opened 2023-03-15
Feature: REST API Test Case

pauloiankoski:
## Description

This Pull Request adds a new `RestApiTestCase` class. The class provides specific methods and helper functions to be used in API tests. This will make it easier and faster to write tests for the Give REST API.

To use the `RestApiTestCase` class, you just need to extend your test from it instead of the standard `TestCase` class. The `RestApiTestCase` class initializes the REST server, creates user accounts for API authentication, provides wrapper functions for creating and dispatching requests, and includes an assertion method for checking WP error responses.

The reason for extending a class instead of using a trait is to keep the codebase clean and avoid duplication of logic. If we were to use a trait, we would need to implement duplicate logic in the standard `TestCase` class and prefix the methods in the trait to avoid naming conflicts with other traits. By extending the `RestApiTestCase` class, we avoid these issues and can easily access its helper methods without having to add additional prefixes or boilerplate code. This results in cleaner, more readable code that is easier to maintain and understand.

To run authenticated tests, we've now got two traits that whip up the users for it:
```php
use HasDefaultWordPressUsers;
use HasDefaultGiveWPUsers;
```

Below is the method signature for the publicly available methods:
```php
/**
 * Wrapper for creating a request.
 *
 * @param string $method     The HTTP method of the request (e.g. GET, POST, etc.).
 * @param string $route      The REST API route to access (e.g. /wp/v2/posts).
 * @param array  $attributes Optional. An array of attributes to set on the request object.
 * @param string $userRole   Optional. The role of the user to authenticate the request (e.g. anonymous, administrator, etc.).
 * 
 * @return WP_REST_Request A new WP_REST_Request object with the specified parameters.
 */
public function createRequest(string $method, string $route, array $attributes = [], string $userRole = 'anonymous'): WP_REST_Request
```
```php
/**
 * Wrapper for dispatching a request.
 *
 * @param WP_REST_Request $request The request object to dispatch to the server.
 * 
 * @return WP_REST_Response The response object returned by the server.
 */
public function dispatchRequest(WP_REST_Request $request): WP_REST_Response
```
```php
/**
 * Asserts that the response is a WP error response with the specified code and status (if provided).
 *
 * @param int|string $code     The expected error code of the WP_Error object.
 * @param mixed      $response The response object to check (can be a WP_REST_Response or WP_Error object).
 * @param int|null   $status   Optional. The expected error status of the WP_Error object (if any).
 * 
 * @return void
 */
protected function assertErrorResponse($code, $response, $status = null)
```

## Testing Instructions

Unit test class sample:
```php
<?php

namespace Give\Tests\Unit\API\Endpoints;

use Give\Tests\RestApiTestCase;
use Give\Tests\TestTraits\HasDefaultWordPressUsers;
use WP_REST_Response;

class TestExample extends RestApiTestCase
{
    use HasDefaultWordPressUsers;

    public function testValidRequest() {
        $request = $this->createRequest('GET', '/wp/v2/users', [], 'administrator');
        $response = $this->dispatchRequest($request);

        $this->assertInstanceOf(WP_REST_Response::class, $response);
        $this->assertEquals(200, $response->get_status());

        $data = $response->get_data();
        $this->assertIsArray($data);
        $this->assertNotEmpty($data);
    }

    public function testInvalidRequest() {
        $request = $this->createRequest('GET', '/wp/v2/unknown');
        $response = $this->dispatchRequest($request);

        $this->assertErrorResponse('rest_no_route', $response, 404);
    }
}
```

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1203466384450656

### DIFF
--- a/tests/RestApiTestCase.php
+++ b/tests/RestApiTestCase.php
@@ -80,46 +80,6 @@ class RestApiTestCase extends TestCase
     }
 
     /**
-     * Wrapper for creating a request
-     *
-     * @unreleased
-     *
-     * @param string $method
-     * @param string $route
-     * @param array  $attributes
-     * @param string $authentication
-     *
-     * @return WP_REST_Request
-     */
-    public function createRequest(
-        string $method,
-        string $route,
-        array $attributes = [],
-        string $authentication = 'anonymous'
-    ): WP_REST_Request {
-        if ( ! in_array($authentication, array_keys(self::$users), true)) {
-            $authentication = 'anonymous';
-        }
-        wp_set_current_user(self::$users[$authentication]);
-
-        return new WP_REST_Request($method, $route, $attributes);
-    }
-
-    /**
-     * Wrapper for dispatching a request
-     *
-     * @unreleased
-     *
-     * @param WP_REST_Request $request
-     *
-     * @return WP_REST_Response
-     */
-    public function dispatchRequest(WP_REST_Request $request): WP_REST_Response
-    {
-        return $this->server->dispatch($request);
-    }
-
-    /**
      * Flushes the WordPress user roles and reloads them from the database.
      *
      * This function ensures that we are testing against the database data, not just in-memory data.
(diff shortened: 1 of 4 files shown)


tests/RestApiTestCase.php
```diff
@@ -0,0 +1,195 @@
...
+     * Create users for API authentication
+     *
+     * @unreleased
+     *
+     * @return void
+     */
+    public static function wpSetUpBeforeClass(WP_UnitTest_Factory $factory)
+    {
+        self::$users = [
+            'anonymous' => 0,
```
jonwaldstein, line 55: What does this user role do?

tests/RestApiTestCase.php
```diff
@@ -0,0 +1,195 @@
...
+            'administrator' => $factory->user->create(['role' => 'administrator']),
+            'editor' => $factory->user->create(['role' => 'editor']),
+            'author' => $factory->user->create(['role' => 'author']),
+            'contributor' => $factory->user->create(['role' => 'contributor']),
+            'subscriber' => $factory->user->create(['role' => 'subscriber']),
```
jonwaldstein, line 60: What's the reason for populating all these users everytime? Are we just anticipating testing for different permissions related to the uer role?

tests/RestApiTestCase.php
pauloiankoski: It does nothing. It's such as not authenticated. We force to set current user to 0 to be sure it's not authenticated while dispatching the request.

tests/RestApiTestCase.php
pauloiankoski: Yeah. It's a more generic approach so the REST API can be tested in different scenarios.

jonwaldstein: I would be inclined to expand on this part a little bit.  In GiveWP we have specific user roles like `give_manager`, `give_donor`, etc. that we might want to test for.  Maybe we just move some of this setup into a trait like `HasDefaultWordPressUsers` and  `HasDefaultGiveWPUsers` so we can opt into this or not.

jonwaldstein (approved):
👍

PR #6745, opened 2023-03-20
Chore: Update readme for Next Gen discovery

taylorfromteamgive:
This is for the feature plugin release. The GiveWP core readme should be updated when the 3.0 feature plugin is live on .org.


JasonTheAdams (approved):
Perfect! Thanks, Taylor!

PR #6751, opened 2023-03-27
Feature: admin donation details api payment fields

pauloiankoski:
## Description

This PR adds payment information fields to the Donation Update endpoint on the admin.

## Affects

GiveWP Rest API

## Pre-review Checklist

-   [x] Includes unit tests

### DIFF
--- a/src/Donations/Endpoints/DonationUpdate.php
+++ b/src/Donations/Endpoints/DonationUpdate.php
@@ -98,7 +98,7 @@ class DonationUpdate extends Endpoint
                 $attrId = $attr::getId();
                 $value = $request->get_param($attrId);
 
-                if (is_null($value) || ! in_array(AttributeUpdatesModel::class, class_implements($attr))) {
+                if (is_null($value) || ! is_a($attr, AttributeUpdatesModel::class, true)) {
                     continue;
                 }
 
--- a/src/Donations/Endpoints/DonationUpdate.php
+++ b/src/Donations/Endpoints/DonationUpdate.php
@@ -8,7 +8,6 @@ use Give\Donations\Endpoints\DonationUpdateAttributes\FeeAmountRecoveredAttribut
 use Give\Donations\Endpoints\DonationUpdateAttributes\FormIdAttribute;
 use Give\Donations\Endpoints\DonationUpdateAttributes\IdAttribute;
 use Give\Donations\Endpoints\DonationUpdateAttributes\StatusAttribute;
-use Give\Donations\Models\Donation;
 use Give\Framework\Exceptions\Primitives\Exception;
 use WP_Error;
 use WP_REST_Request;
@@ -102,10 +101,11 @@ class DonationUpdate extends Endpoint
                     continue;
                 }
 
-                $updatedDonation = $attr::update($value, $donation);
+                $actualValue = $donation->{$attrId};
+                $attr::update($value, $donation);
+                $updatedValue = $donation->{$attrId};
 
-                if (is_a($updatedDonation, Donation::class)) {
-                    $donation = $updatedDonation;
+                if ($actualValue !== $updatedValue) {
                     $updatedFields[] = $attrId;
                 }
             }
--- a/src/Donations/Endpoints/DonationUpdate.php
+++ b/src/Donations/Endpoints/DonationUpdate.php
@@ -3,6 +3,7 @@
 namespace Give\Donations\Endpoints;
 
 use Give\Donations\Endpoints\DonationUpdateAttributes\AmountAttribute;
+use Give\Donations\Endpoints\DonationUpdateAttributes\AttributeUpdatesModel;
 use Give\Donations\Endpoints\DonationUpdateAttributes\CreatedAtAttribute;
 use Give\Donations\Endpoints\DonationUpdateAttributes\FeeAmountRecoveredAttribute;
 use Give\Donations\Endpoints\DonationUpdateAttributes\FormIdAttribute;
@@ -97,7 +98,7 @@ class DonationUpdate extends Endpoint
                 $attrId = $attr::getId();
                 $value = $request->get_param($attrId);
 
-                if (is_null($value)) {
+                if (is_null($value) || ! in_array(AttributeUpdatesModel::class, class_implements($attr))) {
                     continue;
                 }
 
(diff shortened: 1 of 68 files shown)


src/Donations/Endpoints/DonationUpdateAttributes/DonationUpdateAttribute.php
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * @param          $value
+     * @param Donation $donation
+     *
+     * @return Donation|bool
+     */
+    public static function update($value, Donation $donation)
+    {
+        return false;
+    }
```
JasonTheAdams, line 28: I think this can just be an abstract method as nothing really needs to be returned. The model can simply be passed and if its mutated its mutated.

src/Donations/Endpoints/DonationUpdateAttributes/PaymentInformation/FeeAmountRecoveredAttribute.php
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public static function getDefinition(): array
+    {
+        return [
+            'type' => 'number',
+            'required' => false,
+            'minimum' => 0,
+            'sanitize_callback' => function ($param) {
+                return floatval($param);
+            },
```
JasonTheAdams, line 35: We want to avoid turning currency amounts into floats whenever possible. Keep it as a string.

JasonTheAdams (changes requested):
Great work, @pauloiankoski! I left a few notes!

src/Donations/Endpoints/DonationUpdate.php
```diff
@@ -0,0 +1,156 @@
...
+    public function registerRoute()
+    {
+        $this->registerRouteAttributes([
+            IdAttribute::class,
+            StatusAttribute::class,
+            AmountAttribute::class,
+            FeeAmountRecoveredAttribute::class,
+            FormIdAttribute::class,
+            CreatedAtAttribute::class,
+        ]);
```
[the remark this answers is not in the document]
pauloiankoski: Resolved d4948eb2120713dc0b1b23954b12b27c96f8ac96

src/Donations/Endpoints/DonationUpdateAttributes/PaymentInformation/AmountAttribute.php
```diff
@@ -0,0 +1,52 @@
...
+     */
+    public static function update($value, Donation $donation): Donation
+    {
+        $donation->amount = Money::fromDecimal(
+            $value,
+            $donation->amount->getCurrency()
+        );
+
+        return $donation;
+    }
```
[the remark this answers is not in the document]
pauloiankoski: Resolved f02b788976311425f478454accbeaca55de1fe65

src/Donations/Endpoints/DonationUpdateAttributes/PaymentInformation/FeeAmountRecoveredAttribute.php
pauloiankoski: Resolved e3c8e6c82f9bc319df0631ea94b5caa65490da43

src/Donations/Endpoints/DonationUpdateAttributes/DonationUpdateAttribute.php
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * @param          $value
+     * @param Donation $donation
+     *
+     * @return void
+     */
+    public static function update($value, Donation $donation)
+    {
+    }
```
JasonTheAdams, line 27: @pauloiankoski Why not make this an abstract function?

pauloiankoski: In case we have attributes that are read-only (eg: IdAttribute) we would need to include that method as empty as an abstract method requires implementation.

pauloiankoski: Resolved a73180607c598027e3d9f157968f03873e24668f

src/Donations/Endpoints/DonationUpdate.php
```diff
@@ -0,0 +1,144 @@
...
+        $donation = give()->donations->getById($id);
+
+        $updatedFields = [];
+
+        try {
+            foreach ($this->attributes as $attr) {
+                $attrId = $attr::getId();
+                $value = $request->get_param($attrId);
+
+                if (is_null($value) || ! in_array(AttributeUpdatesModel::class, class_implements($attr))) {
```
JasonTheAdams, line 101: @pauloiankoski This reads strangely and accesses more than is necessary. I suggest simplifying with an operator:
```php
if ( $value === null || !$attr instanceof AttributeUpdatesModel::class ) {
```

Also, would it make sense to allow the attribute to determine what it does will a null value? Wouldn't the attribute's validation catch and prevent values it won't accept? If so, the null check isn't necessary.

pauloiankoski: About the operator, that way you propose only works in case `$attr` was an object. In this case, it's just a static class.
CleanShot 2023-03-29 at 21 02 00

CleanShot 2023-03-29 at 21 01 39

--------------

Here, I'm looping through all registered attributes then checking if the user passed a value to it then maybe skip or update it. The validation only prevents the user to send invalid values.

CleanShot 2023-03-29 at 20 56 29

JasonTheAdams: Ah, in that case you can use [is_a](https://www.php.net/manual/en/function.is-a.php): `is_a($attr, AttributeUpdatesModel::class, true)`

pauloiankoski: That worked! 47290827573fa912eed765f98ea7c1626ab80a65

JasonTheAdams: > Here, I'm looping through all registered attributes then checking if the user passed a value to it then maybe skip or update it. The validation only prevents the user to send invalid values.

Ah, so this is intended to allow the front-end to send partial data? As in your screenshot, it would update the status and nothing else?

pauloiankoski: Yeah. That's correct. I just have noticed that the front-end is not passing only the dirty fields but all of them. As I created this route before knowing that, I made it in a way that it can receive partial data.

JasonTheAdams (approved):
Great work, @pauloiankoski! Thank you for taking all my feedback in stride. 😄

PR #6755, opened 2023-03-30
Fix: Update main key unserialization

kjohnson:
## Description

This PR updates how the main key is unserialized when importing donations.

## Testing Instructions

---
  - https://app.asana.com/0/0/1204316125792252

### DIFF
--- a/includes/admin/admin-actions.php
+++ b/includes/admin/admin-actions.php
@@ -709,7 +709,10 @@ function give_donation_import_callback() {
 	$import_setting['dry_run']     = $output['dry_run'];
 
 	// Parent key id.
-	$main_key = maybe_unserialize( $output['main_key'] );
+	$main_key = is_serialized( $output['main_key'] )
+		/** @unreleased Avoid insecure usage of `unserialize` when the data could be submitted by the user. */
+		? @unserialize( trim( $output['main_key'] ), ['allowed_classes' => false] )
+		: $output['main_key'];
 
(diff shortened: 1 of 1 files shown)


includes/admin/admin-actions.php
```diff
@@ -709,7 +709,10 @@ function give_donation_import_callback() {
 	$import_setting['dry_run']     = $output['dry_run'];
 
 	// Parent key id.
-	$main_key = maybe_unserialize( $output['main_key'] );
+	$main_key = is_serialized( $output['main_key'] )
+		/** @unreleased Avoid insecure usage of `unserialize` when the data could be submitted by the user. */
+		? @unserialize( trim( $output['main_key'] ), ['allowed_classes' => false] )
+		: $output['main_key'];
```
jonwaldstein, line 715: Looking at the suggestion and PHP [docs](https://www.php.net/manual/en/function.unserialize.php) I think it might be better to do json_encode for the input `name="main_key"` and do json_decode here

jonwaldstein (approved):

includes/admin/admin-actions.php
kjohnson: Resolving this, for now.

> [...] using allowed_classes is a good enough patch for now (until more time for refactoring/testing can be found)

Benunc (approved):
This passes manual tests (import still works just fine) and automated tests.

PR #6777, opened 2023-04-17
Fix: MoneyPHP support for Bitcoin currencies

### LINKED ISSUE #6744: Bitcoin support in Donations page and All Forms page
## User Story
When Bitcoin is defined as currency for Give, I can't see the donations on the new table (only on the legacy) and I can't access the "All  Forms" page due to a PHP error caused by moneyphp dependency.

## Expected Behavior

Give should be able to handle all the currencies it lists on the Currency tab.

## Steps to Reproduce

1. Go to Give settings on General > Currency and set Bitcoin as currency.
2. Go to a multi step form a make a donation with Bitcoin.
3. Then, go to Give "Donations" page and the new version of the table will show "There was a problem retrieving the donations".
4. Switch to the legacy view, and you will be able to see the donations.
5. Then, try to access the Give "All Forms" page, you will see a PHP error if you have your debugging enabled.

### Error in Donations page:
image

With legacy view, I can see the donations:
image

### Error in "All Forms" page
image

## System Information

<details><summary>Environment</summary>
WordPress: 6.1.1, from offical WP image for Docker (wordpress:php7.4-apache)
PHP: 7.4
Give: 2.52.2
</details>

## Acceptance Criteria

-   [x] Support for Bitcoin in Donations page
-   [x] Support for Bitcoin in All Forms page

JoshuaHungDinh:
Resolves #6744

## Description

This adds a custom aggregated list to include Bitcoin (BTC) in our currency facade. These changes are in relation to MoneyPHP and its ability to find "BTC" within its own list.

## Affects
Currencies

## Visuals

https://user-images.githubusercontent.com/75056371/232568056-d833283f-fc6f-4d7b-b6b3-7f6531e5a445.mov

## Testing Instructions

- Donations->Settings->Currency, Set currency to Bitcoin.
- Visit the Donation & the All Forms page.
- Verify the pages load without errors.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1203760890899570
  - https://app.asana.com/0/0/1204615971928969


src/Framework/Support/Facades/CurrencyFacade.php
```diff
@@ -113,4 +124,20 @@ public function getBaseCurrency(): Currency
...
+     *
+     * @unreleased
+     */
+    private function getCurrenciesList(): Currencies
+    {
+        return new AggregateCurrencies([
+            new ISOCurrencies(),
+            new BitcoinCurrencies(),
+            new CurrencyList([
+                'BTC' => 8,
```
JasonTheAdams, line 139: @JoshuaHungDinh The `BitcoinCurrencies()` doesn't have BTC??? I'm sure the answer is no if your'e adding this, I'm just shocked! 😅

JoshuaHungDinh: @JasonTheAdams Yea, BitcoinCurrencies() uses the official ISO code which is listed as XBT and not BTC.

Fun fact: The International Standards Organization (ISO) dictates that if a currency is not associated with a particular country, it should begin with an “X”. Bitcoin is standardized in the same way as gold, silver and palladium.

JasonTheAdams: Really!? Wow, that's pretty interesting. Learned something pretty interesting!

This does make me wonder, though. Does that mean that elsewhere, when `BTC` is being used as a currency, that's actually incorrect and should be `XBTC`?

JoshuaHungDinh: According to ISO, BTC is not a valid code. I suppose it's accepted worldwide as "BTC" because of its ticker symbol. You do bring up a good point. It is technically incorrect. A better option might be to update our references to the correct currency code rather than creating a custom list to accommodate an incorrect code. If we go that route, I do think we should handle updating the references in a separate PR and I can remove the custom list in this PR. This way we can still keep the aggregated list functionality but I defer to your call.

JasonTheAdams: You read my mind! I'd say let's add the `BitcoinCurrencies` list, remove the `BTC` one, and then correct where we're using BTC as a currency.

JasonTheAdams (approved):
Great work, @JoshuaHungDinh!

rickalday (approved):
It passes manual QA tests

PR #6779, opened 2023-04-19
Feature: add gateway payment and subscription controllers 

jonwaldstein:
## Description

This moves all the gateway _handle_ logic (as it relates to creating payments and subscriptions) into separate controllers.  This will significantly slim down the responsibility of the gateway class itself while giving us more control from the framework side.

## Affects

The gateway api.

## Visuals

N/A

## Testing Instructions

We need to make sure no gateways are affected.  Basically if one gateway using the gateway api works properly than they all should.

## Pre-review Checklist

-   [x] Includes unit tests

---
  - https://app.asana.com/0/0/1204449389516437

### DIFF
--- a/src/Framework/PaymentGateways/Controllers/GatewayPaymentController.php
+++ b/src/Framework/PaymentGateways/Controllers/GatewayPaymentController.php
@@ -68,6 +68,8 @@ class GatewayPaymentController
      */
     protected function handleGatewayCommand(GatewayCommand $command, Donation $donation)
     {
-        (new HandleGatewayPaymentCommand())($command, $donation);
+        $response = (new HandleGatewayPaymentCommand())($command, $donation);
+        
+        $this->handleResponse($response);
     }
 }
--- a/src/Framework/PaymentGateways/Controllers/GatewayPaymentController.php
+++ b/src/Framework/PaymentGateways/Controllers/GatewayPaymentController.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace Give\Framework\PaymentGateways\Controllers;
+
+use Give\Donations\Models\Donation;
+use Give\Framework\Exceptions\Primitives\Exception;
+use Give\Framework\FieldsAPI\Exceptions\TypeNotSupported;
+use Give\Framework\PaymentGateways\Actions\HandleGatewayPaymentCommand;
+use Give\Framework\PaymentGateways\Commands\GatewayCommand;
+use Give\Framework\PaymentGateways\Log\PaymentGatewayLog;
+use Give\Framework\PaymentGateways\PaymentGateway;
+use Give\Framework\PaymentGateways\Traits\HandleHttpResponses;
+use Give\PaymentGateways\Actions\GetGatewayDataFromRequest;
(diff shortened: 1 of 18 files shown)


src/Framework/PaymentGateways/PaymentGateway.php
```diff
@@ -497,7 +270,11 @@ public function callRouteMethod($method, $queryParams)
      */
     private function isFunctionImplementedInGatewayClass(string $methodName): bool
     {
-        $reflector = new ReflectionMethod($this, $methodName);
+        try {
+            $reflector = new ReflectionMethod($this, $methodName);
+        } catch (ReflectionException $e) {
```
JasonTheAdams, line 275: Let's add a log here as this would be quite unexpected.

src/Framework/PaymentGateways/Controllers/GatewaySubscriptionController.php
```diff
@@ -0,0 +1,80 @@
...
+     *
+     * @throws TypeNotSupported
+     * @throws Exception
+     */
+    public function handleGatewayCommand(
+        GatewayCommand $command,
+        Donation $donation,
+        Subscription $subscription
+    ) {
+        (new HandleGatewaySubscriptionCommand())($command, $donation, $subscription);
```
JasonTheAdams, line 78: Making this a single-use Action that terminates feels like a premature abstraction, no? Why not just put the action code here?

src/Framework/PaymentGateways/Controllers/GatewayPaymentController.php
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * @unreleased move logic into action
+     * @since 2.18.0
+     *
+     * @throws TypeNotSupported
+     * @throws Exception
+     */
+    protected function handleGatewayCommand(GatewayCommand $command, Donation $donation)
+    {
+        (new HandleGatewayPaymentCommand())($command, $donation);
```
JasonTheAdams, line 71: Same here. I'm just not following the benefit of breaking them out into terminating actions.

JasonTheAdams (changes requested):
Couple thoughts, but I love how much this simplifies the `PaymentGateway` code!

src/Framework/PaymentGateways/Controllers/GatewaySubscriptionController.php
jonwaldstein: Oh yeah, I meant to actually return commands from those actions so I can write unit tests for them.  I'll update those

src/Framework/PaymentGateways/PaymentGateway.php
jonwaldstein: resolved e237bd6e

src/Framework/PaymentGateways/Controllers/GatewaySubscriptionController.php
jonwaldstein: updated here 32f05b32 & additional tests for command handlers 2dacced7

PR #6812, opened 2023-05-25
Fix: insert/update fee amount recovered using the donation model

pauloiankoski:
## Description

The Donation Amount value was not correctly displayed when the Fee Amount Recovered property was modified using the Donation model. To fix this issue, this Pull Request updates the `_give_fee_donation_amount` meta with the original value of `_give_payment_total`. Then, it updates the `_give_payment_total` with the sum of `_give_fee_amount` and `_give_fee_donation_amount`.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Donations/Repositories/DonationRepository.php
+++ b/src/Donations/Repositories/DonationRepository.php
@@ -316,6 +316,7 @@ class DonationRepository
     }
 
     /**
+     * @unreleased update _give_fee_donation_amount meta with donation amount and add the recovered fee to donation total
      * @since 2.20.0 update amount to use new type, and add currency and exchange rate
      * @since 2.19.6
      */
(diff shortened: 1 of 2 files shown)


src/Framework/PaymentGateways/PaymentGatewayRegister.php
```diff
@@ -25,6 +26,21 @@ public function getPaymentGateways(): array
...
+    public function getNextGenPaymentGateways(): array
+    {
+        return array_filter(
+            $this->gateways,
+            static function ($gateway) {
+                return is_a($gateway, NextGenPaymentGatewayInterface::class, true);
+            }
+        );
+    }
+
```
jonwaldstein, line 43: @pauloiankoski the NextGen terminology is specific to our feature plugin so we don't want this to live in our core gateway registrar. Currently the `NextGenPaymentGatewayInterface` lives in the feature plugin as well.  Eventually we'll rename that and it will all live in core.

For now, you can just add this logic as a local method or a repository method within the feature plugin

jonwaldstein, line 43: @pauloiankoski the NextGen terminology is specific to our feature plugin so we don't want this to live in our core gateway registrar. Currently the `NextGenPaymentGatewayInterface` lives in the feature plugin as well.  Eventually we'll rename that and it will all live in core.

For now, you can just add this logic as a local method or a repository method within the feature plugin

jonwaldstein (changes requested):

src/Framework/PaymentGateways/PaymentGatewayRegister.php
pauloiankoski: Reverted 228518983269eb56f26a89f0a90bec128232500d

jonwaldstein:
We have decided to keep the fee recovery calculation in the add-on during form processing.  We may revisit this later if we feel the model should handle property calculations based on the fee amount.

PR #6813, opened 2023-05-29
Enhancement: add new Checkbox field and deprecate old one

JasonTheAdams:
## Description

Related FFM PR: https://github.com/impress-org/give-form-field-manager/pull/396

I'm aiming to have a set of Fields and Elements baked right into core. Some of them will be in the Form Builder in core, while others will have their blocks introduced in FFM. But the Field API node and Form Template API node will support them in core, so they can be utilized programmatically. This is useful so various add-ons (e.g. Fee Recovery) don't have to reinvent basic types over and over. Moreover, Form Designs will be able to rely on these basic types for styling.

In working on the base types, I realized that the Checkbox field was very much catered to how it works in Form Field Manager — that is, a single "checkbox" can actually have multiple values. I don't like this. A single Checkbox field should represent a single Checkbox input. It also didn't support things like indeterminate values. So I moved the existing Checkbox field to `LegacyNodes/GroupCheckbox` and made a new Checkbox class.

I also did a little bit of typing and other minor tweaks.

## Affects

This does affect FFM and will require changes to continue working with these changes. I tried to make it such that the old version won't completely brick the site, but instead should "only" 😅 break the form. It will be important that folks upgrade both GiveWP and FFM.

## Pre-review Checklist

-   [x] Includes unit tests


JasonTheAdams:
@jonwaldstein Please give this a look when you get a chance.

src/Framework/FieldsAPI/Checkbox.php
```diff
@@ -2,48 +2,112 @@
...
-     *
-     * @return bool
+     * @unreleased
+     */
+    public function indeterminate(): self
+    {
+        $this->defaultValue(null);
+
+        return $this;
+    }
```
jonwaldstein, line 78: indeterminate is a special checkbox's state, not it's value.

jonwaldstein (changes requested):
@JasonTheAdams I left a few comments about the checkbox field.  It looks like your mixing up a checkbox's value property with it's attributes.

jonwaldstein:
@JasonTheAdams personally, I would like to lean more toward how checkboxes work natively to avoid confusion.  I don't see a reason to prevent someone from using a value if that's how they want their checkbox to work.  We can always fallback to a boolean value if not specified.

JasonTheAdams:
@jonwaldstein
fine

JasonTheAdams:
@jonwaldstein This is ready for review again. I did a couple things:

First, I got rid of the whole indeterminate state thing. Turns out, as you indicated, indeterminate state is nothing more than a UI concept. Setting a checkbox as indeterminate doesn't actually affect the underlying checked/value state at all. As such, it doesn't make sense to introduce this on the checkbox level. It only makes sense for more complex UI's that we can handle later if it comes up.

Second, I added distinct `checked` and `value` state to the field, since checkboxes are like options in that the user doesn't set the value, they simply determine _which_ value — in the case of a checkbox it's the value or false.

Third, I overrided the `getDefaultValue` method since we're relying on it in React Hook Form to set up the initial state of things. We actually _shouldn't_ set the `checked` attribute, as RHF will do that for us based on the default value. And since we don't know the value until its set, it couldn't be done in the `checked()` method, as that would mean `value()` would have to always be set first, which is asking for trouble. So I decided it's just easier to ignore the `defaultValue()` method, as it's not really meaningful for a Checkbox, and instead override the getter.

src/Framework/FieldsAPI/Checkbox.php
```diff
@@ -2,48 +2,112 @@
...
+        }
+
+        if (!is_bool($defaultValue) && $defaultValue !== null) {
+            throw new InvalidArgumentException(
+                'Checkbox field only supports boolean or null value'
+            );
+        }
+
+        $this->defaultValue = $defaultValue;
+    }
```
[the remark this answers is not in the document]
JasonTheAdams: Resolved in [4a7603b](https://github.com/impress-org/givewp/pull/6813/commits/4a7603bad3891f702b9d1f639619a19277ebfae0)

src/Framework/FieldsAPI/Checkbox.php
```diff
@@ -2,48 +2,112 @@
...
      *
-     * @return $this
+     * @unreleased
      */
-    public function checked($isChecked = true)
+    public function checked(bool $checked = true): self
     {
-        $default = is_callable($isChecked) ? $isChecked() : $isChecked;
-        $this->defaultValue((bool)$default);
+        $this->defaultValue($checked);
```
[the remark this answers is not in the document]
JasonTheAdams: Resolved in [aabf870](https://github.com/impress-org/givewp/pull/6813/commits/aabf870dac9d944501e2b0aae952aaa11848983a)

src/Framework/FieldsAPI/Checkbox.php
JasonTheAdams: Good catch. This was an interesting lesson for me in exactly _what_ an indeterminate state is! I've removed this whole concept from the field.

src/Framework/FieldsAPI/Checkbox.php
```diff
@@ -2,48 +2,88 @@
 
 namespace Give\Framework\FieldsAPI;
 
+use Give\Framework\Exceptions\Primitives\RuntimeException;
+
 /**
- * @since 2.12.0
+ * A single checkbox input. If supported, an indeterminate value is represented as a null value, as opposed to true or
+ * false value.
```
jonwaldstein, line 9: Do we still need to mention "indeterminate" here?

jonwaldstein (approved):
@JasonTheAdams thanks for the updates, this looks great ✅

src/Framework/FieldsAPI/Checkbox.php
JasonTheAdams: Hah! Nope, good catch. Resolved in 6219e4831

PR #6820, opened 2023-06-05
Feature: add ShowInAdmin trait to fields api

jonwaldstein:
## Description

This adds a new `ShowInAdmin` trait to the fields api.

```
$field->shouldShowInAdmin() // false;
$field->showInAdmin(true);
$field->shouldShowInAdmin() // true;
```

## Affects

Existing fields

## Visuals

N/A

## Testing Instructions

The only way to test would be programmatically using the fields api.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Framework/FieldsAPI/Textarea.php
+++ b/src/Framework/FieldsAPI/Textarea.php
@@ -14,6 +14,8 @@ class Textarea extends Field
     use Concerns\HasLabel;
     use Concerns\HasPlaceholder;
     use Concerns\ShowInReceipt;
+    use Concerns\ShowInAdmin;
+    use Concerns\ShowInAdmin;
     use Concerns\StoreAsMeta;
     use Concerns\HasMaxLength;
     use Concerns\HasMinLength;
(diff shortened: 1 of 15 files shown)


JasonTheAdams:
@jonwaldstein Why not just put the trait on the `Field` base class? I would think even a hidden field can have the option (and would be false by default).

JasonTheAdams:
Hahahah! Nailed it. 😆

jonwaldstein:
@JasonTheAdams great timing, what about `StoreAsMeta` also?  That's being implemented per-field right now along with `ShowInReceipt`

JasonTheAdams:
Yup, I think both of these can live right on the `Field` class. I can't think of a scenario where they wouldn't at least be sensible in every field type. What makes a field a field is that it has a value.

src/Framework/FieldsAPI/Textarea.php
```diff
@@ -13,8 +13,8 @@ class Textarea extends Field
     use Concerns\HasHelpText;
     use Concerns\HasLabel;
     use Concerns\HasPlaceholder;
-    use Concerns\ShowInReceipt;
-    use Concerns\StoreAsMeta;
+
+
```
JasonTheAdams, line 17: Some formatting fun here

JasonTheAdams (approved):
Love how much this cleans up. 👍

PR #6823, opened 2023-06-07
Chore: add traits with auto getters and setters

pauloiankoski:
## Description

This pull request adds new traits to help create classes with multiple properties:
-  `HasGetters` generates `getProperty()` getters for all properties
-  `HasSetters` generates `setProperty('value')` setters for all properties
-  `HasGettersAndSetters` loads both traits, avoiding method name conflicts.

## Affects

N/A

## Pre-review Checklist

-   [ ] Includes unit tests


JasonTheAdams:
Nice, @pauloiankoski! I actually didn't know traits could use traits. That's cool!

As an API, I was imagining we'd do something like this:
```php
<?php

/**
 * @method foo(int $foo): self
 * @method getFoo(): int
 * @method getBar(): string
 * @method baz(string $baz): self
 */
class MyFancyClass {
  use HasGettersAndSetters;

  protected $foo; // can be set and get
  protected $bar; // can be get
  protected $baz; // can be set
  protected $trulyHidden;

  protected $getterProperties = ['bar'];

  protected $setterProperties = ['baz' => 'string'];

  protected $getterAndSetterProperties = ['foo' => 'int'];

}
```

This has a couple benefits:
1. Types can be enforced in the settesr
2. Only properties explicitly set to have getters/setters do.

cc: @jonwaldstein

JasonTheAdams:
I've also considered:
1. Adding `$gettersPrefix` and `$settersPrefix` protected properties, so a class can adjust the prefixes if they choose. By default, I think the getter should have a `get` prefix and setters should just be the property name.
2. Adding a `$settersAreFluent` protected (true by default) so classes can choose whether the setters return the class instance afterwards or not.

src/Framework/FieldsAPI/Concerns/HasSetters.php
```diff
@@ -0,0 +1,31 @@
...
+    public function __call($name, $arguments)
+    {
+        $prefix = substr($name, 0, 3);
+        $property = lcfirst(substr($name, 3));
+
+        if (!property_exists($this, $property)) {
+            throw new BadMethodCallException(sprintf(__('Property %s does not exist', 'givewp'), $name));
+        }
+
+        if ($prefix === 'set') {
```
kjohnson, line 23: The `set` prefix isn't necessary here. We have already established a convention for setters in the Fields API without the prefix.

kjohnson (changes requested):
@pauloiankoski can you add some unit tests for the getter and setter traits (as well as the HasGettersAndSetters trait)?

JasonTheAdams:
Oh yeah, as (genuinely) interesting as it was, @pauloiankoski, to have it split into multiple traits, I don't think it's really necessary if we're going the route of using properties to declare getter/setter proprties.

jonwaldstein:
> Nice, @pauloiankoski! I actually didn't know traits could use traits. That's cool!
> 
> As an API, I was imagining we'd do something like this:
> 
> ```
> <?php
> 
> /**
>  * @method foo(int $foo): self
>  * @method getFoo(): int
>  * @method getBar(): string
>  * @method baz(string $baz): self
>  */
> class MyFancyClass {
>   use HasGettersAndSetters;
> 
>   protected $foo; // can be set and get
>   protected $bar; // can be get
>   protected $baz; // can be set
>   protected $trulyHidden;
> 
>   protected $getterProperties = ['bar'];
> 
>   protected $setterProperties = ['baz' => 'string'];
> 
>   protected $getterAndSetterProperties = ['foo' => 'int'];
> 
> }
> ```
> 
> (Note: scroll to see the entire code snippet.)
> 
> This has a couple benefits:
> 
> 1. Types can be enforced in the settesr
> 2. Only properties explicitly set to have getters/setters do.
> 
> cc: @jonwaldstein

Looking at this, i'm concerned we're get into over-engineering territory and i'm genuinely wondering if it's any more convenient than adding the methods manually.  Between the different property syntaxes, their types, and the docblocks - at a certain point I think the manual methods - being a little more code redundancy - ends up simplifying this whole api.

pauloiankoski:
We have decided to keep manually adding getters and setters to our classes, as this pull request would add more complexity than actually help us. We may revisit this in the future when we support newer versions of PHP, which would provide better support for our intended purpose.

PR #6830, opened 2023-06-21
Refactor: update gateway api adapter for v3 forms

jonwaldstein:
Related: https://github.com/impress-org/givewp-next-gen/pull/203

## Description

This updates how we determine compatibility with gateways for v2 vs v3 forms.  This is achieved with a very simple check for two different methods that is handled by a new `supportsFormVersions()` method on the base gateway class.  

Along with this, offsite gateway compatibility was added to the existing legacy adapter so backwards compatible gateways don't have to know which form they are being used on.

**Updates**
- Added `supportsFormVersions()` method to the base PaymentGateway class & interface
- Added `$supportedFormVersion` param to  & `getPaymentGateways` that will filter gateways that support that specific form version.
- Removed implementation of `LegacyPaymentGatewayInterface` on the base `PaymentGateway` class
- Add 2 optional methods on the base `PaymentGateway` class to be overloaded by gateways including `enqueueScript` and `formSettings`.
- add `'successUrl'`, `'cancelUrl'`, `'failedUrl'` to `$gatewayData` in legacy adapter so gateways (offsite) can just refer to those values and not have to conditionally check the context they're in.

## Affects

- This should not affect existing gateways as we're not enforcing anything that does not already exist on the main gateway class.  
- By removing the `LegacyPaymentGatewayInterface` we are purposely removing the enforced method `getLegacyFormFieldMarkup` for v2 forms.

## Visuals

Screenshot 2023-06-29 at 3 19 40 PM

## Testing Instructions

Make sure all the existing gateways work properly

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1204938037848445


src/Framework/PaymentGateways/PaymentGateway.php
```diff
@@ -57,6 +55,38 @@ public function __construct(SubscriptionModule $subscriptionModule = null)
...
 
+    /**
+     * @unreleased
+     */
+    public static function isDeprecated(): bool
+    {
+        // To help with migration, gateways are considered deprecated until GiveWP 3.0 until they specify otherwise.
+        // TODO: determine if we should hardcode this as true and let gateways specify themselves until GIVE 4.0 or just assume all gateways are compatible with GIVE 3.0
+        return !(defined('GIVE_VERSION') && version_compare(GIVE_VERSION, '3.0.0', '>='));
+    }
```
jonwaldstein, line 66: This is worth a discussion on how to handle migrating gateways to support v3 forms.  In an ideal world, all gateways will be active unless specifying they are deprecated (not compatible with v3 forms).

To be clear, our framework decides how we handle deprecated gateways and for 3.0 if a gateway is marked as deprecated it will not work on v3 forms.  Backwards compatibility for v2 forms is determined by the use of the method `getLegacyFormFieldMarkup`.

You might be thinking: _What about the presence of a gateway script?_(`enqueueScript()` method)
Well, thinking about some gateways like offsite ones that won't necessarily need a js file - we don't want to require it.  So this should not be the source of truth for v3 compatibility

The problem is there are a mix of gateways in the wild and additional add-ons that have not been updated yet.

So we have 2 options.

1. Leave this as true for the duration of 3.0 by default and let gateways mark this as false individually to declare compatibility with v3 forms.  Then by 4.0 this would be false by default.  In other words, _all_ gateways are deprecated for 3.0 until they are updated.
2. Basically do the opposite and let gateways mark themselves as deprecated explicitly.  In other words, we would be assuming all gateways are compatible with 3.0 unless explicitly marked as deprecated.


Personally, I think option 1 would be the safest bet and easiest to manage migrating gateways.  The only real downside is having to overload the `isDeprecated()` method on each gateways that will eventually be false by default.  This would also be necessary for _new_ gateways to overload for the duration of 3.0.

jonwaldstein: @glaubersilva @kjohnson @JasonTheAdams would like to get all your thoughts on this.

jonwaldstein: **Update**

Talked to @JasonTheAdams about this and we came to the conclusion that it would be wise to start versioning our gateway api which just happens to correspond to our form versions.  Eventually these could diverge like v4 forms but gateways on api v3 would still work.  This would mean instead of deprecating gateways, they would announce what gateway api version they support.  This will come from either a version specific interface(s) or a simple array of version numbers.

jonwaldstein: We also want to go with option 1 which means all _existing_ gateways do not work on v3 forms until they explicitly define they do (api version)

src/Framework/PaymentGateways/Contracts/PaymentGatewayInterface.php
```diff
@@ -13,12 +13,13 @@
...
 {
+
     /**
-     * Return a unique identifier for the gateway
+     * Specify which gateway api versions the gateway supports
      *
-     * @since 2.18.0
+     * @unreleased
      */
-    public static function id(): string;
```
JasonTheAdams, line 21: Why are we getting rid of the `id` function?

src/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayAdapter.php
```diff
@@ -123,11 +130,63 @@ public function handleBeforeGateway(array $legacyDonationData, PaymentGateway $r
                 $donation
             );
 
+            /**
+             * Add success, cancel and failed URLs to gateway data.  This will be used in both v2 and v3 forms so gateways can just refer to the gateway data.
+             *
+             * @unreleased
```
JasonTheAdams, line 136: Why are we adding a `@since` tag here? Wouldn't this go on the function itself?

src/Framework/PaymentGateways/PaymentGateway.php
```diff
@@ -57,6 +56,42 @@ public function __construct(SubscriptionModule $subscriptionModule = null)
...
+    public static function supportsApiVersions(): array
+    {
+        $versions = [];
+
+        if (method_exists(static::class, 'getLegacyFormFieldMarkup')) {
+            $versions[] = 2;
+        }
+
+        return $versions;
+    }
```
JasonTheAdams, line 71: Could this simply be a static `$supportedApiVersions = [2]` property? Ideally, it would be great if this method simply knew how to identify a v3 gateway. I'm still considering a `PaymentGatewayV3` interface — even if its empty — that informs this function that it supports v3.

Otherwise, a property seems sufficient.

src/Framework/PaymentGateways/PaymentGatewayRegister.php
```diff
@@ -9,20 +9,31 @@
...
  */
 class PaymentGatewayRegister extends PaymentGatewaysIterator
 {
+    /**
+     * @var string[]
+     */
     protected $gateways = [];
 
     /**
      * Get Gateways
```
JasonTheAdams, line 23: Please update the description to explain how to use the `$apiVersion` parameter.

src/Framework/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayRegisterAdapter.php
```diff
@@ -67,30 +64,52 @@ static function (int $donationId, string $newStatus, string $oldStatus) use (
...
             $gatewaysData[$paymentGateway::id()] = [
-                'admin_label' => $paymentGateway->getName(),
+                'admin_label' => $this->getAdminLabel($paymentGateway),
                 'checkout_label' => $paymentGateway->getPaymentMethodLabel(),
-                'is_visible' => $this->supportsLegacyForm($paymentGateway),
+                'is_visible' => $this->supportsV2Forms($paymentGateway),
             ];
         }
+        $gatewaysData['offline']['admin_label'] .= " (v2)";
+        $gatewaysData['manual']['admin_label'] .= " (v2)";
```
JasonTheAdams, line 84: I think I'd like to also add a notice somewhere on the gateway page explaining what exactly this means. I think it will likely confuse people otherwise.

JasonTheAdams (changes requested):
Good work, @jonwaldstein! Left some more thoughts.

src/Framework/PaymentGateways/PaymentGateway.php
jonwaldstein: The only problem with properties is we can't enforce them in the `PaymentGatewayInterface`

src/Framework/PaymentGateways/Contracts/PaymentGatewayInterface.php
jonwaldstein: I think that was an accident

src/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayAdapter.php
jonwaldstein: Some of the since tags are inline with the filter here but I can move them up

src/Framework/PaymentGateways/PaymentGateway.php
jonwaldstein: So, i've been thinking about this and I think we need to have a further discussion.  Api versions will not always be additions, but more often - changes to the existing api.  A great example of this would be

**V2**
```
public function createPayment(Donation $donation, $gatewayData);
```

**V3**
```
public function createPayment(Donation $donation, array $gatewayData);
```

This brings up a few questions.
1a. Should each version interface specify _all_ the necessary gateway methods?
1b. If so, wouldn't gateways need to both extend the main gateway class and implement an api version?
1c. Along with this, the existing `PaymentGatewayInterface` would need to be v2 but is currently being implemented by the main gateway class which would cause conflicts for additional version interfaces being implemented.  Also, we have to think about the `SubscriptionModuleInterface` that this one extends.
2. Should we just create a new V3 namespace for a new base gateway class that implements the v3 interfaces?

We want to avoid breaking changes to existing gateways but also be able to enforce new api updates when opted in.  From the questionnaire above we can see where things get a little dicey with using interfaces to specify versions. Personally, I'm leaning toward `#2` to simplify things for gateways but that comes with it's own caveats as well.

JasonTheAdams: I was actually having the exact same thought upon reflection yesterday. If we're going to version, it makes sense to me to have an interface that reflects all the necessary methods for that version. A gateway _can_ implement both **if that is possible**. We also have things like the `static::id()` and `$this->getId()` that can be cleaned up (getting rid of the latter).

I'm cool with both having a v2/v3 interface _and_ a v3 namespace with its own base class that implements the v3 interface. The base class ultimately isn't meant to be necessary (implementing code should reference the interface, not the base class), but a super easy starting point that also sets up the subscription module system for us (that really isn't necessary).

src/Framework/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayRegisterAdapter.php
jonwaldstein: I added this message:

Screenshot 2023-06-29 at 3 14 51 PM

jonwaldstein:
@JasonTheAdams this is ready for re-review

includes/admin/settings/class-settings-gateways.php
```diff
@@ -344,58 +344,62 @@ public function render_enabled_gateways( $field, $settings ) {
...
+                    checked('1', $enabled, false),
+                    esc_html($option['admin_label'])
+                );
+                echo '</li>';
+            endforeach;
+            echo '</ul>';
+
+            echo '</div>'; // end gateway-enabled-wrap.
+
+            echo '<div style="padding-top: 1rem;"><p><sup>*</sup>(v2) This gateway is only supported on GiveWP version 2.x.x donation forms (aka v2 forms)</p></div>';
```
JasonTheAdams, line 395: Let's coordinate with Support/docs to have a link folks can click on to learn more about what this means on givewp.com.

src/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayAdapter.php
```diff
@@ -36,12 +39,17 @@ public function getLegacyFormFieldMarkup(
         array $args,
         PaymentGatewayInterface $registeredGateway
     ): string {
-        return $registeredGateway->getLegacyFormFieldMarkup($formId, $args);
+        if (method_exists($registeredGateway, 'getLegacyFormFieldMarkup')) {
+            return $registeredGateway->getLegacyFormFieldMarkup($formId, $args);
+        }
+        
+        return false;
```
JasonTheAdams, line 46: The function return type is `string`, but we're returning `false` here. That will cause a type error. We probably want to return an empty string? Or perhaps throw an exception?

src/PaymentGateways/Gateways/TestGateway/TestGateway.php
```diff
@@ -43,6 +44,15 @@ public function getName(): string
         return __('Test Gateway', 'give');
     }
 
+    /**
+     * @unreleased
+     */
+    public function enqueueScript(int $formId)
+    {
+        // temporary action to enqueue gateway scripts in feature plugin
```
JasonTheAdams, line 52: So is this going to be removed when the Feature Plugin is merged? If so, let's expand this out to explain that and leave a TODO.

src/PaymentGateways/Gateways/TestGateway/TestGatewayOffsite.php
```diff
@@ -32,6 +29,15 @@ class TestGatewayOffsite extends PaymentGateway
         'securelyReturnFromOffsiteRedirect'
     ];
 
+    /**
+     * @unreleased
+     */
+    public function enqueueScript(int $formId)
+    {
+        // temporary action to enqueue gateway scripts in feature plugin
+        do_action('givewp_donation_form_enqueue_test_gateway_offsite_scripts');
```
JasonTheAdams, line 38: Same thing here. Also, we're ultimately merging everything into a single test gateway, right?

JasonTheAdams (changes requested):
I think this is sufficient for now! I left a couple very minor suggestions.

includes/admin/settings/class-settings-gateways.php
JasonTheAdams: I don't think a doc exists at this point, so it may mean adding a task to the backlog and starting coordination with @Benunc.

src/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayAdapter.php
jonwaldstein: you can actually return false on a string type, but I think I can elevate this check up a level

jonwaldstein: moved this up a level to prevent hooking into action a31ffb66

src/PaymentGateways/Gateways/TestGateway/TestGateway.php
jonwaldstein: Yeah we need to setup wp scripts in givewp core so I just want to do that seperately, i'll make a note

jonwaldstein: resolved f4f97558

src/PaymentGateways/Gateways/TestGateway/TestGatewayOffsite.php
jonwaldstein: yeah possibly f4f97558

includes/admin/settings/class-settings-gateways.php
jonwaldstein: will do

src/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayAdapter.php
JasonTheAdams: That's only true if strict types aren't enabled, in which case it will coerce the value.

JasonTheAdams (approved):
Good work! This gives us what we need for an improved UX. 👍 

Since this is in core let's get this to QA before we merge.

Benunc (approved):
This passes all automated tests. Essentially that means that the following gateways have been tested here:

CleanShot 2023-07-05 at 16 20 32

I also tested on Payfast. 

Final tests on the release will catch any other gateways we have automations for, so this is good to merge for now.

jonwaldstein:
@Benunc thanks, I also fixed the issue where some gateway were not showing the label:

Screenshot 2023-07-05 at 5 47 37 PM

PR #6836, opened 2023-06-29
Security: remove uses of payment intent secret and remove existing meta

JasonTheAdams:
## Description

When auditing how API keys are stored within Stripe I found out that we were storing each [Payment Intent's client_secret](https://stripe.com/docs/api/payment_intents/object#payment_intent_object-client_secret). We really shouldn't be storing this as there are better ways of going about the process, and it's not recommended. After further inspection, I realized that it was only being used as a means of identifying a donation (by its secret) within a single function — and that function isn't actually used anywhere! I checked core, Next Gen, and the Stripe add-on.

Therefore, this PR remove that function, all places its being stored, and provides a migration to remove existing secret meta from the database.

## Affects

This is pretty purely a removal, so it shouldn't affect anything. I didn't have to refactor anything to make this possible.

## Testing Instructions

1. Make some donations via Stripe
2. Update to this version of GiveWP
3. Check the DB to make sure there's no donation meta with the `_give_stripe_payment_intent_client_secret` key
4. Make sure Stripe donations still work

## Pre-review Checklist

### DIFF
--- a/src/PaymentGateways/Gateways/Stripe/Migrations/RemovePaymentIntentSecretMeta.php
+++ b/src/PaymentGateways/Gateways/Stripe/Migrations/RemovePaymentIntentSecretMeta.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Give\PaymentGateways\Gateways\Stripe\Migrations;
+
+use Give\Framework\Database\DB;
+use Give\Framework\Migrations\Contracts\Migration;
+
+/**
+ * Removes the secret meta that was unnecessarily stored in the database for donations.
+ *
+ * @unreleased
+ */
+class RemovePaymentIntentSecretMeta extends Migration
+{
+    /**
+     * @inheritDoc
+     */
+    public static function id(): string
+    {
+        return 'remove_payment_intent_secret_meta';
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public static function title(): string
+    {
+        return __('Remove payment intent secret meta', 'give');
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public static function timestamp()
+    {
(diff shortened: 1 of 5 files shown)


src/PaymentGateways/Gateways/Stripe/Migrations/RemovePaymentIntentSecretMeta.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        return __('Remove payment intent secret meta', 'give');
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public static function timestamp()
+    {
+        return strtotime('2020-08-20 00:00:00');
```
kjohnson, line 36: Should this timestamp be updated?

kjohnson (approved):
Query looks good to me - confirmed locally.

Also, a GitHub search shows no uses of the `give_stripe_get_donation_id_by` function across the organization, see https://github.com/search?q=org%3Aimpress-org%20give_stripe_get_donation_id_by&type=code.

jonwaldstein:
@JasonTheAdams what's the status of this?

src/PaymentGateways/Gateways/Stripe/Migrations/RemovePaymentIntentSecretMeta.php
JasonTheAdams: Oops! Yup! Resolved in 917e82149

JasonTheAdams:
Oof, I apparently just dropped the ball on this one. Thanks for the poke, @jonwaldstein! I resolved Kyle's suggestion and sent this off to QA.

rickalday (approved):
Passed manual QA tests

PR #6839, opened 2023-07-03
Fix: Fee recovery amount not set in subscriptions

Genevieve-K:
The fee amount for subscriptions was set in update_subscription_args in class-give-fee-recovery-public.php. But the filter give_recurring_subscription_pre_gateway_args is not applied yet.  This code defines feeAmountRecovered for donations.

Another solution is to apply a filter like 'give_recurring_subscription_pre_gateway_args' in FormData. And the plugin Give Fee Recovery can define this element.

For donations, the function 'give_insert_payment' is still used and the action 'give_insert_payment' fired.

### DIFF
--- a/src/PaymentGateways/DataTransferObjects/FormData.php
+++ b/src/PaymentGateways/DataTransferObjects/FormData.php
@@ -164,6 +164,10 @@ final class FormData
                 $request['post_data']['give_anonymous_donation']
             );
 
+        $isFeeEnabled = isset($request['post_data']['give-fee-mode-enable']) ? 
+        filter_var( $request['post_data']['give-fee-mode-enable'], FILTER_VALIDATE_BOOLEAN ) : false;
+        $self->feeAmountRecovered = ($isFeeEnabled && !empty($request['post_data']['give-fee-amount'])) ? $request['post_data']['give-fee-amount'] : 0;
+
         $self->company = !empty($request['post_data']['give_company_name']) ? $request['post_data']['give_company_name'] : null;
 
         return $self;
@@ -180,6 +184,7 @@ final class FormData
             'status' => DonationStatus::PENDING(),
             'gatewayId' => $this->paymentGateway,
             'amount' => Money::fromDecimal($this->price, $this->currency),
+            'feeAmountRecovered' => Money::fromDecimal($this->feeAmountRecovered, $this->currency),
             'donorId' => $donorId,
             'firstName' => $this->donorInfo->firstName,
             'lastName' => $this->donorInfo->lastName,


kjohnson:
@jonwaldstein since this modifies the Payment Gateway FormData object, can you take a look?

jonwaldstein:
@Genevieve-K thank you for bring this up and creating this PR.  Based on your findings, I'm actually going to add a patch to the fee recovery plugin instead that hooks into the `givewp_subscription_creating` action because all the existing logic for fee-recovery in v2 forms is separated in there so it would cause the least amount of side effects.

I will report back to let you know when it's released. Thank you!

jonwaldstein:
@Genevieve-K this issue has been resolved in the fee-recovery add-on as of version `2.0.0` (just released today).  Thanks again for your help!

jonwaldstein:
@Genevieve-K yes the fix is in fee recovery.

Regarding the filters and actions - great question!  Those all live in a feature plugin [here](https://github.com/impress-org/givewp-next-gen) which holds all the upcoming code for GiveWP 3.0.  We are planning to have this plugin merged into GiveWP core in the near future.

Would love to get your feedback on the upcoming features, learn more here: https://givewp.com/lps/next-generation-donation-forms/

Genevieve-K:
@jonwaldstein If filters and actions are now in a new plug-in, how does it work on my local machine? I don't have this new plug-in installed.

jonwaldstein:
@Genevieve-K you can install the plugin by downloading the current zip from the [landing page](https://givewp.com/lps/next-generation-donation-forms/) or use the feature plugin github repo [releases](https://github.com/impress-org/givewp-next-gen/releases).  Eventually this plugin will be merged into GiveWP core.

Genevieve-K:
@jonwaldstein Thanks you, I will try to download the plug-in. My question is : why does it work without the new add-on?

Genevieve-K:
@jonwaldstein Thank you for your explanations.
I see a problem in Fee Recovery, I cannot suggest a PR. The upgrade AddFeeRecoveryMetaToRenewals does not work, the service provider has not the correct register function.

To change in give-fee-recovery/src/Migrations/ServiceProvider.php:
```
    use Give\Framework\Migrations\MigrationsRegister;

    public function boot()
    {  
       give(MigrationsRegister::class)
            ->addMigration(AddFeeRecoveryMetaToRenewals::class);
    }
```
Could you check please? You can remove the function register() in the AddFeeRecoveryMetaToRenewals class.

jonwaldstein:
@Genevieve-K thanks for bringing this up.  This migration did pass our QA process so it _was_ working at some point.  I wonder if there's something about your data this is not working for which would require more investigation.  Would you be able to submit this to our feedback board so that our support team can look into it? Thanks! https://feedback.givewp.com/

PR #6840, opened 2023-07-03
Fix: Fee recovery amount not set in subscriptions

Genevieve-K:
In Give\PaymentGateways\DataTransferObjects\FormData, I define feeAmountRecovered for donations. Here I copy this value for subscriptions. The fee amount is in the model for donations and subscriptions.

See https://github.com/impress-org/givewp/pull/6839


jonwaldstein:
Thanks again! Please see related comment about patch going in fee recovery plugin https://github.com/impress-org/givewp/pull/6839#issuecomment-1625514349

PR #6857, opened 2023-07-28
Feature: add check for application/json in headers accept

jonwaldstein:
## Description

This adds check for `application/json` in the headers accept in handleHttpResponses.

In v3 forms we will explicitly set the accept header as `application/json` in order to tell the server we want it to respond as json.

## Affects

This is being used in the gateway api and the v3 forms donation related controllers.

## Visuals

N/A

## Testing Instructions

This does not affect any existing functionality.  A related PR that used form-data will rely on this so can be tested alongside.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Framework/PaymentGateways/Traits/HandleHttpResponses.php
+++ b/src/Framework/PaymentGateways/Traits/HandleHttpResponses.php
@@ -68,9 +68,12 @@ trait HandleHttpResponses
      */
     protected function shouldSendAsJson(): bool
     {
-        return isset($_SERVER['CONTENT_TYPE']) && (str_contains(
-                    $_SERVER['CONTENT_TYPE'],
-                    "application/json"
-                ) || str_contains($_SERVER['CONTENT_TYPE'], "multipart/form-data"));
+        if (!isset($_SERVER['CONTENT_TYPE'])) {
+            return false;
+        }
+        
+        $contentType = isset($_SERVER['CONTENT_TYPE']);
+
+        return str_contains($contentType, "application/json") || str_contains($contentType, "multipart/form-data");
     }
 }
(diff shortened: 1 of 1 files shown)


src/Framework/PaymentGateways/Traits/HandleHttpResponses.php
```diff
@@ -60,4 +61,25 @@ public function handleExceptionResponse(\Exception $exception, string $message)
...
+    {
+        if (!isset($_SERVER['CONTENT_TYPE'])) {
+            return false;
+        }
+
+        $contentType = isset($_SERVER['CONTENT_TYPE']);
+
+        return str_contains($contentType, "application/json") || str_contains($contentType, "multipart/form-data");
+    }
 }
```
JasonTheAdams, line 85: @jonwaldstein I think we should check the [Accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) header instead and have the client use that header with `application/json`. Semantically, I don't think it's correct to assume that a `multipart/form-data` content request implies a JSON response. But the Accept header gives the client a way of explicitly specifying the response type.

jonwaldstein: @JasonTheAdams ah yeah thanks for the suggestion.  This was kind of bothering me too beacuse I know that we don't want to necessarily _assume_ the response should be json if using `multipart/form-data` I just figured that these are the two content-types we are using that should return json and didn't think about the accept header.  

So, I think your suggestion is the better approach because it's more of an explicit way for the client to specify what it wants in return (by default the fetch api is literally anything `*/*`).  

To be clear we would do this on the client:
``` 
fetch(url, {
  headers: {
    accept: 'application/json'
  }
})
```
 and on the server check `$_SERVER['HTTP_ACCEPT']` or even  `getallheaders()["Accept"]` for json right?

jonwaldstein: Updated here e9328f4b

JasonTheAdams (approved):
Yeah! Much better!

PR #6862, opened 2023-08-02
Feature: add `@wordpress/scripts` build process for GiveWP 3.0

jonwaldstein:
## Description

This adds `@wordpress/scripts` for building our v3 forms related scripts.  The entry point for these scripts are in `wepback.config.js`. 

## Affects

This only affects new scripts being added to `wepback.config.js`

## Visuals

N/A

## Testing Instructions

- `npm run dev:v3`
- `npm run build:v3`
- `npm run watch:v3`
- `npm run build` runs both mix and wp-scripts for production

## Pre-review Checklist

-   [ ] Includes unit tests


jonwaldstein:
@JasonTheAdams @kjohnson this is ready for review.

cc: @ravinderk this is how we'll be handling v3 related form scripts including new gateways.

.gitignore
```diff
@@ -61,3 +61,8 @@ phpcs.xml
 # POT File
 languages/*
 !languages/README.md
+
+# @wordpress/scripts build directory
+build
+**/build.*
+**/build.asset.*
```
JasonTheAdams, line 68: Should anything be added to the `.distignore` file?

src/Framework/Support/Scripts/Concerns/HasScriptAssetFile.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        $assets = $this->getScriptAsset($path);
+
+        return $assets['version'];
+    }
+
+    /**
+     * @see https://developer.wordpress.org/block-editor/reference-guides/packages/packages-dependency-extraction-webpack-plugin/#wordpress
+     *
+     * @unreleased
```
JasonTheAdams, line 29: Let's add a return typehint here:
```
@return array{dependencies: array<string>, version: int|string}
```

src/Framework/Support/Scripts/Concerns/HasScriptAssetFile.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Give\Framework\Support\Scripts\Concerns;
+
+trait HasScriptAssetFile {
```
JasonTheAdams, line 5: I'm tempted to say this would make a better `Assets` facade, rather than modifying the class the assets are used within. We're really just using this to make these functions available, as it's not even adding any properties, and a facade is better suited for that when we're not really trying to compose anything specific.

src/PaymentGateways/Gateways/TestGateway/testGateway.tsx
```diff
@@ -0,0 +1,50 @@
+import {__} from '@wordpress/i18n';
+//import type {FormData, Gateway} from '@givewp/forms/types';
+
+//TODO: update with types when they are available
+const gateway = {
+    id: 'test-gateway',
+    async beforeCreatePayment(values) {
+        if (values.firstName === 'error') {
+            throw new Error('Failed in some way');
+        }
```
JasonTheAdams, line 10: Oh, this is a nice little tool. Let's make sure this is documented in some way, maybe in the TestGateway class docblock.

JasonTheAdams (changes requested):
Good work, @jonwaldstein! I had a couple minor questions and one suggestion.

src/Framework/Support/Scripts/Concerns/HasScriptAssetFile.php
jonwaldstein: Resolved in 42a79ab1

src/Framework/Support/Scripts/Concerns/HasScriptAssetFile.php
jonwaldstein: Agreed, I updated with a new `ScriptAsset` facade here 42a79ab1

Since this is so specific to the dependency extraction and generated "asset" file, I made this specific to the script asset files hence the `ScriptAsset` name.

.gitignore
jonwaldstein: I don't think so, I assume we want to keep the build folder for distribution since it contains the generated scripts.

src/PaymentGateways/Gateways/TestGateway/testGateway.tsx
jonwaldstein: Yeah it's pretty nifty for testing error messages.  I've been thinking about doing more stuff like this to the test gateway to see different responses. I just added a comment here for now 41d7be7a

jonwaldstein:
@JasonTheAdams ready for re-review

.gitignore
kjohnson: Yeah, the `/build` folder probably needs to be included in the distribution, instead of trying to mess with moving that target directory.

JasonTheAdams: @jonwaldstein For sure! I'm really thinking about if there are any specific files (file maps?) within the build that shouldn't be included in distribution.

JasonTheAdams (approved):
Great work, @jonwaldstein! Nice to have this done for the merge!

.gitignore
jonwaldstein: Gotchya, it appears wp scripts does not generate those file maps so we should be good.  If anything looks weird during distribution we can add to the distignore.

PR #6865, opened 2023-08-03
Feature: add persistence scoping to Fields

### LINKED ISSUE #6861: Field nodes should have flexible storage options
## Details

Currently, our Field nodes are automatically opting into `StoreAsMeta` which makes the assumption that fields will either be storing their value as donor or donation meta.  Although this is nice and simple, it lacks flexibility and we are getting to the point where having options for storage would be very useful.

During our initial discussion, we talked about the concept of storage scopes and closures.

## Additional Context

Currently fields like FeeRecovery need to hook into `do_action('givewp_donate_controller_donation_created', DonateControllerData $formData, Donation $donation)`  or `do_action('givewp_donation_created', Donation $donation)` in order to customize their storage during form processing.  Furthermore, they don't have a way to opt _out_ of the default `StoreAsMeta` so their value will be saved in meta as well. 

Adding a way for the field to specify this logic directly would be a much more intuitive api, something like this (using closure):

```
$field = \GiveFeeRecovery\FormExtension\DonationForm\Fields\FeeRecovery::make('feeRecovery')
// This is just an example, store() does not exist (yet)
    ->store(function($value, $donation){
        $donation->feeAmountRecovered = Money::fromDecimal(
            $value,
            $donation->amount->getCurrency()
        );

        give()->payment_meta->update_meta(
            $donation->id,
            '_give_fee_donation_amount',
            give_sanitize_amount_for_db(
                $donation->amount->formatToDecimal(),
                ['currency' => $donation->amount->getCurrency()]
            )
        );

        $donation->amount = $donation->amount->add($donation->feeAmountRecovered);

        $donation->save();
    });
```

## Acceptance Criteria

JasonTheAdams:
Resolves #6861 

## Description

This PR introduces key features for more flexible storage capabilities in the Field API. It's awesome.

The existing persistence options was inferring the name of meta from the field name, and choosing whether the meta was stored in the donation or donor meta.

First, this introduces a new concept called the "scope" to a Field. A scope can be any string, and is mosts likely to be `donation` or `donor`. When the form is being saved, it will know what to do based on those two scopes (i.e. store them in their respective meta). But if it's any other value, then an add-on can hook in at that point and choose what to do — whether it's store the value or simply have a side-effect.

As a note, the existing `shouldStoreAsDonorMeta` API is fully backwards-compatible, so nothing that's using those methods needs to change.

A `metaKey` can now also be set for the field, so the field name and meta key can be two separate things. We'll likely have the field name be the fallback for donation and donor storage.

## Affects

This is a new API with changes that are only under the hood, so it should be safe. It does affect how a field determines whether it's stored as part of the donor or donation.

## Testing Instructions

Just make sure that custom field meta is still stored in the right place. I believe this is used in the code that persists the old FFM fields.

## Pre-review Checklist

-   [x] Includes unit tests

---
  - https://app.asana.com/0/0/1205225437697504


JasonTheAdams:
This is ready for review, @jonwaldstein!

src/Framework/FieldsAPI/Concerns/HasPersistence.php
```diff
@@ -0,0 +1,98 @@
...
+     */
+    public function getScope(): PersistenceScope
+    {
+        return $this->scope;
+    }
+
+    /**
+     * @unreleased
+     */
+    public function getScopeValue(): string
```
jonwaldstein, line 52: Let me know if i'm understanding how we would implement this...

```
$field->scope('my_custom_scope');
```

Donation Form Context
```
foreach($fields as $field){
  $scope = $field->getScopeValue();
  
  do_action('givewp_donation_form_persistence_scope_$scope}', $value, $donation);
}
```

```
add_action('givewp_donation_form_persistence_scope_my_custom_scope', function($value, $donation){
  // do something
});
```

Some Other Context like options?
```
foreach($fields as $field){
  $scope = $field->getScopeValue();
  
  do_action('givewp_options_persistence_scope_$scope}', $value);
}
```

```
add_action('givewp_options_persistence_scope_my_custom_scope', function($value){
  // do something
});
```

src/Framework/FieldsAPI/Concerns/HasPersistence.php
```diff
@@ -0,0 +1,98 @@
...
+     * @var string|null
+     */
+    protected $metaKey = null;
+
+    /**
+     * @unreleased
+     *
+     * @param string|PersistenceScope $scope
+     */
+    public function scope($scope): self
```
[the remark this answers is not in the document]
JasonTheAdams: Yeah, I thought about that, too. I think I'm learning towards `scope` as it's simple and I can't think of another concept it would clash with. I know I'm talking about persistence elsewhere, but to say the "scope" of a field is "fee recovery" feels sensible to me.

src/Framework/FieldsAPI/Concerns/HasPersistence.php
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * @unreleased updated to use scoping under the hood, no signature change
+     * @since 2.28.0 added types
+     * @since 2.10.2
+     */
+    public function shouldStoreAsDonorMeta(): bool
+    {
+        return $this->scope->isDonor();
+    }
```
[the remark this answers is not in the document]
JasonTheAdams: Yeah, I hear you. The biggest trick at this point is backwards compatibility. GiveWP core and FFM are both assuming the scope is donations by default, so to change that now would cause problems. When getting into settings and such, we can always ignore the scope or else we'd need to apply the scope to the fields.

src/Framework/FieldsAPI/Concerns/HasPersistence.php
JasonTheAdams: @jonwaldstein Yeah, for our donation form persistence, we'll have something like:
```php
public function storeCustomField(Field $field, $value) {
	switch($field->getScopeValue()) {
		case 'donation':
			// store donation meta
			break;

		case 'donor':
			// store donor meta
			break;

		default:
			do_action("givewp_donation_form_persist_field_scope_{$field->getScopeValue()}", $field, $value);
	}
}
```

jonwaldstein: @JasonTheAdams okay yeah that make sense. It's certainly flexible - my main hesitation is that this has the potential of being _too_ flexible - meaning someone can find the name of your intended scope and override it either knowingly or unknowingly.  The scope "custom" is the first thing that comes to mind.  Imagine folks with different fields trying to make custom scopes and each one is using the word "custom" - it would be quite confusing when they don't work properly. Being this is directly related to persisting the field, we want it to be flexible - but not have the potential for the processing to stop working because of a conflict like this.  You could of course argue that it's more of a responsibility on the implementation like the hook name, etc..

One other idea I was thinking about for this api is more like how the validation rules work.  Either you could use the build in strings `$field->scope('donation')` or instead, provide a custom invokable class `$field->scope(new FieldScope())`. Those scopes could implement a simple contract that would receive the `$field`, `$value`, and `$context`/`$data` (extra request data like the Donation model).  I know it sounds a little complex but I like the idea of scopes being more like a singleton.

JasonTheAdams: I've added the file and callback scopes per our discussion. 85dcb734d

rickalday (approved):
Passed manual QA tests

PR #6866, opened 2023-08-07
Feature: Add new field types to the Fields API

pauloiankoski:
## Description

This PR adds a new field for the Fields API (`MultiSelect`) and add changes to Date and Phone fields to adding a new attribute to store the formatting string.

## Pre-review Checklist

-   [ ] Includes unit tests


src/Framework/FieldsAPI/Date.php
```diff
@@ -15,4 +16,29 @@ class Date extends Field
     use Concerns\HasDescription;
 
     const TYPE = 'date';
+
+    /** @var string */
+    protected $dateFormat = '';
```
jonwaldstein, line 21: @pauloiankoski do you think we should add a default format instead of an empty string?

pauloiankoski: Added a default value reenforcing the same default value we have set for the template: `yyyy/mm/dd`

jonwaldstein: Cool, that makes sense.  Should we do the same for `phoneFormat`?

pauloiankoski: Nothing on `phoneFormat` is already an option. In the template, it'll renders a native `<input type="phone" />`.

jonwaldstein (approved):

PR #6875, opened 2023-08-15
Refactor: return the block-node relationships in the donation form schema

pauloiankoski:
## Description

This PR re-implements the block-node relationships array but sharing it through the `givewp_donation_form_schema` hook.

## Pre-review Checklist

-   [ ] Includes unit tests


src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
```diff
@@ -110,7 +114,10 @@ protected function convertInnerBlockToNode(BlockModel $block, int $blockIndex)
         if ($node instanceof Node) {
             $node = $this->mapGenericBlockAttributesToNode($node, $block);
 
-            Hooks::doAction('givewp_donation_form_block_converted_to_node', $node, $block);
+            $this->blockNodeRelationships[$block->clientId] = [
+                'block' => $block,
+                'node' => $node,
+            ];
```
JasonTheAdams, line 120: @pauloiankoski I think we may need to move this to the `__invoke` function, as Sections are also nodes that can be shown/hidden, so we'd want to include them in the node relationships.

JasonTheAdams (changes requested):
@pauloiankoski I had one thought while looking at this, though a minor one.

@jonwaldstein I'd love your thoughts on this. Looking at the FFM conditionals PR with Paulo, I suggested that we supply the node/block map right here and provide it via the hook, since it's clearly not easy (or possible?) to infer a block from the node, and (as in this case) it's useful to refer to the block again when reviewing the form.

src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
pauloiankoski: @JasonTheAdams I didn't include sections on that list because although sections can be shown/hidden, they will always use a field node as a conditional. Also, we don't accept sections to be used as they can't be compared to a value. In other words, we don't need to add sections to that array as those will never be used.

jonwaldstein:
> @jonwaldstein I'd love your thoughts on this. Looking at the FFM conditionals PR with Paulo, I suggested that we supply the node/block map right here and provide it via the hook, since it's clearly not easy (or possible?) to infer a block from the node, and (as in this case) it's useful to refer to the block again when reviewing the form.

@JasonTheAdams yeah, it took a little while for me to understand this - but Paulo walked me through it.  I was on the fence whether the relationship array should be something we supply to an action considering it's only used in form field manager and could potentially be a lot to store in memory.  I thought the block converted action was a good alternative - the downside being you have to build the relationship as a static property - but this certainly does makes it easier from the receiving end.

src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
JasonTheAdams: I can see why we didn't add it for FFM, but from the vantage point of the relationships, Sections are still a block being transformed into a node, so I think it's still worth keeping in the final relationships map.

JasonTheAdams:
@jonwaldstein Yeah, it's an interesting setup. I wasn't a fan of the static properties as, mostly, I'm not a fan of them in general within an instantiated class (as opposed to a static facade). Also, it would've had problems unless we reset it every time the event loop finished, if more than one form was converted at a time.

My thinking here was that in this _exact_ context, we are transforming a block collection into a Fields API form, so having the map is sensible and useful. Also, it really won't take up hardly any memory, as each block/node reference is simply a pointer to the instance in memory — we're not cloning the objects for the map, it's simply another way to find the instances.

I wouldn't do this elsewhere, but I think here it's sensible and useful.

src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
pauloiankoski: Makes sense.

Resolved: 04419f09c63000cfe19d45120c60a6dfdddd9a0b

src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
```diff
@@ -78,17 +82,24 @@ static function ($node) {
...
     {
-        return Section::make($block->getShortName() . '-' . $blockIndex)
+        $node = Section::make($block->getShortName() . '-' . $blockIndex)
             ->label($block->getAttribute('title'))
             ->description($block->getAttribute('description'));
+
+        $this->blockNodeRelationships[$block->clientId] = [
+            'block' => $block,
+            'node' => $node,
+        ];
```
JasonTheAdams, line 100: @pauloiankoski Let's either escalate this to `__invoke` or create a `mapBlockToNode` private function to preserve consistency.

pauloiankoski: d7f8074c972554cc24c31ff497df3c97a0f6e629

Extracted the duplicated code to a private function. Also, rearranged the order of the args of `mapGenericBlockAttributesToNode` so they reflect better the order (block then node).

JasonTheAdams (approved):
Great work, @pauloiankoski! I believe this will be useful in the long run.

I defer to @jonwaldstein as when to merge this, as it looks like the base branch is what's causing the failed static analysis.

jonwaldstein:
> Great work, @pauloiankoski! I believe this will be useful in the long run.
> 
> I defer to @jonwaldstein as when to merge this, as it looks like the base branch is what's causing the failed static analysis.

@pauloiankoski this is good to go, we'll fix the static analysis on the epic branch.

PR #6887, opened 2023-08-17
Feature: Support group node visibility and polymorphism

JasonTheAdams:
### Group Child Visibility

This PR tackles a couple of issues on the journey of adding visibility support for group child nodes — such as `name.lastName`.

First, nodes, no matter where they are in the form AST, can now have visibility conditions. This originally caused issues with triggering an infinite re-render, as the group was re-rendering, which dumped the children, which meant when they rendered their `useEffect` was triggered, which unregistered invisible fields, which caused the whole thing to re-render. The solution to this was to memo the node components. This might sound drastic, but is actually quite sensible due to a core principle of our forms:
> When it comes to the form and its nodes, the props don't change, just the state does.

When the server passes the form structure to the page, it's effectively immutable and never changes. Because of this, applying memo to the node components is perfectly fine as the props aren't meant to change. This doesn't mean the components won't render; state will still trigger a re-render, props won't.

There are actually two happy benefits of this:
- Form performance is improved
- If an input is being watched by another input, it won't lose focus when visibility conditions are triggered

### Polymorphic group children

Next, while working on the `GroupNode` component it was clear that groups only supported fields — elements and groups were simply ignored. Oops. In the process of giving the nodes visibility capabilities, it also opened groups up to child elements and groups.

### Bonus fix

Finally, while messing with form fields via the API, I found that `Form::remove()` removed the node without resetting the underlying array, losing the numeric order of the array. So that was fixed.

## Affects

While making group nodes polymorphic I did change the shape of the `GroupProps` to indicate nodes instead of fields. If groups are used anywhere outside of core, they'll need to update to this. I checked FFM and there weren't any.

## Testing Instructions

Apply visibility conditions to a group child node, something like:
```js
    $form->getNodeByName('lastName')
	->showIf('firstName', '!=', 'John');
```

## Pre-review Checklist

-   [x] Includes unit tests


JasonTheAdams:
@kjohnson @jonwaldstein Since you both helped me today, I'd appreciate a review from either or both of you. Kyle, please let me know if something here doesn't fit what you and @alaca need for Tributes!

JasonTheAdams:
@kjohnson If this looks like it will solve your Tribute needs, then we can merge this!

JasonTheAdams:
> @JasonTheAdams great job! working great. As a side note, your testing example did not submit correctly because the donate controller is expecting a string for lastName - but that can be fixed up the chain.

Hahah! Yeah, I figured it would break something. I don't actually suggest people do that in production. 😆

kjohnson (approved):
This seems to do the job for the Tributes Group!

PR #6892, opened 2023-08-21
Fix: Initialization/handling of nullable values in templates

pauloiankoski:
## Description

This PR addresses two issues associated with nullable values in templates.
- From now on, a template will be initialized with either a `defaultValue` provided through the FieldsAPI or an empty string. Previously, only fields with a `defaultValue` attribute were initialized with that value. This caused other templates to initialize its value attribute as `undefined`.
- I've modified the `onchange` method for the `Date` template, so it now saves the value even when empty. Before this change, the field wouldn't accept an empty string as a valid value, making it impossible to remove a selected date.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/DonationForms/resources/app/utilities/getDefaultValuesFromSections.ts
+++ b/src/DonationForms/resources/app/utilities/getDefaultValuesFromSections.ts
@@ -13,9 +13,7 @@ export default function getDefaultValuesFromSections(sections: Section[]) {
     return reduceFields(
         sections,
         (values, field) => {
-            if (field.defaultValue !== null) {
-                values[field.name] = field.defaultValue;
-            }
+            values[field.name] = field.defaultValue ?? '';
 
             return values;
         },
--- a/src/DonationForms/resources/registrars/templates/fields/Date.tsx
+++ b/src/DonationForms/resources/registrars/templates/fields/Date.tsx
@@ -26,9 +26,7 @@ export default function Date({Label, ErrorMessage, description, dateFormat = 'yy
                         ref={ref}
                         dateFormat={dateFormat}
                         selected={value && parse(value, dateFormat, new window.Date())}
-                        onChange={(date) => {
-                            onChange(format(date, dateFormat));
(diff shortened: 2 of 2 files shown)


src/DonationForms/resources/app/utilities/getDefaultValuesFromSections.ts
```diff
@@ -13,9 +13,7 @@ export default function getDefaultValuesFromSections(sections: Section[]) {
     return reduceFields(
         sections,
         (values, field) => {
-            if (field.defaultValue !== null) {
-                values[field.name] = field.defaultValue;
-            }
+            values[field.name] = field.defaultValue ?? '';
```
jonwaldstein, line 16: @pauloiankoski I had to remember why we did this which conflicts with what your trying to do.  Let me explain:

The default values hold the key / value pairs of all our field names and their default values.  When the fields get registered by RHF, it uses what we define here to register the field.  I think you already understand that part.  We also intentionally exclude the field key in this if there is no default value.

There are 2 tricky pieces to this:
1. We can't assume that every input accepts (or wants) a string for a default value
2. On multi-step forms - each step uses the cascading default values - so its important that default values without a value are omitted.

So I think what we had originally is what we want to keep and should try handling undefined default values on a per-template basis - since currently it seems to only be problematic for controller based fields like the date one.

jonwaldstein (changes requested):

src/DonationForms/resources/app/utilities/getDefaultValuesFromSections.ts
pauloiankoski: Replaced the `Controller` component we had there with a hidden input.

jonwaldstein (changes requested):
@pauloiankoski this is great! one more minor suggestion - and could you please update the PR description to describe the new direction? Thanks!

jonwaldstein (approved):
Great job!

PR #6893, opened 2023-08-21
Fix: default values for conditional fields

### LINKED ISSUE #6891: Conditionally shown fields do not apply default values
## Details

When a field is show conditionally (after initially being hidden) the default value for that field is not applied.

## Expected Behavior

Default values should be applied when a field is conditionally shown, since the

## Steps to Reproduce

1. Add a Radio field
2. Add another Radio Field with a default value
3. Conditionally show the second Radio field based on the first

## Visuals

image

JasonTheAdams:
Resolves #6891 

## Description

This addresses an issue where default values weren't working for fields that were displayed conditionally. When the field finally displayed the default value wasn't applied, so it defaulted to a null value.

The underlying issue here is that default values are actually set when the form first renders, but if a field isn't set to display then its default value isn't set. It's important _not_ to include it in the default values as otherwise it will fail validation before it displays.

The strategy taken here is to check if the field has a non-null and non-undefined value the first time a field is rendered. This does mean an unnecessary check is made for all the fields whose default value was applied, but it should simply skip those and call the `setValue` function. When a field is finally rendered and it has a default value, the default value is applied.

## Affects

This adds a side-effect to all fields in the donation form.

## Testing Instructions

The following can be used to add two radio fields to simulate the issue in the `givewp_donation_form_schema` hook:
```php
$form->getNodeByName('section-2')
    ->append(
        \Give\Framework\FieldsAPI\Radio::make('t-shirt-size')
            ->label('T-Shirt Size')
            ->defaultValue('medium')
            ->showInReceipt()
            ->options(
                ['small', 'Small'],
                ['medium', 'Medium'],
                ['large', 'Large'],
                ['x-large', 'X-Large']
            )
            ->required(),
        \Give\Framework\FieldsAPI\Radio::make('favorite-color')
            ->label('Favorite Color')
            ->defaultValue('red')
            ->showInReceipt()
            ->options(
                ['red', 'Red'],
                ['green', 'Green'],
                ['blue', 'Blue']
            )
            ->showIf('t-shirt-size', '=', 'small')
            ->required()
    );
```

## Pre-review Checklist

### DIFF
--- a/src/DonationForms/resources/app/fields/FieldNode.tsx
+++ b/src/DonationForms/resources/app/fields/FieldNode.tsx
@@ -3,17 +3,34 @@ import {useTemplateWrapper} from '../templates';
 import registerFieldAndBuildProps from '../utilities/registerFieldAndBuildProps';
 import type {FieldProps} from '@givewp/forms/propTypes';
 import memoNode from '@givewp/forms/app/utilities/memoNode';
+import {useEffect} from 'react';
 
 const formTemplates = window.givewp.form.templates;
 
 function FieldNode({node}: {node: Field}) {
-    const {register} = window.givewp.form.hooks.useFormContext();
+    const {register, getValues, setValue} = window.givewp.form.hooks.useFormContext();
     const {errors} = window.givewp.form.hooks.useFormState();
     const Field =
         node.type !== 'hidden'
             ? useTemplateWrapper<FieldProps>(formTemplates.fields[node.type], 'div', node.name)
             : formTemplates.fields[node.type];
     const fieldProps = registerFieldAndBuildProps(node, register, errors);
+
+    /**
+     * Set the default value for the field if it is not already set. This is necessary because the default value is not
+     * applied to the initial form render if it requires conditions to be visible.
+     */
+    useEffect(() => {
+        if (node.defaultValue === undefined) {
+            return;
+        }
+
+        const value = getValues(node.name);
+
+        if (value === undefined || value === null) {
+            setValue(node.name, node.defaultValue);
+        }
+    }, []);
 
     return <Field key={node.name} {...fieldProps} />;
 }


JasonTheAdams:
Mind giving this a look over, @kjohnson? I'm assuming it tackles the issue?

@jonwaldstein Mind giving this small PR a look, too? I'm curious if you can think of a better way to apply the default value with RHF.

JasonTheAdams:
@jonwaldstein An alternative here would be to check and apply the default value in the `useEffect` of `SectionNode` when `showNode` is true. The upside here is I wouldn't have to check the value, as it doesn't matter, and would simply apply the default value if there is one. It also wouldn't have an side-effect for every field. 🤔

JasonTheAdams:
@jonwaldstein I went ahead and tried out the method I described above. It works. The only somewhat shady thing being done at this point is that we're setting the value before the field is registered, which the [setValue](https://react-hook-form.com/docs/useform/setvalue) docs suggest not doing. But that can't really be helped, unless we move this back into the `FieldNode` component.

This is probably six of one and half-a-doze of the other. Both seemed to work just fine without performance problems.

(Kyle, you're welcome to weigh in if you have thoughts. I'm specifying Jon as he and I have spent more time poking at this part of the code. 😄)

src/DonationForms/resources/app/fields/SectionNode.tsx
```diff
@@ -19,10 +19,14 @@ const formTemplates = window.givewp.form.templates;
...
     const showNode = useVisibilityCondition(node.visibilityConditions);
-    const {unregister} = window.givewp.form.hooks.useFormContext();
+    const {unregister, setValue} = window.givewp.form.hooks.useFormContext();
 
     useEffect(() => {
         if (showNode) {
+            if (isField(node) && node.defaultValue !== undefined) {
+                setValue(node.name, node.defaultValue);
+            }
+
```
jonwaldstein, line 29: @JasonTheAdams I'm going to suggest a different strategy that should simplify this:

```
unregister(node.name, {keepDefaultValue: true});
```

When the form first renders, we are handing over the default values for every field - regardless of their visibility conditions - so they already exist in the eyes of RHF.  

When a field then gets unregistered (even initially) it appears RHF is removing it's default value from that initial state.  Using the [unregister](https://react-hook-form.com/docs/useform/unregister) option `keepDefaultValue` we're effectively telling RHF to preserve that default value of that field - and use it during registration (when the field is rendered).

jonwaldstein: [Confirmed](https://github.com/react-hook-form/react-hook-form/blob/5f3df1227c381a9127ab27f7d1d98a9cefee33d5/src/logic/createFormControl.ts#L908), when calling unregister - RHF removes the default value from context using the field name.

When calling register - it [pulls](https://github.com/react-hook-form/react-hook-form/blob/5f3df1227c381a9127ab27f7d1d98a9cefee33d5/src/logic/createFormControl.ts#L257C13-L257C25) the default value from context.

JasonTheAdams: And in swoops @jonwaldstein, taking this PR down to 3 lines. 😆

JasonTheAdams:
@jonwaldstein Ok, well that was ridiculously simple. Lesson learned. Hahah! Mind giving this bad boy a final review?

jonwaldstein (approved):
Perfect!  I tested this locally (including multi-step) and everything is working properly - including conditional validation ⚡

kjohnson (approved):
Awesome!

PR #6895, opened 2023-08-22
Fix: remove excess form builder padding on classic themes

JasonTheAdams:
## Description

I finally figured out why some folks were seeing weird extra padding in the form builder and others weren't. It had to do with which theme was loaded. Apparently, classic themes — that is, not block themes — load an additional CSS file in the builder: `wp-includes/css/dist/edit-post/classic.css`. This was causing extra padding to occur on our blocks.

I thought about dequeuing the style altogether, and even tried, but decided to leave it alone in case it's being relied on in some way we're not aware of. So instead, I just added a small bit of CSS which attempts to undo the specific styles causing issues.

## Affects

Blocks in the form builder

### Before

image

### After

image

## Testing Instructions

1. Activate the Twenty Twenty-One theme
2. View the builder
3. Recoil in horror
4. Update to this PR
5. Check builder again
6. Nod calmly

## Pre-review Checklist


JasonTheAdams:
@JoshuaHungDinh Mind giving this a review? Thanks again for your help!

JoshuaHungDinh (approved):
@JasonTheAdams, scary stuff! I double checked it's looking great now.

PR #6898, opened 2023-08-23
Fix: keep the same meta key and email tag for custom fields after updating the form

glaubersilva:
## Description

This PR fixes a bug that changes the meta key (and tag email) of a custom field when we update the form, even if we not changing anything on the custom field.

For example, if we create a custom field called "Favorite Color", the `meta_donor_favorite_color` displayed on the screen is saved on DB as `favorite_color` the first time. 

After that, when we update the form, the `favorite_color_1` is used instead of the `favorite_color` meta key

But if we use the `meta_donor_favorite_color` as an email tag, the first value set on the `favorite_color` meta key will always be displayed, and the updates stored on `favorite_color_1` will never be displayed on the emails.

For example, if I set my favorite color as BLACK on donation 1, I'll get an email with BLACK in the body message, but if I set my favorite color as RED on donation 2, after a form update, the email will keep saying that my favorite color is  BLACK instead of RED.

## Affects

Meta keys and email tags.

## Visuals

different-meta-keys

colors

## Testing Instructions

1. Create a form with a custom field called "Favorite Color" and set it to be stored as a donor meta key
2. Edit the "New Donation" email and add the meta key created in the previous step as an email tag on the body message - the syntax for that should be something like this: `{meta_donor_favorite_color}`
3. Create a test donation and set the favorite color as BLACK, you should see this same color on the confirmation page and on the "New Donation" email sent to your inbox
4. Update anything on the form created on the first step
5. Repeat step three using another color and **the same user/donor**, then check if the color value changes on the confirmation page and on the "New Donation" email sent to your inbox

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205339263512644
  - https://app.asana.com/0/0/1205198235960347

### DIFF
--- a/src/FormBuilder/resources/js/form-builder/src/supports/field-settings/FieldSettingsHOC.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/supports/field-settings/FieldSettingsHOC.tsx
@@ -61,7 +61,7 @@ const generateEmailTag = (fieldName, storeAsDonorMeta) => {
 function FieldSettingsEdit({attributes, setAttributes, fieldSettings}) {
     const validateFieldName = useFieldNameValidator();
     const [hasFieldNameAttribute, setHasFieldNameAttribute] = useState(attributes.hasOwnProperty('fieldName'));
-    const [isNewField] = useState(hasFieldNameAttribute);
+    const [isNewField] = useState(!hasFieldNameAttribute);
 
     const updateFieldName = useCallback(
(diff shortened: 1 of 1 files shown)


JasonTheAdams (approved):
Oops, yeah, that logically makes sense. I wrote this, and `isNewField` should only be true when the field does not yet have a field name attribute.

rickalday (approved):
Passed manual QA tests.

PR #6899, opened 2023-08-24
Fix: Update checkbox control placeholder style selector for the Terms and Conditions block

JoshuaHungDinh:
## Description

Adds a classname to the CheckboxPlaceholder control for the terms and conditions block. Specific styling for the placeholder is needed for the label and a general label selector was used, resulting in conflicts.

## Affects

Terms and conditions block.

## Visuals

Screen Shot 2023-08-24 at 11 29 23 AM

## Pre-review Checklist


src/FormBuilder/resources/js/form-builder/src/blocks/fields/terms-and-conditions/Edit.tsx
```diff
@@ -234,7 +234,12 @@ function CheckboxPlaceholder({label, linkText, isFormDisplay, agreementText}: Ch
                     border: 'none',
                 }}
             >
-                <CheckboxControl label={label} onChange={null} disabled={true} />
+                <CheckboxControl
+                    className={'give-terms-and-conditions-control'}
+                    label={label}
```
jonwaldstein, line 239: let's make sure to use the `givewp` prefix instead of `give`.

```suggestion
                    className={'givewp-terms-and-conditions-control'}
                    label={label}
```

src/FormBuilder/resources/js/form-builder/src/styles/_components.scss
```diff
@@ -208,10 +208,12 @@ This creates a consistent separator between the tabs and the rest of the sidebar
...
-    font-size: 1rem;
-    font-weight: 500;
-    color: #595959;
+.give-terms-and-conditions-control {
+    .components-checkbox-control__label {
+        font-size: 1rem;
+        font-weight: 500;
+        color: #595959;
+    }
 }
```
jonwaldstein, line 217: Lets actually move this to live inside the terms block folder and just import it into the `Edit.tsx` file.  This is my preferred way of handling component styles with our build process - because importing styles ends up producing the same outcome (one stylesheet) and then the styles get to live alongside their implementation which is easier to see/find.  

I realize our current styles folder conflicts with that mentality but that's the direction I would like to move us toward.

jonwaldstein (changes requested):

src/FormBuilder/resources/js/form-builder/src/styles/_components.scss
JoshuaHungDinh: Awesome that makes sense. When things start to cool down, I'll work on migrating things into there directories & out of inline as well.

src/FormBuilder/resources/js/form-builder/src/App.scss
```diff
@@ -4,3 +4,5 @@
 @import './styles/email-settings';
 @import './styles/components';
 @import './styles/onboarding';
+
+@import './blocks/fields/terms-and-conditions';
```
jonwaldstein, line 8: @JoshuaHungDinh you don't need this import since your importing styles directly into the block's edit component.  it _should_ be included in the final stylesheet.

JoshuaHungDinh: @jonwaldstein, hmm just importing into edit.tsx is not applying the styles for me.

src/FormBuilder/resources/js/form-builder/src/blocks/fields/terms-and-conditions/Edit.tsx
JoshuaHungDinh: Will do!
Resolved: 5b961dbf78f36fc6c56253b7212a7608d3d3b6d1

jonwaldstein (approved):
@JoshuaHungDinh turns out we were not enqueuing the bock registrar styles - I updated that and the block styles work now.

JoshuaHungDinh:
ah got it, thanks @jonwaldstein!

PR #6903, opened 2023-08-25
Feature: Add Stripe account selection as a per-form setting

pauloiankoski:
## Description

This PR introduces the ability to specify an account that differs from the default one on a per-form basis. Includes a new step for the migration process.

### Payment Gateways Block
- **New Attribute**: Added `stripeUseGlobalDefault` and `stripeAccountId` to store settings for Stripe.
- **Stripe Component Enhancements**: 
  - Added a new panel to the inspector sidebar containing three elements: a toggle, a dropdown, and a notice.
  - "Use global default" toggle: Always visible.
  - "Choose account" dropdown: Visible when `useGlobalDefault` is false.
  - "No default account set" notice: Displayed when no global default account is configured and no per-form account is selected.
- **Additional Actions**: When the Donation Form is saved, we update the `give_stripe_per_form_accounts` and `_give_stripe_default_account` meta fields based on user selections. This way, we leverage existing methods to automatically load the appropriate account for each form.

## Visuals

CleanShot 2023-08-24 at 22 33 08
_Not having a global default account_

CleanShot 2023-08-24 at 22 28 14
_Having a global default account_

## Testing Instructions

- Configure two or more Stripe accounts setting one of them as default
- Add a new v3 form
- Open the form and check if your default account is the one being presented on Stripe.
Eg.: _By providing your card information, you allow **GiveWP Test** to charge your card for future payments in accordance with their terms._
- Edit the form setting another account manually through the per-form settings
- Open the form and check if the new account is now being presented as in the example above.

## Pre-review Checklist

-   [x] Includes unit tests

---
  - https://app.asana.com/0/0/1205367244891340


src/PaymentGateways/Gateways/Stripe/StripePaymentElementGateway/StripePaymentElementRepository.php
```diff
@@ -164,7 +165,29 @@ protected function getStripePublishableKey(int $formId): string
...
+            )) {
+            return $defaultAccountId;
+        }
+
+        $allStripeAccounts = give_stripe_get_all_accounts();
+        if (!array_key_exists($stripeSettings['accountId'], $allStripeAccounts)) {
+            return $defaultAccountId;
+        }
+
+        return $stripeSettings['accountId'];
```
jonwaldstein, line 190: @pauloiankoski i'm going to suggest we do this a little differently...  

If this was a brand new feature for Stripe I think this would be fine (although this is the first time a gateway is referencing the blocks which feels a little weird lol), but since this functionality already exists and we're kind of piggy backing off the existing functionality & still have v2 forms - i'd rather make this fully backwards compatible. So here's what i'm thinking:

Hook into `givewp_form_builder_updated` and use the logic you have here to convert the block attributes to the existing give stripe meta keys referenced in `give_stripe_get_connected_account_id()`

Then, we're really just using the block attributes as an adapter for the existing meta and this function `give_stripe_get_connected_account_id($formId)` would still work.

src/FormBuilder/resources/js/form-builder/src/styles/_components.scss
```diff
@@ -155,6 +155,32 @@ This creates a consistent separator between the tabs and the rest of the sidebar
...
+    h4 {
+        color: var(--givewp-grey-900);
+        margin: 0 0 var(--givewp-spacing-2);
+    }
+
+    .components-notice__action.is-link {
+        color: var(--givewp-grey-900);
+        margin-left: 0;
+    }
+}
```
jonwaldstein, line 182: Is this for this settings specifically?  If not it will be applied to all notice components.  If it's specific to this setting I would suggest importing a standalone stylesheet into the block component.

src/FormBuilder/resources/js/form-builder/src/blocks/fields/payment-gateways/settings.tsx
```diff
@@ -13,9 +13,13 @@ const settings: FieldBlock['settings'] = {
     },
     attributes: {
         lock: {remove: true},
+        gatewaysSettings: {
+            type: 'object',
+            default: {},
+        },
```
jonwaldstein, line 19: hmm, I think this might be an early abstraction.  With a big gatewaySettings object, we get ourselves into a big unknown shape because all gateways are different.  We also risk a rogue gateway overriding the entire attribute object and causing issues for other gateways.  In that light, I would prefer specific gateway attributes with clear motives and types like this:

```
stripeUseGlobalDefaults: {
  type: 'boolean'
  default: true
}
```

src/FormBuilder/resources/js/form-builder/src/blocks/fields/payment-gateways/PerFormSettings/index.tsx
```diff
@@ -0,0 +1,56 @@
+import {BlockEditProps} from '@wordpress/blocks';
+import {getFormBuilderWindowData} from '@givewp/form-builder/common/getWindowData';
+import Stripe from './Stripe';
+import {useEffect} from 'react';
+
+export default function PerFormSettings(props: BlockEditProps<any>) {
```
jonwaldstein, line 6: @pauloiankoski this looks very clever! although I think it's a little too abstracted and inflexible for gateways.  This goes back to my comment regarding the gateway specific attributes instead of one big gateway settings attributes.

I think a better approach for this gateway block would be to have gateways use the [registerBlockType](https://developer.wordpress.org/block-editor/reference-guides/filters/block-filters/#blocks-registerblocktype) to add their own specific attributes and settings.  Then all their logic is isolated and lives alongside the rest of their code.  This would also give third party (and add-on) gateways an interface for their own settings.

Very similar to how we interacted with the amount block in currency switcher - by using the filter to add attributes/settings.

For the time being, I would be open to keeping the stripe specific attributes/settings in core and refactoring it later to work as a filter - but i'll leave that decision up to you.

jonwaldstein (changes requested):

pauloiankoski:
@jonwaldstein, I've refactored the entire PR. Here's what I've done:

- Moved logic out of the Stripe component and added an action to update meta keys when the form changes.
  
- Removed the generic attribute from the PaymentGateway block and added attributes through the `blocks.registerBlockType` filter.

- Moved the Stripe component out of the payment gateway block and inserted it via the `editor.BlockEdit` filter.

- Extracted styles from `_components.scss` into a separate SCSS file imported by the Stripe script.

- Updated the Form Migration step to use the new attributes.

- Added a new action to initialize the new attributes when a new form is created.

- Updated unit tests to reflect all changes made in this refactor.

Let me know if you have any questions or need further clarification.

src/PaymentGateways/Gateways/Stripe/Actions/UpdateStripeFormBuilderSettingsMeta.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public function __invoke(DonationForm $form)
+    {
+        $paymentGatewaysBlock = $form->blocks->findByName('givewp/payment-gateways');
+        if (!$paymentGatewaysBlock) {
+            give()->form_meta->delete_meta($form->id, 'give_stripe_per_form_accounts');
+            give()->form_meta->delete_meta($form->id, '_give_stripe_default_account');
+
+            return;
+        }
```
jonwaldstein, line 26: Is there a reason we need to delete this meta?

src/PaymentGateways/Gateways/Stripe/Actions/UpdateStripeFormBuilderSettingsMeta.php
```diff
@@ -0,0 +1,41 @@
...
+
+        [
+            'stripeUseGlobalDefault' => $stripeUseGlobalDefault,
+            'stripeAccountId' => $stripeAccountId
+        ] = $paymentGatewaysBlock->getAttributes() + ['stripeUseGlobalDefault' => true, 'stripeAccountId' => ''];
+
+        if (is_bool($stripeUseGlobalDefault)) {
+            give()->form_meta->update_meta($form->id, "give_stripe_per_form_accounts", $stripeUseGlobalDefault ? 'disabled' : 'enabled');
+        }
```
jonwaldstein, line 35: How about instead of `is_bool`, you do:
```
filter_var($stripeUseGlobalDefault, FILTER_VALIDATE_BOOLEAN)
```
to get the boolean value.

src/PaymentGateways/Gateways/Stripe/Actions/UpdateStripeFormBuilderSettingsMeta.php
```diff
@@ -0,0 +1,41 @@
...
+
+        if (is_string($stripeAccountId)) {
```
jonwaldstein, line 37: We should probably check if `$stripeUseGlobalDefault` is true as well.  Stripe uses prefixes for their ID's so you could even make sure this string includes `acct_`

src/PaymentGateways/Gateways/Stripe/resources/js/StripeAccountPanel.tsx
```diff
@@ -0,0 +1,108 @@
+import {__} from '@wordpress/i18n';
+import {InspectorControls} from '@wordpress/block-editor';
+import {Notice, PanelBody, SelectControl, ToggleControl} from '@wordpress/components';
+import {createInterpolateElement} from '@wordpress/element';
+import {getFormBuilderWindowData} from '@givewp/form-builder/common/getWindowData';
+import './styles.scss';
+
+export default function StripeAccountPanel({
```
jonwaldstein, line 8: I'm thinking we should move this whole resources folder to be inside `Gateways\Stripe\StripePaymentElementGateway`.  That is - if these are specific to the payment element gateway - if so i'd rather keep them together.

src/PaymentGateways/Gateways/Stripe/Actions/EnqueueStripeFormBuilderScripts.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+class EnqueueStripeFormBuilderScripts
+{
+
+    /**
+     * Enqueues the Stripe scripts and styles for the Form Builder.
+     *
+     * @return void
+     */
+    public function __invoke()
```
jonwaldstein, line 18: Same comment here about moving these actions to `Gateways\Stripe\StripePaymentElementGateway`.  It looks specific to payment element

src/FormBuilder/ViewModels/FormBuilderViewModel.php
```diff
@@ -50,6 +50,7 @@ public function storageData(int $donationFormId): array
                 'rewriteSlug' => get_post_type_object('give_forms')->rewrite['slug'],
             ],
             'gateways' => $this->getGateways(),
+            'gatewaysGlobalSettings' => $this->getGatewaysGlobalSettings(),
```
jonwaldstein, line 53: Since we moved all the scripts and logic out to live with Stripe, let's  move the window data too (where you are enqueuing Stripe into form builder):

```
wp_localize_script( 'givewp-stripe-payment-element-form-builder', 'stripePaymentElementGatewaySettings', $this->getGlobalGatewaySettings());
```
```

File: src/Framework/Blocks/BlockModel.php [hunk 11]
```diff
@@ -78,9 +78,7 @@ public function getAttributes(): array
      */
     public function setAttribute(string $name, $value): BlockModel
     {
-        if ($this->hasAttribute($name)) {
-            $this->attributes[$name] = $value;
-        }
+        $this->attributes[$name] = $value;
```
jonwaldstein, line 81: Was this intentional?

src/PaymentGateways/Gateways/Stripe/Actions/AddStripeAttributesToNewForms.php
```diff
@@ -0,0 +1,21 @@
...
+class AddStripeAttributesToNewForms
+{
+
+    /**
+     * Add Stripe attributes to the PaymentGateways block new forms.
+     */
+    public function __invoke(DonationForm $form)
+    {
+        $block = $form->blocks->findByName('givewp/payment-gateways');
+        if ($block) {
```
jonwaldstein, line 17: I think we would want to make sure Stripe is enabled in global settings before setting this attribute.

src/PaymentGateways/Gateways/Stripe/resources/js/StripeAccountPanel.tsx
```diff
@@ -0,0 +1,108 @@
...
+            a: <a href={`${gatewaySettingsUrl}&section=stripe-settings&group=accounts`} />,
+        });
+
+    const hasGlobalDefault = stripeGlobalSettings.default;
+    const hasPerFormDefault = attributes.stripeAccountId;
+    const showGlobalDefaultNotice =
+        !hasGlobalDefault &&
+        (attributes.stripeUseGlobalDefault || (!attributes.stripeUseGlobalDefault && !hasPerFormDefault));
+
+    const useGlobalDefaultHelper = textWithLinkToStripeSettings(
```
jonwaldstein, line 29: Just a heads up the prefix `use` is technically a reserved word in React for hooks.  I think if we had linting setup for react it would catch this from the rules of hooks (this is something we need to add in the future).

I would reccomend changing this function name to `shouldUseGlobalDefaultHelper` to avoid any confusion.

src/PaymentGateways/Gateways/Stripe/resources/js/StripeAccountPanel.tsx
```diff
@@ -0,0 +1,108 @@
...
+}: {
+    attributes: StripeProps;
+    setAttributes: ({}) => void;
+}) {
+    const {gatewaysGlobalSettings, gatewaySettingsUrl} = getFormBuilderWindowData();
+    const stripeGlobalSettings: StripeGlobalSettingsProps = gatewaysGlobalSettings?.stripe_payment_element ?? {};
+
+    const textWithLinkToStripeSettings = (textTemplate: string) =>
+        createInterpolateElement(textTemplate, {
+            a: <a href={`${gatewaySettingsUrl}&section=stripe-settings&group=accounts`} />,
```
jonwaldstein, line 20: This whole url looks like it should come from the window instead of building it here.

jonwaldstein (changes requested):
@pauloiankoski you are a beast for updating this!  everything appears to be working properly and I love this structure.  Just had a few suggestions.

src/PaymentGateways/Gateways/Stripe/Actions/UpdateStripeFormBuilderSettingsMeta.php
pauloiankoski: Not really. Removing it.

src/PaymentGateways/Gateways/Stripe/Actions/UpdateStripeFormBuilderSettingsMeta.php
pauloiankoski: Here I'm just guaranteeing that variable is a boolean as I need to update the meta based on either true or false. Using the filter, the if statement won't be accessed if false.

It's quite counter intuitive but:
$ is true? so 'disabled'
$ is false? so 'enabled'

src/PaymentGateways/Gateways/Stripe/Actions/UpdateStripeFormBuilderSettingsMeta.php
pauloiankoski: Sure.

src/PaymentGateways/Gateways/Stripe/resources/js/StripeAccountPanel.tsx
pauloiankoski: Sure

src/PaymentGateways/Gateways/Stripe/Actions/EnqueueStripeFormBuilderScripts.php
pauloiankoski: Sure

src/FormBuilder/ViewModels/FormBuilderViewModel.php
pauloiankoski: Sure

src/Framework/Blocks/BlockModel.php
pauloiankoski: Yes! I notice that I had added that and now I was not being able to set new attributes programmatically what disallow us and any 3rd-party dev to add custom attributes to a block.

src/PaymentGateways/Gateways/Stripe/Actions/AddStripeAttributesToNewForms.php
pauloiankoski: Sure

src/PaymentGateways/Gateways/Stripe/resources/js/StripeAccountPanel.tsx
pauloiankoski: Sure

src/PaymentGateways/Gateways/Stripe/resources/js/StripeAccountPanel.tsx
pauloiankoski: Sure

src/PaymentGateways/Gateways/Stripe/Actions/UpdateStripeFormBuilderSettingsMeta.php
pauloiankoski: Resolved: https://github.com/impress-org/givewp/pull/6903/commits/b6616ec12999d1cff98f6985b472f994a7e4da1b

src/PaymentGateways/Gateways/Stripe/Actions/UpdateStripeFormBuilderSettingsMeta.php
pauloiankoski: Resolved: https://github.com/impress-org/givewp/pull/6903/commits/404623fb809d02ea78db4d442f93429001398dbb

src/PaymentGateways/Gateways/Stripe/Actions/EnqueueStripeFormBuilderScripts.php
pauloiankoski: Resolved: https://github.com/impress-org/givewp/pull/6903/commits/dc77ec96213d780460efe0875f9502707fcf6a21

src/FormBuilder/ViewModels/FormBuilderViewModel.php
pauloiankoski: Resolved: https://github.com/impress-org/givewp/pull/6903/commits/c5a70ef35e2b5f8d1de15e0061bb2d392ab48bf2

src/PaymentGateways/Gateways/Stripe/Actions/AddStripeAttributesToNewForms.php
pauloiankoski: Resolved: https://github.com/impress-org/givewp/pull/6903/commits/1741124d0c04c863b0174211cf82fb2a162e7887

src/PaymentGateways/Gateways/Stripe/resources/js/StripeAccountPanel.tsx
pauloiankoski: Resolved: https://github.com/impress-org/givewp/pull/6903/commits/3603f59670767ac18835e05beedab50185983736

src/PaymentGateways/Gateways/Stripe/resources/js/StripeAccountPanel.tsx
pauloiankoski: Resolved: https://github.com/impress-org/givewp/pull/6903/commits/c5a70ef35e2b5f8d1de15e0061bb2d392ab48bf2

src/Framework/Blocks/BlockModel.php
jonwaldstein: Ahh good catch!

src/PaymentGateways/Gateways/Stripe/StripePaymentElementGateway/resources/js/StripeAccountPanel.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+export default function StripeAccountPanel({
+    attributes,
+    setAttributes,
+}: {
+    attributes: StripeProps;
+    setAttributes: ({}) => void;
+}) {
+    // @ts-ignore
+    const stripeGlobalSettings: StripeGlobalSettingsProps = window.stripePaymentElementGatewaySettings;
```
jonwaldstein, line 15: Here's a little trick i've been using to scope window data.

Add this before the component and then you can remove `@ts-ignore`

```
declare const window: {
    stripePaymentElementGatewaySettings: StripeGlobalSettingsProps;
} & Window;
```

This is my recommended approach for defining window data because it's a single instance of the window and not a global - which has caused problems with our multi-domain application structure.

If you peek into the form builder getWindowData file you'll see being used.

jonwaldstein (approved):
@pauloiankoski i love how this turned out, excellent job!

I just had one more suggestion but other than that - it's ready for QA ⭐

src/PaymentGateways/Gateways/Stripe/StripePaymentElementGateway/resources/js/StripeAccountPanel.tsx
pauloiankoski: Resolved: a0285f5db0012bc474732e2f374619dc3045c2a7

rickalday (approved):

rickalday (approved):
Passed manual and automated QA tests

PR #6906, opened 2023-08-28
Refactor: Update welcome banner with NextGen 3.0 release design

JoshuaHungDinh:
## Description
I've updated the welcome banner with the new 3.0 release design. The design consists of two links, one internal that leads to the VFB new form page and the other which is an external that opens up a new window for the 3.0 doc page. Two videos have replaced the images & include pause and play actions. The banner will continue to remain closed when exited and should only display on the plugin page.

I felt it would be helpful and more organized to redo the banner with React, as it was previously rendered through a php file. Although not much javascript is needed in this particular banner, using this approach has made it easier to read, debug and ultimately reuse in the future.

To accomplish some of the changes into the react app, I've created a single endpoint to handle the new meta key option ``` givewp_welcome_banner_dismiss ```  which determines if the banner has previously been dismissed. A new window object specifically for the WelcomeBanner to get the necessary information to perform the ajax requests and handle the SVG icons from the assets folder. The other changes are from migrating php into jsx.

I have also removed the previous 3.0 service provider and related content that was used prior to branching off the core plugin and have included it in our main service provider work.

## Additional Context
- 3.0 Doc page is not live yet.
- Updated videos are currently in progress. We have fallback images for the video players until they are completed.

## Affects

Plugin Page

## Visuals

Screen Shot 2023-09-01 at 10 20 39 AM

## Testing Instructions
- Visit Plugin Page
- View Banner
- Play video if available
- Close banner
- Refresh page - verify banner remains closed

## Pre-review Checklist

---
  - https://app.asana.com/0/0/1205399260808646


pauloiankoski (changes requested):
Hey @JoshuaHungDinh! Overall, it looks good. I have a few comments:
-  The close button doesn't dismiss the banner when clicked. Currently, it only triggers the ajax call to the backend. So, you need to refresh the page to see the banner gone.
-  The SVGs of the video player buttons have transparent icons, which looks odd when text moves behind them.
-  Consider giving a more generic name to your option, like `givewp_welcome_banner_dismiss`. Since you're saving the timestamp, it allows us to easily check the option against the release date of any new banner in the future.
-  You might want to update your coding style in your IDE to use PSR12. I noticed some files added to this PR that aren't related to your work. They are here because your IDE reformatted them to a different standard than what most of us are using.

JoshuaHungDinh:
@pauloiankoski 
Resolved:
1. 9252b4df90214bf579fe965f09cbe4a161e01a41
2. 974b97b78f2f30b020a74e55f17d41c637b5e83e
3. 8d9a54f612eb994723b36b9e3eec87d09223f345

pauloiankoski (approved):
Looks good!

rickalday (approved):
Passed manual QA tests.

PR #6908, opened 2023-08-29
Fix: form data converter

pauloiankoski:
## Description

When we pass values to the backend using a `FormData` structure, the ability to send arrays/objects is lost because they are converted to strings. This pull request resolves that issue by preserving arrays/objects in their original form when they are sent to the backend.

CleanShot 2023-08-29 at 11 46 28
_before_

CleanShot 2023-08-29 at 11 45 15
_after_

## Testing Instructions
- Add a MultiSelect field from FFM to your form
- Open that form on the client side
- Select one or two options in that MultiSelect dropdown
- When you submit the form, you must be able to see the selected options in the receipt page.

```php
add_action('givewp_donation_form_schema', static function (DonationForm $form) {
    $field = MultiSelect::make('custom_multiselect')
        ->fieldType('dropdown')
        ->required(true)
        ->showInAdmin(true)
        ->showInReceipt(true)
        ->label('Custom MultiSelect')
        ->rules(new ArrayRule(true))
        ->tap(function ($field) {
            $options = [
                ["Option one", "Option one", false],
                ["Option two", "Option two", false],
                ["Option three", "Option three", false],
            ];
            $field->options(...$options);
        });

    $form->insertAfter('email', $field);
});
```

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/DonationForms/resources/app/utilities/convertValuesToFormData.ts
+++ b/src/DonationForms/resources/app/utilities/convertValuesToFormData.ts
@@ -6,7 +6,7 @@ export default function convertValuesToFormData(values: object): FormData {
     for (const valueKey in values) {
         const value = values[valueKey];
 
-        if (value === 'object') {
+        if (value !== null && value === 'object') {
             for (const objKey in value) {
                 formData.append(`${valueKey}[${objKey}]`, value[objKey]);
             }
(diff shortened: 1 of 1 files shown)


src/DonationForms/resources/app/utilities/convertValuesToFormData.ts
```diff
@@ -3,9 +3,16 @@
  */
 export default function convertValuesToFormData(values: object): FormData {
     const formData = new FormData();
+    for (const valueKey in values) {
+        const value = values[valueKey];
 
-    for (const key in values) {
-        formData.append(key, values[key]);
+        if (value !== null && value === 'object') {
```
jonwaldstein, line 9: Missing `typeof` here right?

```suggestion
        if (value !== null && typeof value === 'object') {
```

jonwaldstein (approved):
@pauloiankoski great catch and fix!

PR #6910, opened 2023-09-01
Fix: Update asset file path

kjohnson:
## Description

This PR enforces a trailing slash on `GIVE_PLUGIN_DIR` instead of adding a leading slash to the relative file path.


JasonTheAdams (approved):
Interesting, looks like this is pretty safe.

PR #6911, opened 2023-09-01
Feature: improve login expired message and clean up storage system

JasonTheAdams:
## Description

The primary goal of this is to provide a more meaningful message to users in the form builder when they got to save and their session has expired. We can later explore a UI to allow the user to log in and not lose their place, but for now we're starting with a good message.

While working on this, I also realized we have some old code for a storage system that no longer works — due to deep WP integration — and cleaned up the storage system a bit.

## Affects

The system for saving, loading, and previewing within the form builder.

## Testing Instructions

Two things to test. First, make sure that the form builder still works, especially saving and previewing forms.

Second, do the following to simulate an expired login session:
1. Open the form builder
2. Open the JS console
3. Press cmd+shift+P to open up a command console and type in "Clear site data", and run the command
4. Make some change in the form
5. Click "update" to save the form
6. An alert should pop up with the new message

## Pre-review Checklist

---
  - https://app.asana.com/0/0/1205405424043724


kjohnson:
The `form-id` query string on the preview URL is being omitted, which is breaking previews.

src/FormBuilder/Routes/RegisterFormBuilderPageRoute.php
```diff
@@ -117,6 +105,12 @@ public function renderPage()
             true
         );
 
+        wp_add_inline_script(
+            '@givewp/form-builder/script',
+            'window.giveStorageData = ' . wp_json_encode($formBuilderViewModel->storageData($donationFormId)) . ';',
```
kjohnson, line 110: This change is breaking the preview URL - something is different about how the data is localized.

kjohnson: `wp_localize_script` passes each value through `html_entity_decode( (string) $value, ENT_QUOTES, 'UTF-8' )` before attaching the data to the window.

kjohnson: This seems to be a conflict with calling `esc_url` up the line.

kjohnson (changes requested):
This is breaking the form preview feature due to a change in how the `storageData` is being localized.

src/FormBuilder/Routes/RegisterFormBuilderPageRoute.php
JasonTheAdams: Resolved in [3f7cff2](https://github.com/impress-org/givewp/pull/6911/commits/3f7cff25a8bc156a87e0f28923acdb431ea7bf4e)

kjohnson (approved):
This is good to go!

rickalday (approved):
Passed manual QA tests.

PR #6915, opened 2023-09-04
Enhancement: Consolidate test gateways

JasonTheAdams:
## Description

There are currently three test gateways in 3.0, which is super confusing. Hahah! There were:
- Test Donations (v2 only)
- Test Gateway
- Test Gateway (offsite)

I tested and realized that Test Gateway is a drop-in replacement for the existing Test Donations gateway, so that can be removed entirely. (Which means less code in `includes`, yay!)

The offsite Test Gateway is really only needed for devs specifically testing how offsite donations work, and is more likely to confuse customers than help them. I think we can add a setting or something in the future if testing offsite gateways comes up more, but for now I simply commented it out from being loaded with a note.

Finally, for the sake of reporting and backwards-compatibility, I changed the Test Gateway to Test Donations and changed its ID to `manual`, which is what the old Test Donations gateway id was. That way, when someone updates to 3.0 there are no surprises or migrations needed, and previous donations made with that gateway will still fit within the new gateway as if nothing had changed.

there can be only one

## Affects

The aforementioned gateways.

## Visuals

image

## Testing Instructions

Make sure that the Test Donations gateway works with v2 **and** v3 forms.

Also, make some test donations before updating to this, and then update, make some more donations, and make sure the reporting sees them as exactly the same.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205444026014417


JasonTheAdams:
@jonwaldstein Since you built the test gateways I'd appreciate it if you could review this.

jonwaldstein (approved):
Looks good, I updated the gateway ID in `testGateway.tsx` as well.

JasonTheAdams:
Awesome! Can you please shoot this off to QA if the tests pass?

rickalday (approved):
Passed manual QA tests

PR #6920, opened 2023-09-06
Feature: update field scopes to support subscriptions

jonwaldstein:
## Description

Currently, we're storing our custom fields and running through field scopes during the `givewp_donate_controller_donation_created` action which makes the most sense.  The tricky part is when the request involves a subscription and you want to modify the subscription model in your scope. 

This adds a third nullable `$subscription` param to the `givewp_donate_controller_donation_created` action.

Below is an example of a callback scope that wants to modify the subscription using the new param:

```
$field->scope(function ($field, $value, $donation, $subscription = null) {
    if (empty($value)) {
        return;
    }
    
    // Update donation model
    
    if ($subscription){
      // Update subscription model
    }
});
```

I also added a has and get method to the `DonateControllerData` DTO so you could do this:

```
add_action('givewp_donate_controller_donation_created', static function($formData, $donation){
if ($formData->has('customField')){
    $formData->get('customField');
}, 10, 2)
```

## Affects

Field scopes

## Visuals

N/A

## Testing Instructions

This would need to be done programmatically using scopes

## Pre-review Checklist

-   [ ] Includes unit tests


JasonTheAdams:
@jonwaldstein Can you provide a concrete example scenario wherein this becomes relevant? It's a bit abstract, so I'm not quite grasping the problem this is aiming to solve.

jonwaldstein:
@JasonTheAdams it's when a field is trying to modify the Subscription model using a scope.  Here's an example using fee recovery.  Without this PR the modified Subscription model is not passed along to the gateway - instead the initial instance of the Subscription model created gets passed along.  It might honestly be easier to talk through - I admit it's a little confusing.

```
$feeRecoveryField->scope(function (FeeRecovery $field, $value, Donation $donation) {
  if (empty($value)) {
        return;
    }

    (new UpdateDonationWithFeeAmountRecovered())($value, $donation);

    if ($subscription = $donation->subscription){
         $subscription->feeAmountRecovered = Money::fromDecimal(
            $value,
            $subscription->amount->getCurrency()
        );

        $subscription->amount = $subscription->amount->add($subscription->feeAmountRecovered);

        $subscription->save();
    }
});
```

src/DonationForms/Listeners/StoreCustomFields.php
```diff
@@ -22,10 +23,10 @@ class StoreCustomFields
      * @return void
      * @throws NameCollisionException
      */
-    public function __invoke(DonationForm $form, Donation $donation, array $customFields)
+    public function __invoke(DonationForm $form, Donation $donation, array $customFields, Subscription $subscription = null)
```
JasonTheAdams, line 26: @jonwaldstein Since we know we're using the subscription value always, let's make it nullable rather than optional:
```php
    public function __invoke(DonationForm $form, Donation $donation, Subscription? $subscription, array $customFields)
```

src/DonationForms/Listeners/StoreCustomFields.php
```diff
@@ -85,14 +86,23 @@ protected function storeAsDonationMeta(int $donationId, string $metaKey, $value)
     /**
      * @since 3.0.0
      */
-    protected function persistFieldScope(Field $field, $value, Donation $donation)
+    protected function storeAsSubscriptionMeta(int $subscriptionId, string $metaKey, $value)
+    {
+        give()->subscription_meta->update_meta($subscriptionId, $metaKey, $value);
+    }
```
JasonTheAdams, line 92: I'm not seeing this used anywhere. Do we want to add a scope for this?

src/DonationForms/Actions/DispatchDonateControllerDonationCreatedListeners.php
```diff
@@ -6,15 +6,18 @@
...
+use Give\Subscriptions\Models\Subscription;
 
 class DispatchDonateControllerDonationCreatedListeners
 {
     /**
      * @since 3.0.0
+     * @throws NameCollisionException
      */
-    public function __invoke(DonateControllerData $formData, Donation $donation)
+    public function __invoke(DonateControllerData $formData, Donation $donation, Subscription $subscription = null)
```
JasonTheAdams, line 18: This can also be nullable instead of optional.

src/DonationForms/Listeners/StoreCustomFields.php
jonwaldstein: ohhh I didn't realize we could do those now. 7.2 for the win!

jonwaldstein: Resolved here 2438b062

src/DonationForms/Listeners/StoreCustomFields.php
jonwaldstein: Added scope in here 2438b062

src/DonationForms/Actions/DispatchDonateControllerDonationCreatedListeners.php
jonwaldstein: Resolved 2438b062

JasonTheAdams (approved):
Looks good! I feel good about these changes!

PR #6925, opened 2023-09-10
Fix: Undefined variable: settings in LegacyFormSettingCompatibility.php

pulsovi:
Resolves 

```
PHP Notice:  Undefined variable: settings in /.../wp-content/plugins/give/src/Form/Template/LegacyFormSettingCompatibility.php on line 144
```

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Form/Template/LegacyFormSettingCompatibility.php
+++ b/src/Form/Template/LegacyFormSettingCompatibility.php
@@ -105,6 +105,8 @@ class LegacyFormSettingCompatibility
      * Note: Only for internal use. This function can be removed or change in future
      *
      * @since 2.7.0
+     *
+     * @unreleased
      */
     public static function migrateExistingFormSettings()
     {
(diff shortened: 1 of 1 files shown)


kjohnson (changes requested):
If `$settings` is an undefined variable, then it is likely a symptom of another problem (and the error brings that problem to light).

That said, instead of returning early, I would prefer to initialize `$settings` as an empty array before the `foreach` on line 139. As written here this prevents the `_give_form_template` meta from being updated.

kjohnson:
Closing as stale. Always happy to revisit in the future.

pulsovi:
> If `$settings` is an undefined variable, then it is likely a symptom of another problem (and the error brings that problem to light).
> 
> That said, instead of returning early, I would prefer to initialize `$settings` as an empty array before the `foreach` on line 139. As written here this prevents the `_give_form_template` meta from being updated.

Sorry for my long absence. I made the recommended change. Can you review it please?

PR #6931, opened 2023-09-12
Fix: Restore focusability of paragraph block/element

### LINKED ISSUE #6930: Paragraph block element inconsistently receives focus
When adding a paragraph element in the Visual Form Builder a value can be set, but re-focusing the paragraph to edit the value is inconsistent.

kjohnson:
Resolves #6930

## Description

This PR replaces the pseudo outline with the native outline property (coupled with outline-offset) to restore focusability of the rich text content editable.

Additionally, a new inspector control was added for a fallback method to edit the content attribute.

---
  - https://app.asana.com/0/0/1205338142988520
  - https://app.asana.com/0/0/1205470381066812


JoshuaHungDinh:
@kjohnson, interesting bug here..

 I think the template will need to use the markup component from Interweave or some other method to render bold and italics for a complete Paragraph block fix. - I can create a separate PR though if you would like to keep this change isolated.

JoshuaHungDinh (approved):

rickalday (approved):
Passed manual QA tests.

PR #6937, opened 2023-09-13
Fix: Change the migration to update gateways settings only when data doesn't exist.

pauloiankoski:
## Description

This PR fix an issue where a migration was updating data when they were already saved.

During the installation of GiveWP, the options `gateways_v3` and `default_gateway_v3` are saved. However, a migration process occurred that incorrectly updated these options. In the future, this migration will only run if the options have not been saved yet.

Additionally, recently added types have been removed from legacy functions.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205494192059508

### DIFF
--- a/includes/gateways/functions.php
+++ b/includes/gateways/functions.php
@@ -23,7 +23,7 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @param int|null $version
  * @return array $gateways All the available gateways
  */
-function give_get_payment_gateways(?int $version = 2): array
+function give_get_payment_gateways($version = 2)
 {
     $suffix = $version === 3 ? '_v3' : '';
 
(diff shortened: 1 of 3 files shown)


jonwaldstein (changes requested):
@pauloiankoski this should work for the install + migration issue but I noticed something else we should revert: _adding types to legacy functions_.

During the new gateways tab the php types were updates in the legacy functions file for `give_get_default_gateway` and others.  I know it's _really_ tempting to add those - but the reality is they're being used throughout the codebase in a loose way that could cause a domino affect of type issues (along with third parties using them without our knowledge).  So to be safe, I would suggest reverting those types and casting / checking for specific types during implementation of these functions in our new code.

pauloiankoski:
@jonwaldstein Fixed on https://github.com/impress-org/givewp/pull/6937/commits/42a7890282f816e391a35113827fe5ce0f8d163a

jonwaldstein (approved):
Thanks for this, I just added one update in our implementation in case there are no enabled gateways otherwise the legacy function falls back to `$enabled_gateways[0]` which index could not exist.  This could also be resolved inside that function but at least our current use case works.

rickalday (approved):
Passed manual and automated QA tests.

PR #6938, opened 2023-09-13
Fix: Escape double quotes in emailTemplateOptions

kjohnson:
## Description

This PR escapes the nested array properties for `FormSettings::$emailTemplateOptions` when converting the form settings to JSON.

## Testing Instructions

1. Create a v2 form
1. Migrate the form to v3
1. Try to load design mode

### DIFF
--- a/src/DonationForms/Properties/FormSettings.php
+++ b/src/DonationForms/Properties/FormSettings.php
@@ -257,11 +257,16 @@ class FormSettings implements Arrayable, Jsonable
                 $this->toArray(),
                 [
                     'goalType' => $this->goalType ? $this->goalType->getValue() : null,
-                    'emailTemplateOptions' => array_map(function($emailTemplateOptions) {
-                        return array_map('addslashes', $emailTemplateOptions);
-                    }, $this->emailTemplateOptions),
+                    'emailTemplateOptions' => array_map([$this, 'addSlashesRecursive'], $this->emailTemplateOptions),
                 ]
             )
         );
+    }
+
+    public function addSlashesRecursive($value)
+    {
+        return is_array($value)
+            ? array_map([$this, 'addSlashesRecursive'], $value)
+            : addslashes($value);
     }
 }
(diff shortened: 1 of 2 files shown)


src/DonationForms/Properties/FormSettings.php
```diff
@@ -234,10 +234,9 @@ public static function fromArray(array $array): self
      */
     public static function fromJson(string $json): self
     {
-        $self = new self();
-        $array = json_decode($json, true);
-
-        return $self::fromArray($array);
```
kjohnson, line 240: `FormSettings::fromArray()` is a static method and creates its own instance of the `FormSettings` object.

src/DonationForms/Properties/FormSettings.php
```diff
@@ -234,10 +234,9 @@ public static function fromArray(array $array): self
...
+        return self::fromArray(
+            json_decode($json, true)
+        );
```
jonwaldstein, line 239: Cool, I just realized `json_decode('', true);` or any invalid json like `json_decode('{"hello": "\\hello"}', true);` returns null.  I think to be safe, we should validate the string as json, an array or just cast the json_decode as an array so it falls back to the default FormSettings object.

jonwaldstein:
@kjohnson im getting this now:

Screenshot 2023-09-13 at 3 01 50 PM

src/DonationForms/Properties/FormSettings.php
kjohnson: My concern is that falling back would be a silent failure.

That said, I think we can get this merged and revisit this part.

jonwaldstein: I'm just concerned that if the settings are corrupt it will break the form and there's no way to fix it.  So we should really be validating the json somewhere

jonwaldstein: I'll open a separate PR for this

jonwaldstein (approved):

PR #6947, opened 2023-09-16
Feature: update amount block default setting styles

JasonTheAdams:
## Description

When selecting the which donation level is the default, there was no visual indicator within the form builder to help clear up what change the radio button was doing. The primary goal of this PR is to add that visual.

As a secondary goal, the styles were cleaned up to use less inline styles and use more classes. The reason this was necessary was because React wasn't properly clearing the outline selected styles when using inline styles. I also switched to using outlines instead of borders, as outlines don't affect the size of the element box.

## Affects

The amount block and how it styles within the form builder

## Visuals

image

## Testing Instructions

Change up the default donation level as well as the default recurring donation frequency and enjoy the visual feedback.

## Pre-review Checklist


jdghinson:
I was thinking of something like this to make it more evident and also if possible add a tooltip to the default radio selector in the block settings. 

Screenshot 2023-09-18 at 1 03 03 PM

jdghinson (approved):
All good

JasonTheAdams:
@jdghinson I tried adding the tooltip and... so strangely

kjohnson (approved):
Looks good to me. Just waiting on the automated tests.

PR #6948, opened 2023-09-18
Fix: Prevent header from disapearing

kjohnson:
## Description

This PR prevents the Editor Skeleton Header component from disappearing.

The underlying issue seems to be that the Editor Skeleton Body has a `min-height` of `100vh`, which was causing the Header to be hidden under certain circumstances.

## Affects

This remove the `min-height` CSS property from the Editor Skeleton Body.

## Testing Instructions

1. Set terms and conditions to show in modal.
1. Go to Design tab and click the Show terms link.
1. When the terms and conditions modal opens, the top bar of the builder **should not** disappear.

---
  - https://app.asana.com/0/0/1205502083833595
  - https://app.asana.com/0/0/1205524417076814

### DIFF
--- a/src/FormBuilder/resources/js/form-builder/src/styles/_block-editor.scss
+++ b/src/FormBuilder/resources/js/form-builder/src/styles/_block-editor.scss
@@ -13,7 +13,6 @@
 }
 
 .interface-interface-skeleton__body {
-    min-height: 100vh;
     width: 100%;
     background-color: var(--givewp-gray-10);
 }


pauloiankoski (approved):
Yep. That works as expected.

jonwaldstein (approved):
@kjohnson looked good to me, I think I added this because of something with design mode but seems like it has been resolved over time.

rickalday (approved):
Passed manual tests

PR #6952, opened 2023-09-18
fix: Prevent fatal errors when redirecting form IDs

kjohnson:
## Description

This PR updates every use of `array_walk` when redirecting Form IDs to avoid fatal errors.

In some cases the `array_walk` was being used incorrectly because arrays were being imploded as comma separated strings.

## Testing Instructions

 Using PHP 8+ add the Form Grid block to a page

---
  - https://app.asana.com/0/0/1205502083833544
  - https://app.asana.com/0/0/1205534412601309


JoshuaHungDinh (approved):
@kjohnson thanks, I see what I was misunderstanding now. I tested this with the FormGrid, DonorWall and MultiFormGoal blocks. All working without error.

rickalday (approved):
Passed manual QA tests.

PR #6953, opened 2023-09-19
Fix: illegal offset

pulsovi:
## Description

FIX : If neither `$formId` nor the Utils\Frontend allows getting a form ID, the return value of `Template->getActiveID` becomes unpredictable and error-prone.

HOW : The PR bailout before call of get_meta, returning `null` if no ID found.

WHY : `null` is the most intuitive value to retrieve in such a situation.

## Affects

It's a fix, it not affects any feature, only bugs.

## Testing Instructions

In a context situation where `Frontend::getFormId()` returns `null`.
Call `Give()->form_meta->__add_meta()` with an improbable value (such "I'm not a template") as the only parameter.
Then call `Template::getActiveID()` without any parameters, the return value is the improbable value you provide to `__add_meta`, which should not be the case.

The PR make it returning `null` instead.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Helpers/Form/Template.php
+++ b/src/Helpers/Form/Template.php
@@ -21,7 +21,13 @@ class Template
      */
     public static function getActiveID($formId = null)
     {
-        return Give()->form_meta->get_meta($formId ?: Frontend::getFormId(), '_give_form_template', true);
+        /*
+         * Bailout: If neither the parameter nor the Utils\Frontend allows to get a form ID
+         * do not call get_meta.
+         * This method return unpredictable values if called with an invalid ID.
+         */
+        if (!($formId = $formId ?: Frontend::getFormId())) return null;
+        return Give()->form_meta->get_meta($formId, '_give_form_template', true);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Helpers/Form/Template.php
```diff
@@ -21,7 +21,13 @@ class Template
      */
     public static function getActiveID($formId = null)
     {
-        return Give()->form_meta->get_meta($formId ?: Frontend::getFormId(), '_give_form_template', true);
+        /*
+         * Bailout: If neither the parameter nor the Utils\Frontend allows to get a form ID
+         * do not call get_meta.
+         * This method return unpredictable values if called with an invalid ID.
+         */
+        if (!($formId = $formId ?: Frontend::getFormId())) return null;
```
kjohnson, line 29: While `null` may be more correct in this situation, I'd rather not introduce a breaking change at this time. Changing return types can cause new and unforeseen problems.

If this is causing an issue in regards to some custom code then I would suggest introducing some defensive programming when calling this method.

PR #6957, opened 2023-09-20
Refactor: normalize currency component

pauloiankoski:
## Description

This Pull Request introduces enhanced consistency in value formatting for the `onValueChange` method in the `CurrencyControl` component by leveraging an internal state. It ensures that all values are normalized before being passed to the method. Furthermore, the `CurrencyControl` component is now globally accessible through `window.givewp.form.components.CurrencyControl`, making it easier for add-ons to utilize this controlled component.

Prior to these changes, input values were stored in the form builder just as they were entered by the user, localized according to each user's locale. This led to unpredictable and inconsistent data formats. For example, one user might edit the Fee Recovery's percentage fee field using a comma, while another could edit only the base amount field using a dot.

To resolve this issue, an internal state has been added to the component. When the input value changes, this internal state will be manually updated with the raw, unformatted value. Simultaneously, the block attribute will be set with the normalized value. This ensures consistent data formatting across different locales and editing stages.

By adopting this approach, we're standardizing the data format, eliminating any inconsistencies, and simplifying the process for future edits and data manipulations.

## Affects

Fee Recovery settings and Donation Levels inputs

## Testing Instructions

- Create a form builder using different locales in your browser (Inspector Tools > Sensors)
- Check that saved data is stored normalized as number

## Pre-review Checklist

-   [ ] Includes unit tests


src/FormBuilder/resources/js/form-builder/src/common/currency.tsx
```diff
@@ -12,6 +13,28 @@ const formatCurrencyAmount = (amount: string) => {
...
+    for (const part of parts) {
+        if (part.type === 'group') {
+            groupSeparator = part.value;
+        } else if (part.type === 'decimal') {
+            decimalSeparator = part.value;
+        }
+    }
+
+    return amount.replaceAll(groupSeparator, '').replace(decimalSeparator, '.');
+}
```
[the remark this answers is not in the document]
pauloiankoski: This is masked input so the user is "forced" to input in its locale. Having spaces for the thousands separator would also be caught by the group separator and removed from the string.

JasonTheAdams:
@pauloiankoski So this normalizes the data. Does it also do the reverse of taking a normalized value and displaying it back to the user in their locale? That is, I write my number, save my form, and reload the page. Does the input now reflect the normalized value? Or as I would expect it in my locale?

pauloiankoski:
@JasonTheAdams Yes! The user sees and edit the amount in its locale while the data is saved in a standard format. The CurrencyInput component handles that for us.

In the example, I wrote using Russian local (spaces for group and comma for decimal), then reloaded the page with Brazilian local (dot for groups and comma for decimal) then finally reloaded the page with my default local en-US (comma for group and dot for decimal).

https://github.com/impress-org/givewp/assets/3921017/1b9b5172-99c6-426a-a9bb-5feef847e9f7

JasonTheAdams (approved):
Great work, @pauloiankoski! This looks so much more rock solid!

src/FormBuilder/resources/js/form-builder/src/common/currency.tsx
```diff
@@ -22,28 +45,50 @@ interface CurrencyControlProps extends CurrencyInputProps {
...
+
+declare const window: {
+    givewp?: {
+        form?: {
+            components?: {
+                CurrencyControl?: typeof CurrencyControl;
+            };
+        };
+    };
+} & Window;
```
JasonTheAdams, line 94: I could be wrong (@jonwaldstein ?), but I don't think this is really doing anything. This trick declares the window structure for this file without affecting the global. So it's really only checking that the assignments above match, but that's also inferred. Doesn't hurt, of course.

jonwaldstein: Yeah, I also wouldn't assign components to the window like this, it gets a little messy and I think would actually keep assigning every render.  Assigning to the window should be elevated outside the component to either the app level or registrar level.

pauloiankoski: Maybe in another context, but we replaced a few `@TS-ignore` with that approach the other day. That's the only reason I brought that here as well.

pauloiankoski: I believe the subjects got mixed up, but I understood both points of view. Firstly, I moved the CurrencyControl component to the components folder. Then, I transferred all registrations of the window object to the index file in the components domain. Lastly, I registered the components object to the "global" window declaration under `getWindowData`.

jonwaldstein: @pauloiankoski great, thanks!

PR #6961, opened 2023-09-21
Feat: Prefill form, allow third party filtering

pulsovi:
## Description

When a WordPress user is logged in and has donations under their belt (i.e. a donor ID), the Give form automatically populates with its information.

But for a donor without a WP user account, nothing is completed.

Additionally, I edit a plugin on my personal site that has session information that could participate in pre-filling if a hook was provided.

## Affects

Form pre filling donor data

## Testing Instructions

Make a donation as a guest user. reload the form.

Before the PR : The form is empty
After the PR : The form contains the first donation data such as the first name and last name of the donor (not credit card number of course).

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/includes/forms/functions.php
+++ b/includes/forms/functions.php
@@ -1055,9 +1055,7 @@ function give_show_login_register_option( $form_id ) {
 /**
  * Get pre fill form field values.
  *
- * Note: Will extract form field values from give_purchase session data.
- *
- * @unreleased Extract values from WP user, logged donor and session donor.
+ * @unreleased Extract values from logged WP user and logged donor.
  *             Allow third party filter values.
  * @since  1.8
  *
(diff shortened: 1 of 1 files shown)


includes/forms/functions.php
```diff
@@ -1106,27 +1117,55 @@ function _give_get_prefill_form_field_values( $form_id ) {
...
+		$logged_in_user_info,
+		$logged_in_donor_info,
+		$donor_from_session_info,
+		$session_info
+	);
 
-	// Output.
-	return wp_parse_args( $give_donor_info_in_session, $logged_in_donor_info );
+	// allow third party plugins filtering the result
+	return apply_filters(
```
kjohnson, line 1163: I don't mind introducing a new filter here, but I would rather not change the functionality to introduce a new  data source (the session). I can see the feature being helpful, but would rather provide this as a code snippet which uses the new filter, which would allow site owners to opt-in to this change of behavior.

kjohnson (changes requested):

kjohnson:
Closing in favor of https://github.com/impress-org/givewp/pull/6989

@pulsovi I decided to add filters where this function is used, since the underlying function is marked as for internal use only. Additionally, this provides context for how the field values are being used.

PR #6962, opened 2023-09-22
Fix: Explicitly set customAmount setting during form migration

kjohnson:
## Description

This PR explicitly sets the `customAmount` setting during the form migration. Previously only `customAmountMin` and `customAmountMax` were being set - with `customAmount` defaulting to enabled.

## Testing Instructions

- Create a v2 form with Custom Amount disabled.
- Migrate the form.
- Custom Amount should be disabled.

---
  - https://app.asana.com/0/0/1205502083833541
  - https://app.asana.com/0/0/1205557954247356


JoshuaHungDinh (approved):
Nice, I also tested the UI & it's working without error.

rickalday (approved):
Passed manual QA tests

PR #6964, opened 2023-09-22
Refactor: Update Feature Notice design

kjohnson:
## Description

- Adds styles to the upgrade button.
- Sets `font-size`, `color` styles for the message.
- Adds bold formatting to `"Add form"` in the message.

## Visuals

Screenshot 2023-09-22 at 15-57-58 Edit Donation Form “My Donation Form” ‹ WordPress — WordPress

---
  - https://app.asana.com/0/0/1205515364894528
  - https://app.asana.com/0/0/1205593292175800

### DIFF
--- a/src/DonationForms/V2/resources/components/Onboarding/Dialogs/FeatureNoticeDialog.tsx
+++ b/src/DonationForms/V2/resources/components/Onboarding/Dialogs/FeatureNoticeDialog.tsx
@@ -1,8 +1,9 @@
-import {__} from '@wordpress/i18n';
+import {__, sprintf} from '@wordpress/i18n';
 import ModalDialog from '@givewp/components/AdminUI/ModalDialog';
 import {CheckVerified, StarsIcon} from '@givewp/components/AdminUI/Icons';
 import Button from '@givewp/components/AdminUI/Button';
 import styles from '../style.module.scss';
+import {Interweave} from "interweave";
 
 
 export default function FeatureNoticeDialog({isUpgrading, isEditing, handleClose}) {
@@ -27,6 +28,14 @@ export default function FeatureNoticeDialog({isUpgrading, isEditing, handleClose
         }
     }
 
+    // @note the <Button/> component does not support the `className` prop.
+    const upgradeButtonStyles = {
+        width: '100%',
+        marginTop: 'var(--givewp-spacing-6)',
+        marginBottom: 'var(--givewp-spacing-4)',
+        backgroundColor: 'var(--wp-blue-blue-50)'
+    }
+
     return (
         <ModalDialog
             isOpen={true}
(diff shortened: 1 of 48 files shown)


src/DonationForms/V2/resources/components/Onboarding/Dialogs/FeatureNoticeDialog.tsx
```diff
@@ -27,6 +28,14 @@ export default function FeatureNoticeDialog({isUpgrading, isEditing, handleClose
         }
     }
 
+    // @note the <Button/> component does not support the `className` prop.
+    const upgradeButtonStyles = {
+        width: '100%',
+        marginTop: 'var(--givewp-spacing-6)',
+        marginBottom: 'var(--givewp-spacing-4)',
+        backgroundColor: 'var(--wp-blue-blue-50)'
+    }
```
pauloiankoski, line 37: Can we add the `className` prop to the `Button` component without any issues? I tested it locally and apparently it worked as expected. This way, we can avoid mixing inline styles with CSS modules in the same file.

src/DonationForms/V2/resources/components/Onboarding/Dialogs/FeatureNoticeDialog.tsx
```diff
@@ -38,12 +47,13 @@ export default function FeatureNoticeDialog({isUpgrading, isEditing, handleClose
...
-                        'give'
-                    )}
-                </p>
+                <Interweave tagName={'p'} className={styles.message} content={
+                    sprintf(
+                        __('GiveWP 3.0 introduces an enhanced forms experience powered by the new Visual Donation Form Builder. The team is still working on add-on and gateway compatibility. If you need to use an add-on or gateway that isn\'t listed, use the "%sAdd form%s" option for now.', 'give'),
+                        '<strong>',
+                        '</strong>'
+                    )
+                } />
```
pauloiankoski, line 56: Just in case you don't wanna import the Interweave library only for this particular usage, we can leverage on WP package instead. (Highly optional though)

```suggestion
                <p className={styles.message}>
                    {createInterpolateElement(
                        sprintf(__('GiveWP 3.0 introduces an enhanced forms experience powered by the new Visual Donation Form Builder. The team is still working on add-on and gateway compatibility. If you need to use an add-on or gateway that isn\'t listed, use the "%sAdd form%s" option for now.', 'give'), '<b>','</b>'),
                        {
                            b: <strong />,
                        }
                    )}
                </p>
```

kjohnson: Awesome. Thanks for the suggestion!

src/DonationForms/V2/resources/components/Onboarding/Dialogs/FeatureNoticeDialog.tsx
kjohnson: @pauloiankoski I think that would be a good refactor, but I'd rather not change shared components in this PR.

pauloiankoski (approved):
@kjohnson We just need to update imports (remove `Interweave` and add `createInterpolatedElement` from `@wordpress/element`) here and we are good to go to QA.

rickalday (approved):
Passed manual QA tests.

PR #6968, opened 2023-09-25
Refactor: Promote form designs to primary editor experience

kjohnson:
## Description

This PR changes the direction of the user experience with regards to the Visual Form Builder by promoting "Design Mode" as the initial interface for editing a form. Additionally, the form schema (blocks) can be edited as an _advanced_ mode.

## Visuals

image

image

image

image

---
  - https://app.asana.com/0/0/1205598761114615


src/FormBuilder/resources/js/form-builder/src/stores/editor-state/index.tsx
pauloiankoski, line 1: Will more data be added to this context in the future? If not, we can replace the complexity of a reducer with a simpler ﻿`useState`.

Just to fulfill my curiosity, why do we create two contexts here instead of passing both values as `﻿value={{state, dispatch}}`?

src/FormBuilder/resources/js/form-builder/src/stores/editor-state/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+            <StoreContextDispatch.Provider value={dispatch}>{children}</StoreContextDispatch.Provider>
+        </StoreContext.Provider>
+    );
+};
+
+const useEditorState = () => useContext<EditorState>(StoreContext);
+
+const useEditorStateDispatch = () => useContext(StoreContextDispatch);
+
+export {EditorStateProvider, useEditorState, useEditorStateDispatch, setEditorMode};
```
pauloiankoski, line 28: Will more data be added to this context in the future? If not, we can replace the complexity of a reducer with a simpler ﻿`useState`.

Just to fulfill my curiosity, why do we create two contexts here instead of passing both values as `﻿value={{state, dispatch}}`?

src/FormBuilder/ServiceProvider.php
```diff
@@ -71,7 +72,8 @@ public function boot()
     protected function setupOnboardingTour()
     {
         add_action('wp_ajax_givewp_tour_completed', static function () {
-            add_user_meta(get_current_user_id(), 'givewp-form-builder-tour-completed', time(), true);
+            $mode = new EditorMode($_POST['mode']);
+            add_user_meta(get_current_user_id(), "givewp-form-builder-$mode-tour-completed", time(), true);
```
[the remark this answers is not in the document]
kjohnson: The `EditorMode` enum will throw an exception if `$_POST['mode']` is invalid.

src/FormBuilder/resources/js/form-builder/src/stores/editor-state/index.tsx
kjohnson: Using the store allows us to use hooks in deeply nested components without prop drilling the state.

As far as the specifics of how this store is laid out... I copy/pasted @jonwaldstein's form state store. 😎

pauloiankoski (approved):
😎

kjohnson:
Building a zip for QA, see https://github.com/impress-org/givewp/actions/runs/6332206599

rickalday (approved):
Passed manual QA tests

PR #6972, opened 2023-09-27
Feature: allow required blocks to move between sections

JasonTheAdams:
## Description

Forms have a number of blocks/fields that are required for the form to function — e.g. donation amount and email. The primary way we've enforced this is to put a lock on the required blocks. This has a couple of flaws:
1. Locked blocks can't be moved out of their parent block, meaning moving these blocks between sections wasn't possible
2. The user can unlock blocks anyway, allowing them to delete the block... making this a moot point

We also have a validation system on the server which checks that the required blocks are present when saving. It's simply not likely that folks will remove things like the name and amount field, which means we're creating friction to do something relatively likely (move blocks between sections) for something unlikely that's prevented another way.

This PR removes the locks on the required fields, allowing them to move freely. It also updates the server error flow to display a Gutenberg model instead of an alert, giving a nicer experience.

## Affects

The required blocks and the block validation system.

## Visuals

image

## Testing Instructions

Move the required blocks around the form builder. Then delete one and try to save. Finally, add it back and attempt to save, making sure it can recover from an error state.

## Pre-review Checklist

---
  - https://app.asana.com/0/0/1205608290013527


JasonTheAdams:
@jonwaldstein This is ready for review!

jdghinson:
Looks good. Let's add the error icon to the header text like the image below to emphasize on it being an error.

Suggestion: Can we also have a cta like "add missing blocks" instead of "Got it" which automatically adds the block to their original/intended position?

Screenshot 2023-09-27 at 8 12 46 AM

jonwaldstein:
@JasonTheAdams i'm getting a bunch of typescript errors on this PR when running `npm run dev`.  is this happening for you?

Screenshot 2023-09-27 at 1 17 23 PM

src/FormBuilder/resources/js/form-builder/src/styles/_components.scss
```diff
@@ -297,3 +297,13 @@ This creates a consistent separator between the tabs and the rest of the sidebar
...
+.give-error-modal {
+    ul {
+        list-style: disc;
+        padding-inline-start: 1rem;
+    }
+
+    &__dismiss-button {
+        float: right;
+    }
+}
```
jonwaldstein, line 309: Whenever possible, i'd like to move away from dumping styles in here (especially components) and start co-locating styles with their component by importing directly.  The final bundle goes to the same place but is much easier to maintain in practice.

JasonTheAdams: Agreed. I actually tried to use a CSS module on a previous PR and the build system threw a complete fit. It seemed to have an issue with both processing the SCSS this way _and_ having CSS modules.

JasonTheAdams:
> @JasonTheAdams Looks good. Let's add the error icon to the header text like the image below to emphasize on it being an error.
All set!
image

> Suggestion: Can we also have a cta like "add missing blocks" instead of "Got it" which automatically adds the block to their original/intended position?
I like the idea of this, but I think it's a bit beyond the scope of this right now. This is also an unlikely scenario. We can improve this later.

src/FormBuilder/resources/js/form-builder/src/styles/_components.scss
JasonTheAdams: Resolved in c68a84d

JasonTheAdams:
> @JasonTheAdams i'm getting a bunch of typescript errors on this PR when running `npm run dev`. is this happening for you?
> 
> Screenshot 2023-09-27 at 1 17 23 PM

Yikes, yeah, you're right @jonwaldstein. None of these have anything to do with what I'm doing, though. Must be from a previous PR. It's weird that these only show up when you run `npm run dev` and not `npm run watc:v3`. 😕

jonwaldstein:
> > @JasonTheAdams i'm getting a bunch of typescript errors on this PR when running `npm run dev`. is this happening for you?
> > Screenshot 2023-09-27 at 1 17 23 PM
> 
> Yikes, yeah, you're right @jonwaldstein. None of these have anything to do with what I'm doing, though. Must be from a previous PR. It's weird that these only show up when you run `npm run dev` and not `npm run watc:v3`. 😕

yeah, it's something I need to look into but basically laravel mix will prevent bundle from typescript errors while wp scripts ignores them.

jdghinson (approved):
Looks good

JasonTheAdams:
Any other feedback, @jonwaldstein, besides the console errors unrelated to this PR?

jonwaldstein:
> Any other feedback, @jonwaldstein, besides the console errors unrelated to this PR?

@JasonTheAdams i'm only getting these build errors on this PR - not develop so that should be resolved before merging

jonwaldstein (approved):

rickalday (approved):
Passed manual QA tests.

PR #6973, opened 2023-09-27
Feature: add FormMeta and PdfSettings steps to the form migration 

glaubersilva:
## Description

This PR adds a new step in the form migration to catch all form meta keys from the old form into the new one.

This PR also adds a new step for the PDF settings so we can make sure that when the user edits a migrated form using the Form Builder, everything works as in the v2 forms.

~~It also adds a new `givewp_form_builder_pdf_settings_default_values` filter, allowing us to set default values when these settings are empty. More details here: https://github.com/impress-org/give-pdf-receipts/pull/253~~

## Affects

Form meta keys for v3 forms and pdf settings in v3 forms.

## Testing Instructions

1. Create a v2 form and change the settings for the PDF receipt in this form
2. Migrate the v2 form from the previous step to a v3 form
3. Check the PDF setting of the v3 form, it should reflect the exact same settings from the v2 form
4. To check if other meta keys (not related to the pdf settings) also were migrated, check it directly on the database on the `wp_give_formmeta` table

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205615499105028
  - https://app.asana.com/0/0/1205573027035196
  - https://app.asana.com/0/0/1205604772387034


glaubersilva:
@jonwaldstein This is ready for review. 

Initially, I created a `PdfSettings` step to migrate only the items related to the PDF Receipt addon, but I refactored it to the `FormMeta` step so we can catch all form meta keys of the old form as you suggested previously.

I didn't find any problem with this approach, but if you think anything else can be broken with that, let me know, so maybe we can roll back it to handle only the meta key related to the PDF Receipts addon.

src/DonationForms/Properties/FormSettings.php
```diff
@@ -243,6 +243,12 @@ public static function fromArray(array $array): self
         $self->multiStepNextButtonText = $array['multiStepNextButtonText'] ?? __('Continue', 'give');
         $self->pdfSettings = isset($array['pdfSettings']) && is_array($array['pdfSettings']) ? $array['pdfSettings'] : [];
 
+        $donationFormId = $_GET['donationFormID'] ?? false;
+        if (count($self->pdfSettings) === 0 && $donationFormId) {
+            $self->pdfSettings = apply_filters('givewp_form_builder_pdf_settings_default_values', [],
+                (int)$donationFormId);
+        }
```
jonwaldstein, line 250: @glaubersilva we don't want to use the `$_GET` super global here as the intention of this class is a model property so it shouldn't be interacting with the request directly.  The form ID should be found upstream where this property lives.

I'm also a little confused why we need a filter here, can you explain and we can see if this is the right solution?

src/FormMigration/Steps/FormMeta.php
```diff
@@ -0,0 +1,29 @@
...
+    public function process()
+    {
+        $oldFormId = $this->formV2->id;
+        $newFormId = $this->formV3->id;
+
+        $oldFormMeta = give()->form_meta->get_meta($oldFormId);
+
+        if ($oldFormMeta && is_array($oldFormMeta)) {
+            foreach ($oldFormMeta as $oldFormMetaKey => $oldFormMetaValue) {
+                $oldFormMetaValue = is_array($oldFormMetaValue) && count($oldFormMetaValue) === 1 ? $oldFormMetaValue[0] : $oldFormMetaValue;
```
jonwaldstein, line 24: what is the check for `is_array` for?

jonwaldstein (changes requested):

src/DonationForms/Properties/FormSettings.php
glaubersilva: Yeah, makes sense! Resolved in bed3b3f7a6df48ff98939d2c7a4cacd19dc409ab

About the filter, we need to use it to be able to set the default values on the PDF Receipt addon side. 

We are using this filter here:

https://github.com/impress-org/give-pdf-receipts/pull/253/files#diff-51e6882460954791c9102a76f0aa70e20e702984407d70dd3c962e8fb3c01414

src/FormMigration/Steps/FormMeta.php
glaubersilva: It's because the `give()->form_meta->get_meta()` returns the values as an array as you can check on the attached screenshot.

form-meta-as-array

glaubersilva:
@jonwaldstein Thanks for the suggestions, I left a few comments about your questions. So this is ready for re-review.

src/DonationForms/Properties/FormSettings.php
jonwaldstein: Hmm, is this just an alternative for a form migration step?  When I brought up the idea of migrating all existing meta - that was intended as a backup for unknown data.  If we know the data that needs to be migrated then lets go ahead and add a migration step.  

As far as this filter goes, i'm not completely convinced it's necessary.  If you need default values for the form builder settings wouldn't you just add those through the localized object in the add-on?  Happy to chat through this one as well.

jonwaldstein: Also, in case you didn't know about these - we have two very useful actions in which you can update the form model either during render or creation:

`givewp_donation_form_schema` and `givewp_form_builder_new_form`

glaubersilva: @jonwaldstein Gotcha your point about the filter, so I removed it and added the default settings to the localized object in the addon. Resolved in 8cdb6323daa2d89032d591c075738fbbcce723c0

About your question:

> Is this just an alternative for a form migration step?

Not necessarily, as these meta keys still are being updated on the addon side (when the form builder gets updated) to keep backward compatibility, so the new forms still need to have these meta keys. Reference: https://github.com/impress-org/give-pdf-receipts/blob/develop/src/FormExtension/FormBuilder/Actions/UpdatePdfSettingsMeta.php

> If we know the data that needs to be migrated then lets go ahead and add a migration step that includes updating the form settings object.

Initially, I also thought this way, but due to the reason I mentioned above I changed my mind. Another reason that I thought would be nice to handle it on the addon side is to be able to use the DTO class that lives there to provide better context about the data which are being handled.

With that in mind, do you think we still need a new step exclusive to the pdf settings? 

If yes, maybe we wouldn't even need that other PR:  https://github.com/impress-org/give-pdf-receipts/pull/253

glaubersilva: As discussed in a call, we decided to close another PR related to this one and I created a new migration step for the Pdf Settings here on this PR.

jonwaldstein (approved):
Thanks for the updates, looks good - ready for QA!

rickalday (approved):
Passed manual QA tests

PR #6978, opened 2023-09-28
Fix: correct install routine when activating in a WordPress multisite.

pauloiankoski:
## Description
This PR addresses an error in the installation process of GiveWP on multisite setups. Previously, only the main site was installed, resulting in fatal errors on existing subsites.

Upon investigating the codebase, it was discovered that the installation method for GiveWP had a conditional check for `is_multisite()` and a `$network_wide` parameter. However, this parameter was never set to `true`, causing improper installation on WP multisite.

In this PR, we enable the parameter when the plugin is activated for the first time, ensuring proper installation of GiveWP on all existing subsites.

## Affects

GiveWP install process

## Testing Instructions

1. Fresh WordPress install with Multiste enabled
2. Create one subsite at My Sites > Network Admin > Add New
3. Navigate to My Sites > Network Admin > Plugins, install GiveWP, and activate it
4. Navigate to My Sites > Network Admin > Sites, scroll to subsite, click "Dashboard" link
5. The subsite dashboard must show and the Donations menu must be present, meaning that Give is activated

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205607759526942

### DIFF
--- a/give.php
+++ b/give.php
@@ -6,7 +6,7 @@
  * Description: The most robust, flexible, and intuitive way to accept donations on WordPress.
  * Author: GiveWP
  * Author URI: https://givewp.com/
- * Version: 2.33.2
+ * Version: 2.33.3
  * Requires at least: 5.0
  * Requires PHP: 7.0
  * Text Domain: give
@@ -316,7 +316,7 @@ final class Give
     {
         // Plugin version.
         if (!defined('GIVE_VERSION')) {
-            define('GIVE_VERSION', '2.33.2');
+            define('GIVE_VERSION', '2.33.3');
         }
 
         // Plugin Root File.
@@ -409,7 +409,7 @@ final class Give
      *
      * Runs on plugin activation and performs initial setup.
      *
-     * @unreleased set network_wide parameter to true, enabling installing in WP multisite
+     * @since 2.33.3 set network_wide parameter to true, enabling installing in WP multisite
      * @since 1.0.0
      */
     public function install()
--- a/give.php
+++ b/give.php
@@ -407,7 +407,7 @@ final class Give
     public function install()
     {
         $this->loadServiceProviders();
-        give_install();
+        give_install(true);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


give.php
```diff
@@ -404,10 +404,18 @@ public function display_old_recurring_compatibility_notice()
...
+     * Runs on plugin activation and performs initial setup.
+     *
+     * @unreleased set network_wide parameter to true, enabling installing in WP multisite
+     * @since 1.0.0
+     */
     public function install()
     {
         $this->loadServiceProviders();
-        give_install();
+        give_install(true);
```
JasonTheAdams, line 418: @pauloiankoski Why even have the parameter if it's going to be fixed as true? I can see the function is already checking for `is_multisite()`, so I'm not sure what more value the parameter provides.

pauloiankoski: That function is called in other places like when creating a new subsite. So, even tho, it's a multisite, we don't want to reinstall Give across all subsites, but only in the new one.

CleanShot 2023-09-28 at 18 40 31

JasonTheAdams: Ahhh, I see. Good catch. 👍

JasonTheAdams (approved):
Thanks for this, @pauloiankoski! Let's make sure QA tests different types of sites — single, multi, and new sub-site installs.

Genevieve-K:
I don't understand this commit. When I activate the plugin on a sub-site, I want it only on this sub-site.
When we activate it on the root site as network enabled, I understand that the plugin is installed on all sites. 
And I have a fatal error: 
`Fatal error: Uncaught TypeError: array_key_exists(): Argument #2 ($array) must be of type array, null given in /var/www/html/wordpress/wp-content/plugins/give/src/PaymentGateways/Gateways/PayPalStandard/Migrations/SetPayPalStandardGatewayId.php:30 `

The support suggests me to delete the give settings and the wp_give tables for the root site. It works only if I changed this $network_wide parameter to false.

PR #6979, opened 2023-09-29
Fix: remove html tags from form title

pauloiankoski:
## Description

In the current implementation, users have the ability to include HTML tags within the form title field.

This Pull Request address this issue by sanitizing the form title value sent during the form update action. The proposed changes ensure that any HTML tags are stripped out from form title when the form is updated, thereby preventing the injection of malicious scripts. Once the backend responds with the updated data, the value of the form title field is replaced with the sanitized version.

Furthermore, the form title column in the Donation Forms List Table has been updated to escape the form title, ensuring a consistent and secure display of the text. In conjunction with this, the form title input field in the header has been styled in accordance with the new design.

## Affects

Donation Forms List Table and Visual Form Builder

## Visuals
CleanShot 2023-10-03 at 14 13 32

## Testing Instructions
1. Add a new v3 form
2. Edit the form title, trying to add HTML tags
3. The tags you add must always be removed whenever publish/update the form

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205643270356557
  - https://app.asana.com/0/0/1205502083833550


src/FormBuilder/resources/js/form-builder/src/common/stripTags.ts
```diff
@@ -0,0 +1,5 @@
+export default function stripTags(text: string): string {
```
JoshuaHungDinh, line 1: This is cool approach to accomplishing the removal of tags,  is it possible to avoid creating dom elements? I tried a couple tags with a regex pattern `` `input.replace(/<[^>]*>/g, '');``` with some success but wanted to know your thoughts about this?

kjohnson:
```js

Comment (kjohnson) [reply src/FormBuilder/resources/js/form-builder/src/common/stripTags.ts:1]: TL;DR Try this:

```js
function stripHtml(dirtyString) {
  const doc = new DOMParser().parseFromString(dirtyString, 'text/html');
  return doc.body.textContent || '';
}
```

--

Chris Coyier has a snippet on CSS-Tricks for stripping tags using regex.

https://css-tricks.com/snippets/javascript/strip-html-tags-in-javascript/

That said, the comments suggest that the regex has issues with special characters (I have not tested this).

It is also noted that using the DOM to render the HTML and then pull `textContent` is open to an XSS attack, since the script tags are executed.

Alternatively, it is suggested that using `DOMParser()` will not execute the script tags, which is safer.

```js
function stripHtml(dirtyString) {
  const doc = new DOMParser().parseFromString(dirtyString, 'text/html');
  return doc.body.textContent || '';
}

// tripHtml("<img onerror="alert(\"could run arbitrary JS here\")"" src="bogus">"); 
```

pauloiankoski: It does look good. I have tested and implement now.

src/FormBuilder/resources/js/form-builder/src/settings/form-summary/index.js
```diff
@@ -17,6 +18,10 @@ const FormSummarySettings = () => {
                     label={__('Title')}
                     value={formTitle}
                     onChange={(formTitle) => dispatch(setFormSettings({formTitle}))}
+                    onBlur={(event) => {
```
kjohnson, line 21: I wondering if there is a good way to centralizer this sanitation.

@jonwaldstein what if we added a new export method to the form state reducer:

```ts
export function setFormTitle(formTitle) {
    return {
        type: UPDATE_SETTINGS,
        settings: {
                formTitle: stripTags(formTitle)
        }
    };
}
```

jonwaldstein (dismissed):
I'm going to suggest a different route for this, _sanitize on the backend_.  

If you look at how a native WordPress post title works, it doesn't try to strip out the tags in the builder.  Instead, it sanitizes on the backend which is the real gatekeeper for malicious activity.  I would suggest we first make sure our backend is properly sanitizing data, then (if necessary) can circle back to the front-end to improve the user experience.

kjohnson:
> it sanitizes on the backend which is the real gatekeeper for malicious activity

That is a great point. And we already have a convention for the server making changes and passing the changes back to the client - see Form Permalinks.

pauloiankoski:
- Removed sanitization from frontend (https://github.com/impress-org/givewp/pull/6979/commits/a2790d4671cc1bd727a0a697a8f792c8c8933b1c)
- Implemented sanitization in the backend before updating settings and form title (https://github.com/impress-org/givewp/pull/6979/commits/65fe83ae07d02c8d9e83b69cd60e88530bd5ce34)
- Added `formTitle` to the list of settings being refreshed in the frontend once the form is saved (https://github.com/impress-org/givewp/pull/6979/commits/19bea2e012cfda827b1247c196f64c1daef8c58d)

kjohnson (approved):
Looks good to me. Tested locally ✅

rickalday (approved):
Passed manual QA tests

PR #6983, opened 2023-09-29
Refactor: Separate "Design Selector" from design tour

kjohnson:
## Description

This PR separates the "Design Selector" from the design guided tour, with the "Design Selector" starting the guided tour (instead of auto-starting the tour).

## Visuals

Screenshot 2023-09-29 at 14-56-04 ‹ WordPress — WordPress

Screenshot 2023-09-29 at 14-30-34 ‹ WordPress — WordPress

Screenshot 2023-09-29 at 15-20-02 ‹ WordPress — WordPress

Screenshot 2023-09-29 at 15-20-10 ‹ WordPress — WordPress

---
  - https://app.asana.com/0/0/1205635546000691


pauloiankoski:
Hey @kjohnson, could you please check this test failure while I keep reviewing the PR?

CleanShot 2023-10-02 at 16 07 50

pauloiankoski (approved):
@kjohnson All this looks good to me. We may need to update the test that broke due to the changes in this pull request tho: https://github.com/impress-org/givewp/pull/6983#issuecomment-1743605455

kjohnson:
> could you please check this test failure 

Yeah, I need to run that by @jonwaldstein. I'm unsetting the design ID for new forms to trigger the design selector, but I want to run that by @jonwaldstein first.

kjohnson:
@pauloiankoski https://github.com/impress-org/givewp/pull/6983/commits/dea1f9efad6ae00fb71af4a048c0f3565c50296f should resolve the test so that we can move forward (and circle back to @jonwaldstein).

kjohnson:
Building for QA: https://github.com/impress-org/givewp/actions/runs/6385283504

pauloiankoski:
> ~@pauloiankoski [dea1f9e](https://github.com/impress-org/givewp/commit/dea1f9efad6ae00fb71af4a048c0f3565c50296f) should resolve the test so that we can move forward (and circle back to @jonwaldstein).~
> 
> Update: That did not work.

@kjohnson `$self->designId = $array['designId'] ?? null;` seems to work as it provides a default value (even tho is null) to the warning and it's the warning that breaks our test.

image
image

kjohnson:
@pauloiankoski that definitely works here, but I would like @jonwaldstein's input before a change a default value like this. We are currently (in `develop`) operating under the assumption that every form model has a design, which is enforced as a form setting with a default value. I don't want to break that assumption and cause deeper issues when something else is expecting the form to always have a design.

kjohnson:
@pauloiankoski I added the `?? null` fallback to prevent the PHP warning - good catch!
See https://github.com/impress-org/givewp/pull/6983/commits/7bfb68378d584049ddd68c0601932713ef98a01a

jonwaldstein:
@kjohnson @pauloiankoski so the donation form actually assumes there is no design id by default.  This was intentional so that forms do not rely on designs and can work as expected if they don't exist.  With this mindset, I would say it's fine to remove any default `designId`.

pauloiankoski (approved):
👏🏻👏🏻👏🏻 Good to go to QA!

kjohnson:
Building for QA: https://github.com/impress-org/givewp/actions/runs/6386223483

rickalday (approved):
Passed manual QA tests

PR #6985, opened 2023-09-30
Fix: Correct the ZIP build process by changing imports

pauloiankoski:
## Description

Our [Generate Plugin Zip](https://github.com/impress-org/givewp/actions/workflows/generate-zip.yml) workflow stopped working at some point. Upon investigating the issue, I discovered that we had imported types in a manner that although works locally, wasn't working in our build process.

This PR modifies the import and usage of those types to ensure they function correctly both locally and during the build process.

## Affects

Modal component

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/FormBuilder/resources/js/form-builder/src/components/modal/index.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/components/modal/index.tsx
@@ -1,10 +1,10 @@
-import {Button, Icon, Modal as GutenbergModal} from '@wordpress/components';
+import {Button, Icon} from '@wordpress/components';
+import GutenbergModal from '@wordpress/components/modal';
 import {info, warning} from '@wordpress/icons';
-import GutenbergModalTypes from 'wordpress__components/Modal';
 import cx from 'classnames';
 import './styles.scss';
 
-interface ModalProps extends GutenbergModalTypes.Props {
+interface ModalProps extends GutenbergModal.Props {
     closeButtonCaption?: string;
 }
 


src/FormBuilder/resources/js/form-builder/src/components/modal/index.tsx
```diff
@@ -1,10 +1,10 @@
-import {Button, Icon, Modal as GutenbergModal} from '@wordpress/components';
+import {Button, Icon} from '@wordpress/components';
+import GutenbergModal from '@wordpress/components/modal';
```
jonwaldstein, line 2: Can we try this instead?

```suggestion
import {Button, Icon, Modal as GutenbergModal} from '@wordpress/components';

```

pauloiankoski: I have tried that before getting to this solution.

CleanShot 2023-10-02 at 13 28 58

kjohnson: Should we import the type directly?

```suggestion
import {Button, Icon} from '@wordpress/components';
import GutenbergModalProps from '@wordpress/components/modal/types';
```

kjohnson: Hm... It looks like we need to import from `@wordpress/components/modal` because of how Gutenberg defines the namespace, which is lost when imported/exported through `@wordpress/components`.

kjohnson (approved):
Good job, @pauloiankoski. Thanks for digging through that.

src/FormBuilder/resources/js/form-builder/src/components/modal/index.tsx
jonwaldstein: The tricky part is that wordpress scripts has no concept of subfolders so  the import needs to be from `@wordpress/components` or else it won't work properly.  I'm thinking we just go back to importing the modal from the main ``@wordpress/components`` package and pull in the types manually from gutenberg.

JasonTheAdams:
Ohh! So that's how I broke it! Wow, good to know! Thanks for fixing that, @pauloiankoski!

PR #6986, opened 2023-10-02
Feature: Show correct notice message on post publish/update

alaca:
## Description

This PR adds logic to display a correct snackbar message when saving the form. On form publish it will show "Form published", and on form update, it will show "Form updated" notice.

## Affects
Form builder snackbar notice message

## Visuals

Screen Capture_select-area_20231002163324

## Testing Instructions

1. create form
2. update the form
3. publish the form
4. edit form and hit update

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205606769216054

### DIFF
--- a/src/FormBuilder/resources/js/form-builder/src/containers/BlockEditorInterfaceSkeletonContainer.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/containers/BlockEditorInterfaceSkeletonContainer.tsx
@@ -16,7 +16,6 @@ import {Button} from "@wordpress/components";
 import {listView, plus} from "@wordpress/icons";
 import {useEditorState} from "@givewp/form-builder/stores/editor-state";
 import EditorMode from "@givewp/form-builder/types/editorMode";
-import {useFormState} from "@givewp/form-builder/stores/form-state";
 
 export default function BlockEditorInterfaceSkeletonContainer() {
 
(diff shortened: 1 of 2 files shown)


src/FormBuilder/resources/js/form-builder/src/containers/BlockEditorInterfaceSkeletonContainer.tsx
```diff
@@ -42,7 +44,11 @@ const DesignEditorSkeleton = () => {
                     showSidebar={showSidebar}
                     toggleShowSidebar={toggleShowSidebar}
                     onSaveNotice={() => {
```
kjohnson, line 46: Can we extract an `onSaveNoticeCallback` function to be shared by the two components? We recently duplicated the interface skeleton to account for design mode, but it could use a refactor for the props that are shared. As a start I think we can extract the notice logic to a shared function.

alaca: @kjohnson the problem with shared function is that we have to use `createSuccessNotice` from `useDispatch('core/notices')` and that has to be inside the functional component. Maybe a better option would be to move all that logic inside the `HeaderContainer` component and remove the `onSaveNotice` callback.

alaca: I just did that https://github.com/impress-org/givewp/pull/6986/commits/d4d41e8850375b48a8814a3933f9c12281ffc87e

kjohnson: Cool. That sounds good!

kjohnson (approved):
Thanks for making that change. Ready to send to QA.

PR #6987, opened 2023-10-02
Fix: Migrated forms now have a draft status until transferred

kjohnson:
## Description

This PR sets the form status of new forms to be `draft` when first migrated and `publish` once transferred.

## Visuals

image

image

## Testing Instructions

1. Migrate a form
2. New form should be draft
3. Transfer the form
4. Form should be publish

---
  - https://app.asana.com/0/0/1205635546000698

### DIFF
--- a/src/FormMigration/Controllers/TransferController.php
+++ b/src/FormMigration/Controllers/TransferController.php
@@ -3,6 +3,7 @@
 namespace Give\FormMigration\Controllers;
 
 use Give\DonationForms\V2\Models\DonationForm;
+use Give\DonationForms\ValueObjects\DonationFormStatus;
 use Give\FormMigration\Actions\GetMigratedFormId;
 use Give\FormMigration\Actions\TransferDonations;
 use Give\FormMigration\Actions\TransferFormUrl;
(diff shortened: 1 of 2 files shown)


src/FormMigration/Controllers/TransferController.php
```diff
@@ -38,6 +39,7 @@ public function __invoke(DonationForm $formV2, TransferOptions $options)
                 wp_trash_post($formV2->id);
             }
 
+            wp_update_post(['ID' => $v3FormId, 'post_status' => DonationFormStatus::PUBLISHED()->getValue()]);
```
pauloiankoski, line 42: Instead of directly setting it as "published," we could use the current status of the v2 form. For example, a form can be private, so the published status may not be appropriate in that case.

kjohnson: Good point! Resolved by https://github.com/impress-org/givewp/pull/6987/commits/e6334d68ba9534f3c30f86873115f0462c798c60.

pauloiankoski (approved):
This is ready to go to QA!

kjohnson:
Building for QA: https://github.com/impress-org/givewp/actions/runs/6385275917

kjohnson:
Building for QA: https://github.com/impress-org/givewp/actions/runs/6386218740

kjohnson:
Building for QA, see https://github.com/impress-org/givewp/actions/runs/6396092387.

rickalday (approved):
Passed manual QA tests

PR #7000, opened 2023-10-05
Refactor: make designId nullable on confirmation receipt

jonwaldstein:
## Description

This allows the designId to be null for the donation confirmation page.  This further ensures our forms do not rely on a design.

## Affects

The donation confirmation 

## Visuals

N/A

## Testing Instructions

- Create a new v3 form and choose a form design on initial load (do not save draft/publish)
- Submit a donation within preview mode and make sure it does not throw an error

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/DonationForms/ViewModels/DonationConfirmationReceiptViewModel.php
+++ b/src/DonationForms/ViewModels/DonationConfirmationReceiptViewModel.php
@@ -102,10 +102,7 @@ class DonationConfirmationReceiptViewModel
 
         $this->enqueueGlobalStyles($primaryColor, $secondaryColor);
 
-        $this->enqueueFormScripts(
-            $this->donation->formId,
-            $formDesignId
-        );
+        $this->enqueueFormScripts($formDesignId);
 
         ob_start();
         wp_print_styles();
@@ -166,7 +163,7 @@ class DonationConfirmationReceiptViewModel
      *
      * @return void
      */
-    private function enqueueFormScripts(int $formId, string $formDesignId)
+    private function enqueueFormScripts(?string $formDesignId)
     {
         $handle = 'givewp-donation-form-registrars';
         wp_enqueue_script(
@@ -196,7 +193,7 @@ class DonationConfirmationReceiptViewModel
         $formDesignRegistrar = give(FormDesignRegistrar::class);
 
         // silently fail if design is missing for some reason
-        if ($formDesignRegistrar->hasDesign($formDesignId)) {
+        if (!empty($formDesignId) && $formDesignRegistrar->hasDesign($formDesignId)) {
             $design = $formDesignRegistrar->getDesign($formDesignId);
 
             if ($design->css()) {


jonwaldstein:

jonwaldstein:
> @jonwaldstein I'm a bit confused as to why we're wanting to build things in such a way that they don't rely on a design?

Alternatively, why would you want our donation form to rely on a design? It's a great _addition_ to the form, but should not be required for the form to be functional.  The way I see it is if for some reason the design was missing from a form either intentionally or not, it should not break the form.

JasonTheAdams:
> Alternatively, why would you want our donation form to rely on a design? It's a great _addition_ to the form, but should not be required for the form to be functional.

I see your point, and appreciate the intent for stability. I don't think this is an obvious assumption, though. So far I've made the assumption that every form has a form design, which has seemed like a reasonable constant. It seems like this came up because forms no longer have a default.

The most compelling case I can think of here is in the case that a form is somehow removed/unloaded, we still want the donation confirmation receipt to render. I'm less convinced that forms should continue to render without a form design.

jonwaldstein:
@JasonTheAdams I agree it's not obvious & likeliness is probably slim with the ux of the visual builder, yet I intentionally made this decision when building the donation form rendering for reasons I stated above (this PR addresses one overlooked part in the confirmation page).  

More to think about: 
From the start, we were envisioning external form designs (from add-ons and third party).  If a form was previously assigned to an external form design and it is removed we don't want the form to break. Maybe that external form design was literally one line of css - why should that prevent the form from being functional?  

These are the types of things that I was thinking about when going this route.

In conclusion, I feel that form designs are optional visual modifiers for a donation form and forms should not rely on them to function properly.

JasonTheAdams:
> From the start, we were envisioning external form designs (from add-ons and third party). If a form was previously assigned to an external form design and it is removed we don't want the form to break. Maybe that external form design was literally one line of css - why should that prevent the form from being functional?

You raise good points. I think my preference would be to have a fallback form design (maybe Classic) rather than no design at all. We do have things, such as payment gateways, that are simply required, and I don't think there's an issue with that. We can be confident that a default form design will always look good, I'm less confident in making sure the default styles are the same.

jonwaldstein:
>I think my preference would be to have a fallback form design (maybe Classic) rather than no design at all. 

I'm not really against having a fallback design, I just want to make sure our form can technically work without one.

>We do have things, such as payment gateways, that are simply required, and I don't think there's an issue with that.

There are some things about our donation form that are integral to its core functionality, gateways being one of them. You can't really make a donation without a gateway.  Some extra css and/or js from a form design though?

Interestingly enough, our donation form doesn't even have a concept of a form design.  They are really just modifiers that get injected before the form.

---

To get back into the problem at hand, the form builder was [updated](https://github.com/impress-org/givewp/pull/6983) to select the form design up front, so the default was removed on first creation of the donation form model.  The confirmation page doesn't have a concept of being previewed, it finds the donation form through the donation ID to determine any settings.

So in order to assign a default form design for the confirmation page, we would either need to hard code that or go back to a default ID on first creation (which I think had some problems with the selector)

kjohnson:
I'll also add that the Donation Forms do still render "without a design" and the styling is mostly just structural. In this case, the Design is extending the base. This was previously a no-design Design or a the Developer design.

And @jonwaldstein raises a good point about external designs, addons, etc. We could still at render fallback to, say, Classic if the specified (or not specified) form design is not found.

src/DonationForms/ViewModels/DonationConfirmationReceiptViewModel.php
```diff
@@ -166,7 +163,7 @@ public function enqueueGlobalStyles(string $primaryColor, string $secondaryColor
      *
      * @return void
      */
-    private function enqueueFormScripts(int $formId, string $formDesignId)
+    private function enqueueFormScripts(?string $formDesignId)
```
kjohnson, line 166: We should add an unreleased tag noting the change in signature.

kjohnson (approved):
Assuming the conversation with Jason is resolved (or will be resolved), I'm approving this PR.

src/DonationForms/ViewModels/DonationConfirmationReceiptViewModel.php
jonwaldstein: Thanks, this is not actually released yet (still 3.0.0)

kjohnson: Ok, then disregard 😎

jonwaldstein:
@JasonTheAdams @kjohnson i'm going to get this merged and then we can re-address the default form design part separately.

rickalday (approved):
Passed manual QA tests

PR #7008, opened 2023-10-06
Fix: errors on the legacy forms/templates when Strip is enabled for v3 forms but not for v2 forms

glaubersilva:
## Description

If you enable Stripe gateway for V3 forms and disable it for v2 forms, then try to make donations using legacy forms/templates, you will notice a few strange behaviors that prevent the donations from getting completed. This PR fixes this bug.

## Affects

Legacy Forms - V2

## Visuals

https://www.loom.com/share/e59231ccf187498da75048cd57751cb9

## Testing Instructions

To replicate the bug, you need to follow these steps:

1. In the `Option-Based Form Editor` tab on the gateway settings, disable all Stripe gateways.
2. In the `Visual Form Builder` tab on the gateway settings, enable the Stripe - Payment Element gateway.
3. Try to make a donation with a legacy form using all 3 templates, you'll see that all of them throw exceptions in the console log and the Classic and the Multi-Steps templates do not let you finish the donation.

To test the fix, just pull this branch and repeat step 3 above.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205699277403068

### DIFF
--- a/includes/gateways/stripe/includes/filters.php
+++ b/includes/gateways/stripe/includes/filters.php
@@ -48,18 +48,20 @@ add_filter( 'give_get_payment_transaction_id-stripe_ach', 'give_stripe_get_payme
 /**
  * This function is used to add Stripe credentials to GiveWP form.
  *
- * @param array  $form_html_tags Form HTML tags.
+ * @unreleased Return $form_html_tags instead of false when Stripe gateway isn't enabled
+ * @since 2.7.0
+ *
  * @param object $form           Form Object.
  *
- * @since 2.7.0
+ * @param array  $form_html_tags Form HTML tags.
  *
  * @return array|bool
  */
 function give_stripe_form_add_data_tag_keys( $form_html_tags, $form ) {
 
 	// Must have a Stripe payment gateway active.
-	if ( ! give_stripe_is_any_payment_method_active() ) {
-		return false;
+    if ( ! give_stripe_is_any_payment_method_active()) {
+        return $form_html_tags;
 	}
 
 	$publishable_key = give_stripe_get_publishable_key( $form->ID );
--- a/includes/gateways/stripe/includes/filters.php
+++ b/includes/gateways/stripe/includes/filters.php
@@ -73,4 +73,4 @@ function give_stripe_form_add_data_tag_keys( $form_html_tags, $form ) {
 	return $form_html_tags;
 }
 
-add_filter('give_form_html_tags', 'give_stripe_form_add_data_tag_keys', 1, 2);
+add_filter('give_form_html_tags', 'give_stripe_form_add_data_tag_keys', 0, 2);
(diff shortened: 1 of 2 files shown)


JasonTheAdams:
Good catch, @glaubersilva! So, from the looks of it, this bug has been around a long time? The fix doesn't look like you're modifying anything remotely recent.

glaubersilva:
@JasonTheAdams What was added recently was the two different tabs for the gateways, so this brought the bug to the surface. Before `3.0` version, if we had all the Stripe gateways disabled (only for v2 forms in this case as the different gateway tabs didn't exist) nothing happened. Now, if Stripe is enabled on a v3 form, it will load a class that ends up loading this specific file that triggers the filter with the wrong return.

JasonTheAdams:
Fascinating! Excellent work! We'll get this reviewed and merged on Monday! 👍

includes/gateways/stripe/includes/filters.php
```diff
@@ -48,18 +48,20 @@ function give_stripe_get_payment_txn_id_fallback( $payment_id ) {
...
+ * @param array  $form_html_tags Form HTML tags.
  *
  * @return array|bool
  */
 function give_stripe_form_add_data_tag_keys( $form_html_tags, $form ) {
 
 	// Must have a Stripe payment gateway active.
-	if ( ! give_stripe_is_any_payment_method_active() ) {
-		return false;
+    if ( ! give_stripe_is_any_payment_method_active()) {
```
jonwaldstein, line 63: I'm debating if we should update the `give_stripe_is_any_payment_method_active` method as well.  It seems to be used in a few other locations, but is synonymously used with the v2 Stripe gateways.

If we left it alone, it will only refer to the option based gateway tab - which is where all the v2 stripe gateways live.  That could avoid loading things that are not necessary for v3 Stripe.

If we updated it, the stripe payment element would be included. Although, that might cause some side effects - like in this situation it would add stripe data to a v2 form even though the Stripe payment element is only used on v3 forms.

Since Stripe Payment Element does use some legacy code like the webhooks, it's possible this function could hinder some overlooked functionality.

🤔

glaubersilva: @jonwaldstein If I understand all that you said, [this commit](344faced1fcaea18a3444c9ef4bf55e747fac864) should prevent unexpected returns for this function. Let me know your thoughts on it.

jonwaldstein: @glaubersilva haha nice, as a note - I prefer not to access superglobals like that in functions because I feel that sort of thing should be elevated outside of the function - closer to the actual request.  

This did spark an idea though, how about we just add a filter to the return value `give_stripe_is_any_payment_method_active` including the `$gateways`, and `$stripePaymentMethods`.

Then if/when we ever need to include the Stripe Payment Element we can just use the filter.  (That's when I would allow reaching for superglobals)

glaubersilva: Resolved in 8c7470136e2b43da0c43b6a64d02063731bfec5f

glaubersilva:

jonwaldstein (approved):
@glaubersilva thanks for the updates, looks good to me!

rickalday (approved):
Passed manual QA tests.

PR #7014, opened 2023-10-09
Fix: Workaround to reset the draggable cursor

kjohnson:
## Description

This PR adds a workaround for the "stuck" draggable cursor when adding blocks to the form.

In the WordPress Block Editor this doesn't seem to be an issue because the Inserter sidebar is closed on block drop.

Looking at the Gutenberg code base the `draggable` component uses a body class to style the cursor when dragging a block. This should be reset on "cleanup", but does not seem to be working.

See https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/draggable/index.tsx#L230

## Testing Instructions

Drag and drop a new block/field into the form.

---
  - https://app.asana.com/0/0/1205681679114246

### DIFF
--- a/src/FormBuilder/resources/js/form-builder/src/App.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/App.tsx
@@ -34,6 +34,15 @@ if (ShortcutProvider === undefined) {
     console.error('ShortcutProvider is undefined.');
 }
 
+/**
+ * This is a workaround for a bug in Gutenberg where the drag cursor gets stuck.
+ *
+ * @unreleased
+ */
+document.addEventListener('dragend', () => {
+    document.body.classList.remove('is-dragging-components-draggable')
+});
+
 export default function App() {
     return (
         <FormBuilderErrorBoundary>
--- a/src/FormBuilder/resources/js/form-builder/src/App.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/App.tsx
@@ -9,6 +9,7 @@ import './App.scss';
 import FormBuilderErrorBoundary from '@givewp/form-builder/errors/FormBuilderErrorBounday';
 import Transfer from '@givewp/form-builder/components/onboarding/transfer';
 import {EditorStateProvider} from "@givewp/form-builder/stores/editor-state";
+import scrollIntoView from 'dom-scroll-into-view';
 
 const {blocks: initialBlocks, formSettings: initialFormSettings} = Storage.load();
 
@@ -40,7 +41,11 @@ if (ShortcutProvider === undefined) {
  * @unreleased
  */
 document.addEventListener('dragend', () => {
-    document.body.classList.remove('is-dragging-components-draggable')
+    // Reset the drag cursor.
+    document.body.classList.remove('is-dragging-components-draggable');
+
+    // Scroll the interface down by 1px to force a repaint and reset the popover position.
+    document.getElementsByClassName('interface-interface-skeleton__body')[0].scrollBy(0,1);
 });
 
 export default function App() {


JasonTheAdams:
@kjohnson Man, that's so weird how they went about this whole thing. In testing this, I did notice that the initial focus of the new field's contextual menu you misaligned:

image

What's weird is that if you move your mouse around the menu, then it will snap itself into the right position.

kjohnson:
> I did notice that the initial focus of the new field's contextual menu you misaligned

Interesting. I'm seeing that too, but not as drastically.

This is probably related to the cleanup process not firing.

image

kjohnson:
> It seems so strange that the GB drag-completion even relies on the state of the sidebar.

I don't think it relies on the state of the sidebar, but rather the cleanup is not processing and closing the sidebar resets the whole thing.

jonwaldstein:
@kjohnson @JasonTheAdams what version of WP & what browser are you using?  I'm not able to reproduce any of these issues locally with WP 6.3.1 & Brave

JasonTheAdams:
> @kjohnson @JasonTheAdams what version of WP & what browser are you using? I'm not able to reproduce any of these issues locally with WP 6.3.1 & Brave

Really? I'm running WP 6.3.1 & Arc (Chromium 117). I saw Matt Cromwell also run into this issue in his [demo video](https://www.loom.com/share/b8416217475d4306a2867b83a9f584a6?sid=3cc320b5-e45b-45b2-ac53-704c6963bae5).

jonwaldstein:
Yes, I cannot replicate.  Below is a fresh local install of [3.0.0-rc.7](https://github.com/impress-org/givewp/releases/tag/3.0.0-rc.7) using Chrome 117

JasonTheAdams:
@jonwaldstein Interesting. What about if you use the latest `release/3.0.0` with the section updates I made?

kjohnson:
@jonwaldstein try it with a field block. I'm able to reproduce the issue with the Company field block but not the Paragraph element block.

jonwaldstein:
@kjohnson you're right, i'm seeing it now with field blocks.  That begs the question though, why is this only happening for field blocks?

jonwaldstein:
Actually, it only happens with core fields - all custom fields including the text field work fine

kjohnson:
This seems to confirm that this is a configuration issue with our own custom blocks and not a larger bug in Gutenberg.

JasonTheAdams:
Try setting that true for a block and see what happens. 🤷‍♂️

kjohnson:
> Try setting that true for a block and see what happens.

```js
supports: {
    // multiple: false,
},
```

Just tried that and it seems to resolve the issue. 🤔 

So we may be back to a Gutenberg bug. 😅

JasonTheAdams:
Wow, fun times. Hahah! Let's make sure to update the GB Issue with our new findings. 😅 👍

kjohnson:
I was unable to reproduce the issue in the WordPress Block Editor using a template custom block with support for multiple disabled. 🤔

jonwaldstein (approved):
Simple enough solution for a very specific bug 👍

JasonTheAdams:
Er, wait, isn't @kjohnson pointing out that this is somehow specific to something we're doing? This workaround also introduces a new (albeit lesser) bug. I'm inclined to do a bit more digging before we push this fix along, especially if's unique to us.

jonwaldstein:
Ohh yeah, I misread that thinking this was a Gutenberg issue - _although_ it still might be.  Agreed, we should keep digging first.

kjohnson:
@JasonTheAdams I was able to correct the popover placement issue by triggering the auto-update of the floating UI.

```js
// Scroll the interface down by 1px to force a repaint and reset the popover position.
document.getElementsByClassName('interface-interface-skeleton__body')[0].scrollBy(0,1);
```

JasonTheAdams (approved):
@kjohnson I just tested this out and it's working better for me as well. For the sake of getting 3.0 out the door without this bug, I'm fine with this workaround. But, and I think you'd agree, we've officially floated into hack territory here. As such, I'd really like to dig deeper (now or a planned time later) to have a better fix in 3.1 or 3.2.

kjohnson:
> But, and I think you'd agree, we've officially floated into hack territory here

Oh, yeah. 100% agree that this is hack-y 😅

kjohnson:
Build for QA: https://github.com/impress-org/givewp/actions/runs/6474809087

rickalday (approved):
Passed manual QA tests

PR #7016, opened 2023-10-09
Hot Fix: prevent failed donations in sites that use proxy services like Cloudflare and gateways that require the original IP address

### LINKED ISSUE #7015: The give_get_ip() method should return a single IP address by default
## User Story

As an advanced user and also a general user, I want the `give_get_ip()` function to return only one IP address by default so we can ensure that payments using gateways that rely on this function's return (expecting only one IP address) don't fail.

## Details

Currently, the `give_get_ip()` function searching for any empty values, packs all the IP addresses back into a comma separated string again.

This can cause some payments to fail.  For example, Stripe + Plaid mandate fails when given multiple IP addresses in a string.

Sites using CloudFlare, by default are going to return two IP addresses in the HTTP_X_FORWARDED_FOR where the first is the user's IP and the second is the CloudFlare proxy IP (e.g. "127.0.0.1,172.70.133.31")

The `give_get_ip()` function should only return one IP address.

## Expected Behavior

In all places in the codebase where this function is used, we are expecting a single value. 

This fact can be checked here: https://github.com/search?q=org%3Aimpress-org%20give_get_ip&type=code

So, is expected that this function returns only one IP address by default.

## Steps to Reproduce

1. Setup Cloudflare and Stripe + Plaid in a test site
2. Try to do a test donation and check it will work as expected
3. The test donation should fail

## Acceptance Criteria

glaubersilva:
Resolves #7015

## Description

This PR prevents the donations from failing making sure that the `give_get_ip()` function will return only a single IP by default. 

So, even if the `$_SERVER['HTTP_X_FORWARDED_FOR']` returns more than one IP split by commas, it will get the first one related to the client and ignore the other ones related to the proxy services.

For reference, check out this: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#syntax

This PR also adds new `$single` and `$header_type` params to the `'give_get_ip'` filter

## Affects

Sites that use proxy services like Cloudflare and gateways that require the original IP address.

## Testing Instructions

1. Setup Cloudflare and Stripe + Plaid in a test site
2. Try to do a test donation and check it will work as expected
3. The test donation should NOT fail

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205692756230214

### DIFF
--- a/give.php
+++ b/give.php
@@ -6,7 +6,7 @@
  * Description: The most robust, flexible, and intuitive way to accept donations on WordPress.
  * Author: GiveWP
  * Author URI: https://givewp.com/
- * Version: 2.33.4
+ * Version: 2.33.5
  * Requires at least: 5.0
  * Requires PHP: 7.0
  * Text Domain: give
@@ -316,7 +316,7 @@ final class Give
     {
         // Plugin version.
         if (!defined('GIVE_VERSION')) {
-            define('GIVE_VERSION', '2.33.4');
+            define('GIVE_VERSION', '2.33.5');
         }
 
         // Plugin Root File.
--- a/includes/gateways/stripe/includes/give-stripe-helpers.php
+++ b/includes/gateways/stripe/includes/give-stripe-helpers.php
@@ -10,9 +10,7 @@
  * @license    https://opensource.org/licenses/gpl-license GNU Public License
  */
 
-use Give\Log\Log;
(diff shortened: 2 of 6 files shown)


jonwaldstein (approved):
Great job! ready for QA 🌈

rickalday (approved):
Passed manual QA tests

PR #7018, opened 2023-10-10
Fix: resolve strange behaviors in amount levels

JasonTheAdams:
## Description

The primary goal of this PR is to resolve this feedback: https://feedback.givewp.com/next-gen/p/donation-levels-should-obey-selections-and-deletions

In working on this, I also found that dragging options around would have a strange result if a lower amount value was moved above a high amount value — it was over-writing the moved amount.

Upon further inspection, I found there were two main issues at play:
1. The `<Draggable>` key was not guaranteed to map to the corresponding value/selection. Since the values can be dragged and deleted, the given row was not guaranteed uniqueness during the lifecycle of the `OptionsList` component, and no value (index, value, selection) could guarantee this without causing too many re-renders.
2. The shared `Options` component only has a single `setOptions` callback, which is too vague. It can be triggered when any events occur (see list below). As such, the `setOptions` was trying to do too much and couldn't make any assumptions based on the event. So it was doing things like doubling the last option, even when an option was removed.
    - Option added
    - Option removed
    - Label changed
    - Value changed
    - Options re-ordered

On top of this, the `Option` component is used in many places (including FFM), and I didn't want to have to break backwards-compatibility.

To handle the key issue, I added an optional `Option.id` property, which is assumed to be unique. If not present, it falls back on previous behavior. In the amount field I'm generating a unique integer for the id. I thought about having the `Options` component apply a unique id, but it's designed to be controlled, so the implementing component manages the options state.

Next, I defined two new event props:
- onAddOption
- onRemoveOption

If these event props are used, they will be preferred in `Options` over the `setOptions` callback, given implementing components a more granular event, allowing them to assume that specific event won't trigger `setOptions`. This helped to clean up the code and prevent some strange behaviors.

## Affects

Mostly the amount levels settings, but it does affect any setting using the `Options` component — though it should be backwards-compatible.

## Testing Instructions

Mess around with the amount levels. Add, remove, change values, drag-n-drop, and so forth. Make sure there are not strange or counter-intuitive behaviors.

## Pre-review Checklist

---
  - https://app.asana.com/0/0/1205662615497141


src/FormBuilder/resources/js/form-builder/src/blocks/fields/amount/inspector/index.tsx
```diff
@@ -121,32 +121,45 @@ const Inspector = ({attributes, setAttributes}) => {
...
+        const newLevel = {
+            id: String(Math.floor(Math.random() * 1000000)),
+            label: formatCurrencyAmount(newLevelValue),
+            value: newLevelValue,
+            checked: false,
+        };
 
-        const checkedLevel = options.filter((option) => option.checked);
+        setDonationLevels([...donationLevels, newLevel]);
+        setAttributes({levels: [...levels, Number(newLevelValue)]});
```
glaubersilva, line 141: If we delete all levels and then add the first one again, this function will not set it as default. Beyond not having a default item set, the drag-and-drop reorder will also not work.

To prevent that we need to add here something like that:

```
if (levels.length === 0) {
	setDonationLevels([
		{
			id: String(Math.floor(Math.random() * 1000000)),
			label: formatCurrencyAmount(newLevelValue.toString()),
			value: newLevelValue.toString(),
			checked: true,
		},
	]);

	setAttributes({defaultLevel: Number(newLevelValue)});
}
```

glaubersilva (changes requested):
@JasonTheAdams Great work! Almost everything worked as expected, except for one little detail which I left a comment about.

As a note, this component isn't the same as FFM, it was copied, pasted, and adapted from there. The addon uses its own component as you can check here: https://github.com/impress-org/give-form-field-manager/blob/develop/src/FormExtension/FormBuilder/Blocks/components/OptionsPanel/index.tsx

JasonTheAdams:
@glaubersilva This is ready for review, again!

src/FormBuilder/resources/js/form-builder/src/components/index.ts
```diff
@@ -1,9 +1,11 @@
...
+import {CurrencyControl} from './CurrencyControl';
+import Options from './OptionsPanel';
 
 export {default as Header} from './header';
 export {SecondarySidebar, Sidebar} from './sidebar';
 
 export default function registerComponents() {
     window.givewp.form.components = window.givewp.form.components || {};
     window.givewp.form.components.CurrencyControl = CurrencyControl;
+    window.givewp.form.components.DraggableOptionsControl = Options;
```
glaubersilva, line 10: The `components` property is a child `givewp` directly and not of the `form` property, so we need to remove the `.form` part of these lines because this is preventing the assets from getting compiled and throwing a few errors in the console. 


image

glaubersilva (changes requested):
@JasonTheAdams I found a problem related to the components mounted on the window global object and left a comment with more details about it.

src/FormBuilder/resources/js/form-builder/src/components/index.ts
JasonTheAdams: @glaubersilva I ended up removing the exporting of components as we're going to go about this a different way by introducing an NPM package of Form Builder components in a later PR.

JasonTheAdams:
@glaubersilva This is ready for review, again!

src/FormBuilder/resources/js/form-builder/src/index.tsx
```diff
@@ -33,7 +32,6 @@ setCategories([
 ]);
 
 registerHooks();
-registerComponents();
```
jonwaldstein, line 36: We can't delete this at the moment, I did a quick search and can see fee recovery using `window.givewp.form.components.CurrencyControl`

Screenshot 2023-11-02 at 4 09 49 PM

src/FormBuilder/resources/js/form-builder/src/components/OptionsPanel/index.tsx
```diff
@@ -28,15 +42,14 @@ export default function Options({currency, multiple, selectable = true, options,
...
-                            currency,
-                            options,
-                            showValues,
-                            multiple,
-                            selectable,
-                            setOptions,
-                            defaultControlsTooltip,
-                        }}
+                        currency={currency}
+                        options={options}
```
jonwaldstein, line 46: I don't know if this is related, but I noticed the donor name honorific options is being displayed as numbers:

Screenshot 2023-11-02 at 4 19 04 PM

jonwaldstein (changes requested):

src/FormBuilder/resources/js/form-builder/src/components/OptionsPanel/index.tsx
JasonTheAdams: Was fixed in #7088

src/FormBuilder/resources/js/form-builder/src/index.tsx
JasonTheAdams: Gooood catch! Reverted in 9401918

JasonTheAdams:
@jonwaldstein This is ready for review again!

jonwaldstein (approved):
Looks good 👍

rickalday:
Passed manual QA tests.

PR #7020, opened 2023-10-10
Refactor: optimize form meta migration query

pauloiankoski:
## Description

While fixing migration bugs, I noticed that the serialized form meta data was being re-serialized during migration, causing fatal errors due to changes in the data structure.

Initially, I considered unserializing the data before inserting it into the database and then serializing it again. However, after discussing it, we decided to completely revamp the way we migrate form metas.

In this Pull Request, we have replaced the previous approach of using a for-loop to store all data in memory and then performing individual insertions. Instead, we now use a single `INSERT INTO` query to streamline the data insertion process, improving both the accuracy and speed of the migration.

## Affects

Form Migration

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205692499999969

### DIFF
--- a/src/FormMigration/Steps/FormMeta.php
+++ b/src/FormMigration/Steps/FormMeta.php
@@ -21,17 +21,21 @@ class FormMeta extends FormMigrationStep
         $formMetaTable = DB::prefix('give_formmeta');
 
         DB::query(
-            "
-                INSERT INTO $formMetaTable (form_id, meta_key, meta_value)
-                SELECT
-                    $newFormId,
-                    meta_key,
-                    meta_value
-                FROM
-                    $formMetaTable
-                WHERE
-                    form_id = $oldFormId
+            DB::prepare(
                 "
+                    INSERT INTO $formMetaTable (form_id, meta_key, meta_value)
+                    SELECT
+                        %d,
+                        meta_key,
+                        meta_value
+                    FROM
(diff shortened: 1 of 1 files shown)


src/FormMigration/Steps/FormMeta.php
```diff
@@ -17,13 +18,20 @@ public function process()
...
+                INSERT INTO $formMetaTable (form_id, meta_key, meta_value)
+                SELECT
+                    $newFormId,
+                    meta_key,
+                    meta_value
+                FROM
+                    $formMetaTable
+                WHERE
+                    form_id = $oldFormId
+                "
```
JasonTheAdams, line 34: Let's be good kids and use `DB::prepare` to insert the `$newFormId` and $oldFormId`. 🍭

JasonTheAdams (approved):
Love this improvement, @pauloiankoski! This will be more reliable and _way_ faster! One minor suggestion, but I'll approve.

src/FormMigration/Steps/FormMeta.php
pauloiankoski: Added here: ddb4fb3e86a412f9c079df0d6bc6bd086d42babb

rickalday (approved):
Passed manual QA tests

PR #7021, opened 2023-10-10
Fix: Prevent warnings from displaying in form iframe

JasonTheAdams:
## Description

This addresses feedback such as: https://feedback.givewp.com/next-gen/p/deprecated-error-notice-when-attempting-to-migrate-a-20-form-to-30

If any kind of deprecation, error, warning, or just plain anything is written to the output buffer, it's currently showing up inside our forms and confirmation receipts. It's fine for things to show up in other places, but we specifically don't want them to show up there, in front of donors.

This clears the output buffer and sets `ini_set('display_errors', 0)` to prevent PHP errors (which supersedes the output buffer). This isn't a perfect system, as any errors/warnings that occur _before_ the `ini_set` will still show up. As such, I put them in the router closures as the earliest point in time we could squelch the errors.

## Visuals

Without this PR:
image

## Testing Instructions

Use the WordPress Beta Test plugin to update to WP 6.4 beta. Then preview the form in the Form Builder (design mode), and make a donation on the front end.

## Pre-review Checklist

### DIFF
--- a/src/DonationForms/Controllers/DonationFormViewController.php
+++ b/src/DonationForms/Controllers/DonationFormViewController.php
@@ -25,6 +25,7 @@ class DonationFormViewController
             $donationForm->settings
         );
 
+        ob_clean();
         return $viewModel->render();
     }
 
@@ -44,6 +45,7 @@ class DonationFormViewController
             $data->formSettings ?: $donationForm->settings
         );
 
+        ob_clean();
         return $viewModel->render(true);
     }
 }
--- a/src/DonationForms/Controllers/DonationConfirmationReceiptViewController.php
+++ b/src/DonationForms/Controllers/DonationConfirmationReceiptViewController.php
@@ -27,6 +27,7 @@ class DonationConfirmationReceiptViewController
             return '';
         }
 
+        ob_clean();
         return (new DonationConfirmationReceiptViewModel($donation))->render();
     }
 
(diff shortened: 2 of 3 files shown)


JasonTheAdams:
@jonwaldstein Mind giving this a look over?

src/DonationForms/Controllers/DonationFormViewController.php
```diff
@@ -25,6 +25,7 @@ public function show(DonationFormViewRouteData $data): string
             $donationForm->settings
         );
 
+        ob_clean();
```
jonwaldstein, line 28: Does calling `ob_clean()` actually do anything? We area already starting and closing the buffer in the render method after this using `ob_start()` and `ob_get_clean()`.

JasonTheAdams: Yeah, if you add `echo 'ARGH!'` on that line and remove the `ob_clean()` it will show up. Double-check me, though.

jonwaldstein: Ah I see, how about we put the `ob_clean()` right before `ob_start()` then in the vewModel render methods to make sure it's always cleared before starting.

JasonTheAdams: I thought about that. The only thing that held me back was the idea that perhaps the `viewModel` shouldn't worry about clearing the previous buffer, as much as capturing its own output. Separations of concerns sort of thing. What do you think?

jonwaldstein: hmm, yeah I guess that makes sense. carry on then lol

jonwaldstein (approved):
👍

rickalday (approved):
Passed manual QA tests

PR #7022, opened 2023-10-10
Feature: Use currency control for goal amount

JasonTheAdams:
## Description

This simple little PR uses the Currency control for the Goal amount, which was previously a number control. So it looks all nice, now! I also increased the debounce delay to 500ms, as 100ms just felt way too short.

## Affects

The donation goal

## Visuals

image

## Testing Instructions

Play around with the donation goal and make sure it looks right and works in the front-end.

## Pre-review Checklist

---
  - https://app.asana.com/0/0/1205076251265772
  - https://app.asana.com/0/0/1205681940003053


src/FormBuilder/resources/js/form-builder/src/settings/donation-goal/index.tsx
```diff
@@ -69,11 +69,11 @@ const DonationGoalSettings = () => {
                         />
                     </PanelRow>
                     <PanelRow>
-                        <NumberControl
+                        <CurrencyControl
```
jonwaldstein, line 72: @JasonTheAdams the goal amount is an arbitrary number that is used throughout every goal type.  In order for a currency control to make sense, you would need to first make sure the goal type refers to a dollar amount.

jonwaldstein (changes requested):

src/FormBuilder/resources/js/form-builder/src/settings/donation-goal/index.tsx
JasonTheAdams: Hah! Oh yeah. Good catch!

JasonTheAdams: Resolved in 5e8217d

jonwaldstein (approved):
Looks good now 👍

rickalday (approved):
Passed manual QA tests.

PR #7025, opened 2023-10-11
Feature: add migration step for Fee Recovery settings

pauloiankoski:
## Description

In this pull request, we add a new step to move Fee Recovery settings from a v2 form to the new  v3 form.

## Affects

Form Migrations

## Testing Instructions

1. Create v2 forms
2. Set Fee Recovery in many different combinations
3. Migrate those forms
4. Confirm your choices in v2 are reflected in the recently created v3 form

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205703084943466

### DIFF
--- a/src/FormMigration/FormMetaDecorator.php
+++ b/src/FormMigration/FormMetaDecorator.php
@@ -455,7 +455,7 @@ class FormMetaDecorator extends FormModelDecorator
     {
         $feeRecoveryStatus = $this->getMeta('_form_give_fee_recovery');
 
-        if ($feeRecoveryStatus === 'disabled') {
+        if (empty($feeRecoveryStatus) || $feeRecoveryStatus === 'disabled') {
             return [];
         }
 
--- a/src/FormMigration/FormMetaDecorator.php
+++ b/src/FormMigration/FormMetaDecorator.php
@@ -5,6 +5,7 @@ namespace Give\FormMigration;
 use Give\DonationForms\V2\Models\DonationForm;
 use Give\DonationForms\ValueObjects\GoalType;
 use Give\FormMigration\Contracts\FormModelDecorator;
+use Give\PaymentGateways\Gateways\Stripe\StripePaymentElementGateway\StripePaymentElementGateway;
 use Give_Email_Notification_Util;
 
 class FormMetaDecorator extends FormModelDecorator
--- a/src/FormMigration/FormMetaDecorator.php
+++ b/src/FormMigration/FormMetaDecorator.php
@@ -487,7 +487,7 @@ class FormMetaDecorator extends FormModelDecorator
         }
 
         return [
-            'useGlobalSettings' => $feeRecoveryStatus === 'global',
+            'useGlobalSettings' => false,
             'feeSupportForAllGateways' => $this->getMeta('_form_give_fee_configuration') === 'all_gateways',
             'perGatewaySettings' => $perGatewaySettings,
             'feePercentage' => (float)$this->getMeta('_form_give_fee_percentage'),
--- a/src/FormMigration/FormMetaDecorator.php
+++ b/src/FormMigration/FormMetaDecorator.php
@@ -460,23 +460,32 @@ class FormMetaDecorator extends FormModelDecorator
 
         return [
             'useGlobalSettings' => $feeRecoveryStatus === 'global',
-            'feeSupportForAllGateways' => give_get_meta(
-                    $this->form->id,
-                    '_form_give_fee_configuration',
-                    true
-                ) === 'all_gateways',
+            'feeSupportForAllGateways' => $this->getMeta('_form_give_fee_configuration') === 'all_gateways',
             'perGatewaySettings' => [],
-            'feePercentage' => (float)give_get_meta($this->form->id, '_form_give_fee_percentage', true),
-            'feeBaseAmount' => (float)give_get_meta($this->form->id, '_form_give_fee_base_amount', true),
-            'maxFeeAmount' => (float)give_get_meta($this->form->id, '_form_give_fee_maximum_fee_amount', true),
-            'includeInDonationSummary' => give_get_meta(
-                    $this->form->id,
-                    '_form_breakdown',
-                    true
-                ) === 'enabled',
-            'donorOptIn' => give_get_meta($this->form->id, '_form_give_fee_mode', true) === 'donor_opt_in',
-            'feeCheckboxLabel' => give_get_meta($this->form->id, '_form_give_fee_checkbox_label', true),
-            'feeMessage' => give_get_meta($this->form->id, '_form_give_fee_explanation', true),
+            'feePercentage' => (float)$this->getMeta('_form_give_fee_percentage'),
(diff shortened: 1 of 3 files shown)


src/FormMigration/FormMetaDecorator.php
```diff
@@ -446,4 +446,67 @@ public function getFormFieldsPlacement(): string
...
+        $feeRecoveryStatus = $this->getMeta('_form_give_fee_recovery');
+
+        if ($feeRecoveryStatus === 'disabled') {
+            return [];
+        }
+
+        $perGatewaySettings = [];
+        $gateways = give_get_ordered_payment_gateways(give_get_enabled_payment_gateways());
+        $gatewaysMap = [
+            'stripe' => 'stripe_payment_element',
```
jonwaldstein, line 464:
```suggestion
            'stripe' => StripePaymentElementGateway::id(),
```

src/FormMigration/Steps/FeeRecovery.php
```diff
@@ -0,0 +1,39 @@
...
+        $feeRecoverySettings = $this->formV2->getFeeRecoverySettings();
+
+        if (empty($feeRecoverySettings)) {
+            if ($feeRecoveryBlock) {
+                $this->fieldBlocks->remove($feeRecoveryBlock->name);
+            }
+
+            return;
+        }
+
```
jonwaldstein, line 26: I'm not sure what's going on here. The form migration should just be concerned with v2 to v3.  I don't think we need to be removing blocks.

src/FormMigration/FormMetaDecorator.php
```diff
@@ -446,4 +446,67 @@ public function getFormFieldsPlacement(): string
...
+    /**
+     * @unreleased
+     */
+    public function getFeeRecoverySettings(): array
+    {
+        $feeRecoveryStatus = $this->getMeta('_form_give_fee_recovery');
+
+        if ($feeRecoveryStatus === 'disabled') {
+            return [];
+        }
```
jonwaldstein, line 459: I would reverse this and check specifically if it's enabled.

```
give_is_setting_enabled( give_get_meta( $form_id, '_form_give_fee_recovery', true );
```

If it's set to global we should just let the global/default GiveWP fee recovery setting determine if it should add the block or not.

jonwaldstein (changes requested):

src/FormMigration/Steps/FeeRecovery.php
pauloiankoski: The Fee Recovery block is automatically included in new forms when the global setting is enabled. If we are migrating a v2 form where Fee Recovery has been disabled, we should remove the block from the form. This is necessary because the new form is added before the migration steps are executed.

src/FormMigration/FormMetaDecorator.php
pauloiankoski: I intended to implement an early return if Fee Recovery is disabled then to preserve all existing options in the new form, even if it is configured to use global settings.

src/FormMigration/Steps/FeeRecovery.php
jonwaldstein: Since the form is being built programmatically, the migration should be bypassing the `givewp_form_builder_new_form` action that default options are hooking into.  If that's _not_ the case then we need to update the migration processing because the migration should not be concerned with those side effects for brand new forms.

src/FormMigration/FormMetaDecorator.php
jonwaldstein: I would just return early if it's not enabled.  I don't see the point in reaching for meta if it's set to global.

src/FormMigration/Steps/FeeRecovery.php
pauloiankoski: In fact, changed the migration on e5f8ca134265bc06d4a0fb2d9d4ebb98bdaaa4b1 and it doesn't include the block automatically.

src/FormMigration/FormMetaDecorator.php
pauloiankoski: Could you confirm that https://github.com/impress-org/givewp/pull/7025/commits/c627f0120cb8ecc94237859c7fc26a25218fd425 is correct?

src/FormMigration/FormMetaDecorator.php
```diff
@@ -446,4 +447,73 @@ public function getFormFieldsPlacement(): string
...
     }
+
+    /**
+     * @unreleased
+     */
+    public function getFeeRecoverySettings(): array
+    {
+        $feeRecoveryStatus = $this->getMeta('_form_give_fee_recovery');
+
+        if ($feeRecoveryStatus === 'disabled') {
```
jonwaldstein, line 458:
```suggestion
        if (empty($feeRecoveryStatus) || $feeRecoveryStatus === 'disabled') {
```

src/FormMigration/FormMetaDecorator.php
```diff
@@ -446,4 +447,73 @@ public function getFormFieldsPlacement(): string
...
+            return [];
+        }
+
+        if ($feeRecoveryStatus === 'global') {
+            return [
+                'useGlobalSettings' => true,
+            ];
+        }
+
```
jonwaldstein, line 467: Just add one more conditional here for:

```
if ($feeRecoveryStatus !== 'enabled') {
      return [];
}
```

I've come to realize its better to assume legacy meta is either empty or disabled by default - so we should check if its specifically enabled instead of assuming it is because it's _not_ disabled.

src/FormMigration/Steps/FeeRecovery.php
```diff
@@ -0,0 +1,31 @@
...
+        if (empty($feeRecoverySettings) || (
+                $feeRecoverySettings['useGlobalSettings'] === true &&
+                !give_is_setting_enabled(give_get_option('give_fee_recovery', 'disabled'))
+            )) {
+            return;
+        }
+
+        $feeRecoveryBlock = BlockModel::make([
+            'name' => 'givewp-fee-recovery/fee-recovery',
+            'attributes' => $feeRecoverySettings,
```
jonwaldstein, line 27: Will the block work properly if only the useGlobalSettings is set to true?  (without setting the other attributes). Lets just make sure of that.

jonwaldstein (changes requested):

src/FormMigration/FormMetaDecorator.php
pauloiankoski: Resolved here: af0ff4719c21c21d87b3d7b2831633795d01faee

src/FormMigration/Steps/FeeRecovery.php
pauloiankoski: No. The block is broken now. Populating the array with the global options now: 7d8baf91a96296bde35c1c6fc2b1c7674e886ee4

jonwaldstein (approved):
🌈

rickalday (approved):
Passed manual QA tests

PR #7026, opened 2023-10-11
Fix:  update initial and 4th step of the block builder tour 

JoshuaHungDinh:
## Description
This PR focuses on enhancing user experience by making necessary adjustments to the Block Builder Tour. The changes address wording on the initial tour modal & a bug causing the block adding sidebar to close when navigating forward or backward, then returning to step 4. 

As per Zeplin the block adding sidebar should remain open after reaching step 4. The singular button that manages the sidebar display, does so in a way much like a toggle. On step 4 initially we had a synchronous call within a promise to click the sidebar handling button. This would cause it toggle it each time the step was revisited. 

To address this we simply added a conditional to only click the button if the button had not been clicked previously,

## Affects

Block Builder tour

## Visuals

https://github.com/impress-org/givewp/assets/75056371/4c396994-9474-4fd7-938e-3323abb0f826

## Asana task/s
https://app.asana.com/0/1203547864809090/1205684297339067

## Testing Instructions

- Verify the block builder tour displays the new text on the initial modal.
- Proceed to step 4 and use both the previous and next button to navigate away from step 4
- Return to step 4 and verify that the side remains open.

## Pre-review Checklist

---
  - https://app.asana.com/0/0/1205701831349179


kjohnson (approved):

rickalday (approved):
Passed manual tests

PR #7029, opened 2023-10-12
Fix: conditionally set gateway id default value

jonwaldstein:
## Description

This conditionally checks sets the donation form's default value for `gatewayId` - making sure it's only set if not empty. 

## Affects

The donation form

## Visuals

Screenshot 2023-10-12 at 1 34 57 PM

## Testing Instructions

- Step through a multi step form without a default gateway
- You should be able to get to the end but not actually submit.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
+++ b/src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
@@ -174,7 +174,7 @@ class ConvertDonationFormBlocksToFieldsApi
                     ->testMode(give_is_test_mode())
                     ->rules(new GatewayRule())
                     ->required()
-                    ->defaultValue($defaultGatewayId);
+                    ->defaultValue(!empty($defaultGatewayId) ? $defaultGatewayId : null);
 
             case "givewp/donation-summary":
                 return DonationSummary::make('donation-summary');


JasonTheAdams:
@jonwaldstein Mind generating a zip from this, putting it on the environment @mathetos was having the issue with, and seeing if it fixes it?

jonwaldstein:
> @jonwaldstein Mind generating a zip from this, putting it on the environment @mathetos was having the issue with, and seeing if it fixes it?

Confirmed it fixes the issue on staging site:

Screenshot 2023-10-12 at 1 57 44 PM

I think this situation will be rare but still good to avoid.

JasonTheAdams:
Nice! Any theories as to _why_ it's happening in the first place? I wonder if what's happening is that the default gateway is set to some gateway that's not active anymore. 🤔

jonwaldstein:
This is the gateway configuration on that install.  Not sure why, but there are no v3 gateways enabled.  It's possible the migration didn't go through for this instance or the gateway config was setup statically like this.

Screenshot 2023-10-12 at 2 09 01 PM

Screenshot 2023-10-12 at 2 09 06 PM

JasonTheAdams:
Interesting. So this fix sets the default as null, which the form takes to mean that it should just set enable whatever the first gateway is?

jonwaldstein:
> Interesting. So this fix sets the default as null, which the form takes to mean that it should just set enable whatever the first gateway is?

Yes, just making sure we're only setting the default gateway when it's defined (or else the empty string gets put through default values and validation route).  The default gateway would only be _undefined_ if _no_ v3 gateways were enabled.

JasonTheAdams (approved):
Strange situation, but at least this fails more gracefully, allowing the donor to proceed.

PR #7030, opened 2023-10-12
Fix: Update FFM migration to support Donation Amount

kjohnson:
## Description

This PR updates the FFM migration to support conditional logic based on the Donation Amount block.

This also resolves an issue with the operator encoded as an HTML entity.

## Visuals

Peek 2023-10-12 15-38

## Testing Instructions

- Add a custom field to a v2 form
- Set a condition for 'Donation Amount is greater than 30'
- Migrate the form

---
  - https://app.asana.com/0/0/1205710844260106


src/Framework/FieldsAPI/Concerns/HasVisibilityConditions.php
```diff
@@ -75,11 +75,15 @@ public function failsVisibilityConditions(array $values): bool
...
      * Set a condition for showing the node.
      *
+     * @unreleased Fixed a bug with encoded HTML entities in the operator, ie > (greater than), < (less than)
      * @since 2.27.3 update to use SimpleConditionSet
      * @since 2.13.0
      */
     public function showIf(string $field, string $operator, $value, string $boolean = 'and'): self
     {
+        // Fixes a bug with encoded HTML entities in the operator, ie > (greater than), < (less than)
+        $operator = html_entity_decode($operator);
```
kjohnson, line 85: This might be better placed elsewhere, but this is the first attempt.

src/FormMigration/Steps/FormFieldManager.php
```diff
@@ -19,6 +19,7 @@ class FormFieldManager extends FormMigrationStep
      * - "field_width" (Field Width)
      * - "css" (CSS Class Name)
      *
+     * @unreleased @Added support for conditions based on Donation Amount.
```
jonwaldstein, line 22:
```suggestion
     * @unreleased added support for conditions based on Donation Amount.
```

src/FormMigration/Steps/FormFieldManager.php
```diff
@@ -28,7 +29,12 @@ public function process()
         if (!$formFields) {
             return;
         }
-        
+
+        $this->fieldBlockRelationships['give-amount'] = [
+            'field' => [], // This property intentionally left empty.
```
jonwaldstein, line 34: Can you explain why this is intentional? i'm just confused what's going on here.

src/FormMigration/Steps/FormFieldManager.php
```diff
@@ -423,40 +429,39 @@ private function insertBlock($block): void
...
-                if ($block->name !== 'givewp/section') {
-                    $block->setAttribute('conditionalLogic', [
-                        'enabled' => false,
-                        'action' => 'show',
-                        'boolean' => 'and',
-                        'rules' => [],
-                    ]);
-                }
+            // Initialize conditional logic support for custom fields.
+            $block->setAttribute('conditionalLogic', [
```
jonwaldstein, line 441: Shouldn't we be checking if conditional logic is enabled first before adding the attribute? I find it strange that we're adding conditional logic that does nothing.

kjohnson: The attribute is an object and doubles as a flag for the block inspector controls. Otherwise you couldn't enable conditional logic for the migrated field.

src/FormMigration/Steps/FormFieldManager.php
kjohnson: Fair point. I'll either update the comment or might be able to just remove it.

src/Framework/FieldsAPI/Concerns/HasVisibilityConditions.php
JasonTheAdams: Hmm. So what, exactly, is sending the html-entity form of the operator? FFM?

src/FormMigration/Steps/FormFieldManager.php
kjohnson: Hopefully this clarifies, see https://github.com/impress-org/givewp/pull/7030/commits/0486e18ffdfbe654c0506120e1470927bba445e7.

src/FormMigration/Steps/FormFieldManager.php
kjohnson: I updated the comment to clarify _why_, see https://github.com/impress-org/givewp/pull/7030/commits/b88d65562ff50ac6e75fe97578d7a7c7cc4ad274.

src/Framework/FieldsAPI/Concerns/HasVisibilityConditions.php
kjohnson: It looks like the `$operator` is being encoded in the preview data sent over the wire from the builder.

JasonTheAdams: I'd trace it back and either prevent it from being encoded when sent, or decode it on the receiving end before it's used in this function.

kjohnson: It looks like the `$operator` is being encoded in the preview data sent over the wire from the builder.

kjohnson: Actually, I think this might be the best place for it, since the data is so nested. This is the central point at which the operator is passed to the validation rules. Otherwise, we'd have to cover multiple entry points to deeply decode HTML entities for this specific case.

jonwaldstein (approved):
Looks good to me, tested it locally and seems to be working as expected 👍

kjohnson:
Building for QA: https://github.com/impress-org/givewp/actions/runs/6512286540

Benunc (approved):
Tested and it works!

src/Framework/FieldsAPI/Concerns/HasVisibilityConditions.php
JasonTheAdams: Let's talk about this more next week. I feel like this is an encroachment of responsibility. The Field API (and especially conditions) shouldn't have the foggiest clue as to what an HTML entity even is. It's removed from that layer. If it's multiple points to correct elsewhere, we can refactor in another way, but I don't think this is the right place to do it, simply the easiest at this point.

PR #7033, opened 2023-10-12
Fix: required fields in following steps no longer break multi-step

jonwaldstein:
## Description

When default values are set to empty strings, they are being added to the default values which affects the multi step validation.  This make sure not to add default values when they are null or empty strings.

Without this PR, custom fields that are required will break the flow of multi-step forms.

Technically this could be avoided by conditionally setting the defaultValue of a Fields api Field to null but that's easier said than done at this point considering its third party and ffm usage.

## Affects

Donation form - specifically multi-step when custom field is required with no default value

## Visuals

N/A

## Testing Instructions

- Create a v3 Form and set to multi step design
- Add a new second section with a custom text field - set to required (no default value)
- Make sure you can navigate through the multi step form.
- Repeat using ffm fields

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205712694264183
  - https://app.asana.com/0/0/1205712930033935

### DIFF
--- a/src/DonationForms/resources/app/utilities/getDefaultValuesFromSections.ts
+++ b/src/DonationForms/resources/app/utilities/getDefaultValuesFromSections.ts
@@ -13,8 +13,10 @@ export default function getDefaultValuesFromSections(sections: Section[]) {
     return reduceFields(
         sections,
         (values, field) => {
-            if (field.defaultValue !== null) {
-                values[field.name] = field.defaultValue;
+            const defaultValue = field.defaultValue;
+
+            if (defaultValue !== null && defaultValue !== '') {
+                values[field.name] = defaultValue;
             }
 
             return values;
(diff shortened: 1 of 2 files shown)


JasonTheAdams (approved):
Hmm, this makes me a bit nervous for multi-select fields wherein one option's value is an empty string, but for the sake of time I'll move this along. We can adjust for that if and when that time comes.

rickalday (approved):
Passed manual QA tests

jonwaldstein:
> Hmm, this makes me a bit nervous for multi-select fields wherein one option's value is an empty string, but for the sake of time I'll move this along. We can adjust for that if and when that time comes.

@JasonTheAdams I tested this with the ffm multi-select block/field (checkbox and dropdown) in different configurations - even specifying empty values and it's working properly:

Screenshot 2023-10-13 at 10 22 19 AM

Screenshot 2023-10-13 at 10 21 33 AM

PR #7043, opened 2023-10-16
Fix: Resolve Matomo compatibility

JasonTheAdams:
## Description

GiveWP 3.0 updated the Faker package, which in turn updated the Psr/Container package. The container interface then conflicts with the [Matomo plugin](https://wordpress.org/plugins/matomo/).

This PR adds Faker to the Strauss list, so it and it's sub-dependencies are prefixed, avoiding conflicts.

In adding Faker to Strauss, I found that the `symfony/deprecation-contracts` package was having issues, as it's a single auto-loaded file. Strauss wasn't updating the file path, so composer was throwing a problem. Since the file is a root-level, conditionally loaded, single function, I decided it was simpler to just have that package excluded from Strauss. That fixed the issue.

## Affects

Faker, which is used by our model factories.

## Testing Instructions

The unit tests should really catch everything here. But we can also use `wp test-give donations` to test making some mock data via the affected factories.

## Pre-review Checklist

---
  - https://app.asana.com/0/0/1205734884251786

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -93,12 +93,18 @@
             "classmap_prefix": "Give_Vendors_",
             "constant_prefix": "GIVE_VENDORS_",
             "packages": [
+                "fakerphp/faker",
                 "stellarwp/validation",
                 "stellarwp/field-conditions",
                 "symfony/polyfill-ctype",
                 "symfony/polyfill-mbstring",
                 "symfony/http-foundation"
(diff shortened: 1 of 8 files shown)


pauloiankoski (approved):
🌈

alexwcoleman (approved):

PR #7046, opened 2023-10-17
Fix: correct Form ID Filtering for Shortcodes and Blocks

pauloiankoski:
## Description

This pull request addresses an issue where users were experiencing unexpected behavior when trying to filter blocks and shortcodes by Form ID. Even after setting the Form ID in the filter, that value was being ignored in the query, returning all values instead of the filtered set. This PR aims to correct this issue to ensure that filtering by Form ID works as expected.

In a previous pull request (https://github.com/impress-org/givewp/pull/6952), we initially replaced `array_walk` with `array_map` to prevent fatal errors. However, this led to an issue. The `array_map` function does not support changing elements by reference, which the `_give_redirect_form_id` function requires. Consequently, `array_map` returned a nullable value, causing the shortcodes to skip any filtering by Form ID.

This pull request partially reverts the changes made in the previous pull request. It reintroduces `array_walk` before the original array is converted to CSV format.

## Affects

Form Grid, Donor Wall, MultiForm Goal and Progress Bar blocks & shortcodes

## Testing Instructions

1. Add the blocks or shortcodes above to a page
2. Filter by Form Id
3. Confirm the output only contains data related to the passed Form Id(s).

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205730825400239


alaca (approved):
Looks good :+1:

matheuswd (approved):

PR #7047, opened 2023-10-18
Refactor: add support to get all gateways on the gateways functions

pauloiankoski:
## Description

After splitting the gateways based on whether they support v2 or v3 forms, we found that it was not possible to get a complete list of all active gateways in Give. This complete list is important for other features, like setting Fee Recovery options for each gateway.

When we made this split, we chose to make sure the `$version` parameter defaults to `2`. This was done to maintain compatibility with older v2 code, ensuring that it would continue to work seamlessly. However, we recently realized that this might not be the best solution. This is because, aside from the donation forms themselves, most other settings require access to the complete list of gateways.

In this pull request, we've changed the default value of the `$version` parameter to `null`. This allows us to combine the lists of v2 and v3 gateways when this parameter has no value defined.

## Affects

Gateways functions

## Testing Instructions

In general, ensure that all code using `give_get_enabled_payment_gateways`, `give_get_payment_gateways` or `give_get_ordered_payment_gateways` are still working as expected.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205749663675382

### DIFF
--- a/includes/gateways/functions.php
+++ b/includes/gateways/functions.php
@@ -23,10 +23,8 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @param int|null $version
  * @return array $gateways All the available gateways
  */
-function give_get_payment_gateways($version = 2)
+function give_get_payment_gateways($version = null)
 {
-    $suffix = $version === 3 ? '_v3' : '';
-
     // Default, built-in gateways
     $gateways = apply_filters(
         'give_payment_gateways',
@@ -43,7 +41,14 @@ function give_get_payment_gateways($version = 2)
         }
     });
 
-    $gatewayLabels = give_get_option('gateways_label' . $suffix, []);
+    $gatewayLabels = [];
+    if (!$version || $version === 2) {
+        $gatewayLabels = array_merge($gatewayLabels, (array)give_get_option('gateways_label', []));
+    }
+
+    if (!$version || $version === 3) {
+        $gatewayLabels = array_merge($gatewayLabels, (array)give_get_option('gateways_label_v3', []));
+    }
 
     // Replace payment gateway checkout label with admin defined checkout label.
     if ($gatewayLabels) {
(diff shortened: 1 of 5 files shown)


includes/gateways/functions.php
```diff
@@ -67,12 +72,18 @@ function give_get_payment_gateways($version = 2)
...
 
-    $enabled = $_POST['gateways' . $suffix] ?? give_get_option('gateways' . $suffix);
+    $enabled = [];
+    if (!$version || $version === 2) {
+        $enabled = array_merge($enabled, $_POST['gateways'] ?? (array)give_get_option('gateways', []));
+    }
+
+    if (!$version || $version === 3) {
+        $enabled = array_merge($enabled, $_POST['gateways_v3'] ?? (array)give_get_option('gateways_v3', []));
+    }
```
jonwaldstein, line 86: Since we're using `array_merge` now, let's make sure we're validating / casting the `$_POST` inputs to an array to prevent a potential type error.

Very verbose example 😄 

```
$gatewaysFromPostRequest = isset($_POST['gateways']) && is_array($_POST['gateways']) ? (array)$_POST['gateways'] : null;
```

includes/gateways/functions.php
```diff
@@ -345,19 +356,30 @@ function give_count_sales_by_gateway( $gateway_id = 'paypal', $status = 'publish
...
+            $_POST['gateways'] ?? (array)give_get_option('gateways', [])
+        );
+    }
+
+    if (!$version || $version === 3) {
+        $gateways_setting = array_merge(
+            $gateways_setting,
+            $_POST['gateways_v3'] ?? (array)give_get_option('gateways_v3', [])
+        );
+    }
```
jonwaldstein, line 382: Same comment here about the `$_POST` request to be resolved as array

jonwaldstein (changes requested):

includes/gateways/functions.php
pauloiankoski: 08fbab2fcdd384ae5c6550d00056edc5ee3ef7dc

includes/gateways/functions.php
pauloiankoski: We don't need that level of verbosity 😆 
If it's already an array, there's no reason to cast it. So casting any value does do the trick for us.

08fbab2fcdd384ae5c6550d00056edc5ee3ef7dc

jonwaldstein (approved):
Looks good, thx buddy

rickalday (approved):
Passed manual QA tests

PR #7048, opened 2023-10-18
Fix: set correct gatewayId when migrating Fee Recovery settings for Stripe

pauloiankoski:
## Description

Currently, when an user migrates a form with Fee Recovery custom settings and has the per-gateway settings option enabled, Stripe options were not being migrated properly.

We have found that the migration script contains an error when mapping gateway ids from v2 to v3: the gateway id for v3 was being used to retrieve the form settings, which obviously will not be found in a v2 form.

This pull request fixes that issue by using the original gateway id to retrieve the settings while using the updated gateway id to set the settings onto the v3 form.

## Affects

Fee Recovery migration

## Testing Instructions

1. Add a v2 form with Fee Recovery
2. Set it to use per-gateway settings
3. Edit Stripe values
4. Migrate the form
5. Confirm that Stripe values were properly migrated to the Fee Recovey block in v3.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205730825400250

### DIFF
--- a/give.php
+++ b/give.php
@@ -6,7 +6,7 @@
  * Description: The most robust, flexible, and intuitive way to accept donations on WordPress.
  * Author: GiveWP
  * Author URI: https://givewp.com/
- * Version: 3.0.1
+ * Version: 3.0.2
  * Requires at least: 6.0
  * Requires PHP: 7.2
  * Text Domain: give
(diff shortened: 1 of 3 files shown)


jonwaldstein (approved):
Good catch 👍

rickalday (approved):
Passed manual QA tests.

PR #7050, opened 2023-10-19
Feature: Confirm publishing

alaca:
## Description

This PR adds a new pre-publish confirmation panel. Now when the site admin tries to publish the form, the confirmation panel will appear with basic info about the form, similar to what  WP does when publishing a page.  Also, the form visibility option is added where the site admin can change the form visibility to private or public.

## Affects
Form builder - form publishing

## Visuals

Screen Capture_select-area_20231019110646

## Testing Instructions

Add a new form and hit publish
Play with visibility options

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1203889335107917
  - https://app.asana.com/0/0/1205755858493631


alaca:
Plugin zip for QA https://github.com/impress-org/givewp/suites/17399325127/artifacts/994411571

kjohnson:
@jdghinson I know that the site name / url is in the design (and matches the WordPress Block Editor), but is it necessary or helpful?

I ask because including this information does introduce some (seemingly unnecessary) overhead. I think it would be good to simplify here.

image

jdghinson:
@kjohnson it's relevant because they get to change the url if they missed something...same applies to the form title

kjohnson:
> @kjohnson it's relevant because they get to change the url if they missed something...same applies to the form title

Sure, the permalink and form title are important, but I don't understand why we need the Site Name and the Site URL.

Also, why don't we re-use the existing Summary settings? Do we need to maintain another version of those same settings for the pre-publish confirmation?

jdghinson:
> > @kjohnson it's relevant because they get to change the url if they missed something...same applies to the form title
> 
> Sure, the permalink and form title are important, but I don't understand why we need the Site Name and the Site URL.
> 
> Also, why don't we re-use the existing Summary settings? Do we need to maintain another version of those same settings for the pre-publish confirmation?

Sure we can take out the site name and site url....plus yeah we can use the existing summary section if it's the same settings

kjohnson (changes requested):
> Sure we can take out the site name and site url....plus yeah we can use the existing summary section if it's the same settings

@alaca let's remove the Site Name / Site URL and replace the Summary section with the existing component. Sorry for creating some additional work here, but I believe the simplification will pay off in terms of future maintenance.

alaca:
@jdghinson can you update the design according to Kyle's suggestions?

src/FormBuilder/resources/js/form-builder/src/components/sidebar/panels/FormPrepublishPanel.tsx
```diff
@@ -0,0 +1,152 @@
...
+                                </strong>
+                            </div>
+
+                            <PanelBody title={__("What's next?", 'give')} initialOpen={true}>
+                                <PanelRow className="givewp-next-gen-prepublish-panel_link">
+                                    <span>
+                                        {__('PAGE ADDRESS', 'give')}
+                                    </span>
+                                    <span>
+                                        <a
```
kjohnson, line 93: This should probably be a button since it is a clickable element, as opposed to using an href for the styles. The Button component has a tertiary style which looks like a link. Also, it has support for an icon component built in.

src/FormBuilder/resources/js/form-builder/src/components/sidebar/panels/FormPrepublishPanel.tsx
```diff
@@ -0,0 +1,152 @@
...
+                                        {__('Close panel', 'give')}
+                                    </Button>
+                                </div>
+                            </div>
+
+                            <div className="givewp-next-gen-prepublish-panel__content_publish">
+                                <strong>
+                                    <a href={permalink}
+                                       target="_blank">{formTitle}
+                                    </a> {__('is now live', 'give')}.
```
kjohnson, line 83: For translations, we need to include a placeholder for reference, otherwise the translayfile has no context of "what" is now live.

Also, we should probably use Interweave to include the a tag mark up in the placeholder.

alaca: Resolved https://github.com/impress-org/givewp/pull/7050/commits/4b14bad6565524f4854f62c0661914796736924c

src/FormBuilder/resources/js/form-builder/src/components/sidebar/panels/FormPrepublishPanel.tsx
alaca: Resolved https://github.com/impress-org/givewp/pull/7050/commits/ac54f4604de4b4fd2291dcaa9255d34f2ab70cd0

kjohnson (approved):

kjohnson:
Build: https://github.com/impress-org/givewp/actions/runs/6737321077

PR #7052, opened 2023-10-19
Feature: save title as pageslug on publish

JoshuaHungDinh:
Resolves Asana task: https://app.asana.com/0/1205428662243164/1205735215345414/f

## Description

This PR saves the form title as the page slug when a form first publishes. After a form has been published the page slug can still be edited, however it is not respective of the form title on any subsequent change.

If the form title is not updated on the first publish the page slug will continue to to work as it previously did, generating the slug trailing with a new number.

I had looked into a central way of dispatching this in the onSave function call - however had inconsistent state changes and the page slug was being overwritten on refresh.  Due to this I've opted to include conditional dispatch call's for the page slug on both form title controls.

Re: https://lw.slack.com/archives/C05QFUQ805V/p1697494770284989?thread_ts=1694639392.259359&cid=C05QFUQ805V

## Visuals

https://github.com/impress-org/givewp/assets/75056371/e5eb541a-b730-4ac5-bbc5-4bd127d91163

## Testing Instructions

- Create a form
- Update the form title.
- Publish the form.
- View new page slug.
- Visit page - verify pageslug works.

## Pre-review Checklist

---
  - https://app.asana.com/0/0/1205799200925606

### DIFF
--- a/src/FormBuilder/resources/js/form-builder/src/containers/HeaderContainer.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/containers/HeaderContainer.tsx
@@ -150,9 +150,7 @@ const HeaderContainer = ({SecondarySidebarButtons = null, showSidebar, toggleSho
                         className={'givewp-form-title'}
                         value={formTitle}
                         onChange={(formTitle) => {
-                            !isPublished &&
-                                formTitle !== 'Donation Form' &&
-                                dispatch(setFormSettings({pageSlug: cleanForSlug(formTitle)}));
+                            !isPublished && dispatch(setFormSettings({pageSlug: cleanForSlug(formTitle)}));
                             dispatch(setFormSettings({formTitle}));
                         }}
                     />
--- a/src/FormBuilder/resources/js/form-builder/src/containers/HeaderContainer.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/containers/HeaderContainer.tsx
@@ -38,6 +38,9 @@ const Logo = () => (
     </div>
 );
 
+/**
+ * @unreleased dispatch page slug from form title on initial publish.
+ */
 const HeaderContainer = ({SecondarySidebarButtons = null, showSidebar, toggleShowSidebar}) => {
     const {blocks, settings: formSettings, isDirty, transfer} = useFormState();
 
--- a/src/FormBuilder/resources/js/form-builder/src/containers/HeaderContainer.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/containers/HeaderContainer.tsx
@@ -76,7 +76,7 @@ const HeaderContainer = ({SecondarySidebarButtons = null, showSidebar, toggleSho
                 dispatch(
                     setFormSettings({
                         formTitle,
-                        pageSlug: pageSlug,
+                        pageSlug,
                     })
                 );
                 dispatch(setIsDirty(false));
--- a/src/FormBuilder/resources/js/form-builder/src/containers/HeaderContainer.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/containers/HeaderContainer.tsx
@@ -147,7 +147,9 @@ const HeaderContainer = ({SecondarySidebarButtons = null, showSidebar, toggleSho
                         className={'givewp-form-title'}
                         value={formTitle}
                         onChange={(formTitle) => {
-                            !isPublished && dispatch(setFormSettings({pageSlug: cleanForSlug(formTitle)}));
+                            !isPublished &&
+                                formTitle !== 'Donation Form' &&
+                                dispatch(setFormSettings({pageSlug: cleanForSlug(formTitle)}));
                             dispatch(setFormSettings({formTitle}));
                         }}
                     />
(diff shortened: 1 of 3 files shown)


src/FormBuilder/resources/js/form-builder/src/containers/HeaderContainer.tsx
```diff
@@ -72,7 +73,12 @@ const HeaderContainer = ({
                 setErrorMessage(error.message);
             })
             .then(({formTitle, pageSlug}: FormSettings) => {
-                dispatch(setFormSettings({formTitle, pageSlug}));
+                dispatch(
+                    setFormSettings({
+                        formTitle,
+                        pageSlug: pageSlug,
```
jonwaldstein, line 79:
```suggestion
                        pageSlug,
```

src/FormBuilder/resources/js/form-builder/src/settings/form-summary/page-slug.js
```diff
@@ -11,71 +11,69 @@ const {
...
 /**
+ * @unreleased pass page slug from parent component to support form title as initial slug.
  * @since 3.0.0
  * @see https://github.com/WordPress/gutenberg/blob/a8c5605f5dd077a601aefce6f58409f54d7d4447/packages/editor/src/components/post-slug/index.js
  */
-const PageSlugControl = () => {
-    const {
-        settings: {pageSlug},
-    } = useFormState();
+const PageSlugControl = ({pageSlug}) => {
```
jonwaldstein, line 18: I noticed this little quirk where if you update the title and click publish - the url below does not reflect the actual slug until you refresh the page or switch modes.  It makes me think the slug is being memoized somewhere (probably in this file). I think we should clean this ux up.
Screenshot 2023-10-23 at 5 36 20 PM

jonwaldstein (changes requested):

src/FormBuilder/resources/js/form-builder/src/containers/HeaderContainer.tsx
```diff
@@ -140,7 +149,12 @@ const HeaderContainer = ({
                     <TextControl
                         className={'givewp-form-title'}
                         value={formTitle}
-                        onChange={(formTitle) => dispatch(setFormSettings({formTitle}))}
+                        onChange={(formTitle) => {
+                            !isPublished &&
+                                formTitle !== 'Donation Form' &&
```
ravinderk, line 154: @JoshuaHungDinh, Are we sure about this string comparison? Will it remain the same in different languages?

src/FormBuilder/resources/js/form-builder/src/settings/form-summary/page-slug.js
```diff
@@ -11,71 +11,69 @@ const {
...
+                renderContent={({onClose}) => (
+                    <div style={{minWidth: 'calc(var(--givewp-sidebar-width) - 48px)'}}>
+                        <div style={{display: 'flex', justifyContent: 'space-between', alignItems: 'center'}}>
+                            <strong style={{fontSize: '14px'}}>{'URL'}</strong>
+                            <Button onClick={onClose}>
+                                <Icon icon={close} size={14}></Icon>
+                            </Button>
+                        </div>
+                        <TextControl
+                            label={'Permalink'}
```
ravinderk, line 56: @JoshuaHungDinh I noticed that a few strings in this component are not translated. Is it intentional?

JoshuaHungDinh: I don't believe so. Resolved: 12f766a92ba4fbacd9645c3ce63d7946a722bfb1

src/FormBuilder/resources/js/form-builder/src/containers/HeaderContainer.tsx
JoshuaHungDinh: Meant to revert that. Thanks! Resolved: d023df110da7d361d8c912dced4a6999e18b98f1

src/FormBuilder/resources/js/form-builder/src/containers/HeaderContainer.tsx
JoshuaHungDinh: Yea I think this can be done better. We don't need to match it in the onChange, if the formTitle is changing then it's redundant.

Because we only need to pass the formTitle to the pageSlugControl when an un-published form has an updated title. I've updated the conditional by checking if `cleanForSlug(formTitle)` matches `pageSlug`. The only time the form title could match the pageSlug is if the formTitle was updated.   Resolved: 4fb4931f139ee44506439d9fcc1d13fc480bbe5a.  (Missed the ```!``` on the isPublished check 3710522661217852fe051c92921d3662b7e8d153)

src/FormBuilder/resources/js/form-builder/src/settings/form-summary/page-slug.js
JoshuaHungDinh: How would you feel about listening for the pageSlug updates? As the props updating however the component is not re-rendering. : 31dd6b7703011c474620edc86ef2c362150061aa bc9dd440a557a2d18d05504e7084beb4ecd443ba

jonwaldstein: yeah that works 👍

jonwaldstein (approved):
Looks good, nice job 👍

rickalday (approved):
Passed manual QA tests.

PR #7059, opened 2023-10-20
Feature: Preview API

alaca:
## Description

This PR introduces a new Preview API that is used to pass the state changes in design mode to the form inside the iframe. Preview API is a simple pub/sub system that hooks up in the form state which makes all changes in design mode show up instantly on the screen without reloading the form.

## Affects

Form Builder
Donation form

## Visuals
Screen Capture_select-area_20231023123856

## Testing Instructions

1. Create a new form
2. Enter design mode
3. Try every  form customization option in the right-hand side panel 
4. Verify that form settings are saved correctly by checking the form on the frontend

## Pre-review Checklist

-   [ ] Includes unit tests

Plugin zip for QA: [https://github.com/impress-org/givewp/suites/17790212597/artifacts/1019602715](https://github.com/impress-org/givewp/suites/17790212597/artifacts/1019602715)

---
  - https://app.asana.com/0/0/1205783384593885
  - https://app.asana.com/0/0/1205269709893095


JasonTheAdams:
Just looking at the gif, @alaca, this looks amazing! Can we remove the debouncing on the text fields so the typing is instantaneous? 😃

alaca:
@JasonTheAdams sure :smiley:

src/DonationForms/resources/app/DonationFormApp.tsx
```diff
@@ -12,70 +12,118 @@ import DonationFormErrorBoundary from '@givewp/forms/app/errors/boundaries/Donat
...
+                        goal: {
+                            ...prevState.goal,
+                            ...data
+                        }
+                    }
+                })
+            })
+
+            subscribe('preview:colors', (data: { [s: string]: string }) => {
+                const [key, value] = Object.entries<string>(data).flat();
```
jonwaldstein, line 80: Let's take this in a slightly different direction. Reason being, right now regardless of preview mode, we're duplicating the form object, loading the subscribe function, and adding a useEffect.  As you can probably guess, I want to be extra careful about preview mode leaking into live mode of the form.

 Instead of adding `preview` as a prop let's actually create an `<AppPreview />` component who's sole purpose is to load the subscribe function and use effect.  It then simply returns `<App form={previewForm} />`.  The form prop would be an optional new prop on the App component so it can come from the server or be overwritten by the prop.
 
 So down below we would do `render(preview ? <AppPreview /> : <App />, root);`

src/DonationForms/resources/app/form/MultiStepForm/index.tsx
```diff
@@ -55,23 +55,23 @@ const convertSectionsToSteps = (sections: Section[], hasFirstStep: boolean) => {
 /**
  * @since 3.0.0
  */
-export default function MultiStepForm({sections, showHeader}: {sections: Section[]; showHeader?: boolean}) {
-    const steps = convertSectionsToSteps(sections, showHeader);
+export default function MultiStepForm({form}) {
```
jonwaldstein, line 58: Let's make sure to add the correct `Form` type for this prop.

src/DonationForms/resources/app/utilities/usePubSub.ts
```diff
@@ -0,0 +1,66 @@
+import {RefObject, useEffect} from 'react';
+
+/**
+ * @unreleased
+ */
+export default function usePubSub() {
```
jonwaldstein, line 6: can you please add some comments/descriptions to this file explaining what each part is doing for our future selves 😄

JasonTheAdams, line 6: Let's call this `useDonationFormPubSub`, for further clarity.

src/DonationForms/resources/app/utilities/usePubSub.ts
```diff
@@ -0,0 +1,66 @@
...
+            }
+        } else {
+            postMessage(message);
+        }
+    }
+
+    return {
+        publish,
+        subscribe
+    }
```
JasonTheAdams, line 65: Let's add an `unsubscribe` method as well so we can handle cleanup for events.

src/DonationForms/resources/app/utilities/usePubSub.ts
JasonTheAdams, line 1: I'm wondering about adding specific functions such as `subscribeToColorsPreview`/`publishColorsPreview` and such, or add an export of constants so static analysis can catch mis-typed keys. A bunch of helper functions is nice, or simply a `PREVIEW_EVENT.COLORS` would be helpful.

src/FormBuilder/resources/js/form-builder/src/settings/donation-goal/index.tsx
```diff
@@ -64,7 +68,15 @@ const DonationGoalSettings = () => {
...
-                            onChange={(goalType) => dispatch(setFormSettings({goalType: goalType}))}
+                            onChange={(goalType: string) => {
+                                dispatch(setFormSettings({goalType}))
+                                publish('preview:goal', {
+                                    type: goalType,
+                                    label: goalType,
+                                    typeIsCount: 'amount' !== goalType,
+                                    typeIsMoney: 'amount' === goalType
+                                }, iframeRef);
+                            }}
```
jonwaldstein, line 79: I'm looking at this thinking there is a lot of contexts to understand about adding a form setting.  We have the form setting json and now window data object to remember.  I would like to explore abstracting the publishing of preview data.  At the very least, we should add some extra functions like `publishPreviewGoal(data: Form['goal'])` and take advantage of typescript for the accepted data.

I'm open to ideas. Part of me thinks this could be a job for the reducer to handle since it's the closest modifier to the form's state but maybe that's a separation of concerns issue.

src/FormBuilder/resources/js/form-builder/src/settings/donation-goal/index.tsx
```diff
@@ -73,7 +85,10 @@ const DonationGoalSettings = () => {
                             label={__('Goal Amount', 'give')}
                             min={0}
                             value={goalAmount}
-                            onChange={debounce((goalAmount) => dispatch(setFormSettings({goalAmount})), 100)}
+                            onChange={debounce((goalAmount: string) => {
```
jonwaldstein, line 88: We can probably get rid of debounce now.

jonwaldstein (changes requested):
@alaca this is amazing! Just had some thoughts and suggestions.  We'll also need to circle back and make sure all settings are accountable.  One I noticed we still need to figure out is the custom css setting.

JasonTheAdams (changes requested):

src/DonationForms/resources/app/utilities/usePubSub.ts
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    const subscribe = (event: string, callback: Function) => {
+        if (!events[event]) {
+            events[event] = [];
+        }
+
+        events[event].push(callback);
+    }
```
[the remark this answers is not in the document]
alaca: Good point. I was thinking about security implications, and yes, using postMessage has security concerns. I don't think having a list of allowed events will help. The attacker could exploit this by loading the form inside an iframe and then sending messages with a malicious payload. The events that we are listening to are "visible" and the attacker won't have any problems seeing the event names. Maybe we could prefix events with something that is coming from the server, like nonce?  

What we can do and that actually makes sense is to check the message origin. But even if we leave it like this, what attacker can do, update the state with something that doesn't belong there? Then it is easier for the attacker to do that from the browser's console. I mean, even if the state is changed, that state cannot be saved to the database on the frontend.

src/DonationForms/resources/app/DonationFormApp.tsx
alaca: Make sense

src/FormBuilder/resources/js/form-builder/src/settings/donation-goal/index.tsx
alaca: I was thinking of adding a helper function, and we should probably do that. I think using reducers will add unnecessary overhead, with little or no benefit.

src/DonationForms/resources/app/utilities/usePubSub.ts
alaca: It is not a problem to add, but I don't think it's necessary since we are using this only in the context of the donation form preview. Even if we add that,  how it will work? I mean, right now we don't have a global state of events, so this will work only in the context of the component where we are subscribed to an event.

JasonTheAdams: It's simply for handling re-rendering in React inside of `useEffect`. We're relying on `preview`, but not including it as a dependency, and if the component re-renders we'd be accumulating subscriptions. We want to return a function in the `useEffect` which handles the downward trigger.

src/DonationForms/resources/app/utilities/usePubSub.ts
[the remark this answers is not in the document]
JasonTheAdams: I'm curious what kind of security vulnerabilities we're anticipating, here? The color properties, for example, are pretty darn safe, due to how it's being applied. Most (or all?) of the values are primitives, so we could potentially just walk the data and nullify anything that's not a primitive. There may be a fairly simple solution, here, as I don't think we're accepting callbacks or any other such thing.

src/DonationForms/resources/app/form/MultiStepForm/index.tsx
alaca: @jonwaldstein we don't have `Form` type, right?

src/DonationForms/resources/app/utilities/usePubSub.ts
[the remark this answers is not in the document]
jonwaldstein: @alaca good points, my main concerns would be XSS and preventing destructive mutations of the data the form relies on (window data).

src/DonationForms/resources/app/form/MultiStepForm/index.tsx
jonwaldstein: We do, `import type {Form} from '@givewp/forms/types';`

src/FormBuilder/resources/js/form-builder/src/settings/donation-goal/index.tsx
jonwaldstein: helper functions would probably suffice.  the reducer I was referring to was the one we are already using for setting the values - but again that might be better seperated.

src/DonationForms/resources/app/form/Header.tsx
```diff
@@ -1,30 +1,28 @@
...
  * @since 3.0.0
  */
-export default function Header() {
+export default function Header({form}) {
+
+    const formatGoalAmount = (amount: number) => {
+        return amountFormatter(form.currency, {
+            maximumFractionDigits: 0,
+        }).format(amount);
+    };
```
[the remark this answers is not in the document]
alaca: Resolved https://github.com/impress-org/givewp/pull/7059/commits/e7b4df07f1abd2895ab2ad83e4f31abcfd55faa1

src/DonationForms/resources/app/form/MultiStepForm/index.tsx
alaca: Resolved https://github.com/impress-org/givewp/pull/7059/commits/a66f19bdf4813abbdaf78d68d128f02020bdfcb1

src/DonationForms/resources/app/utilities/usePubSub.ts
alaca: The initial version was more like a general purpose pub/sub system. Now is all about the donation form in preview mode. I added a bunch of new helper functions for publishing and subscribing to events that use the correct types.

src/DonationForms/resources/app/utilities/usePubSub.ts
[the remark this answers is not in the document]
JasonTheAdams: We'll want to clean up the incoming data. It _should_ only be primitives. So looping through and nullifying functions should help. So long as we're not using `__dangerouslySetInnerHTML` anywhere (which we should be using Interweave, instead), we should be safe with all incoming strings.

src/DonationForms/resources/app/DonationFormApp.tsx
JasonTheAdams: @alaca @jonwaldstein I actually _do_ want a concept of preview within the app. The reason being I want to do more with this in subsequent PRs. For example, in preview mode I don't want it to attempt actual donations. Clicking "donate" should just display the confirmation receipt with the entered data, but not actually attempt to make a donation.

alaca: @JasonTheAdams we can add that back. I didn't delete it. What do you think about having that information in form state? Then we don't have to pass it everywhere as a prop.

JasonTheAdams: Oh yeah, for sure. We want various parts of the form to be able to act a bit differently during preview. Preview mode won't change back and forth, so I'm not worried about it causing unnecessary re-renders.

src/DonationForms/resources/app/utilities/usePubSub.ts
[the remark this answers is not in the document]
alaca: @JasonTheAdams I added the filter here https://github.com/impress-org/givewp/pull/7059/commits/5398b0644ba67054aca4b2be49b071975bc55408

If the passed value is not of primitive type, it will be removed.

src/DonationForms/resources/app/DonationFormApp.tsx
alaca: Resolved https://github.com/impress-org/givewp/pull/7059/commits/238ade6ca2ec1f7e5c082e0d7a8c2a31e61694e3

src/DonationForms/resources/app/utilities/usePubSub.ts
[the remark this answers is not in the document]
JasonTheAdams: @alaca I would think we'd want to do the filtering on the receive end of the message. We can trust our own code that's publishing the data; we can't trust _other_ JS sending messages to the iframe.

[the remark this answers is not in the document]
alaca: Right :man_facepalming:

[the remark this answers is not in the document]
alaca: I think this is it https://github.com/impress-org/givewp/pull/7059/commits/2cd858c28a96b05743dd1482f77d5e1b8f56002a 
:sweat_smile:

src/DonationForms/resources/app/DonationFormApp.tsx
jonwaldstein: I'm fine with a top level preview flag in the form context like @alaca suggested.  I just want to be careful about preview functionality being loaded inside the donation form when it's not in preview mode.

src/FormBuilder/resources/js/form-builder/src/settings/donation-goal/index.tsx
```diff
@@ -82,7 +88,10 @@ const DonationGoalSettings = () => {
                                 label={__('Goal Amount', 'give')}
                                 min={0}
                                 value={goalAmount}
-                                onChange={(goalAmount) => dispatch(setFormSettings({goalAmount}))}
+                                onChange={debounce((goalAmount: number) => {
+                                    dispatch(setFormSettings({goalAmount}))
+                                    publishGoal({targetAmount: goalAmount});
```
jonwaldstein, line 93: We need to add this to the `CurrencyControl` above as well.

jonwaldstein (approved):
I'm so happy about this 😭, great job @alaca!

src/DonationForms/resources/app/DonationFormApp.tsx
```diff
@@ -38,44 +42,106 @@ const initialState = {
...
+            }
+
+            if (data['secondaryColor']) {
+                root.style.setProperty('--givewp-secondary-color', data['secondaryColor']);
+            }
+        })
+
+        subscribeToCss(({customCss}) => {
+            style.innerHTML = customCss;
+        })
```
JasonTheAdams, line 130: @alaca Is there a more specific style node property than `innerHTML`? Is it possible for someone to put  a close tag in the `customCss` variable and then add arbitrary HTML?

alaca: @JasonTheAdams  we can probably use `innerText` instead of `innerHTML`.

alaca: @JasonTheAdams I refactored this to use `innerText`. That should strip all HTML tags, scripts, etc.

alaca: @JasonTheAdams I wasn't happy with the way how we set CSS so I refactored it a bit https://github.com/impress-org/givewp/pull/7059/commits/dd119797739da0beca0137a4d1d82d38a18bfd03

This should be pretty secure :smiley:

JasonTheAdams (commented):
@jonwaldstein @alaca What I believe is my last question: Why are we prop-drilling the form data instead of using a context?

alaca:
@JasonTheAdams that's because the current state is divided and we use two state providers `DonationFormSettingsProvider` and `DonationFormStateProvider`, but we are not passing all the form data that comes from `getWindowData` to providers. The `DonationFormStateProvider` is a bit misleading because it doesn't include the entire form state, just gateway stuff, validation schema, etc., and `DonationFormSettingsProvider` does not include `goal` data and `stats` data required by the `Header` component. The `Header` component was using `getWindowData` before.

JasonTheAdams:
@alaca Hahah! Got it. I'm just a bit concerned that it's not easily available, especially for when we get into form design settings, that could apply anywhere in the node tree. Seems like we need to make sure, one way or the other, it's globally available without having to prop-drill.

Seem sensible to you, @jonwaldstein?

jonwaldstein:
@JasonTheAdams i'm open to adding the form data object into context, but as @alaca pointed out - it's not that straight forward.  This is also specific to the preview api as window data still works as expected for live rendering.

My preference would be to keep the prop drilling for now (since it's really just for the header) and update the context structure (as needed) when we actually introduce a form design api that requires it.  At that time, I would explore breaking out that form data object into pieces that make more sense for the front-end application which could end up being separate context providers.

JasonTheAdams (approved):
I'm SO EXCITED for this!! 🤩

rickalday (approved):
Passed manual tests.

PR #7063, opened 2023-10-22
New: Make donation list table extendable

pulsovi:
Resolves: [GIVE-308]

## Description

I'm developing a plugin on GiveWP. 

My plugin provides a specific GiveWP donation model, like the give-recurring plugin does.
To present in the back office the list of donations made via my plugin, I made an extension of the native donations ListTable.

This `Give\Donations\Endpoints\ListDonations` class is the only one that I have not managed to extend.

This PR proposes minor changes which make this class extensible.

## Affects

Donation ListTable in back-office

## Testing Instructions

In particular, I would like to override the `getWhereConditions` method to add a filter that only keeps my type of donations. 

In the override function, I use `parent::getWhereConditions` to avoid fully rewrite the method. The `protected` keyword prevent me to make this.

Likewise, I would like to change the class pointed to by `$this->listTable` without having to completely rewrite the `handleRequest` function.

The flow of this function does not allow this to be done by wrapping it in an overload. In any case, not using `parent::handleRequest`, you would have to rewrite the method completely.
I therefore propose to extract the offending line in another method that can be overloaded alone.

## Pre-review Checklist

-   [x] Includes unit tests

[GIVE-308]: https://stellarwp.atlassian.net/browse/GIVE-308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Donations/Endpoints/ListDonations.php
```diff
@@ -208,14 +219,15 @@ public function getTotalDonationsCount(): int
...
+     * @unreleased Make the class extendable
      * @since 2.24.0 Remove joins as it uses ModelQueryBuilder and change clauses to use attach_meta
      * @since      2.21.0
      *
      * @param QueryBuilder $query
      *
      * @return array{0: QueryBuilder, 1: array<DonationMetaKeys>}
      */
-    private function getWhereConditions(QueryBuilder $query): array
+    protected function getWhereConditions(QueryBuilder $query): array
```
kjohnson, line 230: I definitely agree that we should make the visibility of this method `protected`.

src/Donations/Endpoints/ListDonations.php
```diff
@@ -151,6 +152,16 @@ public function handleRequest(WP_REST_Request $request): WP_REST_Response
...
     }
 
+    /**
+     * Initialize which list table to use.
+     *
+     * @unreleased
+     */
+    protected function init ()
+    {
+        $this->listTable = give(DonationsListTable::class);
```
kjohnson, line 162: Instead of extracting this to a specific use `init()` method, I would prefer to use dependency injection from the container via the constructor. Currently, we are resolving the class from the container using the `give()` helper method, but moving this dependency to the `__construct()` as a parameter will still use the container, but can be  overloaded by an extending class.

```php
// givewp/src/Donations/Endpoints/ListDonations.php
class ListDonations extends Endpoint {
    public function __construct(DonationsListTable $listTable) {
        $this->listTable = $listTable;
    }
}
```


```php
// my-plugin/my-list-donations.php
class MyListDonations extends MyListDonations {
    public function __construct(MyDonationsListTable $myListTable) {
        $this->listTable = $myListTable;
    }
}
```

kjohnson (changes requested):
@pulsovi thanks for reaching out with the Pull Request. I made a suggestion for how I think we should integrate this change into the Give codebase that allows for the extensibility that you are asking for.

Also, could you share some more detail on what you are building? I'd love to hear about how you are extending GiveWP and how we might make that a better experience.

src/Donations/Endpoints/ListDonations.php
pulsovi: Good morning,

Thank you for your quick response and your interest in my request.

I understood your remarks well. I had thought about this possibility of implementation.

There is a performance issue here. If another API route called, this ListTable will be useless instantiated. This is why I didn't want to instantiate the ListTable during construction.

The instantiation should still be triggered by `handleRequest`, in my opinion. What do you think of this proposal:

```php
// givewp/src/Donations/Endpoints/ListDonations.php
class ListDonations extends Endpoint {
    /**
     * The ListTable to which provide the data
     * @var string
     */
    protected const LIST_TABLE = DonationsListTable::class;

    // ...

    public function handleRequest(WP_REST_Request $request): WP_REST_Response {
        $this->listTable = Give(static::LIST_TABLE);
        $this->request = $request;
        // ...
    }
}
```

```php
// my-plugin/my-list-donations.php
class MyListDonations extends ListDonations {
    protected const LIST_TABLE = MyDonationsListTable::class;
}
```

pulsovi:
Good morning,

What about this topic? Do you agree with my proposal?

pulsovi:
@JoshuaHungDinh sorry for bothering you. What can I do to move this pull request forward?

kjohnson (approved):
I'll package this up and send it through the QA process.

kjohnson:
Build for QA, see https://github.com/impress-org/givewp/actions/runs/7833094361.

rickalday (approved):
Passed automated QA tests.

PR #7073, opened 2023-10-26
Feature: update Giving Tuesday banner

JoshuaHungDinh:
Resolves https://app.asana.com/0/1205251546311816/1205651762721240

## Description

This PR includes the Giving Tuesday banner to be displayed for the sale dates 11/20-11/29. During the sale dates the banner should only display on the Give admin screens and all banners remain closed, when any one of the banners on the admin screen is dismissed. 

The banner has different messages and links for users depending on the type of active pricing plan they currently hold. I've included new methods to check the active licenses of the plugins the user has and used that information to determine their pricing tier eligibility. 

Users with Pro plans will not be shown a banner.

## Links:
- Free users: https://go.givewp.com/bf23
- All others (except pro/agency): https://go.givewp.com/bfup23

## Messages:
- Free account users: Upgrade to a Pricing Plan for Recurring Donations, Fee Recovery, and more
- Basic plan users: Upgrade to a Plus Plan to get all must-have add-ons.
- Plus plan users: Upgrade to Pro and get Peer-to-Peer fundraising.

## Affects
All Give admin screens

## Visuals
Donations list table
Donors list table
Donation Forms list table
Settings page
Reports page
Tools page
Add form page
Add-ons page

## Testing Instructions 

To test :
/src/Promotions/InPluginUpsells/SaleBanners.php 
( line 84 `'startDate' => '2023-11-20 00:00'` can be adjusted to display the banner )

- Verify the banner ONLY displays on our Give admin pages.
- Close one of the banners on any page & verify all the banners remain closed on all pages.
- If you have access to an account with an active license plan - it would be helpful to test that the correct message displays based on the account pricing plan.
- Visit the link to verify it displays the correct page.

## Pre-review Checklist

---
  - https://app.asana.com/0/0/1205849573153024


src/Promotions/InPluginUpsells/SaleBanners.php
```diff
@@ -161,4 +207,78 @@ public static function isShowing(): bool
...
+        if ($licensedPluginSlugs === $basic) {
+            $plan = 'Basic';
+        } elseif ($licensedPluginSlugs === $plus) {
+            $plan = 'Plus';
+        } elseif ($licensedPluginSlugs === $pro) {
+            $plan = 'Pro';
+        }
+
+        return $plan;
+    }
```
pauloiankoski, line 268: Hey @JoshuaHungDinh! What if a user has the Plus license but has also purchased an individual license for another Pro plugin? In such a case, I believe that when comparing the licenses, if they are **equal** to one of the plans, we will consider the user as a Free user.

If my assumption is correct, we would need to modify the comparison approach. Instead of simply checking for equality, we would need to verify if the user has obtained all the licenses associated with that plan. This can be achieved by using the array diff method:

CleanShot 2023-11-01 at 17 00 16

Btw, sorting is not necessary when using `array_diff_assoc`.

pauloiankoski (changes requested):
As a general comment, is there any reason to keep the old banners in the codebase?

JoshuaHungDinh:
> As a general comment, is there any reason to keep the old banners in the codebase?

I don't believe so.. Perhaps we should remove them.

src/Promotions/InPluginUpsells/SaleBanners.php
JoshuaHungDinh: Resolved: 0ab19add5d816fedf4b5415ae9cc2cb2dcb63e41

pauloiankoski (approved):
LGTM! 👍🏻

rickalday (approved):
Passed manual QA tests

PR #7077, opened 2023-10-27
Hot Fix: properly record the Stripe renewals on the Give side

### LINKED ISSUE #7076: Stripe renewals are not recording
In the latest version of GiveWP, subscription renewals from Stripe are not recording.

The workaround, for now, is to manually sync each subscription that has missing renewal records.

https://givewp.canny.io/admin/board/bug-reports/p/stripe-renewals-are-not-recording

glaubersilva:
Resolves #7076

## Description

The webhook events used by the Stripe Element were using the exit in the wrong place and it was preventing the events from the legacy codebase from running.

This PR fixes this problem.

## Affects

Stripe renewals in donations/subscriptions created with v2 forms.

## Testing Instructions

1. Using the latest version of GiveWP and the Recurring Donations add-on, make a recurring donation on a V2 form
2. Spoof a renewal in Stripe by resetting the billing cycle: https://somup.com/c06tcOBP02
3. The renewals should be recorded on the Give side.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205823108730572

### DIFF
--- a/src/PaymentGateways/Gateways/Stripe/StripePaymentElementGateway/Webhooks/Listeners/ChargeRefunded.php
+++ b/src/PaymentGateways/Gateways/Stripe/StripePaymentElementGateway/Webhooks/Listeners/ChargeRefunded.php
@@ -22,6 +22,7 @@ class ChargeRefunded
      *
      * @see https://stripe.com/docs/api/events/types#event_types-charge.refunded
      *
+     * @unreleased Add exit statement only when the event is successfully processed.
      * @since 3.0.0
      *
      * @return void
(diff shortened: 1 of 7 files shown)


jonwaldstein (approved):
@glaubersilva looks great thanks!

JasonTheAdams (commented):
@glaubersilva @jonwaldstein Am I following the logic here that basically we're no longer exiting if the v3 form system _did not_ process the event, which gives the v2 system the opportunity to process it?

jonwaldstein:
> @glaubersilva @jonwaldstein Am I following the logic here that basically we're no longer exiting if the v3 form system _did not_ process the event, which gives the v2 system the opportunity to process it?

Correct, we found the exit statement needed to be called conditionally - only if Stripe Payment Element processed the event.    The original purpose of the action was for Stripe Payment Element to take over the request when need be (because it lived in the seperate next gen repo).  In this scenario we don't want both webhook logics to run - it should be either the legacy Stripe webhook or new one. 

We weighed many different options together and this was the simplest way to resolve the current issue at this time.  When we have a little more time to dedicate to this, we would like to explore creating some kind of proper proxy up the chain and/or merging the logics so they can truly be handled separately to avoid conflict and confusion.

rickalday (approved):
Passed manual QA tests.

PR #7078, opened 2023-10-30
Fix: pass form id by reference in totals shortcode to avoid PHP warning

JoshuaHungDinh:
Resolves https://app.asana.com/0/1205428662243164/1205836070799020

## Description

This PR addresses the following warning 
`Warning: _give_redirect_form_id(): Argument #2 must be passed by reference, value given in /Users/joshuadinh/Local Sites/givewpdeveloper/app/public/wp-content/plugins/givewp/includes/shortcodes.php on line 634 from this line         array_walk($form_ids, '_give_redirect_form_id');
`

## Testing Instructions

- Add totals shortccode to a page.
- Verify no warning displays.
- Example shortcode : `[give_totals ids="1460" message="Thanks to you, we’ve raised {total} across all our congregations combined! Let's keep it up!" progress_bar="true"]`

## Pre-review Checklist

---
  - https://app.asana.com/0/0/1205845725895553

### DIFF
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -632,7 +632,7 @@ function give_totals_shortcode( $atts ) {
 			$form_ids = array_filter( array_map( 'trim', explode( ',', $atts['ids'] ) ) );
 		}
 
-        array_map(
+        $form_ids = array_map(
             static function ($id) {
                 _give_redirect_form_id($id);
 
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -25,6 +25,7 @@ if ( ! defined( 'ABSPATH' ) ) {
  *
  * Displays a user's donation history.
  *
+ * @unreleased pass form id by reference in give_totals shortcode.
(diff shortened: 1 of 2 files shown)


includes/shortcodes.php
```diff
@@ -631,7 +632,14 @@ function give_totals_shortcode( $atts ) {
 			$form_ids = array_filter( array_map( 'trim', explode( ',', $atts['ids'] ) ) );
 		}
 
-        array_walk($form_ids, '_give_redirect_form_id');
+        array_map(
```
pauloiankoski, line 635: When using `array_map`, you must assign the result to a variable. Otherwise, it will have no effect. In this case, please assign it to the variable `$form_ids`.

pauloiankoski (changes requested):

pauloiankoski (approved):
LGTM 👍🏻  QA time!

rickalday (approved):
Passed manual tests

PR #7097, opened 2023-11-10
Fix: provide fallback for donation mode during model queries

### LINKED ISSUE #7096: On certain conditions, the Donations > Donations page will display a fatal error
As an admin, I want to access my Donations page. However, after GiveWP 2.27.3, it breaks when I access it.

The error says:

An error of type E_ERROR was caused in line 50 of the file …give/vendor/myclabs/php-enum/src/Enum.php.

Error message: Uncaught UnexpectedValueException: Value ‘’ is not part of the enum Give\Donations\ValueObjects\DonationMode in /home/huliyoyeho0047/web/ocent-ibis-zala.instawp.xyz/public_html/wp-content/plugins/give/vendor/myclabs/php-enum/src/Enum.php:50

https://givewp.canny.io/admin/board/bug-reports/p/on-certain-conditions-the-donations-donations-page-will-display-a-fatal-error

jonwaldstein:
Resolves #7096 [GIVE-27]

## Description

This resolves an issue where in some cases the donation mode was not stored in the database and the donation model query was causing an error from expecting a value.  This is most notable on the admin list table.  The simplest way to resolve this is to have a static fallback which was updated in this PR.  

Typically we would create a migration for missing data that is expected, but this is unique in that we would have no way of knowing what the actual donation mode was at the time of creation.  Furthermore, in terms of backwards compatibility - this could be a persistent issue from a third party add-on that intercepts or modifies the donation.  So the safest and simplest option is just to have a static fallback value of "live" as that is always the assumption unless specifically defined as "test".

For clarity, in GiveWP 3.0+ forms, we are confident that the payment mode is always stored.

## Affects

- The donation list table when payment mode meta is missing

## Visuals

N/A

## Testing Instructions

- Create a donation without a payment mode
- View the list table and make sure it shows up properly

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-27]: https://stellarwp.atlassian.net/browse/GIVE-27?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Donations/DataTransferObjects/DonationQueryData.php
+++ b/src/Donations/DataTransferObjects/DonationQueryData.php
@@ -124,6 +124,7 @@ final class DonationQueryData
     /**
      * Convert data from object to Donation
      *
+     * @unreleased add fallback for donation mode
      * @since 2.23.0 remove parentId property
      * @since 2.22.0 add support for company field
      * @since 2.20.0 update for new amount property, fee amount recovered, and exchange rate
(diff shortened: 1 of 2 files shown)


JasonTheAdams:
@jonwaldstein Should we also have the Subscription model fallback on live mode?

jonwaldstein:
> @jonwaldstein Should we also have the Subscription model fallback on live mode?

I think that would be proactively wise. Although, I would open a separate PR for that.  

Luckily the subscription table has a payment_mode column so adjusting the list table query (if need be) should be simpler.

JasonTheAdams (approved):
Looks good!

Issue #7064, opened 2023-10-24
[WP_Background_Process] does not support custom `cron_interval` property


nk-o:
We are using the WP_Background_Process library in our plugins (Visual Portfolio, Ghost Kit, etc...), and when your plugin is installed, we see the error:

```
Fatal error: Uncaught TypeError: Unsupported operand types: string * int in .../wp-content/plugins/give/includes/libraries/wp-background-process.php:425
```

This happens when we use the WP_Background_Process with the custom `cron_interval` property. Example from Ghost Kit - https://github.com/nk-crew/ghostkit/blob/58ed4cfe21afb00786603d26983bd07fcff3b2d4/src/classes/class-breakpoints-background.php#L34-L39

To solve this issue, you have to fix your code here - https://github.com/impress-org/givewp/blob/33cffa1cc6a2d64f820411eac1d0b65058d4ed85/includes/libraries/wp-background-process.php#L419-L421

Or use the updated version of the library. Example from the updated library with the fix - https://github.com/deliciousbrains/wp-background-processing/blob/ebbed762c42e70e76573ffbd36e591378ba5f2ee/classes/wp-background-process.php#L648-L650

JoshuaHungDinh:
Hello @nk-o,

Thank you for taking the time to write a detailed report and for bringing it to our attention. I was able to reproduce this error and will discuss it with the team for prioritization.

EuropeanWilderness:
Hi,

we are still running into the error on our website. Once we activate the Give plugin we get the same critical error. Could there be an additional issue. we can give you access to our site to test this yourself.

PR #6856, opened 2023-07-26
Feature: Add WordPressLibraries to manage WP_Background_Process

JoshuaHungDinh:
## Description

This pull request introduces organizational changes and a new directory to handle the deprecated ```WP_Background_Process``` and ```WP_Async_Request``` classes more efficiently. Additionally, this addresses potential conflicts with other plugins using the same classes, specifically a fatal error experienced with the 3rd party plugin "RankMath" and our Give-Zapier plugin.

The primary motivation behind these changes is to eliminate conflicts arising from the loading of the deprecated classes without considering the plugin's scope. By creating a new namespace directory to manager the WP Background Process we can make it easier to manage any future conflicts with third party plugins.

The new namespaced classes can be found with the Framework directory, under WordPressLibraries.

Reference: https://github.com/impress-org/give-zapier/pull/63

## Testing Instructions

Please see referenced PR.

## Pre-review Checklist

-   [x] Acceptance criteria satisfied and marked in related issue
-   [ ] Relevant `@unreleased` tags included in DocBlocks
-   [ ] Includes unit tests
-   [ ] Reviewed by the designer (if follows a design)
-   [x] [Self Review](https://give.gitbook.io/development-manual/devops/github/code-reviews#self-review) of code and UX completed

PR #7099, opened 2023-11-14
Fix: remove and replace deprecated WP_Background_Process & WP_Async_Request

JoshuaHungDinh:
Resolves https://github.com/impress-org/givewp/issues/7064

## Description

This PR removes the deprecated version files of `WP_Background_Process` & `WP_Async_Request`. 
Also replacing `WP_Async_Request` with `WPAsyncRequest` in our `give-async-process`.

In [Feature: Add WordPressLibraries to manage WP_Background_Process #6856](https://github.com/impress-org/givewp/pull/6856) we had forked these files into namespaced versions and imported them into Give_Zapier to avoid fatal error conflicts occurring with 3rd party plugins and our add-on. At the time the decision was to keep the original files within core intact. Because the files are still being loaded we were aware of the possibility that conflicting errors may continue to occur. Recently a developer has reached out to help address the conflict errors within GhostKit, VisualPortfolio and Give due to the mismatched versions of the files being loaded.

## Visuals
Screen Shot 2023-11-15 at 5 41 39 AM

## Testing Instructions
- Install GhostKit &/or VisualPortfolio
- Verify that no fatal errors occur.
## Pre-review Checklist


JasonTheAdams (changes requested):
Nice cleanup work, @JoshuaHungDinh! Do me a small favor and add a description to the `WPAsyncRequest` and `WPBackgroundProcess` classes that tells other devs to _avoid_ use of these classes. They are here only for legacy code purposes, and folks should prefer Action Scheduler, which is also available.

JoshuaHungDinh:
@JasonTheAdams How does this look? 6d859a44eefc1e8f66a61d2f022e954465242b05

JasonTheAdams (approved):
Excellent! Thanks @JoshuaHungDinh!

PR #7107, opened 2023-11-20
Hotfix: prevent lots of extra slashes in the formBuilderSettings meta key which slow down the form builder loading

glaubersilva:
Resolves [GIVE-92]

## Description

This PR prevents extra slashes from being added to the database on each form builder update and also adds a migration to remove extra slashes in forms that already have it saved on DB.

The problem was happening because the `addSlashesRecursive()` method that was added in [this other PR](https://github.com/impress-org/givewp/pull/6938) to escape double quotes in `emailTemplateOptions` (to fix another bug that was happening in migrated forms) was adding extra slashes on DB on each form update.

There is a similar WP core function called `wp_slash()` that says this in the [docs](https://developer.wordpress.org/reference/functions/wp_slash/):

_This should not be used to escape data going directly into an SQL query._

However, before this PR, we were using the data returned by the `addSlashesRecursive()` method in the Donation Forms repository to update the form settings in the database, which was causing problems with the extra slashes.

So, a possible solution would be to edit the `src/DonationForms/Properties/FormSettings.php` file replacing this line:

```
'emailTemplateOptions' => array_map([$this, 'addSlashesRecursive'], $this->emailTemplateOptions),
```

With this one:

```
'emailTemplateOptions' => wp_slash(wp_unslash($this->emailTemplateOptions)),
```

But we can fully remove this line because it's unnecessary and redundant - this was the approach followed on this PR. 

Why?

The `addSlashesRecursive()` method was originally added to solve a problem that was being generated in the `SanitizeDonationFormPreviewRequest` class related to the `wp_kses_post()` function.

However, this problem related to the use of the `wp_kses_post()` function was fixed a month ago while we were porting the PDF Receipts Addons settings to the form builder. 

Here is the reference commit: https://github.com/impress-org/givewp/commit/dd0230fcc9e659957aa52218738a37f7ccd65c8c#diff-17fad574bc74bd58506d3ab524c91e360df7e2b73648367b983040ef6f740f73

The commit above made the `addSlashesRecursive()` method unnecessary and solved (in the source) the problem that this method was solving, but it kept on the codebase generating extra slashes in the database on each form update.

## Affects

The `formBuilderSettings` meta key.

## Visuals

**Before:** 

image

**After:**

image

## Testing Instructions

1. Create a v2 form
2. Migrate the form to v3 (transfer the data)
3. Load design mode, change something, and save the form (refresh the page and do it 2 or 3 times)
4. Check the `formBuilderSettings` meta key on DB and make sure there are no extra slashes there

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-92]: https://stellarwp.atlassian.net/browse/GIVE-92?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/DonationForms/Migrations/CleanMultipleSlashesOnDB.php
```diff
@@ -0,0 +1,69 @@
...
+    public function run()
+    {
+        global $wpdb;
+
+        $formMetaTable = $wpdb->prefix . 'give_formmeta';
+
+        $sql = $wpdb->prepare("SELECT form_id FROM $formMetaTable as fm WHERE
+        (fm.meta_key = 'formBuilderSettings' AND fm.meta_value LIKE '%\\\\\\\\\\\\\\\\\\\\\\\\%')
+        OR
+        (fm.meta_key = 'formBuilderFields' AND fm.meta_value LIKE '%\\\\\\\\\\\\\\\\\\\\\\\\%')");
```
jonwaldstein, line 32: Is this a problem in the formBuilderFields meta as well? I thought it was only in the settings.

src/DonationForms/Migrations/CleanMultipleSlashesOnDB.php
```diff
@@ -0,0 +1,69 @@
...
+     */
+    public function run()
+    {
+        global $wpdb;
+
+        $formMetaTable = $wpdb->prefix . 'give_formmeta';
+
+        $sql = $wpdb->prepare("SELECT form_id FROM $formMetaTable as fm WHERE
+        (fm.meta_key = 'formBuilderSettings' AND fm.meta_value LIKE '%\\\\\\\\\\\\\\\\\\\\\\\\%')
+        OR
```
jonwaldstein, line 31: Are we confident in looking for this many slashes?

src/DonationForms/Migrations/CleanMultipleSlashesOnDB.php
```diff
@@ -0,0 +1,69 @@
...
+        (fm.meta_key = 'formBuilderFields' AND fm.meta_value LIKE '%\\\\\\\\\\\\\\\\\\\\\\\\%')");
+
+        $formIds = $wpdb->get_results($sql, ARRAY_A);
+
+        foreach ($formIds as $formId) {
+            $form = DonationForm::find((int)$formId['form_id']);
+            $settings = (new SanitizeDonationFormPreviewRequest())($form->settings->toArray());
+            $blocks = (new SanitizeDonationFormPreviewRequest())($form->blocks->toArray());
```
jonwaldstein, line 39: Let's pull the functionality of this action directly into the migration.  We always want to limit our production code inside migrations to prevent changes up the chain affecting the migration.

With that in mind, we should attempt to remove the usage of the models here and just try to update the problematic meta by itself so the migration is only touching what it needs to be fixed.

jonwaldstein (changes requested):

src/DonationForms/Migrations/CleanMultipleSlashesOnDB.php
glaubersilva: For some reason, these 24 backslashes became 12 when we passed it to the `$sql` var, and became 6 when it runs on the database _which represents a sequence of only 3 backslashes_ as we need 2 of them to escape just 1 as you can check here: 

image

https://dev.mysql.com/doc/refman/8.0/en/string-literals.html

src/DonationForms/Migrations/CleanMultipleSlashesOnDB.php
glaubersilva: I have the impression that I have seen the problem in the `formBuilderFields` before, but I can't replicate it. 

But maybe is just my brain is lying to me and I only added it on migration because this is also handled in the `SanitizeDonationFormPreviewRequest` class when the form builder loads as you can see on the attached screenshot. 

Not sure if we should keep it "just in case" or if is best to remove it.

image

jonwaldstein: Since migrations directly affect customer data, I think it's important to keep the scope as small as possible and avoid speculating a problem.

We want to make sure our solution solves a specific problem and only touches what it needs.  If there is no evidence or possibility of this affecting `formBuilderFields` than my vote is _do not_ include it.

src/DonationForms/Migrations/CleanMultipleSlashesOnDB.php
glaubersilva: Resolved in 85e822c2bfdc2ed2ba18023fab65d29f824b0658

glaubersilva:
@jonwaldstein Thanks for the improvement suggestions. This is ready for re-review.

jonwaldstein (approved):
@glaubersilva excellent job! this really took a lot of thorough investigation and i'm happy with the slimmer migration scope 🙌 

Just requested one unused piece to be removed, otherwise ready for QA 🌈

src/Framework/Blocks/BlockCollection.php
```diff
@@ -46,6 +46,19 @@ public static function fromJson($blocksJson): self
...
+    public static function fromArray(array $blocksArray): self
+    {
+        $blocks = [];
+        foreach ($blocksArray as $block) {
+            $blocks[] = BlockModel::make($block);
+        }
+
+        return new self($blocks);
+    }
+
```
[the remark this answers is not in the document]
glaubersilva: Resolved in 6ea162619c5b900fe14fb432f6897bcc5eed3530

rickalday (approved):
Passed manual QA tests.

PR #7112, opened 2023-11-21
Add filters when a payment is complete, processing or pending

Genevieve-K:
I need a filter when a payment is accepted, before the confirmation page is displayed. I started the discussion [https://github.com/impress-org/givewp/discussions/7108] to explain my problem. I suggest directly these filters.


jonwaldstein:
From our [discussion](https://github.com/impress-org/givewp/discussions/7108) I do not think this would be a good place to add hooks for your purposes as they would be misleading - depending on how gateways handle the donation - these commands are optional. The most reliable way right now would be to hook into the model actions like `givewp_donation_updated`.  

We will be exploring a more unified approach to determine when a form has been completed in the near future.

PR #7113, opened 2023-11-21
Avoid a fatal error when activating GiveWP for a new site

Genevieve-K:
If GiveWP is activated on a new site, there is a fatal error:  Fatal error: Uncaught TypeError: array_key_exists(): Argument #2 ($array) must be of type array, null given in /var/www/html/wordpress/wp-content/plugins/give/src/PaymentGateways/Gateways/PayPalStandard/Migrations/SetPayPalStandardGatewayId.php:30 Stack trace: #0 

For a new site, no gateway is defined, and $gateways is null.


pauloiankoski (changes requested):
I'd recommend an early return. However, the `$gateways` array is not used in the next two if statements, which would prevent those statements from running.

In this case, I suggest adding an additional condition to that specific statement, such as:
```php
if (is_array($gateways) && array_key_exists('paypal-standard', $gateways)) {
```

Please, let me know if that works.

Genevieve-K:
Yes, that works. I did not pay attention that $gateways will not used after.

PR #7114, opened 2023-11-22
Fix: add distinct to donation forms list query to prevent duplicate results

jonwaldstein:
Resolves [GIVE-109]

## Description

This adds the SQL distinct statement to the donation forms list query to prevent duplicate results.

## Affects

The donation forms list table results

### Before

Screenshot 2023-11-22 at 1 01 46 PM

### After

Screenshot 2023-11-22 at 1 02 06 PM

## Testing Instructions

- Create a few forms and view the donation forms list
- Make sure there are not duplicate results
- Try searching for a title or ID and make sure there are not duplicate results

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-109]: https://stellarwp.atlassian.net/browse/GIVE-109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DonationForms/V2/Endpoints/ListDonationForms.php
+++ b/src/DonationForms/V2/Endpoints/ListDonationForms.php
@@ -149,6 +149,7 @@ class ListDonationForms extends Endpoint
     }
 
     /**
+     * @unreleased added distinct() to the query
      * @since 2.24.0 Refactor to query through the ModelQueryBuilder
      *
      * @return array
@@ -170,7 +171,7 @@ class ListDonationForms extends Endpoint
         $query->limit($perPage)
             ->offset(($page - 1) * $perPage);
 
-        $donationForms = $query->getAll();
+        $donationForms = $query->distinct()->getAll();
 
         if ( ! $donationForms) {
             return [];


JasonTheAdams:
@jonwaldstein I'm curious what the underlying SQL looks like that is returning multiple forms. That usually alludes to something odd in the query, and adding `DISTINCT` compensates for that. This isn't necessarily true, but when the form is the top-most object, it's strange that it _can_ return multiple rows per object.

alaca (approved):
:+1:

alaca:
As @JasonTheAdams mentioned, something is odd here. I tried to reproduce the issue, but I wasn't able to.

jonwaldstein:
@JasonTheAdams @alaca I did look at the SQL at first but nothing looked out of the ordinary so I figured distinct would be a quick win for people frustrated by this.  I'll dig in a little farther and see if I can pin point if it's a query/sql issue.

JasonTheAdams:
Thanks @jonwaldstein! Feel free to drop the resulting SQL here if you think more eyes would help.

alaca:
Here is the generated SQL. 

```sql 
SELECT ID                                                  AS id,
       post_title                                          AS title,
       post_date                                           AS createdAt,
       post_modified                                       AS updatedAt,
       post_status                                         AS status,
       give_formmeta_attach_meta_formEarnings.meta_value   AS formEarnings,
       give_formmeta_attach_meta_formSales.meta_value      AS formSales,
       give_formmeta_attach_meta_donationLevels.meta_value AS donationLevels,
       give_formmeta_attach_meta_setPrice.meta_value       AS setPrice,
       give_formmeta_attach_meta_priceOption.meta_value    AS priceOption,
       give_formmeta_attach_meta_goalOption.meta_value     AS goalOption
FROM wp_posts
         LEFT JOIN wp_give_formmeta give_formmeta_attach_meta_formEarnings
                   ON ID = give_formmeta_attach_meta_formEarnings.form_id AND
                      give_formmeta_attach_meta_formEarnings.meta_key = '_give_form_earnings'
         LEFT JOIN wp_give_formmeta give_formmeta_attach_meta_formSales
                   ON ID = give_formmeta_attach_meta_formSales.form_id AND
                      give_formmeta_attach_meta_formSales.meta_key = '_give_form_sales'
         LEFT JOIN wp_give_formmeta give_formmeta_attach_meta_donationLevels
                   ON ID = give_formmeta_attach_meta_donationLevels.form_id AND
                      give_formmeta_attach_meta_donationLevels.meta_key = '_give_donation_levels'
         LEFT JOIN wp_give_formmeta give_formmeta_attach_meta_setPrice
                   ON ID = give_formmeta_attach_meta_setPrice.form_id AND
                      give_formmeta_attach_meta_setPrice.meta_key = '_give_set_price'
         LEFT JOIN wp_give_formmeta give_formmeta_attach_meta_priceOption
                   ON ID = give_formmeta_attach_meta_priceOption.form_id AND
                      give_formmeta_attach_meta_priceOption.meta_key = '_give_price_option'
         LEFT JOIN wp_give_formmeta give_formmeta_attach_meta_goalOption
                   ON ID = give_formmeta_attach_meta_goalOption.form_id AND
                      give_formmeta_attach_meta_goalOption.meta_key = '_give_goal_option'
WHERE post_type = 'give_forms'
  AND post_status IN ('publish', 'draft', 'pending', 'private', 'upgraded')
```

jonwaldstein:
Okay I found the problem, duplicate meta keys.

The query & underlying SQL is fine.  It's that there are some form meta keys that are duplicated (will have to figure out why)

Screenshot 2023-11-22 at 1 29 05 PM

alaca:
Yep, that's it.

jonwaldstein:
@alaca if we can figure out where the problem originates, remove it and add a migration - that would be my preference.  

However, I realize that could take some time that we might not currently have available - so I think simply using distinct for now would still be a good solution for the interim. 

@JasonTheAdams to your point, we _are_ compensating by using distinct, but it seems to be warranted until we can figure out the underlying problem.

JasonTheAdams:
I happened upon this video and it helped me understand our predicament: https://youtu.be/5hsl47I3svw?si=c7hP-jglwkD7YQuh

The issue is that our `LEFT JOIN` meta lines will create a duplicate line if there's more than one meta line. That means this is more than just an issue with our list table, but could happen _any_ time we query a donation and join the meta. We may consider putting a `LIMIT 1` on the `attachMeta` queries. 🤔

alaca:
I'm not sure we can just add LIMIT to the JOIN query, at least not in the way we are writing queries now. We have to use JOIN because we need data from the meta table, but we might try a slightly different approach - JOIN with subquery. Then we can limit the result of the subquery.

JasonTheAdams:
@alaca Yeah, I had a feeling a LIMIT can't be applied to a join in this way. I think it could be worth experimenting with a sub-query join instead of a table join. It's definitely a gnarly bug of how `attachMeta` currently works, which can easily throw us off in other places. As it is, if we apply `DISTINCT` here we're just compensating for a (now) known query builder bug. I'd rather fix it at the source.

alaca:
I agree with Jon, we should probably release this as a hot fix. 

Also, I saw the task for replacing the Give QB with the QB package from Stellar, maybe it would be a good idea to resolve this issue there first. I also have ideas on how to improve the QB, mainly insert, update, and delete methods.

JasonTheAdams:
Since adding `DISTINCT` won't adversely affect the fix we're discussing, I'm onboard with releasing this as a hot fix tomorrow — not today, as it's not breaking any donations, and it's not worth introducing risk. Also, since 3.1.2 is effectively merged, there's no way to do a hot fix without releasing all of that as well.

@alaca I'm ok with including this as part of the Stellar DB work, but that puts it in a much higher priority for me. We now have a known bug that affects _any_ model attaching meta. Mind you, I don't think duplicate meta happens too often, and this only affects `QueryBuilder::getAll()` (since `QueryBuilder::get()` only grabs the first row). Still, not something we want to ignore as it's difficult to predict where it will occur.

jonwaldstein:
@JasonTheAdams sounds good, i'll get this into QA and create a new issue to tackle the QB part.

PR #7054, opened 2023-10-19
Refactor: replace the Editor component (react-quill) with WP ClassicEditor component ( TinyMCE) 

glaubersilva:
Resolves https://stellarwp.atlassian.net/jira/software/c/projects/GIVE/boards/382?selectedIssue=GIVE-55 

## Description

This PR adds a new `ClassicEditor` component (extracted from the Custom PDF Builder modal for v3 forms from the PDF Receipts addon) that uses the TinyMCE editor shipped with WordPress under the hood and uses it in all places where the Editor component (that uses the react-quill package under the hood) was being used.

This PR also mounts this new component in the global window object to allow the addons also use this same component preventing code duplication.

## Affects

- Email message content in the email settings of V3 forms.
- The Agreement Text content in the Terms & Consent block of V3 forms.
- The Donation Instructions in the Offline Donations Gateway.

## Visuals

image

image

image

## Testing Instructions

1. Create or edit a v3 form;
2. Go to the email settings and play around with the email message content of the email notifications.
3. Add a Terms & Consent block, go to the Agreement Text content, and play around with the editor options. 
4. Go to the Offline Donations gateway settings and play around with the donation instructions content of the gateway.

## Pre-review Checklist

-   [ ] Acceptance criteria satisfied and marked in related issue
-   [ ] Relevant `@unreleased` tags included in DocBlocks
-   [ ] Includes unit tests
-   [ ] Reviewed by the designer (if follows a design)
-   [ ] [Self Review](https://give.gitbook.io/development-manual/devops/github/code-reviews#self-review) of code and UX completed

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205606769216041

PR #7122, opened 2023-11-29
Refactor: remove ClassicEditor from window.givewp.form.components and use the new @givewp/form-builder-library npm package

glaubersilva:
Resolves [GIVE-140](https://stellarwp.atlassian.net/browse/GIVE-140)

## Description

The `ClassicEditor` was [previously](https://github.com/impress-org/givewp/pull/7054) mounted in the window global object to prevent code duplication in the addons. 

But it also introduces a few problems like forcing bumping the minimum Give core version in the addons, and other problems related to possible changes of the components in the core that can break functionalities in the addons. 

~~In the future, a JS package will be released to share the components between the addon without major issues~~

~~But for now, was decided to remove this component from the global window object and fork it in each addon that needs to use it.~~ 

So, this PR removes this component from the global window object and also replaces some components with the ones shipped with the `@givewp/form-builder-library` npm package.

This PR also changes the `OptionsPanel` component used in the donations levels and name title prefixes to the version shipped with the new package.

## Affects

- The `window.givewp.form.components` property.
- Email message content in the email settings of V3 forms.
- The Agreement Text content in the Terms & Consent block of V3 forms.
- The Donation Instructions in the Offline Donations Gateway.
- Donation levels in the Donation Amount and Levels block.
- Name title prefixes in the Donor Name block.

## Testing Instructions

1. Create or edit a v3 form;
2. Go to the email settings and play around with the email message content of the email notifications.
3. Add a Terms & Consent block, go to the Agreement Text content, and play around with the editor options.
4. Go to the Offline Donations gateway settings and play around with the donation instructions content of the gateway.
5. Go to the donation levels in the Donation Amount and Levels block and play around with the options/levels.
6. Go to the name title prefixes in the Donor Name block and play around with the options/prefixes.

To test the addons that were using the ClassicEditor component mounted in the window global object, check out these two PRs:

https://github.com/impress-org/give-tributes/pull/324

https://github.com/impress-org/give-form-field-manager/pull/416

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-140]: https://stellarwp.atlassian.net/browse/GIVE-140?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jonwaldstein:
@glaubersilva is any currently released add-on using this component?  technically this was released in the latest version of GiveWP so we need to be extra careful about removing it.

glaubersilva:
@jonwaldstein Only the FFM and Tributes addon was using it, but the PRs that implemented it still weren't released, so we are safe. 

The other addon that uses a component with the same name is the PDF Receipts, but it also uses its own version of this component. 

You can double-check it here: https://github.com/search?q=org%3Aimpress-org%20ClassicEditor&type=code

jonwaldstein:
@glaubersilva just to be safe, let's hold off merging this until the add-ons are _not_ referencing this component through the window on their develop branches.

glaubersilva:

src/DonationForms/resources/app/fields/SectionNode.tsx
```diff
@@ -47,12 +47,16 @@ function SectionNode({node}: {node: Node}) {
 
     if (isField(node) && fieldTemplateExists(node)) {
         if (node.type === 'gateways') {
+            // @ts-ignore
```
jonwaldstein, line 50: I'm seeing alot of these `@ts-ignore` additions that is concerning.  These should _only_ be used as a last resort - and considering they were not needed previously tells me there is some kind of internal problem we should be fixing up the chain.

can we remove them?

src/FormBuilder/resources/js/form-builder/src/blocks/fields/amount/inspector/index.tsx
```diff
@@ -14,8 +14,8 @@ import periodLookup from '../period-lookup';
 import RecurringDonationsPromo from '@givewp/form-builder/promos/recurring-donations';
 import {getFormBuilderWindowData} from '@givewp/form-builder/common/getWindowData';
 import {useCallback, useState} from '@wordpress/element';
-import Options from '@givewp/form-builder/components/OptionsPanel';
-import {OptionProps} from '@givewp/form-builder/components/OptionsPanel/types';
+import {OptionsPanel} from '@givewp/form-builder-library';
+import {OptionProps} from '@givewp/form-builder-library/build/OptionsPanel/types';
```
jonwaldstein, line 18: FYI you can [import as type](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export) so it gets excluded during runtime

```suggestion
import type {OptionProps} from '@givewp/form-builder-library/build/OptionsPanel/types';
```

jonwaldstein (changes requested):

src/DonationForms/resources/app/fields/SectionNode.tsx
glaubersilva: After installing version `1.0.0` of the new package, the IDE started to highlight all places that use the node attribute and it also started preventing the `npm run dev` from finishing the compile process due this error:

```
TS2322: Type { node: Field; } is not assignable to type IntrinsicAttributes Property node does not exist on type IntrinsicAttributes
``` 

So now, after updating the package for the `1.1.2` version, I tried to remove these `@ts-ignore` tags but the problem keeps happening even with the new package as you can check in the attached screenshot. 

image

src/FormBuilder/resources/js/form-builder/src/blocks/fields/amount/inspector/index.tsx
glaubersilva: Resolved in 460514e85889540592879c6dc3612d84a90124af

src/DonationForms/resources/app/fields/SectionNode.tsx
glaubersilva: @jonwaldstein One more thing, if we execute the `npm run dev:v3` the problem doesn't happen. So, may this is happening due to this: 

> ...running npm run dev on GiveWP runs both the Laravel Mix dev and WP Scripts dev. Laravel Mix has strict TypeScript checking in place, while WP Scripts does not. It's annoying. Usually, the best route is to simply address the TS warnings/errors, using `@ts-ignore` when needed.

**Source:** https://github.com/impress-org/form-builder-library/pull/1#issuecomment-1836378810

Any thoughts?

jonwaldstein: Okay, i'd like to explore the problem more.  Although Laravel mix points out typescript errors while wp-scripts does not - that alone does not warrant adding `@ts-ignore` as it defeats the purpose of using typescript all together.  

I would only recommend using `@ts-ignore` when resolving the TS error is _not_ possible - say from an external library.  However, if typescript is pointing out an error in our code than we should be resolving that error.

I do understand certain dependencies can add conflicting typescript errors but those should be _addressed_ - not be _ignored_, so we can figure out _why_ for the long-term.

jonwaldstein: resolved [c471a2a](https://github.com/impress-org/givewp/pull/7122/commits/c471a2a892497d004799daf21b90bbd2ae0d9474)

jonwaldstein (approved):
Beautiful! ready for QA 🌈

rickalday (approved):
Passed manual QA tests.

PR #7126, opened 2023-12-05
Feature: Add password update to donor dashboard

kjohnson:
## Description

This PR adds a password update form to the Donor Dashboard.
 
## Visuals

Screenshot 2023-12-05 at 17-04-41 Donor Dashboard – WordPress

### DIFF
--- a/src/DonorDashboards/Tabs/EditProfileTab/PasswordRoute.php
+++ b/src/DonorDashboards/Tabs/EditProfileTab/PasswordRoute.php
@@ -38,7 +38,7 @@ class PasswordRoute extends RouteAbstract
     /**
      * Handles password update.
      *
-     * @since 2.10.0
+     * @unreleased
      *
      * @param WP_REST_Request $request
      *
--- a/src/DonorDashboards/Tabs/EditProfileTab/PasswordRoute.php
+++ b/src/DonorDashboards/Tabs/EditProfileTab/PasswordRoute.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace Give\DonorDashboards\Tabs\EditProfileTab;
+
+use Give\DonorDashboards\Helpers\SanitizeProfileData as SanitizeHelper;
+use Give\DonorDashboards\Profile as Profile;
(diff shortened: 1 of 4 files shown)


src/DonorDashboards/Tabs/EditProfileTab/PasswordRoute.php
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Give\DonorDashboards\Tabs\EditProfileTab;
+
+use Give\DonorDashboards\Helpers\SanitizeProfileData as SanitizeHelper;
+use Give\DonorDashboards\Profile as Profile;
+use Give\DonorDashboards\Tabs\Contracts\Route as RouteAbstract;
+use WP_REST_Request;
+
+/**
+ * @since 2.10.0
```
JoshuaHungDinh, line 11: Same question about the tags from the other PR

JoshuaHungDinh:
One thing I stumbled across is that if you reset your password and then immediately go to log out your session will hang. But the intended password functionality is working correctly.

src/DonorDashboards/Tabs/EditProfileTab/PasswordRoute.php
kjohnson: Good 'ol copy and pase 😎 
Resolved, see https://github.com/impress-org/givewp/pull/7126/commits/89fb91b87b567583eafc40f1b9e572bfb9a07d4d

kjohnson:
> One thing I stumbled across is that if you reset your password and then immediately go to log out your session will hang. But the intended password functionality is working correctly.

Good catch. After password update the cookie is invalidated, so we need to refresh the page.

Resolved, see https://github.com/impress-org/givewp/pull/7126/commits/5733227dfa3f47d9cd7c91d212ee547432def959

JoshuaHungDinh (approved):
Left a comment to address on the original GIVE-29, but otherwise they are both ready for QA. Thanks!

rickalday (approved):
Passed manual QA tests

PR #7129, opened 2023-12-06
Feature: add Mailchimp add-on to the form migration process

JoshuaHungDinh:
Resolves: [GIVE-136](https://stellarwp.atlassian.net/browse/GIVE-136)
## Description

This PR includes a form migration step for the new Mailchimp block migrating from a v2 form to a v3 form. I also removed the commented section of supported addons for mailchimp so that it will display on our list when upgrading.

 Cases when the block should be added.
- While Mailchimp is globally enabled on all forms, if the form is not explicitly disabled from the per form setting migrating should add the block.
- While Mailchimp is not globally enabled if it is explicitly enabled on the per form setting than migrating should add the block.

## Testing Instructions
- Use [New Mailchimp Block 
](https://github.com/impress-org/give-mailchimp/pull/84)
- Create a v2 form -> update the settings -> migrate the form to verify the new settings and block are added if they match one of the above cases. If not verify the block in not added to the new form.
- Verify Mailchimp icon is displayed as a supported addon.

## Pre-review ChecklistT

[GIVE-136]: https://stellarwp.atlassian.net/browse/GIVE-136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/FormMigration/Steps/MailchimpSettings.php
```diff
@@ -0,0 +1,73 @@
...
+class MailchimpSettings extends FormMigrationStep
+{
+    /**
+     * @unreleased
+     */
+    public function process(): void
+    {
+        $prevFormId = $this->formV2->id;
+
+        $isFormEnabled = $this->getFormMetaValue($prevFormId, '_give_mailchimp_enable') === 'true';
```
kjohnson, line 20: I would suggest adding these calls to the `FormMetaDecorator`, which is a more declarative approach.

For example, you can call `$this->formV2->getDonationGoalAmount()`, which encapsulates where/how that value is stored.

In this case, it could be `$this->formV2->isMailchimpEnabled()`.

Also, instead of directly comparing against `true`, consider the function helper `give_is_setting_enabled()`.

src/FormMigration/Steps/MailchimpSettings.php
```diff
@@ -0,0 +1,73 @@
...
+        $isFormDisabled = $this->getFormMetaValue($prevFormId, '_give_mailchimp_disable') === 'true';
+        $isGloballyEnabled = give_get_option( 'give_mailchimp_show_checkout_signup') === 'on';
+
+        if ($isFormDisabled || (!$isGloballyEnabled && !$isFormEnabled)) {
```
kjohnson, line 24: I would suggest moving the "if this should migrate" login in the `canHandle` method of the migration step, which is an optional method that runs before the migration step is processed. By default it just returns true, but you can define the method and add your logic there.

The result is a `process` method which only does processing, like making the block and inserting the block.

src/FormMigration/Steps/MailchimpSettings.php
```diff
@@ -0,0 +1,73 @@
...
+        return $meta === '' ? null : $meta;
+    }
+
+    /**
+     * @unreleased
+     */
+    private function getAttributes($prevFormId): array
+    {
+        return   [
+            'label'            => $this->getFormMetaValue($prevFormId, '_give_mailchimp_custom_label') ??
```
kjohnson, line 52: These would also be good additions to the `FormMetaDecorator`.

kjohnson (commented):
There is some re-organization that I would suggest considering.

src/FormMigration/Steps/MailchimpSettings.php
JoshuaHungDinh: Much cleaner. a0161125118cc5af042dc2b0087be48f697359db

src/FormMigration/Steps/MailchimpSettings.php
JoshuaHungDinh: Ok, got it. Thanks! a0161125118cc5af042dc2b0087be48f697359db

src/FormMigration/Steps/MailchimpSettings.php
JoshuaHungDinh: We are just talking about the form meta values here correct? f8a30efecf86d0c0a15de8ecb42590bdd60ac82b

kjohnson (approved):

kjohnson:
Build for QA: https://github.com/impress-org/givewp/actions/runs/7200690861

rickalday (approved):
Passed manual QA tests

PR #7131, opened 2023-12-06
Fix: duplicate meta keys

alaca:
Resolves [GIVE-137]

## Description
This PR resolves the issue of duplicate donation forms. The problem was in the logic of the `get_earnings` method combined with the donation transfer logic which resulted in the meta key being duplicated on every page load. That also affected the donation forms list table which was displaying duplicated results. 

The issue is resolved by using `update_post_meta` instead of `add_post_meta` function, and by adding a migration for deleting duplicated meta keys (`_give_form_earnings`, `_give_form_sales`) and setting the correct values for  `_give_form_earnings` is added. 

## Affects

Donation Forms list table
Donation Form data transfer

## Testing Instructions

The best way to test this is on an existing site that has this problem.

Create a new site by using Local or something similar.
Install All-In-On WP Migration plugin.
Download the site backup and import it (https://drive.google.com/file/d/16SIlaCQ-1nMvneFpYvdLStA2_ovtQ6vG/view?usp=sharing).
After import, verify that the problem exists by checking the donation forms list table, or even better by querying the `give_formmeta` table. Something like: 
```sql 
select form_id, meta_key from wp_give_formmeta where form_id = FORM_ID and ( meta_key = '_give_form_sales' or  meta_key = '_give_form_earnings');
```

Now delete or disable the existing Give plugin and install Give from GitHub repo.
Switch to this branch `fix/duplicate-meta-keys`
Check the donation forms list table.

## Pre-review Checklist

[GIVE-137]: https://stellarwp.atlassian.net/browse/GIVE-137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DonationForms/Migrations/RemoveDuplicateMeta.php
+++ b/src/DonationForms/Migrations/RemoveDuplicateMeta.php
@@ -49,6 +49,7 @@ class RemoveDuplicateMeta extends Migration
 
         foreach ($posts as $post) {
             $formEarnings = give_get_meta($post->ID, '_give_form_earnings');
+            $formSales = give_get_meta($post->ID, '_give_form_sales');
 
             // Update meta with duplicate keys only
             if (count($formEarnings) > 1) {
@@ -59,6 +60,15 @@ class RemoveDuplicateMeta extends Migration
                 $earnings = $this->getEarningsFromDonations($post->ID);
                 give_update_meta($post->ID, '_give_form_earnings', $earnings);
             }
+
+            if (count($formSales) > 1) {
+                // Delete all meta
+                give_delete_meta($post->ID, '_give_form_sales');
+
+                // Get sales count from donations
+                $sales = $this->getSalesCountFromDonations($post->ID);
+                give_update_meta($post->ID, '_give_form_sales', $sales);
+            }
         }
     }
 
@@ -67,9 +77,9 @@ class RemoveDuplicateMeta extends Migration
      *
      * @param $formId
      *
-     * @return int
+     * @return float | int
      */
-    private function getEarningsFromDonations($formId): int
+    private function getEarningsFromDonations($formId)
     {
         $donations = Db::table('posts')
             ->attachMeta(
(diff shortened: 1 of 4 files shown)


alaca:
@jonwaldstein good catch. Although it's not necessary, we should update it. Resolved!

src/DonationForms/Migrations/RemoveDuplicateMeta.php
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Get sales count from donations by Form ID
+     *
+     * @param $formId
+     *
+     * @return int
+     */
+    private function getSalesCountFromDonations($formId): int
+    {
+        return Db::table('posts')
```
jonwaldstein, line 109:
```suggestion
        return DB::table('posts')
```

src/DonationForms/Migrations/RemoveDuplicateMeta.php
```diff
@@ -0,0 +1,122 @@
...
+            ->attachMeta(
+                'give_donationmeta',
+                'ID',
+                'donation_id',
+                [DonationMetaKeys::FORM_ID(), 'formId']
```
jonwaldstein, line 114: With these value objects you will either need to use `DonationMetaKeys::FORM_ID()->getValue()` or `DonationMetaKeys::FORM_ID` to get the value or else it will be an object.

```suggestion
                [DonationMetaKeys::FORM_ID, 'formId']
```

src/DonationForms/Migrations/RemoveDuplicateMeta.php
```diff
@@ -0,0 +1,122 @@
...
+     */
+    private function getEarningsFromDonations($formId)
+    {
+        $donations = Db::table('posts')
+            ->attachMeta(
+                'give_donationmeta',
+                'ID',
+                'donation_id',
+                [DonationMetaKeys::FORM_ID(), 'formId'],
+                [DonationMetaKeys::AMOUNT(), 'amount'])
```
jonwaldstein, line 90:
```suggestion
                [DonationMetaKeys::FORM_ID, 'formId'],
                [DonationMetaKeys::AMOUNT, 'amount'])
```

jonwaldstein (changes requested):

src/DonationForms/Migrations/RemoveDuplicateMeta.php
alaca: @jonwaldstein good catch! Not sure how  I missed that :sweat_smile: 
Resolved https://github.com/impress-org/givewp/pull/7131/commits/b29729ff110b24c13adeb2e963861371d985c025

jonwaldstein (approved):
@alaca great job! ready for QA 🌈

rickalday (approved):
Passed manual QA test

PR #7134, opened 2023-12-07
Fix: Prevent duplicate "Donate Now" button on legacy form template

kjohnson:
## Description

This PR prevents a duplicate "Donate Now" button on the legacy form template, in which the modal launch button is being rendered when the display option is not set to modal.

Specifically, this issue is resolved by reverting the `fullForm` display option to the original `onpage`.

### DIFF
--- a/src/DonationForms/Blocks/DonationFormBlock/resources/app/index.tsx
+++ b/src/DonationForms/Blocks/DonationFormBlock/resources/app/index.tsx
@@ -5,7 +5,7 @@ import IframeResizer from 'iframe-resizer-react';
 import '../editor/styles/index.scss';
 
 type DonationFormBlockAppProps = {
-    formFormat: 'fullForm' | 'newTab' | 'modal' | string;
+    formFormat: 'onpage' | 'newTab' | 'modal' | string;
     dataSrc: string;
     embedId: string;
     openFormButton: string;
--- a/includes/forms/functions.php
+++ b/includes/forms/functions.php
@@ -1587,7 +1587,7 @@ function give_get_default_form_shortcode_args() {
 		'show_goal'             => true,
 		'show_content'          => '',
 		'float_labels'          => '',
-        'display_style'         => 'fullForm',
+        'display_style'         => 'onpage',
         'continue_button_title' => __('Donate now', 'give'),
 
 		// This attribute belong to form template functionality.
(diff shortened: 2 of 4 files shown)


src/DonationForms/Blocks/DonationFormBlock/resources/app/index.tsx
```diff
@@ -5,7 +5,7 @@ import IframeResizer from 'iframe-resizer-react';
 import '../editor/styles/index.scss';
 
 type DonationFormBlockAppProps = {
-    formFormat: 'fullForm' | 'newTab' | 'modal' | string;
+    formFormat: 'onpage' | 'newTab' | 'modal' | string;
```
jonwaldstein, line 8: I see we're replacing all the `fullForm` values with `onpage` - however, i'm concerned people could be directly using `fullForm` - what if instead we just added `onpage` to these formFormat options instead for replacing?

kjohnson: The intended result is the same. v2 used `onpage` and v3 supports either `onpage` or `fullForm` (because `fullForm` was introduced later, as a rename of `onpage`).

kjohnson: We could conditionally add it as an option if the selected value is `fullForm`, so that they can change the value (at which point it would not be a future option).

kjohnson: Although, to reproduce the issue you would have to, while using the buggy version, add a donation form block, update the display style to something else, and then update the display style again to `fullPage`. Otherwise, the default value when the block is created is still `onpage`, which is correct. So, I'm not sure how common this is.

jonwaldstein: (Thinking out load)

If a block was previously stored with the `fullForm` value - that means it is stored in post content as such. So when rendering it would still use `fullForm` unless manually changed.  In retrospect, `onpage` should be the default when rendering (when `display_style` attribute is _not_ one of the expected options)

kjohnson: It would be simple enough to have a support response of "update, remove the block, and re-add the block".

jonwaldstein: I wouldn't go down that path, that would be a huge nuisance for customers.  Honestly, this might be a non-issue as v3 forms don't actually use `fullForm` or `onPage` - they treat those as the default.  And it looks like v2 forms work similarly.

jonwaldstein (approved):

PR #7149, opened 2023-12-14
Feature: add receipt api for fields

jonwaldstein:
Resolves: [GIVE-180]

### Preface

For fields that either don't store their value in meta (or at all), use a custom scope, or the field attributes just does not correlate to what should be displayed on the confirmation page - there is currently no way to modify the label & value.

At first I was going to simply add an action like `givewp_donation_confirmation_receipt` (which we still can) but then realized how removed from context the action is as it relates to field data.  For example, you will often need to go through the form schema to find the field to use field attributes/settings.  Where as giving fields a first-class api has the benefit of staying in context with the rest of the field settings/schema and likely will reduce logic & overhead.

### The solution

This gives fields a way to optionally & programmatically display their label/value in the confirmation page.  Currently the `receiptValue(Closure $value)` api is a simple closure that (in the context of a donation form) will pass the `$field` and `$donation` model.  I thought about adding a string option like scopes do but thought this would be the simpler way to go for now (since I also added filters).

For example:
```
$field = Select::make(FundField::FUND_ID)
->tap(function ($field) use ($block) {
    $this->setFieldOptions($field, $block);
})->scope(function (Select $field, $value, Donation $donation) {
    if (!empty($value)) {
        give(FundsRevenueRepository::class)->updateDonationFund($donation->id, $value);
    }
})
->showInReceipt()
->receiptLabel(__('Fund', 'give-funds'))
->receiptValue(function (Select $field, Donation $donation) {
    $fund = give(Funds::class)->getByDonationId($donation->id);
    if (!$fund) {
        return null;
    }

    return $fund->getTitle();
});
```

I also added a couple filters so folks could modify the values if need be:
- `apply_filters('givewp_donation_confirmation_page_field_value_for_{$fieldName}', $value, $field, $donation)`
- `apply_filters('givewp_donation_confirmation_page_field_label_for_{$fieldName}', $value, $field, $donation)`

These would be most useful for customers using add-ons that add their own receipt labels/values and might want to modify that (maybe for language, syntax, or something else).  You could also use these filters _instead_ of the fields api if need be - either way it's an easy win.

## Affects

- The donation confirmation page

### Api in action! ("Fund")

Screenshot 2023-12-14 at 3 09 09 PM

## Testing Instructions

- Add a field programmatically to the form
- use the new api for your receipt values

```
add_action('givewp_donation_form_schema', static function (DonationForm $form) {
    $field = Text::make('custom_text_field_example')
        ->showInAdmin()
        ->showInReceipt()
        ->scope('my_custom_scope_stored_somewhere_fancy')
        ->label('Whats your favorite color?')
        ->receiptLabel('Your favorite color is:')
        ->receiptValue(function (Text $field, Donation $donation) {
            // here you could do something fancy like get the value from a custom table
            // or derive from your field, etc.
            return "Blue!";
        });

    $form->insertAfter('email', $field);
});
```

Or with filters:

```
add_filter('givewp_donation_confirmation_page_field_label_for_custom_text_field_example', static function ($label, $field, $donation) {
    return 'Your new favorite color is:';
}, 10, 3);

add_filter('givewp_donation_confirmation_page_field_value_for_custom_text_field_example', static function ($label, $field, $donation) {
    return 'Blue';
}, 10, 3);

add_action('givewp_donation_form_schema', static function (DonationForm $form) {
    $field = Text::make('custom_text_field_example')
        ->showInAdmin()
        ->showInReceipt()
        ->scope('my_custom_scope_stored_somewhere_fancy')
        ->label('Whats your favorite color?');

    $form->insertAfter('email', $field);
});
```

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-180]: https://stellarwp.atlassian.net/browse/GIVE-180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Framework/Receipts/Actions/GenerateConfirmationPageReceipt.php
```diff
@@ -46,26 +49,56 @@ protected function getCustomFields(Donation $donation): array
...
+                sprintf("givewp_donation_confirmation_page_field_label_for_%s", $field->getName()),
+                $field->hasReceiptLabel() ? $field->getReceiptLabel() : $field->getLabel(),
+                $field,
+                $donation
             );
+
+            if (!empty($label)){
+                 $receiptDetails[] = new ReceiptDetail(
+                    $label,
+                    $value ?? ''
```
kjohnson, line 99: What is the use case for adding a label to the Receipt, but not a value?

jonwaldstein: It's a good question, I think it has to do with the option to _show in receipt_ - where the expectation is it will show up as a line item - not necessarily meaning it has a value.

JasonTheAdams: Is this basically just handling scenarios where the value is null but we want the value to always be a string, or something? I'm confused, too. Don't all fields have a corresponding value?

jonwaldstein: Circling back on this for more context.  This actually comes down to a UX decision - when a form has fields that are not required, should they still show up in the receipt as label & empty value?  This is currently how it works but we can talk about this as a team.

kjohnson: Update: via Slack we decided to maintain the status quo and possibly revisit if necessary.

kjohnson (approved):

rickalday (approved):
Passed manual QA tests.

PR #7155, opened 2023-12-22
State PHP requirement

szepeviktor:
## Description

Helps people using [Composer](https://github.com/szepeviktor/composer-managed-wordpress) to install WordPress.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,6 +5,7 @@
     "description": "GiveWP - a WordPress Donation Plugin and Fundraising Platform",
     "homepage": "https://givewp.com/",
     "require": {
+        "php": "^7.2 || ~8.0.0",
         "composer/installers": "~1.2",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -5,6 +5,7 @@
     "description": "GiveWP - a WordPress Donation Plugin and Fundraising Platform",
     "homepage": "https://givewp.com/",
     "require": {
+        "php": "^7.2 || ~8.0.0",
```
jonwaldstein, line 8: We do already have a composer `platform` config set to php 7.2

szepeviktor: Okay. What should I do now?

jonwaldstein: If you agree this addition is not necessary since we already have the platform config, we can just close the PR

szepeviktor: All right!

PR #7159, opened 2023-12-26
Fix: allow users to type into the state/county field input

pauloiankoski:
Fix necessary for [GIVE-12]

## Description

While working on Gift Aid 3.0 compatibility, we encountered an issue with the `State` field (referred to as `County` for Gift Aid in the United Kingdom). The problem is that users are unable to enter any text in this field. In the case of countries like the USA, the state field is a dropdown menu with available options and it works correctly. However, for countries like the UK, it should be a text input, and that's where the error occurs.

Upon further investigation, I discovered that the text input had a fixed value of `''` and did not update when typing.

This pull request resolves the issue by making the value of the field dynamic, updating according to the current value of the `state` state.

## Affects

Billing Address block

## Visuals

|CleanShot 2023-12-26 at 18 09 34|CleanShot 2023-12-26 at 18 10 24|
|----|----|

## Testing Instructions

1. Add a v3 form
2. Add the Billing Address block
3. Open the form
4. Select United Kingdom in the country field
5. Type into the County field

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-12]: https://stellarwp.atlassian.net/browse/GIVE-12?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DonationForms/resources/registrars/templates/groups/BillingAddress.tsx
+++ b/src/DonationForms/resources/registrars/templates/groups/BillingAddress.tsx
@@ -43,6 +43,7 @@ async function getStates(url, country) {
 /**
  * This component is used to dynamically update the state field based on the country value
  *
+ * @unreleased Set current state value to the state input field
  * @since 3.0.0
  */
 function StateFieldContainer({
--- a/src/DonationForms/resources/registrars/templates/groups/BillingAddress.tsx
+++ b/src/DonationForms/resources/registrars/templates/groups/BillingAddress.tsx
@@ -182,16 +182,13 @@ function StateFieldContainer({
             <label>
                 <Label label={stateLabel ?? __('State', 'give')} required={stateRequired} />
 
-                {statesLoading ? (
-                    <input type="text" disabled={true} value={__('Loading...', 'give')} />
-                ) : (
-                    <input
-                        type="text"
-                        value={state}
-                        onChange={updateStateValue}
-                        aria-invalid={fieldError ? 'true' : 'false'}
-                    />
-                )}
+                <input
+                    type="text"
+                    onChange={updateStateValue}
+                    aria-invalid={fieldError ? 'true' : 'false'}
+                    placeholder={statesLoading ? __('Loading...', 'give') : ''}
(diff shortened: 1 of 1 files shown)


src/DonationForms/resources/registrars/templates/groups/BillingAddress.tsx
```diff
@@ -185,7 +187,7 @@ function StateFieldContainer({
                 ) : (
                     <input
                         type="text"
-                        value={''}
+                        value={state}
```
JoshuaHungDinh, line 190: I got this warning in the console. What are your thoughts on removing the value prop of the input all together to keep it uncontrolled? Also above this line is a conditional text input with a value of `loading..` I don't see an attempt to load values here. If we remove it as well the console warnings no longer display, states still can load when available and it functions as a text input when needed.

Screen Shot 2024-01-02 at 12 08 46 PM

pauloiankoski: Good catch @JoshuaHungDinh! I simplified that on 41310a61dc4b5135381674d3d602ae1f786e8b94.

The _Loading..._ text is displayed while checking/loading options based on the current country selection. In some cases, that input is replaced with a pre-populated dropdown. Before the commit above, since we were setting a value for that _Loading..._ text, we had to set something else later. I moved that text to be a placeholder in the field instead so that problem does not happen.

JoshuaHungDinh (approved):

src/DonationForms/resources/registrars/templates/groups/BillingAddress.tsx
JoshuaHungDinh: Great idea!

Everything seems to be working smoothly. I just noticed that we can remove line 69 (unused variable). Could you double-check that, and once confirmed, this will be all set for QA.

rickalday (approved):
Passed manual QA tests.

PR #7169, opened 2024-01-04
Feature: Design Settings

kjohnson:
Resolves GIVE-110

## Description

This PR introduces a new tab experience for the Design sidebar, allowing users to manage their form designs between two setting types `General` & `Styles`. It also introduces new design settings to the Donation Form, including Background Images, Branding Logos, Section & Input  field label Styles. 

To accomplish styling we are attaching top-level classnames to the root element. New styling is then achieved with the BEM naming convention Base styles have been applied and for some settings ex. HeaderImage its up to each form design to handle how they interpret and display the setting.

A new method  for subscribing to settings has been added to the dynamic app preview function. The main purpose is to handle attaching the various classnames to the dom for preview. For some of the pseudo elements the structure breaks when changing or deleting so additional logic was added to reconstruct the classnames in specific cases.

### WIP:
An additional framework will be included at a later date to handle opting in to specific styles.. At which time the cover header image, section and logo styling will also be included. 

Note: The "Cover" style for the background image requires that the `<Title/>` and `<Description/>` components be wrapped by a `<div/>` to align the text together. This is accomplished by the `<HeaderTextWrapper/>` component, which "collapses" to a `<Fragment/>` when the wrapping `<div/>` is not necessary. This maintains backwards compatibility with existing styles because the default behavior is not changed.

For both "Cover" and "Logo" the images are created with pseudo elements `:before` / `:after`& attached via css variables.

## Affects

Design settings / right sidebar

## Visuals

https://github.com/impress-org/givewp/assets/75056371/04015471-84c6-4e62-87f7-06fd8a59320e

https://github.com/impress-org/givewp/assets/75056371/8da3a7f3-b978-40f3-9950-59d044292d26

https://github.com/impress-org/givewp/assets/75056371/a01a24aa-ad7b-46ec-aa84-6b02fd5a3115

https://github.com/impress-org/givewp/assets/75056371/d693a888-f801-4484-a87a-b9c355e7fbab

## Testing Instructions

- The new design sidebar. It should be able to switch between tabs, as well as open and close according to the various controls in the build settings environment.
- Test each form header image style. : above, center and background for both MultiStep & Classic form designs. While testing this be sure to test with the other header styles to see if the layout breaks. Also test uploading, removing and switching between images.
- Test both input styles for MultiStep & Classic form designs. This includes the active styling applied when an input field is selected and the error styles applied when submitting a form without a required field.


kjohnson:
Build for QA, see https://github.com/impress-org/givewp/actions/runs/7615998108.

kjohnson:
Build 2 for QA, see https://github.com/impress-org/givewp/actions/runs/7701281813

JoshuaHungDinh (approved):
Looks good.

kjohnson:
Build 3 for QA, see https://github.com/impress-org/givewp/actions/runs/7716725040.

rickalday (approved):
Passed manual QA tests.

PR #7166, opened 2024-01-03
Feature: add Two Panel (steps) option to Form design drop-down

jonwaldstein:
Resolves [GIVE-153]

## Description

This adds a new "Two Panel (steps)" option to the form layout dropdown list.  Currently this does not include any styles, just the option and is equivalent to multi-step.

## Affects

- Form design dropdown

## Visuals

Screenshot 2024-01-09 at 2 12 58 PM

## Testing Instructions

- Try selecting "Two Panel (steps)" from the form design list.

## Pre-review Checklist

-   [ ] Acceptance criteria satisfied and marked in related issue
-   [x] Relevant `@unreleased` tags included in DocBlocks
-   [ ] Includes unit tests
-   [ ] Reviewed by the designer (if follows a design)
-   [ ] [Self Review](https://give.gitbook.io/development-manual/devops/github/code-reviews#self-review) of code and UX completed

[GIVE-153]: https://stellarwp.atlassian.net/browse/GIVE-153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7173, opened 2024-01-05
Feature: style Two-Column layout

pauloiankoski:
Resolves [GIVE-200]

Depends on https://github.com/impress-org/givewp/pull/7166

## Description

This pull-request styles the new "Two column" form layout. Also, performs some adjusts to the form builder canvas when in the design tab.

## Affects

Two-Column layout and Design Tab in the Form Builder

## Visuals

CleanShot 2024-01-05 at 19 09 36
_Form Builder_

CleanShot 2024-01-05 at 19 08 58
_Desktop_

CleanShot 2024-01-05 at 19 11 16
_Tablet_

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-200]: https://stellarwp.atlassian.net/browse/GIVE-200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JasonTheAdams:
@pauloiankoski Can we please use this opportunity to make the canvas take the full width available in design mode and center the form? It looks odd to have a large blank space on the left that we know will never be filled with a sidebar? The width is a bit an artifact at this point.

jonwaldstein:
@JasonTheAdams while I agree thats probably a better UX, making that change would affect the entire form builder and comes with a lot of side effects (we've tried to do this before).  That makes it out of scope for this project which is focused on just building a new form layout.

JasonTheAdams:
@jonwaldstein Bah! Fine! 😆

tantrum

pauloiankoski:
@JasonTheAdams A change specific for the design tab is not a big deal tho. However, I kinda feel it unbalanced:

Also, when switching back to the Build tab, the form canvas will be in a different position which looks weird:

CleanShot 2024-01-05 at 19 27 30

JasonTheAdams:
Mmm, thanks for the gif, @pauloiankoski. That's really helpful. So is the form currently right-aligned in design mode such that it can continue to extend left? Or is there some weird invisible boundary?

I'm curious what @jdghinson thinks, too.

pauloiankoski:
Currently, the form has a fixed width and is centered since the sidebar is floating with `position: fixed`. In that GIF, I changed the canvas width to be 100% - `var(--givewp-sidebar-width)` when the sidebar is present so that the canvas ends where the sidebar starts.

JasonTheAdams:
Got it. Thanks for playing with that, Paulo. I'll agree with Jon on this and say let's clean that up in a subsequent PR if @jdghinson feels it's merited. Ideally, it wouldn't be a fixed with and would fill the available space. But that's not important for this PR.

JasonTheAdams:
@jonwaldstein

you win this time

src/DonationForms/resources/app/DonationFormApp.tsx
```diff
@@ -59,8 +59,10 @@ function App({form}: { form: DonationForm }) {
         return (
             <DonationFormSettingsProvider value={{...form.settings,...donationFormNodeSettings}}>
                 <DonationFormStateProvider initialState={initialState}>
-                    {!form.design?.includeHeaderInMultiStep && form.settings.showHeader && <Header form={form} />}
-                    <MultiStepForm form={form} />
+                    <div className="givewp-donation-form__container">
```
jonwaldstein, line 62: There is already a div that wraps the header and form with a classname "givewp-donation-form" - is there any compelling reason to add this as well or could you achieve the styling with the existing div?

src/DonationForms/FormDesigns/TwoPanelStepsFormLayout/css/_sections.scss
```diff
@@ -0,0 +1,16 @@
+.givewp-donation-form__steps {
+    @media screen and (max-width: 71.5rem) {
```
jonwaldstein, line 2: Where does this `71.5rem` value come from?

pauloiankoski: That's the space necessary for the two columns plus extra margins:
`0.5rem (form pad) + 35rem (col1) + 0.5rem (gap) + 35rem (col2) + 0.5rem (form pad) = 71.5rem`

src/DonationForms/resources/app/DonationFormApp.tsx
pauloiankoski: I needed that for responsive as with the detached header, I had to created a attached-like style for the form and I couldn't achieve that using the form div because I also needed an extra div for the form padding.

src/DonationForms/FormDesigns/TwoPanelStepsFormLayout/css/_sections.scss
jonwaldstein: oh interesting, so is that basically your calculation for when to break for tablet?

pauloiankoski: Exactly. If we don't have space enough for the two columns, so break it in a single one.

src/DonationForms/resources/app/DonationFormApp.tsx
jonwaldstein: Okay, my main hesitation is (although it doesn't seem like much) it's _technically_ a breaking change for styles because a customer could be using custom css and targeting the current structure.  I think it would be worth a little exploring if we can achieve with css grid - otherwise this would be an update I would want to make to default form as well so the base structure is the same across forms layouts.

pauloiankoski: I can remove that container then rely on the media queries to manage the margins. I'll let you know when that's ready for review.

pauloiankoski: @jonwaldstein refactored that on 582746161c232c167310ed781f91573a01188088

jonwaldstein: Great, i think that is a safer approach for now, thanks!

jonwaldstein (approved):
Looks good, nice job!

PR #7171, opened 2024-01-04
Refactor: default multi-step form layouts will have header detached from form

jonwaldstein:
Resolves [GIVE-201]

## Description

This updates the default multi-step enabled form layouts to have the header _detached_ from the form. To include the header as the first step, form layouts will now have an opt-in property on their class called: `$includeHeaderInMultiStep`.  

This approach will allow more flexibility for form layouts to display and style both the header and multi-step forms.

## Affects

- Currently this affects the existing multi-step form design (_under-the-hood_) however everything should work as expected.

## Visuals

Screenshot 2024-01-05 at 10 34 52 AM

### With `$includeHeaderInMultiStep` as false

Screenshot 2024-01-04 at 2 41 29 PM

## Testing Instructions

- Pull branch and build assets
- Test the existing multi-step form - especially with the "show header" setting
- If you want to get fancy - programatically update the `$includeHeaderInMultiStep` to false in `MultiStepFormDesign` and see the default header positioning.

## Pre-review Checklist

-   [ ] Acceptance criteria satisfied and marked in related issue
-   [x] Relevant `@unreleased` tags included in DocBlocks
-   [ ] Includes unit tests
-   [ ] Reviewed by the designer (if follows a design)
-   [ ] [Self Review](https://give.gitbook.io/development-manual/devops/github/code-reviews#self-review) of code and UX completed

[GIVE-201]: https://stellarwp.atlassian.net/browse/GIVE-201?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7180, opened 2024-01-10
Epic: Two Panel (steps) form layout

jonwaldstein:
Resolves [GIVE-150]

## Description

This epic adds a Two Panel (steps) form layout to the visual form builder.

Includes the following PRs:
- #7171
- #7166
- #7173
- #7184
- #7192 
- #7258

## Affects

- Existing multi-step forms (under the hood)
- New form design/layout 

## Visuals

Screenshot 2024-01-15 at 3 33 23 PM

## Testing Instructions

**New layout**

- Create a v3 form
- Go to design mode and select the new "Two Column (steps)" form design/layout
- Publish and view the form
- The layout should match the design

**Existing multi-step layout**
- Make sure the existing multi-step layout works as expected

**"Header" settings**
- On multi-step and two panel make sure the show heading settings works as expected
- Verify the Header Image works as expected

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-150]: https://stellarwp.atlassian.net/browse/GIVE-150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JasonTheAdams:
@jonwaldstein @pauloiankoski This is what I'm seeing when I select the layout. 😬 

_ GiveWP Test _ WordPress

JasonTheAdams:
@jdghinson I'm not loving the card layout for this — i.e. the fact that the header just looks like a card beside the form. They don't look connected in any way, which doesn't as clearly associate them to me. It feels like two completely separate things. It also doesn't look good without an image or long description in the header — but I'm guessing the design settings introduces that?

cc: @jonwaldstein

jonwaldstein:
@JasonTheAdams what you are seeing in the builder preview is the current expected behavior.  It's a limitation of the form builder preview and is something we have been talking about improving (this is part of the whole VB prep project).  That is not something we can solve for this specific project as we are only introducing a new layout to an existing framework.  Although, I agree the form builder's preview experience lacks some good UX, the most important part is what happens on the donor facing front-end so we have to pick our battles.

jonwaldstein:
@JasonTheAdams also confirming the header image field is being included in design settings - this layout is not responsible for adding that setting.

JasonTheAdams:
> @JasonTheAdams what you are seeing in the builder preview is the current expected behavior. It's a limitation of the form builder preview and is something we have been talking about improving (this is part of the whole VB prep project). That is not something we can solve for this specific project as we are only introducing a new layout to an existing framework. Although, I agree the form builder's preview experience lacks some good UX, the most important part is what happens on the donor facing front-end so we have to pick our battles.

@jonwaldstein I totally get what you're saying. To be honest, I'm not comfortable releasing this feature like this. It just looks bad, and I don't want that to be someone's impression. They don't know we're going to follow up and fix it.

jonwaldstein:
@JasonTheAdams i'm assuming you don't like the overlap of the sidebar right? @jdghinson and I were just discussing this and agree that should be improved along with the fluidity of the canvas - more like the native WP editor. However, that's an existing experience right now - and frankly has been that way since we launched 3.0.  If you are looking on a normal sized browser that's not in full screen mode - which would be a common scenario - you are going to get a sidebar overlap.  The good news is you can also toggle the sidebar to close it - it's not the best experience but it does offer a way to view the entire preview.

I think releasing this prior to form builder improvements would be a call @angelablake should make.  I personally think people would be more excited to use this on their website right now (as requested) and won't care as much that the sidebar overlaps inside the form builder preview because the majority of value is what donors actually see and use - not necessarily the admin.

Screenshot 2024-01-16 at 4 32 29 PM

angelablake:
@JasonTheAdams I'm not sure I understand what doesn't look right. To me, this looks exactly like [the design](https://www.figma.com/file/T1w9MaCGJ6Z1UTVp4ek4WJ/New-donation-form?type=design&node-id=1179-5117&mode=design&t=urxW8cjDEUrwTFi0-0), just without the header image, which will be released with Design Settings. 

Design Settings are supposed to release before the Two Panel layout, so we should have a chance to see what it will really look like before releasing. However, we could consider adding a default header image, so it doesn't just look like a card before the admin adds an image themselves. @jdghinson may have already accounted for that.

The only other difference is that the design was done on a white background, and the Visual Builder's canvas has a gray background, which does make the two columns look more "separate." Not sure how much we want to try and control background color.

JasonTheAdams:
> @JasonTheAdams i'm assuming you don't like the overlap of the sidebar right? @jdghinson and I were just discussing this and agree that should be improved along with the fluidity of the canvas - more like the native WP editor. However, that's an existing experience right now - and frankly has been that way since we launched 3.0. If you are looking on a normal sized browser that's not in full screen mode - which would be a common scenario - you are going to get a sidebar overlap. The good news is you can also toggle the sidebar to close it - it's not the best experience but it does offer a way to view the entire preview.
> 
> I think releasing this prior to form builder improvements would be a call @angelablake should make. I personally think people would be more excited to use this on their website right now (as requested) and won't care as much that the sidebar overlaps inside the form builder preview because the majority of value is what donors actually see and use - not necessarily the admin.

I understand and appreciate that technically this bug has been in place for a while — i.e., the form builder is not properly responsive. That said, we've not had any form layouts as wide as this, so on a typical laptop the design mode has looked fine. But now, on my 16" screen at full width, and even if I go full screen, it just looks bad with the design settings going over the form.

I suppose I'm channeling my inner Devin here when I say that I'm just not a fan of the initial impression this gives the user. It would be one thing if this was tucked away, but it's literally the first thing they see when making a new two-panel form. It is Angela's call, but I wanted to be transparent that I personally wouldn't release it like this.

JasonTheAdams:
> @JasonTheAdams I'm not sure I understand what doesn't look right. To me, this looks exactly like [the design](https://www.figma.com/file/T1w9MaCGJ6Z1UTVp4ek4WJ/New-donation-form?type=design&node-id=1179-5117&mode=design&t=urxW8cjDEUrwTFi0-0), just without the header image, which will be released with Design Settings.
> 
> Design Settings are supposed to release before the Two Panel layout, so we should have a chance to see what it will really look like before releasing. However, we could consider adding a default header image, so it doesn't just look like a card before the admin adds an image themselves. @jdghinson may have already accounted for that.
> 
> The only other difference is that the design was done on a white background, and the Visual Builder's canvas has a gray background, which does make the two columns look more "separate." Not sure how much we want to try and control background color.

@angelablake You're totally right that it looks like the design. Admittedly, I haven't seen it, yet. I'm not sure if I was supposed to. I'm cool with it going out like this, I just wanted to be honest that I think it would look better as a single frame with a separator, rather than two side-by-side cards with a gap in between that the background color pokes through.

angelablake:
@JasonTheAdams Ah, I thought you saw the Two Panel designs. I get what you're saying, but I think we can release the current design and look out for customer feedback.

If the canvas responsiveness is the main hold back for release, I'd like to know what kind of time and effort that would involve before forming a strong opinion. @jonwaldstein is that something that could easily be fit in alongside development of the second Two Panel design (tabs)? We're still waiting on Design Settings anyway ... how long of a delay might we be looking at if we wanted to fix the canvas responsiveness as well?

JasonTheAdams:
@jonwaldstein I'm betting this hasn't been tested out with the modal display mode of the form block/shortocde. When working on the Illuminate site I noticed that the modal is not responsive, and that the iframe has an inline style of `width: 32.5rem`. I get the sense we could never get the width resizing working properly (looks like @JoshuaHungDinh built it out). This is going to be a challenge with the modal for different layout sizes if we don't figure it out.

rickalday (approved):
Passed manual QA tests.

PR #7186, opened 2024-01-16
Feature: add Gift Aid migration step

pauloiankoski:
Resolves [GIVE-225]

## Description

This pull request adds a migration step to migrate Gift Aid data from a v2 to a v3 form.

## Affects

Donation Form migration

## Testing Instructions

1. Create a v2 form with Gift Aid
2. Migrate it to v3
3. The new form must have the Gift Aid block with all previous settings

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-225]: https://stellarwp.atlassian.net/browse/GIVE-225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/FormMigration/FormMetaDecorator.php
```diff
@@ -684,4 +683,29 @@ private function isDefaultFund(int $fundId): bool
...
+    public function getGiftAidSettings(): array
+    {
+        $giftAidStatus = $this->getMeta('give_gift_aid_enable_disable');
+
+        if (empty($giftAidStatus) || ($giftAidStatus !== 'enabled' && $giftAidStatus !== 'global')) {
+            return [];
+        }
+
+        return [
+            'useGlobalSettings' => $giftAidStatus === 'global',
```
kjohnson, line 699: I would caution against bundling like this in the `FormMetaDecorator`. These methods are intended to be human readable abstractions of the technical implementation and to decouple the shape of the data from how it is stored and used.

A few things to consider as a side effect of bundling the settings:
- Couples the block attributes shape with the database storage implementation
- Creates an un-typed return array which is prone to creating bugs when refactoring
- Returns different data depending on the value of the Gift Aid status

kjohnson: I know it may seem like extra, unnecessary work to create the many methods, but it has advantages in future maintenance (which is always harder than the initial write).

kjohnson (changes requested):

src/FormMigration/FormMetaDecorator.php
pauloiankoski: Refactored on https://github.com/impress-org/givewp/pull/7186/commits/479836b7c882cbdff8709affdc044d4d9dd485a3

src/FormMigration/FormMetaDecorator.php
```diff
@@ -684,4 +683,86 @@ private function isDefaultFund(int $fundId): bool
...
+    {
+        return __('Tell me more', 'give-gift-aid');
+    }
+
+    /**
+     * @unreleased
+     */
+    public function getGiftAidModalHeader(): string
+    {
+        return __('What is Gift Aid?', 'give-gift-aid');
```
kjohnson, line 734: If these are static strings (possibly new settings?) and not meta of the form then they can be added directly to the migration step process method.

kjohnson (approved):
One nitpick for the Modal Header and Link Text attributes, but other than that this is ready for QA. Thanks for making those changes in the FormMetaDecorator.

src/FormMigration/FormMetaDecorator.php
pauloiankoski: Refactored on https://github.com/impress-org/givewp/pull/7186/commits/26e3e881070eb1f484126a2ebd149a8ee4edefe7

rickalday (approved):
Passed manual QA tests.

PR #7194, opened 2024-01-19
Fix: Remove superfluous allow_url_fopen check to avoid critical error from falsy value

JeromeBeckett:
## Issue
On PHP >= 8.0, when `allow_url_fopen` is `0`, a critical error blocks the rest of the page loading on the System Info page after the TLS values.

On Tools > System Info:
"TSL Connection" shows "Connection uses" (blank)
"TLS Rating" shows "There has been a critical error on this website..."

## Description
If the config value is false, `$tls_check` remains false, and then `json_decode()` fails, returning `null`. Then `property_exists()` is called with the null value, which triggers the critical error.

I've simply added checks to see if `$tls_check` is false, and if so, display text prompting the user to enable `allow_url_fopen` instead of attempting to decode the non-existent body text.

I considered attempting to use cURL as a fallback if the user has it enabled but figured this may be overkill. Also I think the fallback message is appropriate but we may want to amend the wording.

## Affects
Avoids error on System Info page.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/includes/admin/tools/views/html-admin-page-system-info.php
+++ b/includes/admin/tools/views/html-admin-page-system-info.php
@@ -207,12 +207,14 @@ $give_updates = Give_Updates::get_instance();
 				$tls_check = wp_remote_get( 'https://www.howsmyssl.com/a/check' );
 			}
 
-			if ( ! is_wp_error( $tls_check ) ) {
-				$tls_check = json_decode( wp_remote_retrieve_body( $tls_check ), false );
+            if ( ! $tls_check ) {
+                echo __( 'Please enable' ) . ' <b>allow_url_fopen</b> ' . __( 'in your php.ini file to determine this value.' );
+            } else if ( ! is_wp_error( $tls_check ) ) {
+                $tls_check = json_decode( wp_remote_retrieve_body( $tls_check ), false );
 
-				/* translators: %s: SSL connection response */
-				printf( __( 'Connection uses %s', 'give' ), esc_html( $tls_check->tls_version ) );
-			}
+                /* translators: %s: SSL connection response */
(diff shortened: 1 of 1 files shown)


jonwaldstein:
@JeromeBeckett thanks for the PR! our team will review this.

@alaca can you please review this one?

alaca (changes requested):
Hi Jerome, thank you for the PR. They are much appreciated. 
I reviewed our problem here, and from what I can tell, there is a logical problem in our code. Mainly, setting the `$tls_check` to `false`, and then checking if `$tls_check` is an instance of `WP_Error` using the `is_wp_error` function. That will fail because `$tls_check` will always be `false` if `allow_url_fopen` is disabled. The `wp_remote_get` function uses Curl anyway, so it will work regardless of the `allow_url_fopen` value set in PHP.ini. 

The solution would be to remove the conditional check for the `allow_url_fopen` value and set the `$tls_check` to whatever is returned from `wp_remote_get`. 

Example: 

```php
			$tls_check = wp_remote_get( 'https://www.howsmyssl.com/a/check' );

			if ( ! is_wp_error( $tls_check ) ) {
				$tls_check = json_decode( wp_remote_retrieve_body( $tls_check ), false );

				/* translators: %s: SSL connection response */
				printf( __( 'Connection uses %s', 'give' ), esc_html( $tls_check->tls_version ) );
			}
```

JeromeBeckett:
Thanks for reviewing the request, @alaca. I did not know `wp_remote_get` uses cURL, that makes things easier. I have applied the suggested changes.

alaca (approved):
Looks good :+1: 

One more thing, could you please rename the PR title to something that better describes the problem you solved here? Thank you!

JeromeBeckett:
No problem - updated.

PR #7204, opened 2024-01-25
Fix: Update nonce validation to match when translated

Genevieve-K:
As an administrator, I would like to update donations with a bulk action. The result is:  
`We're unable to recognize your session. Please refresh the screen to try again; otherwise contact your website administrator for assistance. `

The nonce verification failed.
In class-wp-list-table.php, the nonce is created from `'bulk-' . $this->_args['plural']`. In French this plural is "formulaires". And the verification uses `'bulk-forms'`.

## Testing Instructions
Test bulk actions  in other languages as English

### DIFF
--- a/includes/admin/payments/class-payments-table.php
+++ b/includes/admin/payments/class-payments-table.php
@@ -812,7 +812,7 @@ class Give_Payment_History_Table extends WP_List_Table {
             return;
         }
 
-        give_validate_nonce($_GET['_wpnonce'] ?? '', 'bulk-forms');
+        give_validate_nonce($_GET['_wpnonce'] ?? '', 'bulk-'.$this->_args['plural']);
 
         foreach ($ids as $id) {
 			// Detect when a bulk action is being triggered.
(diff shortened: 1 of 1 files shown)


Genevieve-K:
There is also a nonce check in admin-actions.php. But here `$this->_args['plural'] `is not none.  We have to find another solution. The result is that the notice is not displayed.

kjohnson:
@Genevieve-K I think it would be best if the nonce verification was not dependent on the translation. Could you instead update the localized nonce to be `bulk-forms` untranslated?

Genevieve-K:
Yes it's better. I didn't understand that the action is defined in class-payments-table.php.

kjohnson (approved):
From what I can tell, the `plural` argument passed to the [WP_List_Table](https://developer.wordpress.org/reference/classes/wp_list_table/) class is used as a class name on the table and as part of the nonce for bulk actions. Changing this value to be a non-translatable string **should not** result in any untranslated labels, so we should be good here.

```
*     @type string $plural   Plural value used for labels and the objects being listed.
*                            This affects things such as CSS class-names and nonces used
*                            in the list table, e.g. 'posts'. Default empty.
```

kjohnson:
Build for QA, see https://github.com/impress-org/givewp/actions/runs/7728329239.

rickalday (approved):
Passed manual QA tests.

PR #7207, opened 2024-01-29
Feature: add event handler classes and Action Scheduler helper class

glaubersilva:
## Description

This PR adds a few classes that usually get repeated through different gateway add-ons. The main goal is preventing code duplication through different projects as occurs currently.

To archive it, the following classes were added to the core:

1) Event handler classes that are commonly used in gateway integrations. Usually, these classes are used to handle async events (Action Scheduler async actions) created when the gateways send webhook notifications. Also, unit tests were added for each one of these classes. For more details, check out this comment: https://github.com/impress-org/givewp/pull/7207#discussion_r1517687589

_*** A similar approach already was used previously in the Recurrion Add-on to handle async subscription events, you can check it here:_ https://github.com/impress-org/give-recurring/tree/develop/src/PaymentGatewayModules/EventHandlers

**Edit:** the subscription event handlers from the link above were moved to Core through this PR. For more details, check out this comment: https://github.com/impress-org/givewp/pull/7207#discussion_r1516218907

2) A helper class called `AsBackgroundJobs` was added to encapsulate common functions that use Action Scheduler under the hood to handle async jobs. Also, the `vendor/woocommerce/action-scheduler` was added in the `scanDirectories` of the `phpstan.neon` file to prevent errors in the Static Analysis GitHub Actions related to missing Action Scheduler functions. 

## Affects

It will let the gateway integrations faster and prevent code duplication through different add-ons.

## Testing Instructions

You can run the unit tests or test it through a gateway implementation that uses all these classes, like this one: https://github.com/impress-org/give-mollie/pull/47

## Pre-review Checklist

-   [ ] Includes unit tests


tests/Unit/PaymentGateways/EventHandlers/DonationCancelledTest.php
```diff
@@ -0,0 +1,43 @@
...
+class DonationCancelledTest extends TestCase
+{
+    use RefreshDatabase;
+    
+    /**
+     * @unreleased
+     *
+     * @throws Exception
+     */
+    public function testShouldSetStatusToCancelled()
```
jonwaldstein, line 24: I love that you added unit tests! Let's keep this going and make these complete by asserting the opposite of what is expected like `testShouldNotSetStatusToCancelled` or more specific like `testShouldNotSetStatusToCancelledWhenRecurring`.  

Since each event handler has additional logic to prevent the status from being updated - we want to add tests to make sure those scenarios are working properly 😄

src/Helpers/AsBackgroundJobs.php
```diff
@@ -0,0 +1,124 @@
...
+namespace Give\Helpers;
+
+use ActionScheduler;
+
+/**
+ * @see https://actionscheduler.org/
+ *
+ * @unreleased
+ */
+class AsBackgroundJobs
```
jonwaldstein, line 12: Can we find another place to put this instead of the generic helpers file?  Maybe we add a new folder in framework called `ActionScheduler` where we put stuff like this.

src/PaymentGateways/Actions/UpdateDonationStatus.php
```diff
@@ -0,0 +1,121 @@
...
+use Give\Donations\Models\Donation;
+use Give\Donations\Models\DonationNote;
+use Give\Donations\ValueObjects\DonationStatus;
+use Give\Framework\PaymentGateways\Log\PaymentGatewayLog;
+use Give\Subscriptions\ValueObjects\SubscriptionStatus;
+
+/**
+ * @unreleased
+ */
+class UpdateDonationStatus
```
jonwaldstein, line 15: Let's actually move this file and all the event handler files into the Framework/PaymentGateways folder as these are more of framework level actions.  The top level PaymentGateways directory is more about the specific gateways and the one inside framework is for shared logic.

src/Helpers/AsBackgroundJobs.php
```diff
@@ -0,0 +1,124 @@
...
+     *
+     * @param string $hook     The hook to trigger.
+     * @param array  $args     Arguments to pass when the hook triggers.
+     * @param string $group    The group to assign this job to.
+     * @param bool   $unique   Whether the action should be unique.
+     * @param int    $priority Lower values take precedence over higher values. Defaults to 10, with acceptable values falling in the range 0-255.
+     *
+     * @return int The action ID. Zero if there was an error scheduling the action.
+     */
+    public static function enqueueAsyncAction(
```
jonwaldstein, line 25: Just looking at how we're using this, it might make more sense to create a facade, then all the methods will be static by default but can still be mockable

jonwaldstein (changes requested):

tests/Unit/PaymentGateways/EventHandlers/DonationCancelledTest.php
glaubersilva: Resolved in 45a24096bac0d5515d8ffdbc7bfa80de2ba5ebe2

src/PaymentGateways/Actions/UpdateDonationStatus.php
glaubersilva: Resolved in 22eb283dd0d7805e1d4b43dea2b2da149f9ac5b8

The funny thing is that I started to develop it inside the framework folder, and then I changed it to keep similar to the recurring addon where similar classes live here: https://github.com/impress-org/give-recurring/tree/feature/mollie-submodule-GIVE-297/src/PaymentGatewayModules/EventHandlers

So, this makes me wonder if we should move these classes from the Recurring add-on to the Framework folder as well. This way, we can provide in the framework all necessary event handlers to implement a gateway, which can be useful for 3rd party integrations.

src/Helpers/AsBackgroundJobs.php
glaubersilva: Resolved in 75665bf1789d8c7801602105ea20f3b896b05c43 and 0efdd386c483557d1cd5f41ce61b2e565158947f

src/Helpers/AsBackgroundJobs.php
glaubersilva: Resolved in 0efdd386c483557d1cd5f41ce61b2e565158947f

glaubersilva:

After my last updates, this PR started to get some Static Analysis errors related to this  code line: https://github.com/impress-org/givewp/blob/develop/src/ServiceProviders/LegacyServiceProvider.php#L5

But I didn't touch on this, so I am not sure if has some relation with the facade implementation I did, maybe I'm missing something here.

src/PaymentGateways/Actions/UpdateDonationStatus.php
jonwaldstein: @glaubersilva I like that idea! let's pull in the subscription event handlers into the framework folder as you suggested.

glaubersilva: Great! I just moved all subscription event handlers and their proper tests to Core.

glaubersilva:
@jonwaldstein After merging [this other PR](https://github.com/impress-org/givewp/pull/7283) into this one, the static analysis errors are gone. Let me know if there is anything else.

src/Framework/PaymentGateways/EventHandlers/DonationCancelled.php
```diff
@@ -0,0 +1,34 @@
...
+class DonationCancelled
+{
+    /**
+     * @unreleased
+     *
+     * @throws Exception
+     */
+    public function __invoke(string $gatewayTransactionId, string $message = '', $skipRecurringDonations = false)
+    {
+        $donation = give(DonationRepository::class)->getByGatewayTransactionId($gatewayTransactionId);
```
jonwaldstein, line 22: It looks like some of the event handlers are using the DonationRepository class while others are using the service container name directly like `give()->donations` - I prefer the latter as its one less import - can we just tighten up that consistency please?

```suggestion
        $donation = give()->donations->getByGatewayTransactionId($gatewayTransactionId);
```

src/Framework/PaymentGateways/EventHandlers/DonationAbandoned.php
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * @unreleased
+ */
+class DonationAbandoned
+{
+    /**
+     * @unreleased
+     * @throws Exception
+     */
+    public function __invoke(string $gatewayTransactionId, string $message = '', $skipRecurringDonations = false)
```
jonwaldstein, line 18: I just noticed these event handlers only receive a `$gatewayTransactionId` rather than a Donation ID.  I think i lost sight of the context here and could use your help explaining:
1. What are we considering to be an `Event`?
2. Are EventHandlers intended to be used directly by action scheduler / webhook?

src/Framework/PaymentGateways/EventHandlers/DonationCancelled.php
glaubersilva: Resolved in 57775d55735b3629354c76be8883a0e991637568

src/Framework/PaymentGateways/EventHandlers/DonationAbandoned.php
jonwaldstein: @glaubersilva thanks for the clarification, i'm tracking now 😄 

Yes lets go with your suggestion and move these into a new `Webhooks` folder to give these EventHandlers more context - i'm sure there will be more webhook related functionality we can add to that folder in the future like validation helpers and url generators.

glaubersilva: Resolved in a5c3f58f4c06130591d0e3717b4e9b2c3b8fdde7

src/Framework/PaymentGateways/Actions/UpdateDonationStatus.php
```diff
@@ -0,0 +1,121 @@
...
+use Give\Donations\Models\Donation;
+use Give\Donations\Models\DonationNote;
+use Give\Donations\ValueObjects\DonationStatus;
+use Give\Framework\PaymentGateways\Log\PaymentGatewayLog;
+use Give\Subscriptions\ValueObjects\SubscriptionStatus;
+
+/**
+ * @unreleased
+ */
+class UpdateDonationStatus
```
jonwaldstein, line 15: @glaubersilva should we move these actions into the webhooks folder as well? they seem related to those as updating a status can already be done with the models.  It seems these status updates include notes and gateway specific information like transaction ID

glaubersilva: @jonwaldstein I agree, makes more sense! 

Resolved in 67561161c97f1945ba0fe253e6283e58f2328b61

jonwaldstein: thanks man!

jonwaldstein (approved):
@glaubersilva you are a rock star 🎸 ⭐

rickalday (approved):
Passed manual QA tests

PR #7209, opened 2024-01-29
Format donation count output in progress bar component

### LINKED ISSUE #7208: Multi-form goal block should have formatted donation number total
## User Story

As an admin, I want to ensure all frontend numbers have similar formatting to be consistent.

## Details

The Multi-goal Block currently does not format the "Donation" field

## Acceptance Criteria

-   [ ] The donation data is properly formatted

DevinWalker:
Resolves #7208 [GIVE-2050]

## Description

This resolves the issue with the "Donation" count total not being formatted properly. 

I'm using `give_format_amount()` for this, even though it's a total, not an amount because I didn't see any other generic non-currency-based way to do it. This seems to work fine, though.

## Affects

Progress bar component with donation counts.

## Visuals

*Before*

*After*

## Pre-review Checklist

-   [x] Includes unit tests

[GIVE-2050]: https://stellarwp.atlassian.net/browse/GIVE-2050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/MultiFormGoals/resources/views/progressbar.php
```diff
@@ -32,7 +32,7 @@
         </div>
         <div part="stat-count" class="give-progress-bar-block__stat">
             <div part="stat-count-value"><?php
-                echo esc_html($this->getDonationCount()); ?></div>
+                echo esc_html($this->getFormattedDonationCount()); ?></div>
```
jonwaldstein, line 35: I think we could simplify this by using [number_format](https://www.php.net/manual/en/function.number-format.php) directly instead of `give_format_amount`:

```suggestion
                echo esc_html(number_format($this->getDonationCount())); ?></div>
```

Screenshot 2024-02-01 at 4 08 35 PM

jonwaldstein: Hey @DevinWalker coming back 10 months later to this lol any response? 😄

kjohnson: Using `give_format_amount` still takes into account currency related settings.

image

jonwaldstein: I guess the question becomes, is that something we want to be taken into account to display the count/sum of is that really only useful for displaying currency amounts?

kjohnson: Grouping separators are still important for non-currency amounts.

It would probably be better to use the browser locale settings, but to @DevinWalker's point the `give_format_amount` is the closest function to re-use.

kjohnson: Perhaps more importantly, he is using `give_format_number` here so that it matches the other values, which are using the currency settings and not browser's locale.

So using the browser's locale would be correct in isolation, but matching the other values is probably more important in this case.

jonwaldstein: @kjohnson I agree, in v3 forms we should be using the browser `Intl.NumberFormat` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat#using_locales

For the purposes of v2 I think this is fine 👍

rickalday (approved):
Passed QA tests

PR #7213, opened 2024-01-31
Feature: Add event data modeling

kjohnson:
Resolves [GIVE-286](https://stellarwp.atlassian.net/browse/GIVE-286?atlOrigin=eyJpIjoiZjQ2NjRmNzM2MTIwNGU5N2ExNjQyYWY1YWU1NzgyZGMiLCJwIjoiaiJ9)

## Description

This PR registers Models, Repositories, Migrations, Factories, and REST Endpoints for Events.

## Affects

Adds a new `paginate` helper to the `ModelQueryBuilder` class.

[GIVE-286]: https://stellarwp.atlassian.net/browse/GIVE-286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[GIVE-289]: https://stellarwp.atlassian.net/browse/GIVE-289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/EventTickets/Actions/RegisterEventsPostType.php
pauloiankoski, line 1: One issue we face on Give is that Donations are currently a custom post type instead of being handled as an entity with our own custom tables. I believe we should move towards having Events as a custom table to prevent potential future issues.

Edit: I see we are creating the custom table later here. What's the purpose of this custom post type? Is it a leftover?

src/EventTickets/Migrations/CreateEventTicketsTable.php
```diff
@@ -0,0 +1,56 @@
...
+class CreateEventTicketsTable extends Migration {
+    /**
+     * @inheritdoc
+     */
+    public static function id() {
+        return 'give-events-create-events-tickets-table';
+    }
+
+    public static function title() {
+        return 'Create give_events table';
```
pauloiankoski, line 22: Maybe a small typo here. I believe you meant to say "_Create give_event_tickets table_" here.

src/EventTickets/Migrations/CreateEventsTable.php
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public function run() {
+        global $wpdb;
+
+        $table   = $wpdb->give_events;
+        $charset = DB::get_charset_collate();
+
+        $sql = "CREATE TABLE $table (
+			id INT UNSIGNED NOT NULL,
+			form_id INT NULL,
```
pauloiankoski, line 43: We don't need to assign forms to events because we currently have a one-to-many relationship, so it may not be necessary.

src/EventTickets/Migrations/CreateEventsTable.php
```diff
@@ -0,0 +1,57 @@
...
+			description TEXT NULL,
+			start_datetime DATETIME NULL,
+			end_datetime DATETIME NULL,
+			ticket_close_datetime DATETIME NULL,
```
pauloiankoski, line 47: What's the purpose of this column? I couldn't find any mention to it.

src/EventTickets/Migrations/CreateEventsTable.php
```diff
@@ -0,0 +1,57 @@
...
+			PRIMARY KEY  (id)
+		) $charset";
```
pauloiankoski, line 49: I believe we are missing columns like `created_at` and `updated_at` (same in the other new tables).

Jeffrey has added a Timezone field. Will we include it in the formatted DateTime object before saving? We will need to retrieve that information to display it in the field template.

src/EventTickets/Migrations/CreateEventTicketsTable.php
```diff
@@ -0,0 +1,56 @@
...
+        $charset = DB::get_charset_collate();
+
+        // id BINARY(16) DEFAULT (UUID_TO_BIN(UUID())),
+        $sql = "CREATE TABLE $table (
+			id INT UNSIGNED NOT NULL AUTO_INCREMENT,
+			event_id INT UNSIGNED NOT NULL,
+			ticket_type_id INT UNSIGNED NOT NULL,
+			date_created DATETIME NOT NULL,
+			PRIMARY KEY  (id)
+		) $charset";
```
pauloiankoski, line 48: I think we need to record which donation generated this ticket in order to retrieve the donor's information.

src/EventTickets/Routes/GetEventTicketTypeTickets.php
```diff
@@ -0,0 +1,85 @@
...
+    public function handleRequest(WP_REST_Request $request)
+    {
+        $tickets = EventTicket::findByTicketType($request->get_param('ticket_type_id'))
+            ->paginate(
+                $request->get_param('page'),
+                $request->get_param('per_page')
+            )
+            ->getAll();
+
+        return new WP_REST_Response(
```
pauloiankoski, line 76: The REST API for ListTable returns an instance of a ListTable object. This object includes headers, columns, and formatted values that can be printed by the React app. Unless these routes were not intended to be used on that.

pauloiankoski (commented):
Hey @kjohnson, I have some comments that I think we should discuss further at this initial stage.

src/EventTickets/Actions/RegisterEventsPostType.php
kjohnson: The database design here uses a Custom Post Type + Custom Table. This avoids the common pitfalls of using CPTs by joining on a single table (as opposed to using post meta which requires `n` joins).

This is similar to the direction that Peer-to-peer has recently taken to better support SEO and page customization. Events are similar to Campaigns (and posts) in that each has a public representation on the site - as opposed to tickets which are more like comments are to posts. For that reason, a CPT makes sense for an Event but not for Tickets or Ticket Types.

Further, not using a CPT for tickets or ticket types avoids table bloat (which is a problem with donations and form submissions).

src/EventTickets/Migrations/CreateEventTicketsTable.php
kjohnson: Good catch - should be "Create give_event_tickets table".

src/EventTickets/Migrations/CreateEventsTable.php
kjohnson: Good point - I'll remove the `form_id` property and column.

kjohnson: Resolved, see [8b39068](https://github.com/impress-org/givewp/pull/7213/commits/8b39068012f9ec06c177574ec2c5779f526531d4).

src/EventTickets/Migrations/CreateEventsTable.php
kjohnson: > Last date & time tickets will be available

We need a way to (optionally) disable ticket sales independently from the event date.

For example, "Tickets are available up until one week before the event."

src/EventTickets/Migrations/CreateEventsTable.php
kjohnson: I'll add the `created_at` and `updated_at` columns for the Tickets and Ticket Types tables (Events inherits this from the CPT).

Re: Timezone: I'm unsure how we should handle that and would rather add that later when we have a chance to better understand the best practice for handling timezones.

kjohnson: Resolved, see [a053c51](https://github.com/impress-org/givewp/pull/7213/commits/a053c51bf765277bfa85ce38cde81cfcd64d23cd).

src/EventTickets/Routes/GetEventTicketTypeTickets.php
kjohnson: Let's include these REST Endpoints as is for regular data lookup and I'll add List Table specific endpoints in a new PR.

src/EventTickets/Migrations/CreateEventTicketsTable.php
kjohnson: I'll add a `donation_id` property and column.

kjohnson: Resolved, see [381fbaa](https://github.com/impress-org/givewp/pull/7213/commits/381fbaa73e94fbd9e03d2f165d742caa477ad560).

src/EventTickets/Routes/GetEventTicketTypeTickets.php
kjohnson: See [GIVE-289](https://stellarwp.atlassian.net/browse/GIVE-289?atlOrigin=eyJpIjoiN2QyYmJhYzJjYTBmNGFjMWE3YTVkNGIyNjVjNDIxODciLCJwIjoiaiJ9).

src/EventTickets/Migrations/CreateEventsTable.php
pauloiankoski: In that sense, we could also have a start date: `sales_start_datetime` and `sales_end_datetime`. But, once this is not part of the current scope and could easily be added later, I'd be inclided to drop it for now and implement that column we really decide to have that feature. What do you think?

kjohnson: It is low cost to leave the column here as compared to introducing an additional database migration to add it. Since the "spec" for this feature is unclear let's just leave it and move on.

src/EventTickets/Actions/RegisterEventsPostType.php
kjohnson: I removed the Custom Post Type, see [f44ffe5](https://github.com/impress-org/givewp/pull/7213/commits/f44ffe519cf8adc4cf3a0c6ffeb9caed51059708).

src/EventTickets/Routes/GetEventTicketTypeTickets.php
JasonTheAdams, line 1: @kjohnson Maybe the namespace should be `RestRoutes` as not to confuse it with view routes?

src/EventTickets/Factories/EventFactory.php
```diff
@@ -0,0 +1,34 @@
...
+        $start_datetime = $this->faker->dateTimeThisYear('+6 months');
+
+        return [
+            'id' => wp_insert_post([
+                'post_type' => 'give_event',
+                'post_title' => $title,
+                'post_content' => '[]', // Empty block editor content for potential event page
+                'post_excerpt' => $description,
+                'post_status' => 'publish',
+            ]),
```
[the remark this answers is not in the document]
kjohnson: Yup - missed that in a previous change.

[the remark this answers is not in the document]
kjohnson: Resolved, see [641e4bd](https://github.com/impress-org/givewp/pull/7213/commits/641e4bd4fe5e989dcea897549d9ec392b98d32b6).

PR #7220, opened 2024-02-06
Feature: Add endpoint for Event List Table

kjohnson:
Resolves GIVE-289

### DIFF
--- a/src/EventTickets/Routes/GetEventsListTable.php
+++ b/src/EventTickets/Routes/GetEventsListTable.php
@@ -1,0 +1,143 @@
+<?php
+
+namespace Give\EventTickets\Routes;
+
+use Give\EventTickets\Models\Event;
+use WP_REST_Request;
+use WP_REST_Response;
+
+/**
+ * @unreleased
+ */
+class GetEventsListTable
+{
+    /**
+     * @var string
+     */
+    protected $endpoint = 'events-tickets/events/list-table';
+
+    /**
+     * @inheritDoc
+     */
+    public function registerRoute()
+    {
+        register_rest_route(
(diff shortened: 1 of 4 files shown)


src/EventTickets/Routes/GetEventsListTable.php
```diff
@@ -0,0 +1,143 @@
...
+     */
+    protected function formatColumnTitle(Event $event): string
+    {
+        return sprintf(
+            '<a href="%s" aria-label="%s">%s</a>',
+            admin_url("edit.php?post_type=give_forms&page=give-events&view=overview&id=$event->id"),
+            __('View event details', 'give'),
+            $event->title
+        );
+    }
```
pauloiankoski, line 128: Hey @kjohnson, are you just scaffolding the endpoint for the Events List Table? If so, we are good to go. Otherwise, we need to stick with the `ListTable` structure that separate some concerns and allows 3rd-party users to manage columns.

Check the files below:
```
Give\Donations\Endpoint\ListDonations.php
Give\Donations\ListTable\DonationsListTable.php
Give\Donations\ListTable\Columns\*.php
```

kjohnson: We don't need to introduce extensibility yet. A simple array mapper will work for now.

pauloiankoski (approved):
We are good to go for now, but we will need to revisit this later to ensure it matches the format of the other ListTable endpoints.

kjohnson:
> we will need to revisit this later to ensure it matches the format of the other ListTable endpoints.

It does match the same format as the other endpoints. The response returned from the server is in the same format, regardless of how the data is formed.

PR #7227, opened 2024-02-13
Feature: Add eager loading of related models

kjohnson:
## Description

This PR introduces a Eager Loader to support loading a models relationship with a single additional query (as opposed to an additional query per model).

Example:

```php
// Given 1 event this runs 2 queries.
// Given 10 events this runs 11 queries.
// Given 100 events this runs 101 queries.
$events = Event::query()->getAll();
foreach($events as $event);
    $events->tickets; // Additional query for each event
}
```

```php
// Given 1 events this runs 2 queries.
// Given 10 events this runs 2 queries
// Given 100 events  this runs 2 queries.
$events = Event::withTickets()->getAll();
foreach($events as $event);
    $events->tickets; // No additional query - already cached
}
```

## Affects

The existing `ModelQueryBuilder` is not affected. Instead the `EagerLoader` class decorates the underlying query builder instance.

## Visuals

```php
class Event extends Model
{
    protected $relationships = [
        'tickets' => Relationship::HAS_MANY,
    ];

    public static function withTickets()
    {
        return new EagerLoader(self::class, 'tickets', Ticket::class, 'event_id');
    }
```

## Testing Instructions

Add a `with*()` method to a Model class and query the model with an eager loaded relationship.

### DIFF
--- a/src/Framework/Models/EagerLoader.php
+++ b/src/Framework/Models/EagerLoader.php
@@ -1,0 +1,127 @@
+<?php
+
+namespace Give\Framework\Models;
+
+use ReflectionClass;
+
+/**
+ * @unreleased
+ */
+class EagerLoader
+{
+    /**
+     * @unreleased
+     * @var string
+     */
+    protected $modelClass;
+
+    /**
+     * @unreleased
+     * @var string
+     */
+    protected $relationshipKey;
+
+    /**
+     * @unreleased
+     * @var ModelQueryBuilder
+     */
+    protected $modelQuery;
+
+    /**
+     * @unreleased
+     * @var ModelQueryBuilder
+     */
+    protected $eagerLoadedQuery;
+
+    /**
+     * @unreleased
+     * @var string
+     */
+    protected $foreignKey;
+
+    /**
+     * @var mixed
+     */
+    protected $foreignAttribute;
+
+    public function __construct($modelClass, $relationshipKey, $eagerLoadedModelClass, $foreignKey, $foreignAttribute = null)
+    {
+        $this->modelClass = $modelClass;
+        $this->modelQuery = $modelClass::query();
+        $this->relationshipKey = $relationshipKey;
+        $this->eagerLoadedQuery = $eagerLoadedModelClass::query();
+        $this->foreignKey = $foreignKey;
+        $this->foreignAttribute = $foreignAttribute ?? $foreignKey;
+    }
+
+    /**
+     * @unreleased
+     */
+    public function __call($name, $arguments)
+    {
+        $this->modelQuery->$name(...$arguments);
+        return $this;
+    }
+
+    /**
+     * @unreleased
+     *
+     * This method wraps the `get()` method of the underlying ModelQueryBuilder.
+     * It uses the results to query the related models and pre-set the cachedRelations property.
+     */
+    public function get()
+    {
+        $model = $this->modelQuery->get();
+
+        $eagerLoadedModels = $this->eagerLoadedQuery
+            ->where($this->foreignKey, $model->id)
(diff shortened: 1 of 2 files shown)


src/Framework/Models/EagerLoader.php
```diff
@@ -0,0 +1,127 @@
...
+     * @var string
+     */
+    protected $foreignKey;
+
+    /**
+     * @var mixed
+     */
+    protected $foreignAttribute;
+
+    public function __construct($modelClass, $relationshipKey, $eagerLoadedModelClass, $foreignKey, $foreignAttribute = null)
```
jonwaldstein, line 47: Let's add types to these params and a docblock for the construct.

src/Framework/Models/EagerLoader.php
```diff
@@ -0,0 +1,127 @@
...
+    {
+        $this->modelClass = $modelClass;
+        $this->modelQuery = $modelClass::query();
```
jonwaldstein, line 50: I think we might need to make sure the model has the `query()` method in some way as it comes from interfaces like `ModelCrud` and `ModelReadOnly`

src/Framework/Models/EagerLoader.php
```diff
@@ -0,0 +1,127 @@
...
+        return $this;
+    }
+
+    /**
+     * @unreleased
+     *
+     * This method wraps the `get()` method of the underlying ModelQueryBuilder.
+     * It uses the results to query the related models and pre-set the cachedRelations property.
+     */
+    public function get()
```
jonwaldstein, line 72: Can you make sure the IDE is aware of the return types on these? Jason did something fancy with generic in the model query builder that worked pretty well.  Basically each get and getAll can either return model instance(s) or null

jonwaldstein (changes requested):
@kjohnson this is really cool!

JasonTheAdams:
When this is done, let's make sure it finds it's way to https://github.com/stellarwp/models

At some point, we should switch to using that library. We would just have to also swap out our DB/QueryBuilder at the same time since they're dependent.

src/Framework/Models/EagerLoader.php
kjohnson: I think I figured out the class string typing.

src/Framework/Models/EagerLoader.php
kjohnson: I decided to check against the `Model::class` as a sub-class.

src/Framework/Models/EagerLoader.php
kjohnson: Yeah, I was able to replicate what Jason did.

src/Framework/Models/EagerLoader.php
kjohnson: Resolved, see https://github.com/impress-org/givewp/pull/7227/commits/69707995eddb9f276cc69aa9c51272f72cc207e4.

src/Framework/Models/EagerLoader.php
kjohnson: Resolved, see https://github.com/impress-org/givewp/pull/7227/commits/69707995eddb9f276cc69aa9c51272f72cc207e4.

src/Framework/Models/EagerLoader.php
kjohnson: Resolved, see https://github.com/impress-org/givewp/pull/7227/commits/69707995eddb9f276cc69aa9c51272f72cc207e4.

jonwaldstein (approved):
@kjohnson thanks for the updates, looks good to me - nice job!

PR #7233, opened 2024-02-15
Refactor: Convert entity properties to camel case

pauloiankoski:
## Description

The main goal of this pull request is to convert entity properties to camel case.

## Affects

Models, Repositories, Factories and Migrations of Event, EventTickets and EventTicketTypes

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/EventTickets/Models/EventTicketType.php
+++ b/src/EventTickets/Models/EventTicketType.php
@@ -30,7 +30,7 @@ class EventTicketType extends Model implements ModelCrud /*, ModelHasFactory */
         'label' => 'string',
         'description' => 'string',
         'price' => Money::class,
-        'totalTickets' => 'int',
+        'maxAvailable' => 'int',
         'createdAt' => DateTime::class,
         'updatedAt' => DateTime::class,
     ];
--- a/src/EventTickets/Models/EventTicketType.php
+++ b/src/EventTickets/Models/EventTicketType.php
@@ -106,7 +106,7 @@ class EventTicketType extends Model implements ModelCrud /*, ModelHasFactory */
     /**
      * @unreleased
      *
-     * @return ModelQueryBuilder<Event>
+     * @return ModelQueryBuilder<EventTicketType>
      */
     public static function query(): ModelQueryBuilder
     {
@@ -131,7 +131,7 @@ class EventTicketType extends Model implements ModelCrud /*, ModelHasFactory */
      */
     public function eventTickets(): ModelQueryBuilder
     {
-        return give(EventTicketRepository::class)->queryByEventId($this->id);
+        return give(EventTicketRepository::class)->queryByEventId($this->eventId);
     }
 
     /**
--- a/src/EventTickets/Models/EventTicketType.php
+++ b/src/EventTickets/Models/EventTicketType.php
@@ -29,13 +29,13 @@ class EventTicketType extends Model implements ModelCrud /*, ModelHasFactory */
      */
     protected $properties = [
         'id' => 'int',
-        'event_id' => 'int',
+        'eventId' => 'int',
         'label' => 'string',
         'description' => 'string',
         'price' => Money::class,
-        'max_available' => 'int',
-        'created_at' => DateTime::class,
-        'updated_at' => DateTime::class,
+        'totalTickets' => 'int',
+        'createdAt' => DateTime::class,
+        'updatedAt' => DateTime::class,
     ];
 
     /**
@@ -123,7 +123,7 @@ class EventTicketType extends Model implements ModelCrud /*, ModelHasFactory */
      */
     public function event(): ModelQueryBuilder
     {
-        return give(EventRepository::class)->queryById($this->event_id);
+        return give(EventRepository::class)->queryById($this->eventId);
     }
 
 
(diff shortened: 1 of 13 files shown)


kjohnson (changes requested):
Be careful not to include so many changes under the title of "convert entity properties to camel case" when converting the entity properties to camel case is a minor change of the pull request.

> Renaming max_available to totalTickets to better indicate that it represents the total amount of tickets available before any sales start.

Let's stick with `maxAvailable` because we need to differentiate between how many are available, how many are currently available, and how many have been sold. The person managing the event may think of it as "total tickets" when creating the event, but as a technical concept we need to separate this concept from the total tickets sold. The "total" is ambiguous and is used to describe multiple concepts in this feature.

> Removing ticket_close_datetime as there are no plans to have that data in the near future. The functionality will be controlled by endDateTime and manual disabling. Adding a column that won't be used seems unnecessary. We can consider adding it in the future if needed while keeping our feature clean.

No final decision has been made as to whether or not we will ship this feature. It is mentioned in the discovery document, but not the UI design. That does not mean that a decision has been made to not include it just because it isn't present in the design files.

> The price property format was changed from using the formatToMinorAmount method to the formatToDecimal method.

Currency values should be stored in the minor amount of the currency - not the display format of a decimal. Decimals lead to floating point math issues and require additional parsing.

pauloiankoski:
>Be careful not to include so many changes under the title of "convert entity properties to camel case" when converting the entity properties to camel case is a minor change of the pull request.

Sure, I apologize for that. Initially, it was meant to be a small task, but then I ended up reviewing all of them.

>Let's stick with maxAvailable because we need to differentiate between how many are available, how many are currently available, and how many have been sold. The person managing the event may think of it as "total tickets" when creating the event, but as a technical concept we need to separate this concept from the total tickets sold. The "total" is ambiguous and is used to describe multiple concepts in this feature.

The term `maxAvailable` is ambiguous to me too. It could be interpreted as the maximum number of tickets an individual can purchase or the maximum number available at any given moment, which might change over time as tickets are sold.

>No final decision has been made as to whether or not we will ship this feature. It is mentioned in the discovery document, but not the UI design. That does not mean that a decision has been made to not include it just because it isn't present in the design files.

In fact, we don't have a decision either to use it or not to use it and that's exactly why I still believe it's better for our product to add it when necessary rather than leaving it there unused for any uncertain amount of time. Shipping our product with an unused column gives a sense of unfinished work, even while in Beta.

>Currency values should be stored in the minor amount of the currency - not the display format of a decimal. Decimals lead to floating point math issues and require additional parsing.

Our models store values as decimals, including Donations, Subscriptions, and certain add-ons such as Fee Recovery. All values are stored using `formatToDecimal` and instantiated with `fromDecimal`. I don't think we need to change this exclusively for the EventTicketTypes.

pauloiankoski:
@kjohnson Let's reconvene here because we may need an additional property in our model to retrieve the number of tickets already sold. This property could be dynamic and count the number of tickets generated for a specific `ticketTypeId`. This discussion also impacts the `maxAvailable` one.

kjohnson:
> Our models store values as decimals, including Donations, Subscriptions, and certain add-ons such as Fee Recovery. All values are stored using formatToDecimal and instantiated with fromDecimal. I don't think we need to change this exclusively for the EventTicketTypes.

These are stored as decimals for backwards compatibility.

Moving forward we should be storing currency as minor amounts, see the Revenue table as an example.

kjohnson:
> Initially, it was meant to be a small task, but then I ended up reviewing all of them.

Please revert the unrelated changes so that we can merge the change to camel case.

kjohnson:
> Let's reconvene here because we may need an additional property in our model to retrieve the number of tickets already sold.

We can address adding new model properties in another PR.

> [W]e may need an additional property in our model to retrieve the number of tickets already sold. This property could be dynamic and count the number of tickets generated for a specific ticketTypeId. This discussion also impacts the maxAvailable one.

We don't need a model property for this and can instead update the query to include a count of related tickets for real-time results (and avoid syncing issues).

> Adding to this, we need to have a status column on ticket types, setting if they are enabled or disabled.

We can use the `ticket_close_datetime` column and store `now()` as the close date.

src/EventTickets/Models/EventTicketType.php
```diff
@@ -29,13 +26,13 @@ class EventTicketType extends Model implements ModelCrud /*, ModelHasFactory */
...
         'id' => 'int',
-        'event_id' => 'int',
+        'eventId' => 'int',
         'label' => 'string',
         'description' => 'string',
         'price' => Money::class,
-        'max_available' => 'int',
-        'created_at' => DateTime::class,
-        'updated_at' => DateTime::class,
+        'totalTickets' => 'int',
```
kjohnson, line 33:
```suggestion
        'maxAvailable' => 'int',
```

src/EventTickets/Repositories/EventTicketTypeRepository.php
```diff
@@ -25,9 +23,10 @@ class EventTicketTypeRepository
      * @var string[]
      */
     private $requiredProperties = [
-        'event_id',
-        'created_at',
-        'updated_at',
+        'eventId',
+        'label',
+        'price',
+        'max_available',
```
kjohnson, line 29:
```suggestion
        'maxAvailable',
```

kjohnson (approved):

PR #7249, opened 2024-02-21
Feature: add form featured image migration step

glaubersilva:
Resolves [GIVE-348]

## Description

This PR adds header image migrations for the v2 form templates. Each template uses different data as the headerImage based on the available settings. The default behavior of the designSettingsImageStyle is set to background when an image supplied, so setting an image style is not necessary unless you want to update it.

### Legacy template
- Uses the featured image as the default

### MultiStep template
- Uses the `introduction` image as the default
- Uses the featured image from the WP default setting as a fallback 
- Sets the image style to `center` as this most closely resembles the current way it is displayed on v2 forms

### Classic template 
- Uses the `headerBackgroundImage` as the default
- It does NOT use the featured image from the WP default setting as a fallback 

## Affects

All 3 form templates during the migration process.

## Visuals

In the below attached screenshots you can see that the `Classic` template doesn't use the featured image from the WP default setting as a fallback:

classic-template-settings

image

## Testing Instructions

Create a v2 form for each of the 3 form templates

**Legacy**

1. Set a featured image and migrate to a v3 form, the featured image previously set should be visible with the _background_ image style in the header

**MultiStep**

1. Set a featured image and migrate to a v3 form, the featured image previously set should be visible with the _center_ image style
2. Set the `introduction` image and migrate to a v3 form, the image previously set should be visible with the _center_ image style
3. Set both the `introduction` image and a featured image, the `introduction` image previously set should be visible with the _center_ image style

**Classic** 

1. Set a featured image and migrate to a v3 form, the featured image previously set should _NOT_ be visible in the header **(this template doesn't use the featured image from the WP default setting as a fallback)**
2. Set a `headerBackgroundImage` and migrate to a v3 form, the header image previously set should be visible with the _background_ image style in the header

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-348]: https://stellarwp.atlassian.net/browse/GIVE-348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/FormMigration/FormMetaDecorator.php
+++ b/src/FormMigration/FormMetaDecorator.php
@@ -5,7 +5,6 @@ namespace Give\FormMigration;
 use Give\DonationForms\V2\Models\DonationForm;
 use Give\DonationForms\ValueObjects\GoalType;
 use Give\FormMigration\Contracts\FormModelDecorator;
-use Give\Helpers\Form\Template as FormTemplateUtils;
 use Give\PaymentGateways\Gateways\Stripe\StripePaymentElementGateway\StripePaymentElementGateway;
 use Give_Email_Notification_Util;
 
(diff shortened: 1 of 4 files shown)


src/FormMigration/FormMetaDecorator.php
```diff
@@ -761,4 +763,27 @@ public function getGiftAidDeclarationForm(): string
...
+            // Sequoia Template (Multi-Step)
+            $featuredImage = $templateSettings['introduction']['image'];
+        } elseif ( ! empty($templateSettings['visual_appearance']['header_background_image'])) {
+            // Classic Template - it doesn't use the featured image from the WP default setting as a fallback
+            $featuredImage = $templateSettings['visual_appearance']['header_background_image'];
+        } elseif ( ! isset($templateSettings['visual_appearance']['header_background_image'])) {
+            // Legacy Template or Sequoia Template without the ['introduction']['image'] setting
+            $featuredImage = get_the_post_thumbnail_url($this->form->id, 'full');
+        } else {
+            $featuredImage = '';
```
JasonTheAdams, line 784: Nit: I recommend using `null` to represent a non-value instead of an empty string. An empty string is still a thing, while `null` is a special value that specifically indicates "nothing".

tests/Feature/FormMigration/TestFormMetaDecorator.php
JasonTheAdams, line 1: Yay for tests! 🙌

JasonTheAdams (approved):
This looks good from a code perspective, @glaubersilva! I didn't get a chance to test it, and only had one small nitpick. I'll approve since the change I'm requesting is so small.

src/FormMigration/FormMetaDecorator.php
glaubersilva: Resolved in add96a3c1627653656abbc44c014bfea0d5e2639

PR #7251, opened 2024-02-22
Feature: Generate Tickets for Donation

kjohnson:
Resolves [GIVE-370](https://stellarwp.atlassian.net/browse/GIVE-370)

## Description

This PR generates tickets for each ticket type purchased during the donation process.

Note: This also resolves a small bug with the `created_at` and `updated_at` timestamps when creating a Ticket.

## Testing Instructions

Select a number of tickets when making a donation.

The generated tickets should be added to the `give_event_tickets` table.

[GIVE-370]: https://stellarwp.atlassian.net/browse/GIVE-370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/EventTickets/Actions/ConvertEventTicketsBlockToFieldsApi.php
```diff
@@ -34,11 +38,12 @@ public function __invoke(BlockModel $block, int $formId)
                     ->ticketTypes($ticketTypes);
 
                 $eventTicketsField->scope(function (EventTickets $field, $value, Donation $donation) {
-                    if (empty($value)) {
-                        return;
-                    }
 
-                    // (new UpdateDonationWithEventTickets())($value, $donation, $field);
```
pauloiankoski, line 41: Besides generating the tickets here, I think we also need to update the donation amount to include the amount paid in tickets like in this draft, similar to what we do in Fee Recovery, otherwise, the donation amount will be shown with an total amount smaller than what have been paid:

```php
$donation->amount = $donation->amount->add(new Money(
    array_sum(array_column(json_decode($value), 'amount')),
    $donation->amount->getCurrency()
));

$donation->save();
```

src/EventTickets/DataTransferObjects/TicketPurchaseData.php
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * @unreleased
+     */
+    public static function fromFieldValueObject(stdClass $object): self
+    {
+        $self = new self();
+
+        $self->quantity = $object->quantity;
+        $self->ticketType = EventTicketType::find($object->ticketId);
```
pauloiankoski, line 41: In case you think it's worth to be added, we also have the `amount` attribute being passed. It's `quantity * ticketPrice` and it's passed as minor amount.

src/EventTickets/Actions/ConvertEventTicketsBlockToFieldsApi.php
kjohnson: It seems potentially dangerous to update a Donations amount. Should the full amount not be set before the donation is created?

src/EventTickets/DataTransferObjects/TicketPurchaseData.php
kjohnson: I don't think we need to use the amount here and I would be cautious to use numerical values passed by the client which can be altered.

src/EventTickets/Actions/ConvertEventTicketsBlockToFieldsApi.php
kjohnson: Does the field scope process before the payment method is charged the donation? In that case, ok, we will need to update the donation amount.

Does that cause any issue with payment intents?

kjohnson: Hm... If the scope runs before the payment gateway is charged, then we'll need to delay the ticket generation until after the charge.

src/EventTickets/DataTransferObjects/TicketPurchaseData.php
pauloiankoski: Right! I've noticed the difference in my approach here compared to the Fee Recovery feature. For Fee Recovery, I introduced a custom validation rule to detect any modifications made by the user in the frontend. This backend validation ensures any changes are caught, preventing the processing of the donation if discrepancies are found. This method enhances our certainty level at this point. However, without such validation, we would need to re-calculate using backend data to ensure it's correct instead.

src/EventTickets/Actions/ConvertEventTicketsBlockToFieldsApi.php
pauloiankoski: Yeah, the scope is more concerned about preparing the donation for being processed (paid). Btw, I think I like the idea of relying on donation status change hooks to create and delete tickets.

kjohnson: We'll need to create the tickets here, because this is where the form data is available.

The status update hook won't have that unless we store it somewhere.

@pauloiankoski how about we merge this and consider how to support a ticket status in a new PR? Jason has suggested using the donation status as the ticket status and I think that could work.

src/EventTickets/DataTransferObjects/TicketPurchaseData.php
kjohnson: Here I think we just need to re-fetch the ticket amount from the database, which would prevent any price tampering from the client side.

kjohnson:
@pauloiankoski I'll open a new PR for updating the donation amount.

pauloiankoski (approved):

PR #7254, opened 2024-02-23
Feature: Add endpoint for event forms

kjohnson:
Resolves [GIVE-363](https://stellarwp.atlassian.net/browse/GIVE-363)

## Description

This PR adds a new WP Rest Route for listing forms related to an event as a GET request.

Note: A event related form is identified by the contents of the JSON `formBuilderFields` column. JSON related MySQL functionality is not introduced until `MySQL 5.7`, but GiveWP supports back to `MySQL 5.5`. For this reason, this endpoint uses `whereLike` queries to determine if a form contains an event block.

## Visuals

image

[GIVE-363]: https://stellarwp.atlassian.net/browse/GIVE-363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/EventTickets/Routes/GetEventForms.php
+++ b/src/EventTickets/Routes/GetEventForms.php
@@ -1,0 +1,80 @@
+<?php
+
+namespace Give\EventTickets\Routes;
+
+use Give\API\RestRoute;
+use Give\DonationForms\Models\DonationForm;
+use Give\EventTickets\Models\Event;
+use WP_REST_Request;
+use WP_REST_Response;
+
+/**
+ * @unreleased
+ */
+class GetEventForms implements RestRoute
+{
+    /** @var string */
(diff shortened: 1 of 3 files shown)


pauloiankoski (commented):
This PR is working as expected. However, as we may implement the List Table component to consume this endpoint, it'll require some changes. Do you think we can merge it as is for now then update it in a new PR or do it whenever we get to that point?

kjohnson:
@pauloiankoski this can be merged as a stand alone REST endpoint and we can add another route for `GetEventFormsListTable` if we need to.

pauloiankoski (approved):

PR #7258, opened 2024-02-26
refactor: accommodate header image for 2-panel layout

JoshuaHungDinh:
Resolves [GIVE-150]

## Description

This PR includes the necessary changes to accommodate the header image while using the 2-panel layout.

Additional changes include:
- Remove margin from container that was clipping the border-radius on the form.
- Remove media max-width on the header panel as it was applying in the canvas but not on the frontend resulting in inconsistent views.
- Replace border-radius in design settings stylesheet with `--givewp-rounded-8`

## Affects

HeaderImage design setting

## Visuals

https://github.com/impress-org/givewp/assets/75056371/51874de2-7377-46ce-9ed7-253db0e6c016

## Testing Instructions
- Create a form in the visual form builder
- Set layout to two-panel
- Upload a header image
-  Test all 3 image styles
- Toggle the various header settings to ensure the layout does not break
- Verify previous layouts do not break

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-150]: https://stellarwp.atlassian.net/browse/GIVE-150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DonationForms/resources/styles/design-settings/image.scss
+++ b/src/DonationForms/resources/styles/design-settings/image.scss
@@ -98,7 +98,7 @@
                     }
 
                     .givewp-layouts-goal {
-                        margin-top: -1.35rem;
+                        margin-top: -1.25rem;
                     }
                 }
             }
--- a/src/DonationForms/resources/styles/design-settings/image.scss
+++ b/src/DonationForms/resources/styles/design-settings/image.scss
@@ -107,8 +107,9 @@
 }
 
 /* Multi Step */
-.givewp-donation-form__steps-body {
+.givewp-donation-form__steps-body, .givewp-donation-form > .givewp-layouts-header {
     padding: 0;
+    background: #fff;
 
     .givewp-layouts-multiStepForm, .givewp-layouts-header__templates-ms {
         padding: 2rem 2rem 0 2rem;
@@ -216,3 +217,9 @@
     }
 }
 
+/* Two-Panel */
+.givewp-donation-form > .givewp-layouts-header {
+    .givewp-layouts-header__templates-ms {
+        padding-bottom: 2rem;
+    }
+}
--- a/src/DonationForms/resources/styles/design-settings/image.scss
+++ b/src/DonationForms/resources/styles/design-settings/image.scss
@@ -25,8 +25,8 @@
                     img {
                         object-fit: cover;
                         min-width: 100%;
-                        border-top-left-radius: 0.5rem;
-                        border-top-right-radius: 0.5rem;
+                        border-top-left-radius: var(--givewp-rounded-8);
+                        border-top-right-radius: var(--givewp-rounded-8);
                     }
                 }
             }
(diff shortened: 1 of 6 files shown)


src/DonationForms/resources/styles/design-settings/image.scss
```diff
@@ -216,3 +218,28 @@
...
+
+                img {
+                    position: absolute;
+                    left: 0;
+                    bottom: 0;
+                    height: 100%;
+                    min-height: 100%;
+                    border-radius: var(--givewp-rounded-8);
+                }
+            }
```
JasonTheAdams, line 242: @JoshuaHungDinh So the `.givewp-layouts-headerImage` isn't applied to the actual `img` tag? I'm guessing not because it depends on the image position?

JoshuaHungDinh: It's not being applied to the image tag because of the use of `withTemplateWrapper`. The image position is a dynamic classname that wraps around the header template so that we can easily create the ideal html structure for each position.

src/DonationForms/resources/styles/design-settings/image.scss
```diff
@@ -216,3 +218,28 @@
     }
 }
 
+/* Two-Panel */
+.givewp-donation-form > .givewp-layouts-header {
+    .givewp-layouts-header__templates-ms {
```
jonwaldstein, line 223: @JoshuaHungDinh head up, we'll want to circle back to the default header template classnames so they are not multi-select specific as referring to here `givewp-layouts-header__templates-ms` since right now if you were to create a _new_ form design that is _not_ multi-step it would still have this classname.

We don't have to solve that here, just wanted to bring that up before releasing the header image and I can help do that part

src/DonationForms/resources/styles/design-settings/image.scss
```diff
@@ -216,3 +218,28 @@
     }
 }
 
+/* Two-Panel */
```
jonwaldstein, line 221: I think we would want to move any two panel specific styles outside of the base stylesheet and into its own styles.

src/DonationForms/resources/styles/design-settings/image.scss
JoshuaHungDinh: Yea, I'd like a quick chat about this when you get back.

src/DonationForms/resources/styles/design-settings/image.scss
JoshuaHungDinh: Yes, I think we should be looking at it as the base image stylesheet. 
Resolved: c18b103411730e4bfcc078af9e1afdad87df1830

jonwaldstein (approved):
great job man!

PR #7259, opened 2024-02-28
Fix: Update Stripe statement descriptor usage per API changes

JasonTheAdams:
## Description

Stripe made changes to how statement descriptors are used in Payment Intents for cards: https://support.stripe.com/questions/use-of-the-statement-descriptor-parameter-on-paymentintents-for-card-charges 

This breaking change affects new Stripe accounts, but prevents donations from working, as such it’s high priority.

Note: I did not change the descriptor for anything other than Payment Intents with `card` as the `payment_method_types` as that's the only scenario affected by the change.

## Affects

Donations processing a credit card using Stripe

## Testing Instructions

1. Test a v2 form using a credit card with Stripe
2. Test a v3 form using a credit card with Stripe

In both cases, check the descriptor in the Stripe dashboard to make sure it's present and looks good.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/includes/gateways/stripe/includes/give-stripe-helpers.php
+++ b/includes/gateways/stripe/includes/give-stripe-helpers.php
@@ -842,7 +842,7 @@ function give_stripe_is_source_type( $id, $type = 'src' ) {
 /**
  * This helper function is used to process Stripe payments.
  *
- * @unreleased switch to using statement_descriptor_suffix per Stripe API changes (https://support.stripe.com/questions/use-of-the-statement-descriptor-parameter-on-paymentintents-for-card-charges)
+ * @unreleased remove descriptor as Stripe automatically adds it, per Stripe API changes (https://support.stripe.com/questions/use-of-the-statement-descriptor-parameter-on-paymentintents-for-card-charges)
  * @since 2.33.0 no longer store the payment intent secret
  * @since 2.5.0
  *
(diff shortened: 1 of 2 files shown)


JasonTheAdams:
@jonwaldstein When you get a chance please review this hotfix, thanks!

includes/gateways/stripe/includes/give-stripe-helpers.php
```diff
@@ -930,7 +931,7 @@ function give_stripe_process_payment( $donation_data, $stripe_gateway ) {
 					'amount'               => $stripe_gateway->format_amount( $donation_data['price'] ),
 					'currency'             => give_get_currency( $form_id ),
 					'payment_method_types' => [ 'card' ],
-					'statement_descriptor' => give_stripe_get_statement_descriptor(),
+                    'statement_descriptor_suffix' => give_stripe_get_statement_descriptor(),
```
jonwaldstein, line 934: fyi, looks like both of these can be modified using the `give_stripe_create_intent_args` filter right now (if need be)

jonwaldstein (approved):
Nice! We should also make sure the other stripe gateways work properly during QA especially including checkout as I see a nested `'payment_intent_data' => ['statement_descriptor']` in there re: `CreateCheckoutSession`

includes/gateways/stripe/includes/give-stripe-helpers.php
JasonTheAdams: Yup! True!

JasonTheAdams:
@jonwaldstein After chatting with Stripe, the change here is that the statement descriptor — as set in the "Public details" settings — is automatically added for cards. It looks like it's also added for Checkout Sessions. So we don't need to add the suffix.

In fact, I'm thinking I may just remove the descriptor entirely for cards. Our `get_stripe_get_statement_descriptor` function just grabs the descriptor from the Stripe settings. So applying it as the suffix makes a redundant descriptor:
image

JasonTheAdams:
Yeah, removing the suffix for cards works better:
image

rickalday (approved):
Passed manual and automated QA tests.

PR #7260, opened 2024-02-28
Feature: Add Create Event modal to Events list table

pauloiankoski:
Resolves [GIVE-336]

## Description
This pull request introduces a modal that allows users to create events directly from the Events List Table.

### Changes to Existing Code
- The `Create Event` button in the `EventsTicketsListTable` component has been replaced with the `CreateEventModal` component. This new component houses both the button and all associated logic.
- A `wrapperClassName` property has been added to the `ModalDialog` component, enabling us to apply distinct styles to modals based on various contexts.
- The route handler for `DeleteEventsListTable` has been corrected. Previously, the delete method was implemented incorrectly. Additionally, its class has now been included in the service provider, as it was omitted earlier, making it unavailable in the API.

### Notes
- Due to our new API Routes, adjustments are needed to ensure compatibility with List Tables. Currently, a temporary hack is in place. However, as I plan to refactor the API Routes for other list tables soon, I anticipate removing this hack in the near future.
- A simple validation was implemented while there's no decision about how those will be handled on these modals.
- For this PoC, the date-time field is currently implemented with a native HTML5 input. We may enhance this by replacing it with a React Dates component in the future.

## Visuals
CleanShot 2024-02-28 at 13 03 07

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-336]: https://stellarwp.atlassian.net/browse/GIVE-336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Views/Components/AdminUI/ModalDialog/index.tsx
```diff
@@ -13,21 +13,22 @@ export interface ModalProps {
...
+export default function Modal({
+    title,
+    icon,
+    children,
+    insertInto,
+    handleClose,
+    isOpen = true,
+    showHeader = true,
+    showCloseIcon = true,
+    wrapperClassName = '',
```
kjohnson, line 28: Do we need to modify the shard component to add a wrapper class, or could we just target styles on the contents of the modal (add a wrapping div in the content)?

I would prefer to not make changes to other domains / shared components unless absolutely necessary (or at least after some attempt to keep the changes inside of the domain).

pauloiankoski: The modal is added with `CreatePortal` directly to the `body` of the page, so it's not a child of any of my components. To avoid styling conflicts with other `DialogModal` components on the page, I thought a better approach optionally setting a custom class to that component.

kjohnson: Can we make this work by wrapping the content of the modal with a div and a custom class?

kjohnson:
> The route handler for DeleteEventsListTable has been corrected. Previously, the delete method was implemented incorrectly. Additionally, its class has now been included in the service provider, as it was omitted earlier, making it unavailable in the API.

I cherry-picked the commit for the Delete handler into the epic branch, which cleans up this changeset to focus on the modal.

src/Views/Components/AdminUI/ModalDialog/index.tsx
kjohnson: We may ultimately need to update the shared component, but I would like to consider another way first.

pauloiankoski: Unfortunately, we can't achieve the same result as the changes I'm applying here are not for its content but changing the modal `max-width` and styling the header.

kjohnson (approved):

PR #7261, opened 2024-02-28
Feature: Add route for event details page

pauloiankoski:
Resolves [GIVE-358]

## Description

This pull request introduces specific conditions to determine when to display the Event Tickets List Table versus the Event Details Page, based on the `id` query parameter. It also lays the groundwork for the future Event Details Page by setting up a minimal structure for its rendering.

## Visuals
CleanShot 2024-02-28 at 14 28 52

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-358]: https://stellarwp.atlassian.net/browse/GIVE-358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/EventTickets/ServiceProvider.php
```diff
@@ -90,6 +91,7 @@ private function registerEventTicketsAdminPage(): void
     {
         Hooks::addAction('admin_menu', RegisterEventsMenuItem::class, '__invoke', 15);
         Hooks::addAction('admin_enqueue_scripts', EnqueueListTableScripts::class);
+        Hooks::addAction('admin_enqueue_scripts', EnqueueEventDetailsScripts::class);
```
kjohnson, line 94: What if we called the enqueue actions from within the render method of the submenu page registration? Then we wouldn't need the `isShowing()` check and the logic of which script gets enqueued is co-located with the page render.

```php
class RegisterEventsMenuItem
{
    public function render()
    {
        get_query_var('id')
            ? give(EnqueueEventDetailsScripts::class)()
            : give(EnqueueListTableScripts::class)();
        
        echo '<div id="give-admin-event-tickets-root"></div>';
    }
}
```

kjohnson: Here is a quick PR showing / explaining my suggestion, see https://github.com/impress-org/givewp/pull/7262.

src/EventTickets/Actions/RegisterEventsMenuItem.php
```diff
@@ -26,6 +29,18 @@ public function __invoke()
      */
     public function render()
     {
+        if(isset($_GET['id'])) {
+            $event = Event::find(absint($_GET['id']));
+
+            if (!$event) {
+                wp_die(__('Event not found', 'give-event-tickets'), 404);
```
kjohnson, line 36: As a note, we can probably improve this user experience by redirecting to the list table when an event does not exist. The tricky part here is that we are mid request and may not be able to do a server side redirect, but we could localize some self instantiating javascript to change the window location.

kjohnson: What we have here is fine for now, but something to consider in a separate PR.

kjohnson (approved):
I think all that is left here is to adjust the List Table script to look for a different query parameter for the create modal deep-link, but that can be a separate refactoring PR.

PR #7262, opened 2024-02-29
Refactor: Co-location submenu scripts with render

kjohnson:
## Description

This PR co-locates submenu scripts with the render method.

The benefit here is that the logic of which script is getting enqueued is in a single place with the render of the submenu, which is where you would look to debug an issue - as opposed to having multiple enqueue actions that may or may not enqueue a script.

This also greatly simplifies the `isShowing()` logic, since most of the check is no longer necessary. 

And the enqueue actions themselves are now only responsible for enqueuing scripts and styles - and not responsible for determining if the script should or should not be enqueued.

### DIFF
--- a/src/EventTickets/Actions/RegisterEventsMenuItem.php
+++ b/src/EventTickets/Actions/RegisterEventsMenuItem.php
@@ -2,6 +2,9 @@
 
 namespace Give\EventTickets\Actions;
 
+use Give\EventTickets\Models\Event;
+use Give\EventTickets\Repositories\EventRepository;
+
 /**
  * @unreleased
  */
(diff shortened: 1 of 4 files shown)


pauloiankoski:
I've been using `&id=new` to trigger the create event modal when being referred by an external page (link on the Event block, for instance). What do you suggest for that?

src/EventTickets/Actions/RegisterEventsMenuItem.php
```diff
@@ -26,6 +29,18 @@ public function __invoke()
      */
     public function render()
     {
+        if(isset($_GET['id'])) {
```
kjohnson, line 32: @pauloiankoski we can update the check to account for the re-use of the id parameter.

```suggestion
        if(isset($_GET['id']) && 'new' !== $_GET['id']) {
```

kjohnson, line 32: @pauloiankoski alternatively, we could use an additional query parameter such as `?new`, instead of re-using the `?id` query parameter.

kjohnson: This would probably be my suggestion, to use a different query parameter for deep-linking the create modal.

pauloiankoski: Either option works for me. Let's go with whatever is your preference then.

kjohnson: I think it'll be easier to follow if we use a different query parameter for the deep-link.

pauloiankoski: In that case, we are done here and I'll implement the changes on my PR.

pauloiankoski (approved):

PR #7272, opened 2024-03-02
Feature: Add modal to edit an event

pauloiankoski:
Resolves [GIVE-361]

## Description

This PR brings a refactor for the `CreateEventModal` component extracting most of its logic to a generic `EventFormModal` component that can handle either create or update an event, then implement that new component into the Event section in the Event Details page.

## Visuals
CleanShot 2024-03-02 at 20 05 06

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-361]: https://stellarwp.atlassian.net/browse/GIVE-361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/EventTickets/resources/admin/components/EventFormModal/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+    return date;
+};
+
+/**
+ * Format a given date to be used in datetime inputs
+ *
+ * @unreleased
+ */
+const getDateString = (date: Date) => {
+    if (!date) {
```
kjohnson, line 25: I think if we are calling `getDateString` and not passing a date then we want an error to occur, so that we know what the problem is. Otherwise, we run the risk of creating downstream issues that are harder to diagnose.

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+    return dateWithOffset.toISOString().slice(0, -8);
+};
+
+/**
+ * Event Form Modal component
+ *
+ * @unreleased
+ */
+export default function EventFormModal({isOpen, handleClose, apiSettings, title, event}: EventModalProps) {
+    const API = new EventTicketsApi(apiSettings);
```
kjohnson, line 41: Instead of passing in the `apiSettings` to the component, should we be passing in the `EventTicketsApi` object itself?

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+            endDateTime: getDateString(
+                event?.endDateTime?.date ? new Date(event?.endDateTime?.date) : getNextSharpHour(2)
+            ),
+        },
+    });
+
+    const onSubmit: SubmitHandler<Inputs> = async (inputs) => {
+        try {
+            inputs.startDateTime += ':00';
+            inputs.endDateTime += ':00';
```
kjohnson, line 63: It feels like a red flag to be appending a specific format, like `:00` to date time values. Is there a better way to handle this?

pauloiankoski: Proposal: 502c7d0fecbb1415c5124d327cad3ad4c79b8557

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+ * Format a given date to be used in datetime inputs
+ *
+ * @unreleased
+ */
+const getDateString = (date: Date) => {
+    if (!date) {
+        return;
+    }
+
+    const offsetInMs = date.getTimezoneOffset() * 60 * 1000;
```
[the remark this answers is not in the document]
pauloiankoski: Ms is the abbreviation of milliseconds.

Refactored: c766aaa8831decbec01dc6624ff2ca441568aa00

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
pauloiankoski: Proposal: 21a5f9cb6264aff6e435a5394dfc16915a22020d

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
pauloiankoski: This component is shared between Events ListTable and Event Details page. Each has distinct apiSettings being passed so I thought to keep each of them managing those. For instance, in the list table we have the `GiveEventTickets` object while in the details page we have the `GiveEventTicketsDetails`.

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
kjohnson: Is this not already a required parameter? Do we need an extra check here with the typing?

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
kjohnson: Could we use [Date.prototype.toISOString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) to simplify this formatting? Create a `new Date` object from the string and then format to an ISO date string.

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
pauloiankoski: What's your suggestion here?

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
pauloiankoski: When I use `toISOString`, it converts my time to UTC instead of respecting my current timezone, resulting in a wrong time being saved.

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
kjohnson: Can we just remove the check? Does the compiled typescript not already check this?

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
pauloiankoski: I think is important to note that this behavior is due to the input we are using. Whenever we implement a new DateTime selector, this may become not relevant anymore.

kjohnson: What about using `getDateString()` from above? Would that work?

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
pauloiankoski: Oh, finally got what you meant here.

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
pauloiankoski: It seems to work but I had to do one change on that method, keep the seconds in the string. See 5450ecb53233cbeb6d8ab652f4d2da69ae22ea25

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
pauloiankoski: d49c81ccfb1338ddae3b183ab8cb83cfe8a832c4

src/EventTickets/resources/admin/components/EventFormModal/index.tsx
kjohnson: Okay - that works.

Lastly, since we are using a magic number to slice the date string, can you leave a comment that explains what the magic number represents?

Better yet, we could extract a declarative function, something like, `removeTimezoneFromDateString()`, so that the function name describes the intention without needing to add a comment.

pauloiankoski: Extracted: 3ea38282880161d01cdbb626aad01ea0efae2988

kjohnson (approved):

PR #7273, opened 2024-03-03
Feature: Add create/edit tickets modal

pauloiankoski:
Resolves [GIVE-366] and [GIVE-368]

## Description

This PR adds the ability to add and edit a ticket type using a modal.

#### Note:
- Didn't yet implement currency input, avoiding complexity at this point. We do have a shared component but that was designed to work through `@wordpress/scripts`. To make it work here, we may need to port the custom functions and use `react-currency-input-field` directly instead.

## Visuals
CleanShot 2024-03-03 at 19 19 02

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-366]: https://stellarwp.atlassian.net/browse/GIVE-366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GIVE-368]: https://stellarwp.atlassian.net/browse/GIVE-368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/EventTickets/resources/admin/components/TicketTypeFormModal/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+            title: ticketData?.title || '',
+            description: ticketData?.description || '',
+            price: ticketData?.price || null,
+            capacity: ticketData?.capacity || null,
+        });
+    }, [ticketData, reset]);
+
+    const onSubmit: SubmitHandler<Inputs> = async (data) => {
+        try {
+            data.price = data.price * 100;
```
kjohnson, line 37: This assumes a non-zero decimal currency. Do we have a money library in JS that we can use to format this to a minor amount?

pauloiankoski:
```js
data.price = accounting.formatNumber(data.price, 2, '', '');
```
What do you think?

kjohnson: I thought we were using something like [dinero.js](https://dinerojs.com/)

```js
const price = Dinero({ amount: 5000, currency: 'EUR' })
```

kjohnson: How are we handling it in the Donation Form?

kjohnson: Looks like we are using `Intl.NumberFormat` for the Donation Form.

kjohnson: For backing into the minor amount from a locale formatted input we may need to check the currency for the configured precision and then move the decimal point (if necessary) from there.

pauloiankoski: It seems that even using Dinero, we keep needing to manually multiply by 100 as it doesn't have a method to convert decimal number to amount.

```js
const priceInMinorAmount = Dinero({amount: Math.round(data.price * 100), currency: currencyCode}).getAmount();
data.price = priceInMinorAmount;
```

pauloiankoski: What if instead we change the API endpoint to receive the price as a decimal then you handle Money amounts from there?

kjohnson: That is an option - and may be the best option at this point.

kjohnson: Wait. I think we are using a React Currency Input somewhere else that we could use here.

kjohnson: I'm getting ahead of myself. For now let's take the easiest path forward and we can consider refactoring after QA.

pauloiankoski: Well. Our input field currently only accept integers so I think that keep things as they are now is fine. Later, we'll replace that input with the React Currency Input Field component and many of this may change.

kjohnson: That sounds good.

kjohnson (approved):

PR #7275, opened 2024-03-04
Feature: Add event details view model

kjohnson:
Resolves [GIVE-404](https://stellarwp.atlassian.net/browse/GIVE-404)

## Description

This PR adds the `EventDetails` View Model for formatting data to be localized to the event details page.

Additionally, a new `forms` relationship has been added to the `Event` model.

## Testing Instructions

```php
class EnqueueEventDetailsScripts
{
    public function __invoke(Event $event)
    {
        $eventDetails = new EventDetails($event);

        $data = [
            //
            'event' => $eventDetails->exports(),
        ];
    }
}
```

[GIVE-404]: https://stellarwp.atlassian.net/browse/GIVE-404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/EventTickets/ViewModels/EventDetails.php
+++ b/src/EventTickets/ViewModels/EventDetails.php
@@ -3,6 +3,7 @@
 namespace Give\EventTickets\ViewModels;
 
 use Give\DonationForms\Models\DonationForm;
+use Give\EventTickets\DataTransferObjects\EventTicketTypeData;
 use Give\EventTickets\Models\Event;
 use Give\EventTickets\Models\EventTicketType;
 
@@ -34,7 +35,7 @@ class EventDetails
         return array_merge($this->event->toArray(), [
 
             'ticketTypes' => array_map(function (EventTicketType $ticketType) {
-                return $ticketType->toArray();
+                return EventTicketTypeData::make($ticketType)->toArray();
             }, $this->event->ticketTypes),
 
             'forms' => array_map(function (DonationForm $form) {
--- a/src/EventTickets/ViewModels/EventDetails.php
+++ b/src/EventTickets/ViewModels/EventDetails.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Give\EventTickets\ViewModels;
+
+use Give\DonationForms\Models\DonationForm;
+use Give\EventTickets\Models\Event;
+use Give\EventTickets\Models\EventTicketType;
+
+/**
+ * @unreleased
+ */
+class EventDetails
+{
+
+    /**
+     * @unreleased
+     * @var Event
+     */
+    protected $event;
+
+    /**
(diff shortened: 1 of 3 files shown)


pauloiankoski (commented):
Left a comment. Other than that, we just need to call the new class into the `EnqueueEventDetailsScripts` class.

kjohnson:
> we just need to call the new class into the EnqueueEventDetailsScripts class, unless you do think it must be part of this PR.

I would suggest localizing the data using this View Model when you refactor the client code. This PR is just introducing the class (with tests).

src/EventTickets/ViewModels/EventDetails.php
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * @unreleased
+     */
+    public function exports(): array
+    {
+        return array_merge($this->event->toArray(), [
+
+            'ticketTypes' => array_map(function (EventTicketType $ticketType) {
+                return $ticketType->toArray();
```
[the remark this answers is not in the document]
kjohnson: Resolved, see [94b88e6](https://github.com/impress-org/givewp/pull/7275/commits/94b88e629db962b9cf565742cb854c039d476836).

pauloiankoski (approved):

src/EventTickets/ViewModels/EventDetails.php
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * @unreleased
+     */
+    public function exports(): array
+    {
+        return array_merge($this->event->toArray(), [
+
+            'ticketTypes' => array_map(function (EventTicketType $ticketType) {
+                return EventTicketTypeData::make($ticketType)->toArray();
+            }, $this->event->ticketTypes),
```
pauloiankoski, line 39: Shouldn't this be calling the method instead?

src/EventTickets/ViewModels/EventDetails.php
```diff
@@ -0,0 +1,47 @@
...
+
+            'forms' => array_map(function (DonationForm $form) {
+                return ['id' => $form->id, 'title' => $form->title];
+            }, $this->event->forms),
```
pauloiankoski, line 43: Same from above here.

src/EventTickets/ViewModels/EventDetails.php
kjohnson: Calling `->ticketType` uses a caching layer added by model relationships. The relationship is queried and cached for later calls.

Calling `->ticketType()` as a method returns a `ModelQueryBuilder`, which is called under the hood when getting the relationship as a property.

PR #7276, opened 2024-03-05
Fix: remove required token key breaking PayPal connections

JasonTheAdams:
## Description

It looks as though PayPal no longer provides an `app_id` when sending the token data to GiveWP, which the system is currently requiring and throwing a fatal error if not found. This PR removes the check for that key. Also, as far as I can tell, that key is never referenced at any future point, so it looks completely safe to remove.

A team member from PayPal confirmed that:
> `app_id` is superfluous and can be safely removed from validation

## Affects

The PayPal onboarding flow, specifically when a user is returned from connecting to PayPal.

## Testing Instructions

Try connecting to both a sandbox account and live account. Make a donation to make sure nothing is broken down the line.

## Pre-review Checklist


jonwaldstein (approved):
🥇

rickalday (approved):
Passed QA tests.

PR #7282, opened 2024-03-07
Feature: Update donation amount to include ticket cost

kjohnson:
Resolves [GIVE-419](https://stellarwp.atlassian.net/browse/GIVE-419)

## Description

This PR update the amount of the donation to include the cost of purchased tickets.

Note: This runs **after** the donation object is created and **before** the payment gateway charges the payment method.

[GIVE-419]: https://stellarwp.atlassian.net/browse/GIVE-419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/EventTickets/Actions/ConvertEventTicketsBlockToFieldsApi.php
+++ b/src/EventTickets/Actions/ConvertEventTicketsBlockToFieldsApi.php
@@ -11,6 +11,7 @@ use Give\EventTickets\Models\EventTicketType;
 use Give\EventTickets\Repositories\EventRepository;
 use Give\Framework\Blocks\BlockModel;
 use Give\Framework\FieldsAPI\Exceptions\EmptyNameException;
+use Give\Framework\Support\ValueObjects\Money;
 
 class ConvertEventTicketsBlockToFieldsApi
 {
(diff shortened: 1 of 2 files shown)


pauloiankoski (approved):
Looks good to me. Not directly related to this but anyway related yet: in the admin, viewing the donation details, when we have an amount recovered by Fee Recovery, there's different way to show the donation amount details. I wonder if that's not something we may address also.

JasonTheAdams:
Perhaps this is already happening, but I've wondered if we should restructure the form inputs themselves to handle multiple methods of affecting the final amount. Historically, we've assumed that the amount block has the single `amount` hidden input, and that is the amount that is ultimately charged. The only exception is Fee Recovery, which I suspect works like this and adds the final value on the server. But this means that, within the form, any component that wants to know the total amount has to know how to sum up all the parts; donation summary gets around this by everything else registering with it. 

In short, would it be worth it to have a `amountTotal` hidden input that all amount inputs affect, which is then used by components like donation summary, and is the final amount sent to the server?

This would mean we'd need to add validation to the total amount to make sure it reflects the restrictions (e.g. minimum donation amount). It could be a bit complicated since, for things like tickets, someone could conceivably add a bunch of tickets and then manually set the total amount, resulting in tickets for less... 🤔

JasonTheAdams:
I could totally be circling back as to why this method _is_ a good idea. Hahah! Just wanted to share what's been on my mind as we've discussed adding multiple amount inputs.

kjohnson:
> for things like tickets, someone could conceivably add a bunch of tickets and then manually set the total amount, resulting in tickets for less... 🤔

I think this is the important part for Event Tickets. We need that separate for validation.

PR #7283, opened 2024-03-07
Fix: Remove prefixed phpstan class

kjohnson:
## Description

This has apparently been here for a while, but hasn't caused an issue because it is only declared in a `use` statement. But for some reason, an Event Tickets PR is starting to throw an error during static analysis.

### DIFF
--- a/src/ServiceProviders/LegacyServiceProvider.php
+++ b/src/ServiceProviders/LegacyServiceProvider.php
@@ -2,7 +2,6 @@
 
 namespace Give\ServiceProviders;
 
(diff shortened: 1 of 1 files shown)


JasonTheAdams:
@kjohnson If Exception is being used, should we add a `use` for the GiveWP Exception primitive?

kjohnson:
@JasonTheAdams a low level `Exception` is not being used. In this file we have more specific exception classes.

JasonTheAdams:
Thanks for finding and fixing that, @kjohnson! 💪

PR #7291, opened 2024-03-12
Refactor: Change unlimited capacity concept from 0 to null

pauloiankoski:
Resolves [GIVE-435]

## Description

Before this pull request, we used to save 0 for the capacity field if left empty. Moving forward, we will store it as null.

This PR includes many changes to support nullable capacity:
-  TicketType Model now verifies null before converting to an integer.
-  TicketTypeRepository no longer requires capacity for validation during insert/update.
-  Create/Update Ticket Type routes now accept null or an integer for capacity, without requiring the argument.
-  SalesCount and SalesAmount columns on Events table now display "Unlimited" if any ticket type has a nullable capacity.
-  Ticket Type edit form now preserves the capacity value without attempting to convert it.
-  Ticket Type table on Event details page now handles null to display "Unlimited."
-  Ticket Type data being passed to the field template will now return either null or a number for the `ticketsAvailable` prop.
-  Field template checks for null or a number greater than 0 for `ticketsAvailable` before showing the quantity input; otherwise, it displays a simple "Sold Out" label.
-  Field template now only shows remaining tickets when the ticket type's capacity is not unlimited.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/EventTickets/ListTable/Columns/SalesAmountColumn.php
+++ b/src/EventTickets/ListTable/Columns/SalesAmountColumn.php
@@ -50,6 +50,10 @@ class SalesAmountColumn extends ModelColumn
                 'price' => $ticketType->price,
                 'capacity' => $ticketType->capacity,
             ];
+
+            if (is_null($ticketType->capacity)) {
+                $maxCapacitySales = __('Unlimited', 'give');
+            }
         }
 
         $salesTotal = array_reduce(
@@ -59,14 +63,19 @@ class SalesAmountColumn extends ModelColumn
             },
             new Money(0, give_get_currency())
         );
-        $maxCapacitySales = array_reduce($ticketTypes, function (Money $carry, $ticketType) {
-            return $carry->add($ticketType['price']->multiply($ticketType['capacity']));
-        }, new Money(0, give_get_currency()));
 
-        $salesPercentage = $maxCapacitySales->formatToMinorAmount() > 0 ? max(
-            min($salesTotal->formatToMinorAmount() / $maxCapacitySales->formatToMinorAmount(), 100),
-            0
-        ) : 0;
+        $salesPercentage = $salesTotal->getAmount() > 0 ? 100 : 0;
+
+        if ( ! isset($maxCapacitySales)) {
+            $maxCapacitySales = array_reduce($ticketTypes, function (Money $carry, $ticketType) {
+                return $carry->add($ticketType['price']->multiply($ticketType['capacity']));
+            }, new Money(0, give_get_currency()));
+
+            $salesPercentage = $maxCapacitySales->formatToMinorAmount() > 0 ? max(
+                min($salesTotal->formatToMinorAmount() / $maxCapacitySales->formatToMinorAmount(), 100),
+                0
+            ) : 0;
+        }
 
         $template = '
             <div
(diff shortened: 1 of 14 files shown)


src/EventTickets/Routes/CreateEventTicketType.php
```diff
@@ -62,9 +62,13 @@ public function registerRoute()
...
                         'type' => 'integer',
-                        'required' => true,
-                        'sanitize_callback' => 'absint',
-                        'validate_callback' => 'rest_is_integer',
+                        'required' => false,
+                        'sanitize_callback' => function ($value) {
+                            return $value === 'null' ? null : absint($value);
+                        },
+                        'validate_callback' => function ($value) {
+                            return $value === 'null' || rest_is_integer($value);
```
jonwaldstein, line 70: Why would the value be a string of `'null'`?

src/EventTickets/ListTable/Columns/SalesAmountColumn.php
```diff
@@ -59,14 +63,19 @@ function (Money $carry, $eventTicket) use ($ticketTypes) {
...
+        if ( ! isset($maxCapacitySales)) {
+            $maxCapacitySales = array_reduce($ticketTypes, function (Money $carry, $ticketType) {
+                return $carry->add($ticketType['price']->multiply($ticketType['capacity']));
+            }, new Money(0, give_get_currency()));
+
+            $salesPercentage = $maxCapacitySales->formatToMinorAmount() > 0 ? max(
+                min($salesTotal->formatToMinorAmount() / $maxCapacitySales->formatToMinorAmount(), 100),
+                0
+            ) : 0;
+        }
```
jonwaldstein, line 78: This is the type of logic that makes me nervous without a unit test 😅

kjohnson (commented):
Is such a large refactor required to support Unlimited Capacity? I thought it was just a bug to fix to support `0` as Unlimited? I'm not sure we need to refactor to using `null`, which comes with its own set of problems.

kjohnson:
> I have appended the "quantity per ticket in the donation summary" change to this pull request as it is a simple and standalone modification

While this is stand alone and simple, I think it still should be submitted separately. A separate PR would associate 1:1 with a Jira ticket and could quickly be reviewed, approved, and merged.

This is particularly beneficial when the PR to which it is attached is sizable, as it requires the reviewer the separate the context and review multiple changes.

jonwaldstein:
Call summary (Jon & Paulo)

- This PR introduces a concept of _unlimited capacity_ that was not complete before
- We dicussed nuances to having an _unlimited capacity_ and we think a better direction would be a new boolean column specifically for this that overrides the capacity value like `capacityUnlimited`
- We will circle back to this PR when we have the time
- For now we will create a new PR that simply requires a capacity value

kjohnson:
No discussion here? I understand a call can be efficient, but the discussion started here as text.

What is wrong with using `0` as the flag for unlimited, which is fairly common practice for this sort of feature.

jonwaldstein:
@kjohnson the call summary is where we left the discussion, its not a definitive direction and could use more time dedication that we simply don't have at this moment.

jonwaldstein:
Closing for now as we do not have plans to introduce this functionality at the moment.  (closing confirmed by @pauloiankoski 😄)

PR #7292, opened 2024-03-12
Feature: prevent some events from being deleted

pauloiankoski:
Resolves [GIVE-422]

## Description

This PR adds an initial implementation for preventing events that has tickets generated from being deleted. Two main actions has been taken: (1) Added an alert when clicking the delete button and (2) added a validation to the API route to ignore those events with tickets.

### To do:
In the future, add an informative modal to explain why the event cannot be deleted.

## Visuals
CleanShot 2024-03-11 at 21 37 06
CleanShot 2024-03-12 at 15 00 09

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-422]: https://stellarwp.atlassian.net/browse/GIVE-422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/EventTickets/ListTable/Columns/SalesCountColumn.php
+++ b/src/EventTickets/ListTable/Columns/SalesCountColumn.php
@@ -21,7 +21,7 @@ class SalesCountColumn extends ModelColumn
      */
     public static function getId(): string
     {
-        return 'sales-count';
+        return 'salesCount';
     }
 
     /**
--- a/src/EventTickets/resources/admin/components/EventTicketsListTable/EventTicketsRowActions.tsx
+++ b/src/EventTickets/resources/admin/components/EventTicketsListTable/EventTicketsRowActions.tsx
@@ -11,6 +11,8 @@ export function EventTicketsRowActions({item, setUpdateErrors, parameters}) {
     const showConfirmModal = useContext(ShowConfirmModalContext);
     const {mutate} = useSWRConfig();
 
+    const salesCount = parseInt(item?.salesCount?.match(/^\d+/)[0], 10);
+
     const fetchAndUpdateErrors = async (parameters, endpoint, id, method) => {
         const response = await eventTicketsApi.fetchWithArgs(endpoint, {ids: [id]}, method);
         setUpdateErrors(response);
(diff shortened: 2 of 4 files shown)


src/EventTickets/resources/admin/components/EventTicketsListTable/EventTicketsRowActions.tsx
```diff
@@ -32,13 +34,15 @@ export function EventTicketsRowActions({item, setUpdateErrors, parameters}) {
...
-                actionId={item.id}
-                displayText={__('Delete', 'give')}
-                hiddenText={item.title}
-                highlight
-            />
+            {salesCount === 0 && (
+                <RowAction
+                    onClick={confirmModal}
+                    actionId={item.id}
+                    displayText={__('Delete', 'give')}
```
kjohnson, line 41: @jdghinson what are your thoughts on the UX of removing the "delete" button when an event cannot be deleted?

My concern is that it could be confusing if there is no explanation.

src/EventTickets/ListTable/Columns/SalesCountColumn.php
```diff
@@ -21,7 +21,7 @@ class SalesCountColumn extends ModelColumn
      */
     public static function getId(): string
     {
-        return 'sales-count';
+        return 'salesCount';
```
kjohnson, line 24: As a note, the IDs don't have to be updated. We can use array syntax in JavaScript when accessing this key.

src/EventTickets/resources/admin/components/EventTicketsListTable/EventTicketsRowActions.tsx
pauloiankoski: Reverted that and implemented a native alert for now (agreed with Jeffrey). https://github.com/impress-org/givewp/pull/7292/commits/9bb51df75b653b3d5d476850a059afa2c4918868

jonwaldstein (approved):

src/EventTickets/ListTable/Columns/SalesCountColumn.php
pauloiankoski: This has only been renamed as we are using camelCase for others.

PR #7298, opened 2024-03-12
Feature: Add Event Beta Feedback Popup

kjohnson:
Resolves [GIVE-420](https://stellarwp.atlassian.net/browse/GIVE-420)

## Description

This PR adds the feedback popup component to the Events List Table admin page.

The popup is automatically shown or hidden based on the local storage value stored under `givewpEventTicketsBetaFeedback`. Clicking the "Provide Feedback" link or close icon will set the local storage to prevent the popup from automatically opening on the next page load.

Note: This is mostly ported from `peer-to-peer`, but has been updated to remove the use of `@wordpress/components` due to a compiling error with that library.

## Visuals

image

[GIVE-420]: https://stellarwp.atlassian.net/browse/GIVE-420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pauloiankoski (commented):
In general, it looks good to me. Currently, I would move the `Feedback` folder into the `components` folder where all components live. In the future, if we choose to reuse this component elsewhere, we will likely need to refactor it by extracting all inline CSS.

What do you think about targeting the feedback link to a new tab?

kjohnson:
> I would move the Feedback folder into the components folder where all components live.

Since the feedback popup is a temporary feature I placed it on a higher level, since it will not be long lived.

kjohnson:
> What do you think about targeting the feedback link to a new tab?

We could open in a new tab. I'm not sure if that is necessary. We can ask @angelablake .

angelablake:
Yep, please open the feedback board in a new tab. That's the preferable behavior for any link that takes the user outside their WordPress site.

PR #7310, opened 2024-03-14
Fix: WP edit bulk action

alaca:
## Description

This PR resolves the issue that happens when using edit bulk action to edit WP posts or pages. The problem was that our action to get the v3 form edit URL was also triggered by the WP edit bulk action. Since WP sends an array of IDs for edit bulk action and we don't need that, the issue is resolved by adding a conditional check where we check if the `$_GET['post']` variable is an array.

## Affects

WP edit bulk action for posts, pages, etc.

## Testing Instructions

Go to posts or pages, select them all, and try to edit them using the edit bulk action. Everything should work as expected. Also, keep an eye on the error log. Servers that run PHP8 will throw an error, so make sure you test this on a server that runs PHP8. 

## Pre-review Checklist

### DIFF
--- a/src/FormBuilder/Routes/EditFormRoute.php
+++ b/src/FormBuilder/Routes/EditFormRoute.php
@@ -19,10 +19,12 @@ class EditFormRoute
     public function __invoke()
     {
         if (isset($_GET['post'], $_GET['action']) && 'edit' === $_GET['action']) {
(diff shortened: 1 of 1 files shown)


jonwaldstein (approved):
Looks good to me, ready for QA 🌈

rickalday (approved):
Passed manual QA tests.

PR #7321, opened 2024-03-21
Refactor: Improve Event Tickets API validations

pauloiankoski:
Resolves [GIVE-371]

## Description

This pull request improves Event Tickets API routes by enhancing permission checks and validations to prevent invalid requests from reaching the handler. Additionally, it includes unit tests that cover various edge cases, that ran through the actual API system with the `RestApiTestCase`.

## Affects

Event Tickets API routes

## Pre-review Checklist

-   [x] Includes unit tests

[GIVE-371]: https://stellarwp.atlassian.net/browse/GIVE-371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/EventTickets/Routes/CreateEvent.php
+++ b/src/EventTickets/Routes/CreateEvent.php
@@ -6,6 +6,7 @@ use Give\API\RestRoute;
 use Give\EventTickets\Models\Event;
 use WP_REST_Request;
 use WP_REST_Response;
+use WP_REST_Server;
 
 /**
  * @since 3.6.0
@@ -28,7 +29,7 @@ class CreateEvent implements RestRoute
             $this->endpoint,
             [
                 [
-                    'methods' => 'POST',
+                    'methods' => WP_REST_Server::CREATABLE,
                     'callback' => [$this, 'handleRequest'],
                     'permission_callback' => function () {
                         return current_user_can('publish_give_payments');
--- a/src/EventTickets/Routes/CreateEvent.php
+++ b/src/EventTickets/Routes/CreateEvent.php
@@ -17,6 +17,9 @@ class CreateEvent implements RestRoute
 
     /**
      * @inheritDoc
+     *
+     * @unreleased Set the permission callback to "publish_give_payments" and description's sanitize callback to "textarea".
+     * @since 3.6.0
      */
     public function registerRoute()
     {
--- a/src/EventTickets/Routes/CreateEvent.php
+++ b/src/EventTickets/Routes/CreateEvent.php
@@ -28,7 +28,7 @@ class CreateEvent implements RestRoute
                     'methods' => 'POST',
                     'callback' => [$this, 'handleRequest'],
                     'permission_callback' => function () {
-                        return current_user_can( 'manage_options' );
+                        return current_user_can('publish_give_payments');
                     }
                 ],
                 'args' => [
--- a/src/EventTickets/Routes/CreateEvent.php
+++ b/src/EventTickets/Routes/CreateEvent.php
@@ -32,7 +32,7 @@ class CreateEvent implements RestRoute
                     'methods' => WP_REST_Server::CREATABLE,
                     'callback' => [$this, 'handleRequest'],
                     'permission_callback' => function () {
-                        return current_user_can('publish_give_payments');
+                        return current_user_can('edit_give_forms');
                     }
                 ],
                 'args' => [
--- a/src/EventTickets/Routes/CreateEvent.php
+++ b/src/EventTickets/Routes/CreateEvent.php
@@ -40,7 +40,7 @@ class CreateEvent implements RestRoute
                     'description' => [
                         'type' => 'string',
                         'required' => false,
-                        'sanitize_callback' => 'sanitize_text_field',
+                        'sanitize_callback' => 'sanitize_textarea_field',
                     ],
                     'startDateTime' => [
                         'type' => 'string',
--- a/src/EventTickets/Routes/DeleteEventTicketType.php
+++ b/src/EventTickets/Routes/DeleteEventTicketType.php
@@ -5,7 +5,6 @@ namespace Give\EventTickets\Routes;
 use Give\API\RestRoute;
 use Give\EventTickets\Models\EventTicketType;
 use Give\Framework\Exceptions\Primitives\Exception;
-use WP_Error;
 use WP_REST_Request;
 use WP_REST_Response;
 use WP_REST_Server;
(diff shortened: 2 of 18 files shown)


src/EventTickets/Routes/DeleteEventTicketType.php
```diff
@@ -49,36 +55,25 @@ public function registerRoute()
...
     {
         $ticketType = EventTicketType::find($request->get_param('ticket_type_id'));
 
         $salesCount = $ticketType->eventTickets()->count();
 
         if ($salesCount > 0) {
-            return new WP_REST_Response([
-                'message' => __('This ticket type has been sold and cannot be deleted.', 'give'),
-            ], 400);
+            return new WP_Error(
```
kjohnson, line 68: What is the benefit in changing this from a `WP_REST_Response` to a `WP_Error`?

src/EventTickets/Routes/GetEventForms.php
```diff
@@ -28,7 +31,9 @@ public function registerRoute()
                 [
                     'methods' => 'GET',
                     'callback' => [$this, 'handleRequest'],
-                    'permission_callback' => '__return_true',
+                    'permission_callback' => function () {
+                        return current_user_can('read');
```
kjohnson, line 35: I would think that this endpoint is either public (no permissions needed), or only readable by an administrator. Checking against the `read` capability suggests that the user must be authenticated, but any authenticated user can access the information, which doesn't sound correct.

src/EventTickets/Routes/DeleteEventTicketType.php
```diff
@@ -30,14 +34,16 @@ public function registerRoute()
...
+                    },
                 ],
                 'args' => [
                     'ticket_type_id' => [
                         'type' => 'integer',
                         'sanitize_callback' => 'absint',
-                        'validate_callback' => function ($eventId) {
-                            return EventTicketType::find($eventId);
+                        'validate_callback' => function ($ticketTypeId) {
+                            return give(EventTicketTypeRepository::class)->queryById($ticketTypeId)->count() > 0;
```
kjohnson, line 46: What is the benefit of validating against the count, as opposed to `EventTicketType|null`? Using the facade, if the entity does not exist then it returns null, which should fail the validation.

Also, if we are to validate against the count, comparing against zero is not necessary, since a positive integer is already `truthy` and zero is `falsey`.

src/EventTickets/Routes/CreateEvent.php
```diff
@@ -25,10 +29,10 @@ public function registerRoute()
             $this->endpoint,
             [
                 [
-                    'methods' => 'POST',
+                    'methods' => WP_REST_Server::CREATABLE,
                     'callback' => [$this, 'handleRequest'],
                     'permission_callback' => function () {
-                        return current_user_can( 'manage_options' );
+                        return current_user_can('publish_give_payments');
```
kjohnson, line 35: Since we don't have a specific permission for this dataset, does it make sense to authorize against a payment capability? Does it suffice to fallback to a simple "is admin" (`manage_options`) check?

src/EventTickets/Routes/GetEventForms.php
pauloiankoski: That's a valid point. I suggest restricting this to admins who can manage forms and posts. What level of capability do you think is appropriate for this? Should we limit it to `manage_options` or consider a lower level like `edit_posts`?

src/EventTickets/Routes/CreateEvent.php
pauloiankoski: I think the answer in the second comment applies to this as well.

src/EventTickets/Routes/DeleteEventTicketType.php
pauloiankoski: I only noticed the issue when writing unit tests. The validation callback was returning `null` for invalid ticket type ids without halting execution. The reason is that the `validate_callback` checks for a false return from that method, and `null` is not considered false.

CleanShot 2024-03-26 at 17 36 03

A solution would be check against `null` like in the snippet bellow. What do you think?
```php
'validate_callback' => function ($ticketTypeId) {
    return EventTicketType::find($ticketTypeId) !== null;
},

kjohnson: Yeah, I think it would be simpler to check against `null` and leverage our facade.

As an aside, we might consider a `::findOrFail($id)` method in the future (another good Friday project).

src/EventTickets/Routes/GetEventForms.php
kjohnson: Using `manage_options` is a common fallback for "is administrator", which I usually prefer over re-using `edit_posts` (when it isn't a `post`).

src/EventTickets/Routes/DeleteEventTicketType.php
kjohnson: Interesting point. In that case, should this check be moved to the attribute `validate_callback`?

src/EventTickets/Routes/GetEventForms.php
pauloiankoski: I'm concerned that users with the `give_manager` role can't manage events using `manage_options`. I'm exploring a capability like `manage_give_settings` that could meet this need for both Administrators and Give Managers.

kjohnson: Ok, if you find a better capability then let's use that.

src/EventTickets/Routes/DeleteEventTicketType.php
pauloiankoski: Changed all to be like `Model::find($entityId) !== null` on ef40760a21e94d8b887460706f6b97d63be4915a

src/EventTickets/Routes/DeleteEventTicketType.php
pauloiankoski: I do agree. Moved it to the `validation_callback` while keeping the custom message.

src/EventTickets/Routes/GetEventForms.php
pauloiankoski: After checking that other menu items use the `edit_give_forms` to even being displayed, the permission checks were updated to use that capability: ef40760a21e94d8b887460706f6b97d63be4915a

kjohnson (approved):

PR #7328, opened 2024-03-25
Feature: add mailchimp form migration unit tests

JoshuaHungDinh:
Resolves # [GIVE-447]

## Description
This PR includes unit tests on the form migration process for the Mailchimp add-on. Three tests were added that check when migrating a form that uses the global settings, migrating a form that has per-form settings & migrating a form without meta.

The tests revealed some areas of improvement within the Mailchimp migration process & adjustments were made to accommodate them.
- The double optin setting is limited to global settings through`give_mailchimp_double_opt_in`. The method checking the form meta for this option in the decorator has been removed.
- In the decorator the meta key for `_give_mailchimp_send_donation` has been updated to `_give_mailchimp_send_donation_data`
- Previously an array was used as the default for subscriber tags. The subscriber tag setting is limited to form meta through `_give_mailchimp_tags`. A null value should explicitly be used as the default for this. While Migrating a form from the global settings the previous array value would add a blank tag in the subscriber tag list.

Additionally the Nullish coalescing operator (??) was used to determine which value should be used inside the getAttributes method of the migration. This was keeping some values from properly accessing the global option as a fallback. I've added the global options inside the decorator as the default for the getMeta method and have included `give_is_setting_enabled` to help determine the boolean values.

## Affects
Mailchimp Migration Process

## Visuals

https://github.com/impress-org/givewp/assets/75056371/7fda7d88-d99a-4b24-94f6-ea7555efddca

https://github.com/impress-org/givewp/assets/75056371/f0b8016a-c830-450e-b11c-13f96c8fcdeb

## Testing Instructions
- Set your global settings
- add a v2 form -> then migrate the form without editing the form settings.
- Verify the v3 form has the correct settings set from the global options.
- Create another v2 form -> update the per-form settings on the right -> then migrate the form
- Verify the v3 form has the correct settings set from the form options.
- Enable globally than disable per form settings the block should not be added.

## Pre-review Checklist

-   [x] Includes unit tests

[GIVE-447]: https://stellarwp.atlassian.net/browse/GIVE-447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/FormMigration/FormMetaDecorator.php
+++ b/src/FormMigration/FormMetaDecorator.php
@@ -529,8 +529,7 @@ class FormMetaDecorator extends FormModelDecorator
      */
     public function getMailchimpLabel()
     {
-        $value = $this->getMeta('_give_mailchimp_custom_label');
-        return $value === '' ? null : $value;
+        return $this->getMeta('_give_mailchimp_custom_label',  give_get_option('give_mailchimp_label', __('Subscribe to newsletter?')));
     }
 
     /**
@@ -538,7 +537,7 @@ class FormMetaDecorator extends FormModelDecorator
      */
     public function getMailchimpDefaultChecked(): bool
     {
-        return $this->getMeta('_give_mailchimp_checked_default');
+        return $this->getMeta('_give_mailchimp_checked_default', give_get_option('give_mailchimp_checked_default', true));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


jonwaldstein (approved):
@JoshuaHungDinh great job man! I really appreciate how much effort you put into the PR description, those details are so helpful - thank you! 🙌

PR #7331, opened 2024-03-26
Feature: add phone number to donor profile

glaubersilva:
Resolves [GIVE-531]

## Description

This PR adds a phone input to the donor profile page. The [International Telephone Input JS library (sponsored by Twilio)](https://github.com/jackocnr/intl-tel-input) was used to make sure that the numbers were stored in the proper format regardless of the donor country.

The big advantage of using this JS library is that we can validate the numbers for all countries before submitting the form and store the donor phone number using the international format - which can be useful in the future if we decide to implement some kind of integration with 3rd party software to send SMS to donors because services like that (For example Twilio) require numbers in the international format.

The whole functionality was encapsulated in the `give_get_intl_tel_input()` function, this way we can reuse it in other places that require a phone input using only vanilla JS. 

~~In a subsequent PR, we also can create a version of this phone input in our form builder library to allow the use as a React Component.~~  **Edit:** The library already provides a [React Component](https://intl-tel-input.com/examples/react-component.html) that we can use when working in the Visual Form Builder or any other UI that uses React.

## Affects

The donor profile page.

## Visuals

image

image

## Testing Instructions

1. Go to the donor profile page;
2. Edit the donor details;
3. Try to save an empty phone number, it should work;
4. Try to save a valid phone number, it should work;
5. Try to save an invalid phone number, it should NOT work.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-531]: https://stellarwp.atlassian.net/browse/GIVE-531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/includes/admin/donors/donor-actions.php
+++ b/includes/admin/donors/donor-actions.php
@@ -9,9 +9,9 @@
  * @since       1.0
  */
 
-// Exit if accessed directly.
 use Give\Donors\Models\Donor;
 
+// Exit if accessed directly.
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
--- a/includes/admin/donors/donor-actions.php
+++ b/includes/admin/donors/donor-actions.php
@@ -127,7 +127,7 @@ function give_edit_donor( $args ) {
 	$donor->update_meta( '_give_donor_company', sanitize_text_field( $args['give_donor_company'] ) );
 
     $donorModel = Donor::find($donor->id);
-    $donorModel->phone = sanitize_text_field($args['give_donor_phone_number']);
+    $donorModel->phone = sanitize_text_field($args['give_donor_phone_number--full_number']);
     $donorModel->save();
 
 	// If First name of donor is empty, then fetch the current first name of donor.
--- a/includes/admin/donors/donor-actions.php
+++ b/includes/admin/donors/donor-actions.php
@@ -9,8 +9,6 @@
  * @since       1.0
  */
 
-use Give\Donors\Models\Donor;
-
 // Exit if accessed directly.
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
@@ -128,9 +126,15 @@ function give_edit_donor( $args ) {
 	// Save company name in when admin update donor company name from dashboard.
 	$donor->update_meta( '_give_donor_company', sanitize_text_field( $args['give_donor_company'] ) );
 
-    $donorModel = Donor::find($donor->id);
-    $donorModel->phone = sanitize_text_field($args['give_donor_phone_number--international-format']);
-    $donorModel->save();
+    /**
+     * Fires after using the submitted data to update the donor metadata.
+     *
+     * @param array $args     The sanitized data submitted.
+     * @param int   $donor_id The donor ID.
+     *
+     * @unreleased
+     */
+    do_action('give_admin_donor_details_updating', $args, $donor->id);
 
 	// If First name of donor is empty, then fetch the current first name of donor.
 	if ( empty( $donor_info['first_name'] ) ) {
--- a/includes/admin/donors/donor-actions.php
+++ b/includes/admin/donors/donor-actions.php
@@ -19,11 +19,13 @@ if ( ! defined( 'ABSPATH' ) ) {
 /**
  * Processes a donor edit.
  *
+ * @unreleased Add support to the "phone number" field
+ * @since      1.0
+ *
  * @param array $args The $_POST array being passed.
  *
- * @since 1.0
- *
  * @return array|bool $output Response messages
+ * @throws Exception
  */
 function give_edit_donor( $args ) {
 
--- a/includes/admin/donors/donor-actions.php
+++ b/includes/admin/donors/donor-actions.php
@@ -127,7 +127,7 @@ function give_edit_donor( $args ) {
 	$donor->update_meta( '_give_donor_company', sanitize_text_field( $args['give_donor_company'] ) );
 
     $donorModel = Donor::find($donor->id);
-    $donorModel->phone = sanitize_text_field($args['give_donor_phone_number--full_number']);
+    $donorModel->phone = sanitize_text_field($args['give_donor_phone_number--international-format']);
     $donorModel->save();
 
 	// If First name of donor is empty, then fetch the current first name of donor.
--- a/includes/admin/donors/donor-actions.php
+++ b/includes/admin/donors/donor-actions.php
@@ -19,7 +19,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 /**
  * Processes a donor edit.
  *
- * @unreleased Add support to the "phone number" field
+ * @unreleased Add support to the "phone" field
  * @since      1.0
  *
  * @param array $args The $_POST array being passed.
--- a/includes/admin/donors/donor-actions.php
+++ b/includes/admin/donors/donor-actions.php
@@ -10,6 +10,8 @@
  */
 
 // Exit if accessed directly.
+use Give\Donors\Models\Donor;
+
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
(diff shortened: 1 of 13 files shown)


src/Donors/Migrations/addPhoneColumn.php
```diff
@@ -0,0 +1,60 @@
...
+namespace Give\Donors\Migrations;
+
+use Give\Framework\Database\Exceptions\DatabaseQueryException;
+use Give\Framework\Migrations\Contracts\Migration;
+use Give\Framework\Migrations\Exceptions\DatabaseMigrationException;
+
+/**
+ * @unreleased
+ */
+class addPhoneColumn extends Migration
```
jonwaldstein, line 12:
```suggestion
class AddPhoneColumn extends Migration
```

src/Donors/ServiceProvider.php
```diff
@@ -55,6 +57,10 @@ public function boot()
 
         $this->addCustomFieldsToDonorDetails();
         $this->enforceDonorsAsUsers();
+
+        give(MigrationsRegister::class)->addMigrations([
+            addPhoneColumn::class,
```
jonwaldstein, line 62:
```suggestion
            AddPhoneColumn::class,
```

src/Donors/ServiceProvider.php
```diff
@@ -8,8 +8,10 @@
 use Give\Donors\CustomFields\Controllers\DonorDetailsController;
 use Give\Donors\Exceptions\FailedDonorUserCreationException;
 use Give\Donors\ListTable\DonorsListTable;
+use Give\Donors\Migrations\addPhoneColumn;
```
jonwaldstein, line 11:
```suggestion
use Give\Donors\Migrations\AddPhoneColumn;
```

src/Donors/Migrations/addPhoneColumn.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        global $wpdb;
+
+        $donorsTableName = "{$wpdb->prefix}give_donors";
+
+        try {
+            maybe_add_column(
+                $donorsTableName,
+                'phone',
+                "ALTER TABLE `$donorsTableName` ADD COLUMN `phone` varchar(50) NOT NULL DEFAULT '' AFTER `email`"
```
jonwaldstein, line 29: This is nit-picky haha but can we change this to be after `name`?  I just like the idea of the required columns like email and name to be first and optional ones like phone to be later in the order. 

```suggestion
                "ALTER TABLE `$donorsTableName` ADD COLUMN `phone` varchar(50) NOT NULL DEFAULT '' AFTER `name`"
```

jonwaldstein: Also, is there any particular reason you went with `varchar(50)` for the phone number?

src/Donors/Migrations/addPhoneColumn.php
```diff
@@ -0,0 +1,60 @@
...
+     * @throws DatabaseMigrationException
+     */
+    public function run()
+    {
+        global $wpdb;
+
+        $donorsTableName = "{$wpdb->prefix}give_donors";
+
+        try {
+            maybe_add_column(
```
jonwaldstein, line 26: I think this is the first time i've seen `maybe_add_column` used in our migrations haha! I actually think it's a good use case to check if the column already exists 👍

src/Donors/Repositories/DonorRepository.php
```diff
@@ -119,6 +120,7 @@ public function insert(Donor $donor)
             'date_created' => Temporal::getFormattedDateTime($dateCreated),
             'user_id' => $donor->userId ?? 0,
             'email' => $donor->email,
+            'phone' => $donor->phone ?? '',
```
jonwaldstein, line 123: Since the phone number is optional, I think we want to use the isset approach  here instead so by default it's just an empty column:

```
if (isset($donor->phone)) {
    $args['phone'] = $donor->phone
}
```

src/Donors/Models/Donor.php
```diff
@@ -50,6 +52,7 @@ class Donor extends Model implements ModelCrud, ModelHasFactory
         'firstName' => 'string',
         'lastName' => 'string',
         'email' => 'string',
+        'phone' => 'string',
```
jonwaldstein, line 55: Please update `TestDonor` & `TestDonorRepository` to reflect this change.

jonwaldstein (changes requested):

src/Donors/Migrations/addPhoneColumn.php
glaubersilva: Resolved in 078a7d30bf7b851381f77f1b4785a2a5e0315de7

src/Donors/ServiceProvider.php
glaubersilva: Resolved in 078a7d30bf7b851381f77f1b4785a2a5e0315de7

src/Donors/ServiceProvider.php
glaubersilva: Resolved in 078a7d30bf7b851381f77f1b4785a2a5e0315de7

src/Donors/Migrations/addPhoneColumn.php
glaubersilva: Resolved in fdfd6147a061b95246cd1ada9eca1e33ca1e5ac2

About the `varchar(50)`, I decide to use this value after checking the replies in this thread: https://stackoverflow.com/questions/723587/whats-the-longest-possible-worldwide-phone-number-i-should-consider-in-sql-varc

I believe this size should let this field future-proof in all possible scenarios discussed on this thread, but I'm open to suggestions.

src/Donors/Repositories/DonorRepository.php
glaubersilva: Resolved in 1002d0d2c812a529db132d79d60538dce2bbca12

src/Donors/Models/Donor.php
glaubersilva: Resolved in 2fd2065c9df9a2ef13924b2211ed650a3707b847 and f7a4c8528b62ba170796ad49d75b029c4601b182

glaubersilva:

src/Donors/Migrations/addPhoneColumn.php
jonwaldstein: Sounds good, i'm cool with that just wanted to clarify how you arrived there 😄

includes/admin/donors/donor-actions.php
```diff
@@ -124,6 +128,10 @@ function give_edit_donor( $args ) {
 	// Save company name in when admin update donor company name from dashboard.
 	$donor->update_meta( '_give_donor_company', sanitize_text_field( $args['give_donor_company'] ) );
 
+    $donorModel = Donor::find($donor->id);
+    $donorModel->phone = sanitize_text_field($args['give_donor_phone_number--international-format']);
+    $donorModel->save();
+
```
jonwaldstein, line 134: @glaubersilva the one part i'm debating here is if using the model should be used directly in the legacy code as it's only used for that one update and could have a side effect.  

Another option is we could create a new action here that we can hook into and run the update with the model.  This would also give us more flexibility for adding more fields in the future.  Something like 
```
do_action('give_admin_donor_details_updating', $args, $donor->id);
````
```
add_action('give_admin_donor_details_updating', function($args, $donorId){
    $donorModel = Donor::find($donor->id);
    $donorModel->phone = sanitize_text_field($args['give_donor_phone_number--international-format']);
    $donorModel->save();
}, 10, 2)
```

glaubersilva: @jonwaldstein Thanks for the suggestion! I liked this idea because using the Donor model in the legacy code also bothered me.  

Resolved in 6fc5a89eab240dc2f4cfab6b6741171ed0fc7886 and c36a9fb73ea86e76d2c9a86b6916d04631cabfc2

jonwaldstein: Great minds think alike! ✊

jonwaldstein (approved):
@glaubersilva great job man ☎️ ! ready for QA 🌈

PR #7335, opened 2024-03-29
Feature: add custom alt tags to "HeaderImage"

JoshuaHungDinh:
Resolves # [GIVE-568]

## Description

This introduces a new setting that allows users to include custom alt tag information for the HeaderImage setting. This is done by passing attachment.alt to the onChange handler inside the media-library component. If no alt tag information is available the alt tag will use the form title as a fallback.

The media-library component will assign an empty string if the information is not filled out, so an "empty" check was added to the formSettings for the frontend and the Logical OR (||) was used for design preview to ensure the form title properly renders as a fallback.

## Affects
HeaderImage
FormSettings

## Testing Instructions
- Create a v3 form
- Add a an image for the Header, while selecting an image on the right of the media uploader you should see an input to add alt tag information. Assign custom alt tag text.
- Right click HeaderImage in design preview to open up the console and view the img tag, you should see the alt property is assigned your custom text. Also verify on the frontend page.
- Go back to the builder and remove the alt tag information. 
- View the element in the console for both design preview & on the frontend. Verify that the alt tag has been replaced by the form title.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/DonationForms/Properties/FormSettings.php
+++ b/src/DonationForms/Properties/FormSettings.php
@@ -189,6 +189,12 @@ class FormSettings implements Arrayable, Jsonable
     public $designSettingsImageUrl;
 
     /**
+     * @unreleased
+     * @var string
+     */
+    public $designSettingsImageAlt;
+
+    /**
      * @since 3.4.0
      * @var string
      */
@@ -314,6 +320,7 @@ class FormSettings implements Arrayable, Jsonable
         ) ? $array['pdfSettings'] : [];
 
         $self->designSettingsImageUrl = $array['designSettingsImageUrl'] ?? '';
+        $self->designSettingsImageAlt = ! empty($array['designSettingsImageAlt']) ? $array['designSettingsImageAlt'] : $self->formTitle;
         $self->designSettingsImageStyle = ! empty($array['designSettingsImageStyle']) ? new DesignSettingsImageStyle(
             $array['designSettingsImageStyle']
         ) : DesignSettingsImageStyle::BACKGROUND();
--- a/src/DonationForms/resources/app/form/Header.tsx
+++ b/src/DonationForms/resources/app/form/Header.tsx
@@ -32,7 +32,7 @@ export default function Header({form}: {form: DonationForm}) {
                     form.settings?.designSettingsImageUrl && (
                         <HeaderImageTemplate
                             url={form.settings?.designSettingsImageUrl}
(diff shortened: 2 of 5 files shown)


src/DonationForms/Properties/FormSettings.php
```diff
@@ -314,6 +320,7 @@ public static function fromArray(array $array): self
         ) ? $array['pdfSettings'] : [];
 
         $self->designSettingsImageUrl = $array['designSettingsImageUrl'] ?? '';
+        $self->designSettingsImageAlt = ! empty($array['designSettingsImageAlt']) ? $array['designSettingsImageAlt'] : $self->formTitle;
```
kjohnson, line 323: This could be simplified to a [shorthand ternary](https://stitcher.io/blog/shorthand-comparisons-in-php#shorthand-ternary-operator) (`?:`). That is, if the array key exists but is an empty string or null (otherwise you'll get undefined index notice).

kjohnson: Note that the shorthand ternary behaves differently that the [null coalescing operator](https://stitcher.io/blog/shorthand-comparisons-in-php#null-coalescing-operator) (`??`).

kjohnson (commented):
@JoshuaHungDinh could you add a test for the `FormSettings` for the fallback value?

src/DonationForms/Properties/FormSettings.php
JoshuaHungDinh: Good to know next time. I seem to have had mixed the FormSettings and the HeaderTemplate conditionals. The Nullish coalescing operator in fact does work for the FormSettings. At the time of testing I did not have the proper conditional in the template & had confused the two.  c73d19901923eebe6d55405710308baf5b399846

Unless Im mistaken I couldn't find FormSetting Unit Tests but I would be happy to make them for the values & defaults. How would you feel if I started a new PR for this as I'd include tests for all the settings?

kjohnson: While I often suggest moving additional work in a separate PR, in the case of a test I think it would be important to add it as a part of the PR. I see that the class does not have a test yet,so if you'd like to jump on a pairing session I'd be happy to work together on adding the test.

kjohnson (approved):

rickalday (approved):
Passed manual QA tests.

PR #7340, opened 2024-04-03
Feature: add phone number block/field to the visual form builder and donation forms

glaubersilva:
Resolves [GIVE-532]

## Description

This PR adds a new Phone block to the visual form builder to allow admins to set up a form to collect the donor's phone numbers.  When the donation is finished, the phone number is displayed on the confirmation page under the "Donor Details" section, it also is displayed in the edit donation screen in the wp-admin area.

Also, this PR prevents a donor with a phone number saved previously from getting this value overwritten in cases where a new donation is made with a new number - instead of overwriting the previous value, we save the new phone number only in the donation meta. ~~instead of overwriting the previous value, we save the new phone number in the "additional phones" property.~~

## Affects

- The Donor model
- The Donation Form model
- The Visual Form Builder

## Visuals

image

image

image

image

## Testing Instructions

1. Create a v3 form;
2. Add the new Phone block;
3. Make a test donation;
4. On the confirmation page, check if the phone is visible in the donor details;
5. On the donation edit page, check if the phone is displayed in the donor details section.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-532]: https://stellarwp.atlassian.net/browse/GIVE-532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JasonTheAdams:
Yo, this looks amazing, @glaubersilva! 🤩

includes/misc-functions.php
```diff
@@ -2624,117 +2624,3 @@ function give_get_page_by_title(string $page_title, string $output = OBJECT, str
 
     return get_post($pages[0], $output);
 }
-
- /**
- * @see https://github.com/jackocnr/intl-tel-input
- *
- * @unreleased
- */
-function give_get_intl_tel_input(string $value, string $id, string $class = '', string $name = ''):string {
```
jonwaldstein, line 2633: was this something that already existed or did you introduce it recently?  If it was here previously I would rather mark it as deprecated in case a third party is using.

src/DonationForms/Rules/PhoneIntlInputRule.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return new self();
+    }
+
+    /**
+     * @unreleased
+     */
+    public function __invoke($value, Closure $fail, string $key, array $values)
+    {
+        $errorMap = IntlTelInput::getErrorMap();
```
jonwaldstein, line 35: What kind of error would happen here? Let's add a description to the doc block here

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
```diff
@@ -1,15 +1,86 @@
...
+
+        const css = document.createElement('link');
+        css.href = intlTelInputSettings.cssUrl;
+        css.rel = 'stylesheet';
+        document.body.appendChild(css);
+
+        const script = document.createElement('script');
+        script.src = intlTelInputSettings.scriptUrl;
+        script.async = true;
+        script.onload = () => {
```
jonwaldstein, line 40: I'm a little confused on what's going on here within the context of a react component.  Maybe i'm missing something, but It feels like reaching into the dom for an input we already have access to directly is unnecessary.  I would expect the intl input to supply it's own css and external scripts without reaching into the dom.

src/Donors/Models/Donor.php
```diff
@@ -55,6 +56,7 @@ class Donor extends Model implements ModelCrud, ModelHasFactory
         'phone' => 'string',
         'prefix' => 'string',
         'additionalEmails' => ['array', []],
+        'additionalPhones' => ['array', []],
```
jonwaldstein, line 59: Ah, I see your attempting to add the `additionalPhones` just like `additionalEmails` works.  Honestly, I would like to prevent this for now.  The `additionalEmails` key is already problematic and should have been accomplished differently than how it works now.  What we have is really a workaround to compensate for the legacy storage of additional emails.

For this scope, we should only be supporting one phone number for a donor.  So let's remove the additional phone numbers for now and can circle back to a better approach if/when the request for that comes up.

src/FormBuilder/resources/js/form-builder/src/blocks/fields/phone/Edit.tsx
```diff
@@ -0,0 +1,66 @@
...
+                </BaseControl>
+            </div>
+
+            <InspectorControls>
+                <PanelBody title={__('Field Settings', 'give')} initialOpen={true}>
+                    <PanelRow>
+                        <Label label={label} setAttributes={setAttributes} />
+                    </PanelRow>
+                    <PanelRow>
+                        <Required isRequired={required} setAttributes={setAttributes} />
```
jonwaldstein, line 60: This required toggle was not working for me:

Screenshot 2024-04-11 at 1 04 00 PM

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
```diff
@@ -1,15 +1,86 @@
...
-
 import {PhoneProps} from '@givewp/forms/propTypes';
+import {useEffect} from 'react';
+import InputMask from 'react-input-mask';
 
-export default function Phone({Label, ErrorMessage, placeholder, description, phoneFormat, inputProps}: PhoneProps) {
+/**
+ * @unreleased
+ */
+export default function Phone({
```
jonwaldstein, line 8: Let's make sure these updates do not cause a conflict with form field manager as I believe this is a shared template.

jonwaldstein (changes requested):

includes/misc-functions.php
glaubersilva: This was introduced in [this PR](https://github.com/impress-org/givewp/pull/7331) which was released in the Give 3.7 to be used in the donor profile page, so it's pretty recent. So, do you think still is necessary to deprecate it instead of just removing it?

src/DonationForms/Rules/PhoneIntlInputRule.php
glaubersilva: Resolve in 1581db8b4be06cc61fc3284aedf3417539c98290

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
glaubersilva: I also was expecting to use their react component, but last week it wasn't being imported from `'intl-tel-input/react'` as demonstrated in their [demo page](https://intl-tel-input.com/examples/react-component.html). 

I tried it again today, as they released a few new versions fixing bugs since last week, now the import is working but the component still is pretty buggy and the autoPlaceholder feature isn't working.

So, this was a solution I found to make it work for now, but we can try to update it again in the near future as this library has a very active release frequency as you can see here: https://github.com/jackocnr/intl-tel-input/releases

src/Donors/Models/Donor.php
glaubersilva: Resolved in f214cb7890523c071038f559a9a5eb48901424ba

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
glaubersilva: Yep! I did a few tests and the template keeps working as before in phone inputs created with FFM as you can see in the attached screenshot. In fact, I think would be a good upgrade to change the FFM phone input to also use the `intl-tel-input` library.

image

glaubersilva:

includes/misc-functions.php
jonwaldstein: ah okay, in that case I think its safe to delete

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
jonwaldstein: sounds good, thanks for checking!

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
jonwaldstein: oh yeah, that looks much easier to maintain! ideally we would use that but I trust you if you say it's buggy - I would suggest creating an issue on their repo with what you are finding is not working as expected.

For my feedback I was really wondering why we couldn't just just refs instead of ID's and render script and style tags instead of using createElement like:

```
<input type={'hidden'} placeholder={placeholder} {...inputProps} /><input id={intlTelInputId} type={'text'} aria-invalid={fieldError ? 'true' : 'false'} />
<script src={intlTelInputSettings.scriptUrl} />
```

Also, it looks like we're using a CDN for this library, any reason to use that over bundling the library? if we bundle the library then we would not even need to include script/style tags as we can access from node_modules

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
```diff
@@ -1,15 +1,89 @@
...
+    phoneFormat,
+    inputProps,
+    fieldError,
+    intlTelInputSettings,
+}: PhoneProps) {
     const FieldDescription = window.givewp.form.templates.layouts.fieldDescription;
+    const {useFormContext} = window.givewp.form.hooks;
+    const {setValue, setError, trigger} = useFormContext();
+
+    const intlTelInputId = inputProps.name + '_intl_tel_input';
```
jonwaldstein, line 22: Regarding my previous comment about Refs instead of IDs you should be able to do:

`const inputRef = useRef(null)`

then likely within useEffect:
`const const intlTelInputId = inputRef.current`

actually while I was typing this I found this codepen that shows an example of this approach with this library from the maintainer: https://codepen.io/jackocnr/pen/RwePbZg

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
glaubersilva: Thanks for the clarification and explanations about the `useRef`

About this:

> Also, it looks like we're using a CDN for this library, any reason to use that over bundling the library? if we bundle the library then we would not even need to include script/style tags as we can access from node_modules

Initially, I was using it because the `utilsScript` is pretty big, and serving it through the CDN is faster. However, it's not necessary for the other scripts, so I changed it to import from `node_modules`.

So, it was resolved in 8226c6dc942a22ff20fd024c8e359827901e5b76 and 4a7c5b3945a68d416dcac2f52da5037c4ff9fc8c 

However, the `utilsScript` kept being loaded through the CDN. Even because it is not possible to import it from `node_modules` as explained here: https://github.com/jackocnr/intl-tel-input/issues/980#issuecomment-541060928

_Do you think we should implement one of the solutions discussed in the link above to load it without using the CDN?_

Another important thing to notice is that to make the `utilsScript` properly load in the Form Builder and the Form Builder Preview, was necessary to use a timeout. Without this timeout, problems related to the autoPlaceholder feature were happening.

In the form preview, also was necessary to do this: https://github.com/impress-org/givewp/pull/7340/files#diff-cbb7d030d97960550cc8faf9e98582fbd7772afe2fc33d82c79b7b178102e528R43-R50

About this: 

>  I would suggest creating an issue on their repo with what you are finding is not working as expected.

When I made the changes that I mentioned previously I realized that the problems I was facing using the React component were the same related to timeout and form preview that I mentioned above. For some reason, they did not happen before when loading all scripts through the CDN, but it started after I changed it.

So, I was able to make it work as expected in the Donations Form:

```
import IntlTelInput from 'intl-tel-input/react';
import 'intl-tel-input/build/css/intlTelInput.css';
<IntlTelInput
      initOptions={{
          utilsScript: intlTelInputSettings.utilsScriptUrl,
          initialCountry: intlTelInputSettings.initialCountry,
          showSelectedDialCode: intlTelInputSettings.showSelectedDialCode,
          strictMode: intlTelInputSettings.strictMode,
          i18n: intlTelInputSettings.i18n,
      }}
/>
```


But in Form Builder and Form Builder preview, I wasn't able to delay the `utilsScript` loading, so I kept using the `useRef` approach instead of using the React component.

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
glaubersilva: Resolved in 8226c6dc942a22ff20fd024c8e359827901e5b76 and 4a7c5b3945a68d416dcac2f52da5037c4ff9fc8c

glaubersilva:

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
jonwaldstein: @glaubersilva okay sounds like you experimented a lot with this.  My ideal solution would be to use their react component (for less maintenance) and not have to do anything weird to account for the form builder.  I'd really like to understand _why_ the timeout is needed and figure out if there is something we can tweak without compensating.  I would also be in favor of letting the autocomplete functionality be limited at this time as a trade-off. 

Let me know what you think, we could also do some some pair programming as well 😄

glaubersilva: @jonwaldstein Gotcha, no problem! I changed it to use the React component.

Resolved in ed12a8edef73e664e3e2842abf7a7cda3502e754 and ba59a6e82fe87617b21009cbecd98e8632786195 and 134d68c775f17c3df4fcdaa6700072d87821233f

About this:

> I'd really like to understand why the timeout is needed...

Even using the React component still was necessary to use a timeout inside a `useEffect` to fix these undesired behaviors in the Form Builder:

1. The autoPlaceholder feature doesn't work.
2. The autoPlaceholder feature doesn't use the proper padding left in the first render.

I left a few comments in the mentioned commits above with these details, but you also can check them in the attached screenshots:

image
image

About this:

> ...and figure out if there is something we can tweak without compensating.

Seems the problem happens due to a delay in the load of the `utilsScript` that only happens in the Form Builder, but I didn't investigate it.  

In fact, I don't think it's a problem from the form builder that we can tweak, but a problem related to how the library loads the `utilsScript` in certain scenarios. 

In the form builder, we are adding the component to a page already loaded, and the `utilsScript` seems to have difficulty getting loaded in this case. But in the Donations Form, it works as expected.

src/FormBuilder/resources/js/form-builder/src/blocks/fields/phone/index.tsx
```diff
@@ -0,0 +1,13 @@
+import settings from './settings';
+
+import {FieldBlock} from '@givewp/form-builder/types';
+
+/**
+ * @unreleased
+ */
+const phone: FieldBlock = {
+    name: 'givewp/phone',
```
jonwaldstein, line 9: Let's actually change this to be specific to donor phone as it's not a generic phone block like ffm

```suggestion
    name: 'givewp/donor-phone',
```

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
```diff
@@ -1,15 +1,102 @@
...
+    inputProps,
+    fieldError,
+    intlTelInputSettings,
+}: PhoneProps) {
     const FieldDescription = window.givewp.form.templates.layouts.fieldDescription;
+    const {useFormContext} = window.givewp.form.hooks;
+    const {setValue, setError, trigger} = useFormContext();
+
+    const isIntlTelInput = intlTelInputSettings.initialCountry && intlTelInputSettings.utilsScriptUrl;
+    const isFormBuilderPreview = window.location.href.indexOf('about:srcdoc') > -1;
```
jonwaldstein, line 25: Just to give you a high level concept, our fields _should_ not have to think about being in the context of the form builder preview.  The builder preview should be treated like the live form.  I realize this is sometimes easier said than done as you had to make a workaround for a preview issue specifically.  

My question for you is can we apply this workaround without having to check if we're in the form builder preview?  Meaning the workaround effectively applied to the input regardless of preview or live mode.

src/FormBuilder/resources/js/form-builder/src/blocks/fields/phone/settings.tsx
```diff
@@ -0,0 +1,46 @@
...
+import Edit from './Edit';
+import {Icon} from '@wordpress/icons';
+import {Path, SVG} from '@wordpress/components';
+
+/**
+ * @unreleased
+ */
+const settings: FieldBlock['settings'] = {
+    ...defaultSettings,
+    title: __('Phone', 'give'),
```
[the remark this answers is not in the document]
glaubersilva: Nice catch! I agree it would be better. However, the same thing is happening with the email block. So, thinking about keeping the consistency, I believe that we need to update the email block name as well. What are your thoughts on it?

image

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
jonwaldstein: Okay cool

- Re: `utilsScript` let's go with that
- Re: fullscreen - can we do the opposite and set fullscreen to false for both?

src/FormBuilder/resources/js/form-builder/src/blocks/fields/phone/settings.tsx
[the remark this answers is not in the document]
jonwaldstein: Ah yeah, we can circle back to that one day haha!

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
glaubersilva: Well, thinking better, the best solution probably would be to keep the `useFullscreenPopup` property set to `false` in all scenarios and use the dropdown in the mobile as well - check the attached screenshot for reference. This way would be possible to fully remove all references to the form builder.

image

src/FormBuilder/resources/js/form-builder/src/blocks/fields/phone/settings.tsx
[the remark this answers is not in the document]
glaubersilva: Resolved in 8401b98cda3228c20ed8c98152745717c78c2c57

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
glaubersilva: Resolve in 0c0287ee2fc10840f34276a416fc88ca0fe19592

src/FormBuilder/resources/js/form-builder/src/blocks/fields/phone/index.tsx
glaubersilva: Resolved in 8401b98cda3228c20ed8c98152745717c78c2c57

glaubersilva:
@jonwaldstein I did implement the last changes suggested and this is ready for re-review.

src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
```diff
@@ -126,6 +129,7 @@ protected function convertInnerBlockToNode(BlockModel $block, int $blockIndex)
     }
 
     /**
+     * @unreleased Add "givewp/phone" block
```
jonwaldstein, line 132:
```suggestion
     * @unreleased Add "givewp/donor-phone" block
```

jonwaldstein (approved):
Great job man!  Ready for QA 🌈

src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
glaubersilva: Resolved in 3abf3e032628efca199098bae5ad0dbf8c7ce8f7

rickalday (approved):
Passed manual QA tests.

PR #7342, opened 2024-04-04
Hotfix: Resolve incompatibility issue with WP 6.5

pauloiankoski:
Resolves [GIVE-583]

## Description
This pull request introduces a hotfix for an incompatibility issue between Visual Form Builder and WordPress 6.5. Previously, any modification to a block resulted in a re-render. Upon investigation, I discovered this was caused by a function designed to ensure all blocks reside within a section block. However, this function was triggering a change to the form state on every modification. This pull request now verifies whether a change causes a block to be placed outside of a section block; if not, it simply ignores the remainder of the function.

## Affects
Visual Form Builder

## Visuals
CleanShot 2024-04-04 at 15 14 45

## Testing Instructions
1. Create/Open a form in the VFB
2. Edit any block attribute
3. Confirm the block remains selected
4. Move a block to be placed outside of a section block
5. Confirm a section block is automatically created and the moved block is placed within that new section

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-583]: https://stellarwp.atlassian.net/browse/GIVE-583?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/FormBuilder/resources/js/form-builder/src/containers/BlockEditorContainer.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/containers/BlockEditorContainer.tsx
@@ -10,12 +10,19 @@ import {__} from "@wordpress/i18n";
 
 
 /**
+ * @unreleased Early return dispatchFormBlocks if there are no unwrapped blocks.
  * @since 3.0.0
  */
 export default function BlockEditorContainer() {
     const {blocks} = useFormState();
     const dispatch = useFormStateDispatch();
     const dispatchFormBlocks = (blocks) => {
+        const hasUnwrappedBlocks = blocks.some((block) => block.name !== 'givewp/section');
+
+        if (!hasUnwrappedBlocks) {
+            return;
+        }
+
         dispatch(setFormBlocks(blocks.map((block) => {
             return block.name == 'givewp/section'
                 ? block
--- a/src/FormBuilder/resources/js/form-builder/src/containers/BlockEditorContainer.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/containers/BlockEditorContainer.tsx
@@ -10,7 +10,7 @@ import {__} from "@wordpress/i18n";
 
 
 /**
- * @unreleased Early return dispatchFormBlocks if there are no unwrapped blocks.
+ * @since 3.6.2 Early return dispatchFormBlocks if there are no unwrapped blocks.
  * @since 3.0.0
  */
 export default function BlockEditorContainer() {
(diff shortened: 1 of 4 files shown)


src/FormBuilder/resources/js/form-builder/src/containers/BlockEditorContainer.tsx
```diff
@@ -10,12 +10,19 @@ import {__} from "@wordpress/i18n";
...
  */
 export default function BlockEditorContainer() {
     const {blocks} = useFormState();
     const dispatch = useFormStateDispatch();
     const dispatchFormBlocks = (blocks) => {
+        const hasUnwrappedBlocks = blocks.some((block) => block.name !== 'givewp/section');
+
+        if (!hasUnwrappedBlocks) {
+            return;
+        }
```
jonwaldstein, line 24: @pauloiankoski nice find! I wonder why this matters? 

However, we will still need to dispatch the blocks in order for the block schema to persist into state or else you won't be able to actually update the form.  I think this is all we would need to do:

```suggestion
        if (!hasUnwrappedBlocks) {
            return dispatch(setFormBlocks(blocks));
        }
```

src/FormBuilder/resources/js/form-builder/src/containers/BlockEditorContainer.tsx
```diff
@@ -34,7 +41,7 @@ export default function BlockEditorContainer() {
     parseMissingBlocks(blocks);
 
     return (
-        <BlockEditorProvider value={blocks} onInput={dispatchFormBlocks} onChange={dispatchFormBlocks}>
```
jonwaldstein, line 37: Was `onInput` causing an issue?

jonwaldstein: I think `onInput` is responsible for the interacting with block setting inputs, so this may actually be needed to update the form

src/FormBuilder/resources/js/form-builder/src/containers/BlockEditorContainer.tsx
pauloiankoski: Implemented on 90970eca9a0345bbaf636b4d5e363cbc257a5ec6

jonwaldstein: Cool cool, actually should we be returning this or breaking it out into an if / else?  I don't believe the prop callbacks this is used by are expecting to receive anything.

pauloiankoski:
```js
updateParent( blocks, {
	selection: {
		selectionStart: getSelectionStart(),
		selectionEnd: getSelectionEnd(),
		initialPosition:
			getSelectedBlocksInitialCaretPosition(),
	},
	undoIgnore,
} );
```
Where `updateParent` is a variable where the callback function is stored. It means the return of that callback function is never used, so it's like `void` is expected.

jonwaldstein:
Build for QA: https://github.com/impress-org/givewp/actions/runs/8560018627

PR #7356, opened 2024-04-19
Feature: Notification Handler endpoints

alaca:
## Description
This PR adds endpoints for handling notifications for the new `NotificationPlaceholder` component 
https://github.com/impress-org/form-builder-library/pull/8

## Pre-review Checklist


give.php
```diff
@@ -58,6 +58,7 @@
 use Give\Form\Templates;
 use Give\FormBuilder\ServiceProvider as FormBuilderServiceProvider;
 use Give\FormMigration\ServiceProvider as FormMigrationServiceProvider;
+use Give\NotificationsHandler\ServiceProvider as NotificationHandlerServiceProvider;
```
jonwaldstein, line 61: Looks like we are giving `NotificationsHandler` it's own top level domain in the src folder.  I'm wondering if this should actually be in the `FormBuilder` domain or if we do want to make this more general like `Notifications` to leave room for other notification type apis.

src/NotificationsHandler/Routes/DismissNotification.php
```diff
@@ -0,0 +1,59 @@
...
+use Give\API\RestRoute;
+use WP_REST_Request;
+use WP_REST_Response;
+
+/**
+ * Dismiss notification route
+ *
+ * @unreleased
+ */
+class DismissNotification implements RestRoute
```
jonwaldstein, line 14: Let's write a test for this 😄 

@pauloiankoski made a `RestApiTestCase` for rest api testing specifically - but you can also do a unit test for `handleRequest` and mock the response like i've done [here](https://github.com/impress-org/givewp/blob/abc1aa7bd0b21d0d301edcfa9046310b64795e04/tests/Feature/Controllers/FormBuilderResourceControllerTest.php#L21)

give.php
alaca: Good point. For me, this is not entirely the Form Builder thing, but I guess it would make sense to move this to the FB domain.

alaca: On another thought, I think having Notifications TLD might be a better option.

PR #7361, opened 2024-04-23
Refactor: Merge block level attributes

pauloiankoski:
## Description

This pull-request merges all separated attributes we currently have for the levels, their descriptions and defaultValue.

## Affects
Amount block level attribute structure

## Testing Instructions
Remove then re-add the Amount block, then confirm that everything still works as before.

## Pre-review Checklist

-   [ ] Acceptance criteria satisfied and marked in related issue
-   [ ] Relevant `@unreleased` tags included in DocBlocks
-   [ ] Includes unit tests
-   [ ] Reviewed by the designer (if follows a design)
-   [x] [Self Review](https://give.gitbook.io/development-manual/devops/github/code-reviews#self-review) of code and UX completed

…indexing

PR #7364, opened 2024-04-24
Epic: add donation amount level descriptions

JoshuaHungDinh:
Resolves: [GIVE-578]

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/FormBuilder/BlockModels/DonationAmountBlockModel.php
+++ b/src/FormBuilder/BlockModels/DonationAmountBlockModel.php
@@ -60,23 +60,22 @@ class DonationAmountBlockModel
     }
 
     /**
+     * @unreleased Changed the return type to an array of OptionsProps
      * @since 3.0.0
      *
-     * @return float[]
+     * @return array ['label' => string, 'value' => string, 'checked' => bool][]
      */
     public function getLevels(): array
     {
-        return array_map(static function($level) {
-            return (float)filter_var($level, FILTER_SANITIZE_NUMBER_FLOAT,FILTER_FLAG_ALLOW_FRACTION);
-        }, $this->block->getAttribute('levels'));
+        return $this->block->getAttribute('levels');
     }
 
     /**
-     * @since 3.0.0
+     * @unreleased
      */
-    public function getDefaultLevel(): ?float
+    public function isDescriptionEnabled(): bool
     {
-        return (float)filter_var($this->block->getAttribute('defaultLevel'), FILTER_SANITIZE_NUMBER_FLOAT,FILTER_FLAG_ALLOW_FRACTION);
+        return $this->block->getAttribute('descriptionsEnabled') === true;
     }
 
     /**
--- a/src/DonationForms/Actions/ConvertDonationAmountBlockToFieldsApi.php
+++ b/src/DonationForms/Actions/ConvertDonationAmountBlockToFieldsApi.php
@@ -56,18 +56,24 @@ class ConvertDonationAmountBlockToFieldsApi
 
             /** @var Amount $amountNode */
             $amountNode = $group->getNodeByName('amount');
-            $defaultLevel = $block->getDefaultLevel() > 0 ? $block->getDefaultLevel() : 10;
             $amountNode
                 ->label($block->getLabel())
-                ->levels(...$block->getLevels())
-                ->allowLevels($block->getPriceOption() === 'multi')
                 ->allowCustomAmount($block->isCustomAmountEnabled())
-                ->fixedAmountValue($block->getSetPrice())
-                ->defaultValue(
-                    $block->getPriceOption() === 'set' ?
-                        $block->getSetPrice() : $defaultLevel
-                )
                 ->rules(...$amountRules);
+
+            $priceOptions = $block->getPriceOption();
+            if ($priceOptions === 'multi') {
+                ['levels' => $levels, 'checked' => $checked] = $this->prepareLevelsArray($block);
+
+                $amountNode
+                    ->allowLevels()
+                    ->levels(...$levels)
+                    ->defaultValue($checked);
+            } else {
+                $amountNode
+                    ->fixedAmountValue($block->getSetPrice())
+                    ->defaultValue($block->getSetPrice());
+            }
 
             /** @var Hidden $currencyNode */
             $currencyNode = $group->getNodeByName('currency');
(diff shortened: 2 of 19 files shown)


rickalday (approved):
Passed manual QA tests.

src/DonationForms/resources/registrars/templates/fields/Amount/DonationAmountLevels.tsx
```diff
@@ -23,24 +24,51 @@ export default function DonationAmountLevels({
...
+    const groupedLevels = levels.reduce(
+        (acc, level) => {
+            const key = level.label ? 'labeled' : 'unlabeled';
+            acc[key].push(level);
+            return acc;
+        },
+        {labeled: [], unlabeled: []}
+    );
+
+    const allLevels = [...groupedLevels.labeled, ...groupedLevels.unlabeled];
```
jonwaldstein, line 36: @pauloiankoski it looks like these are untyped which could cause some confusion during the implementation below since we were previously using the types levels array.  If you add types for `labeled` and `unlabeled` it should result in type definitions for allLevels

src/FormBuilder/BlockModels/DonationAmountBlockModel.php
```diff
@@ -60,23 +60,22 @@ public function getLabel(): string
...
      * @since 3.0.0
      *
-     * @return float[]
+     * @return array ['label' => string, 'value' => string, 'checked' => bool][]
      */
     public function getLevels(): array
     {
-        return array_map(static function($level) {
-            return (float)filter_var($level, FILTER_SANITIZE_NUMBER_FLOAT,FILTER_FLAG_ALLOW_FRACTION);
-        }, $this->block->getAttribute('levels'));
```
jonwaldstein, line 71: @JoshuaHungDinh @pauloiankoski was removing this filtering intentional? Remember the `getAttribute()` method just returns whatever is in the database underneath that attribute key.  That means we are relying on the data to provide the type which sounds fine in theory but is not something I recommended with blocks.  This can be problematic which is one of the reasons I created the [BlockType](https://github.com/impress-org/givewp/blob/becc18a76b646945c5a62b698609eaca733b72bc/src/Framework/Blocks/BlockType.php) so we can validate and cast our attributes to the data source we expect.

I would recommend casting the `levels` attribute in the shape you are expecting and to continue using the filter_var approach when necessary

src/DonationForms/resources/registrars/templates/fields/Amount/DonationAmountLevels.tsx
JoshuaHungDinh: Resolved: 3a5fc46567692475a0c01bed07c840421ceae193

src/FormBuilder/BlockModels/DonationAmountBlockModel.php
pauloiankoski: Would you mean like this?
CleanShot 2024-04-29 at 19 37 00

About the BlockType, did I get it correctly that it may be a replacement to the custom BlockModels we have now? I got little bit confused since they have a lot of similarities.

jonwaldstein: @pauloiankoski yes on your example, that's what I was imagining 👍 

Re: BlockType, yes the custom BlockModel we have for the donation amount will eventually be replaced by a BlockType.  That was my first attempt before creating a system to be re-used.  I plan on working on that migration during a fun friday 😄

pauloiankoski: Resolved on https://github.com/impress-org/givewp/pull/7364/commits/48ba42be90668b55354a11d4bba37bfc478deb9c

JoshuaHungDinh: addressing a test: 99322ff8163936c17299142b60fffb7acba5dfbe

jonwaldstein:
@pauloiankoski @JoshuaHungDinh I tried to run build and i'm getting an error, do we need another form builder library release?

Screenshot 2024-05-02 at 11 14 24 AM

pauloiankoski:
@jonwaldstein Indeed. We have implemented the max length in the description field. Without updating the package, the PR might break.

jonwaldstein:
@pauloiankoski @JoshuaHungDinh @rickalday 

I'd like to add an additional QA test to confirm the [donation level schema migration](https://github.com/impress-org/givewp/pull/7361) for backwards compatibility.

1. Before using this PR, create a v3 form with donation amount levels
2. Switch to this PR and confirm the existing donation amount levels are working properly

We need to confirm _existing donation amount blocks_ are migrated to the new schema successfully. This is _very_ important to make sure or else it could significantly impact existing donation amount fields.

Side note @pauloiankoski I believe we can write a unit test for the `run()` method in `UpdateDonationLevelsSchema` to assert this as well 😄 

I ran into this issue that could very well be a js cache'ing problem on my end - but it made me want to circle back and confirm the migration.

Screenshot 2024-05-02 at 11 50 58 AM

pauloiankoski:
> Side note @pauloiankoski I believe we can write a unit test for the `run()` method in `UpdateDonationLevelsSchema` to assert this as well 😄

@jonwaldstein Test implemented on ca0b514cf820d2c5a3b3330dd4e40e09ca877f32

src/FormBuilder/BlockModels/DonationAmountBlockModel.php
```diff
@@ -60,23 +60,28 @@ public function getLabel(): string
...
      *
-     * @return float[]
+     * @return array ['label' => string, 'value' => string, 'checked' => bool][]
      */
     public function getLevels(): array
     {
         return array_map(static function($level) {
-            return (float)filter_var($level, FILTER_SANITIZE_NUMBER_FLOAT,FILTER_FLAG_ALLOW_FRACTION);
+            return [
+                'label' => (string)filter_var($level['label'] ?? '', FILTER_SANITIZE_STRING),
```
jonwaldstein, line 72: @pauloiankoski heads up `FILTER_SANITIZE_STRING` is [deprecated](https://www.php.net/manual/en/filter.constants.php#constant.filter-sanitize-string) in php 8.1 - they are recommending to use [htmlspecialchars()](https://www.php.net/manual/en/function.htmlspecialchars.php) instead

JoshuaHungDinh: Resolved: https://github.com/impress-org/givewp/commit/2f368c85e5774c4b28e0b9346d5b4a2981d17107

PR #7368, opened 2024-04-25
Feature: Form goal UI

alaca:
## Description

This PR adds new form goal options.  

Screen Capture_select-area_20240426134518

## Pre-review Checklist


src/FormBuilder/ViewModels/FormBuilderViewModel.php
```diff
@@ -110,6 +113,25 @@ public function getGoalTypeOption(
...
     }
 
+    /**
+     * @unreleased
+     */
+    public function getGoalProgressOption(
+        string $value,
+        string $label,
+        string $description,
+        bool $isCustom = false
```
kjohnson, line 123: The `isCustom` flag seems redundant here. Since there are only two options, should we instead just do comparisons against the option name/value?

src/FormBuilder/resources/js/form-builder/src/components/DatePicker/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+        onSelect(convertJsDateToMySQLDate(date));
+        setIsVisible(false);
+    };
+
+    const checkDate = (date: Date) => {
+        if (invalidDateBefore) {
+            return date < new Date(invalidDateBefore);
+        }
+
+        if (invalidDateAfter) {
```
kjohnson, line 61: Is there a situation where both `invalidDateBefore` and `invalidDateAfter` have values? What would be the expected outcome?

src/FormBuilder/ViewModels/FormBuilderViewModel.php
alaca: Yes, you are right. I'll simplify this.

src/FormBuilder/resources/js/form-builder/src/components/DatePicker/index.tsx
alaca: Setting both parameters will give you a selectable date range, and the user can only select dates within that range. I added this to prevent users from selecting an end date that is before the start date.

src/FormBuilder/ViewModels/FormBuilderViewModel.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/7368/commits/abeee7fb7856a00517b5b23d0c8c3be450e62ad0

src/FormBuilder/resources/js/form-builder/src/components/DatePicker/index.tsx
kjohnson: Ok, I think I misunderstood how this was being used.

kjohnson:
@alaca we should probably base this PR on an epic branch, instead of `develop`.

alaca:
@kjohnson you are right, do we have an epic branch?

kjohnson:
@alaca I created and pushed `epic/time-based-form-goals-GIVE-565`.

kjohnson (approved):
LGTM

alaca:
@kjohnson there was an error in date range checking, I fixed it in the last commit https://github.com/impress-org/givewp/pull/7368/commits/247c7176793818100ea6ab66bfb49375bf268a01

kjohnson:
@alaca Ok, that makes sense.

PR #7369, opened 2024-04-25
Fix: Made string translatable in payment-gateway.php file

DAnn2012:
Fix: Made string translatable in payment-gateway.php file


jonwaldstein (approved):
@DAnn2012 thank you for the contribution 😄

PR #7374, opened 2024-04-27
Fix: Made string translatable in inspector.js file

DAnn2012:
Fix: Made string translatable in inspector.js file


jonwaldstein (approved):
@DAnn2012 thank you for the contribution 😄

PR #7376, opened 2024-04-29
Refactor: Update goal/stats to use Donation Query

kjohnson:
Resolves GIVE-677

## Description

This PR updates the donation goal and the donation stats for the Donation Form to use a new `DonationQuery`, which supports a date range and accounts for recovered fees.

## Affects

- Adds a new `DonationQuery` class to account for recovered fees.
- Adds new goal settings to the `DonationFormGoalData` DTO.
- Updates the `getTotalRevenue` method with the `DonationQuery`.
- Adds a new `getTotalRevenueForDateRange` method to the repository.
- Updates the `getTotalNumberOfDonations` method with the `DonationQuery`.
- Adds a new `getTotalNumberOfDonationsForDateRange` method to the repository.
- Updates the `formStatsData` array to use the `DonationQuery`.

### DIFF
--- a/src/DonationForms/DonationQuery.php
+++ b/src/DonationForms/DonationQuery.php
@@ -1,0 +1,79 @@
+<?php
+
+namespace Give\DonationForms;
+
+use Give\Framework\QueryBuilder\JoinQueryBuilder;
+use Give\Framework\QueryBuilder\QueryBuilder;
+
+/**
+ * An opinionated Query Builder for GiveWP donations and meta fields.
+ *
+ * @unreleased
+ *
+ * Example usage:
+ * (new DonationQuery)
+ *     ->form(1816)
+ *     ->between('2024-02-00', '2024-02-23')
+ *     ->sumIntendedAmount();
+ */
+class DonationQuery extends QueryBuilder
+{
+    /**
+     * @unreleased
+     */
+    public function __construct()
+    {
+        $this->from('posts', 'donation');
+    }
+
+    /**
+     * An opinionated join method for the donation meta table.
+     * @unreleased
+     */
+    public function joinMeta($key, $alias)
+    {
+        $this->join(function (JoinQueryBuilder $builder) use ($key, $alias) {
+            $builder
+                ->leftJoin('give_donationmeta', $alias)
+                ->on('donation.ID', $alias . '.donation_id')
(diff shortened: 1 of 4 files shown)


src/DonationForms/DonationQuery.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * An opinionated whereBetween method for the completed date meta field.
+     * @unreleased
+     */
+    public function between($startDate, $endDate)
+    {
+        $this->joinMeta('_give_completed_date', 'completed');
+        $this->whereBetween('completed.meta_value', $startDate, $endDate);
```
alaca, line 62: I didn't test it on my end, but I'm very surprised that this works since the `meta_value` type is `longtext` and not `datetime` :thinking:

src/DonationForms/DonationQuery.php
```diff
@@ -0,0 +1,79 @@
...
+     * Returns a calculated sum of the intended amounts (without recovered fees) for the donations.
+     * @unreleased
+     * @return int|float
+     */
+    public function sumIntendedAmount()
+    {
+        $this->joinMeta('_give_payment_total', 'amount');
+        $this->joinMeta('_give_fee_donation_amount', 'intendedAmount');
+        return $this->sum(
+            'COALESCE(intendedAmount.meta_value, amount.meta_value)'
```
alaca, line 76: Hmmm, I'm not sure I fully understand this. 
`_give_payment_total` has an amount that includes fees, and `_give_fee_donation_amount` doesn't include fees. 
`_give_fee_donation_amount` is set only if fee recovery is checked. Correct me if I'm wrong, but COALESCE will return both  `_give_payment_total` and `_give_fee_donation_amount` if the values are not null. So, if both are set, then the calculation is wrong, or I'm missing something?

src/DonationForms/DonationQuery.php
alaca: Tested this on my end and it works :sweat_smile: 
Interesting

src/DonationForms/DonationQuery.php
alaca: Hmm, it works as expected.

alaca (approved):
I had some concerns regarding dates, but it looks like it works as expected. Also, COALESCE is doing its magic.

PR #7379, opened 2024-05-02
Fix: sanitize shortcode atts

alaca:
## Description
This PR is an addition to https://github.com/impress-org/givewp/pull/7375
In the previous PR, sanitization is added to input args only, but that won't be enough if the malicious code is already saved in the database. To fix this, sanitization of the data before output is also added.

## Affects

`give_form` shortcode

## Testing Instructions

Install and activate Give
Create a form using  "Legacy Form" template
Create a new Post and add a shortcode
`[give_form id="13721" display_style='abcdef"onmouseover=alert(123) b=']`
Replace id with form id
Open the post and move the mouse on the button

## Pre-review Checklist

### DIFF
--- a/includes/class-give-donate-form.php
+++ b/includes/class-give-donate-form.php
@@ -838,6 +838,11 @@ class Give_Donate_Form {
 		// Remove empty class names.
 		$form_classes_array = array_filter( $form_classes_array );
 
+        /**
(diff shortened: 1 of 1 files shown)


rickalday (approved):
Passed manual QA tests

kjohnson (approved):
LGTM

PR #7380, opened 2024-05-02
Feature: Update all form goal blocks/shortcodes to respect the new v3 fields/queries

alaca:
## Description

This PR updates all form goal blocks/shortcodes to respect the new v3 fields/queries. 
`[give_goal]` shortcode is updated and now it supports two new arguments `start_date` and `end_date` where the user can set a date range for the goal. 

## Affects

- Multi-Form Goal block
- [give_goal]
- [give_multi_form_goal]
- [give_totals]

## Pre-review Checklist

### DIFF
--- a/src/MultiFormGoals/ProgressBar/Model.php
+++ b/src/MultiFormGoals/ProgressBar/Model.php
@@ -2,6 +2,7 @@
 
 namespace Give\MultiFormGoals\ProgressBar;
 
+use Give\DonationForms\DonationQuery;
 use Give\ValueObjects\Money;
 
 class Model
@@ -132,14 +133,18 @@ class Model
     /**
      * Get raw earnings value for Progress Bar
      *
+     * @unreleased use DonationQuery
      * @since 2.9.0
      */
     public function getTotal(): string
     {
-        $query = new Query($this->getForms());
-        $results = $query->getResults();
+        $amount = 0;
 
-        return Money::ofMinor($results->total, give_get_option('currency'))->getAmount();
+        foreach($this->getForms() as $formId) {
+            $amount += (new DonationQuery())->form($formId)->sumIntendedAmount();
+        }
+
+        return $amount;
     }
 
     /**
--- a/src/MultiFormGoals/ProgressBar/Model.php
+++ b/src/MultiFormGoals/ProgressBar/Model.php
@@ -138,13 +138,7 @@ class Model
      */
     public function getTotal(): string
     {
-        $amount = 0;
-
-        foreach($this->getForms() as $formId) {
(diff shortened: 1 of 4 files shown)


src/MultiFormGoals/ProgressBar/Model.php
```diff
@@ -132,14 +133,18 @@ public function getDonationRevenueResults()
...
      * @since 2.9.0
      */
     public function getTotal(): string
     {
-        $query = new Query($this->getForms());
-        $results = $query->getResults();
+        $amount = 0;
 
-        return Money::ofMinor($results->total, give_get_option('currency'))->getAmount();
+        foreach($this->getForms() as $formId) {
```
kjohnson, line 143: @alaca what do you think about updating the query to accept multiple form IDs? We could update the `where` to be a `whereIn`.

templates/shortcode-goal.php
```diff
@@ -36,6 +38,25 @@
 $show_text           = isset( $args['show_text'] ) ? filter_var( $args['show_text'], FILTER_VALIDATE_BOOLEAN ) : true;
 $show_bar            = isset( $args['show_bar'] ) ? filter_var( $args['show_bar'], FILTER_VALIDATE_BOOLEAN ) : true;
 
+/**
+ * @unreleased use DonationQuery to get donation amounts
+ */
+$form_income = 0;
+
+if ($args['start_date'] === $args['end_date']) {
```
kjohnson, line 46: I might suggest instantiating the query and setting the `->form()` value and then conditionally setting a `between()` value, as opposed to duplicating the `new DonationQuery` code.

On that note, I've considered adding a `->tap($callback)` method to the `DonationQuery` class which would be great for this.

For example:

```php
(new DonationQuery)
    ->form(1)
    ->tap(function($query) {
       if($startDate && $endDate) {
           $query->between($startDate, $endDate);
        }
    })
    ->sumIndendedAmount()
;
```

src/MultiFormGoals/ProgressBar/Model.php
alaca: That's a great idea. I added a new method `forms` where you can pass an array of form IDs.  See https://github.com/impress-org/givewp/pull/7380/commits/35eb5346024fafd85e9ee7653346fb08b698ba1e

templates/shortcode-goal.php
alaca: Yeah, I like that.

alaca: I refactored it a bit https://github.com/impress-org/givewp/pull/7380/commits/7ea042ea03cb5ec63a085cb1a53262ec23dc14d1

If you could add the `tap`  method, that would be great.

kjohnson (approved):

PR #7383, opened 2024-05-03
Chore: update base enum class with forked version from myclabs

jonwaldstein:
Resolves: [GIVE-760]

## Description

We have a php 8.1 incompatibility conflict with the myclabs/php-enum class.  Unfortunately we are currently tied to php 7.2 at the moment and cannot update to their latest package that [resolves the conflict](https://github.com/myclabs/php-enum/pull/145).  

```
Return type of MyCLabs\Enum\Enum::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in vendor/myclabs/php-enum/src/Enum.php on line 246
```

Furthermore, this package is [not really maintained anymore](https://github.com/myclabs/php-enum?tab=readme-ov-file#native-enums-and-migration) since enums were introduced in php 8.1

So for more control over stability and compatibility with GiveWP, this PR forks the base enum class from myclabs including the unit tests and updates our Enum to extend it.

## Affects

This technically affects our value objects that extend the enum class, however it's still the same logic with the namespace being our own so the risk is very low.

- Fee recovery add-on is currently using the myclabs library loaded from give core https://github.com/impress-org/give-fee-recovery/pull/336

## Visuals

N/A

## Testing Instructions

Unit tests and static analysis should take care of the bulk testing.  We will want to run through QA regression testing as well.

## Pre-review Checklist

-   [x] Includes unit tests

[GIVE-760]: https://stellarwp.atlassian.net/browse/GIVE-760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Framework/Support/ValueObjects/BaseEnum.php
```diff
@@ -0,0 +1,324 @@
+<?php
+
+namespace Give\Framework\Support\ValueObjects;
+
+/**
+ * @unreleased remove stringable
+ *
+ * This is a fork of the myclabs/php-enum library 1.8.4.
+ * In GiveWP we have php constraints that do not allow us to use the latest version of this package.
```
JasonTheAdams, line 9: Let's detail what the constraints are so we know when we're no longer bound by them.

JasonTheAdams (approved):
Ah, makes sense. Left one very minor suggestion so I'll approve.

src/Framework/Support/ValueObjects/BaseEnum.php
jonwaldstein: Sounds good cea4bf4b

rickalday (approved):
Passed QA tests

PR #7385, opened 2024-05-06
Refactor: Simplify goal data to use query builder directly

kjohnson:
Instead of supporting new repository methods, this PR simplifies the goal data to use the new `DonationQuery` directly.

## Description

I noticed that each update required a check for all-time vs custom and then an additional `*ForDateRange()` method. This instantiates the query and conditionally sets the start/end values.

Each type of value then completes the query.

`GoalType::DONATIONS()` calls `$query->count()`.
`GoalType::DONORS()` calls `$query->countDonors()`.
`GoalType::AMOUNT()` calls `$query->sumIntendedAmount()`.


alaca (approved):
Looks good :+1:

PR #7402, opened 2024-06-01
Fix: Made strings translatable in html-admin-page-system-info.php file

DAnn2012:
Fix: Made strings translatable in html-admin-page-system-info.php file


jonwaldstein (approved):
@DAnn2012 thanks for the contribution!

PR #7417, opened 2024-06-20
Fix fatal error in give-export-donations-exporter.php

Genevieve-K:
The meta 'razorpay_donation_response' is an array. Must be serialized to avoid a fatal error: 

`TypeError: Uncaught TypeError: substr(): Argument #1 ($string) must be of type string, array given in /var/www/html/wordpress/wp-content/plugins/give/includes/admin/tools/export/give-export-donations-exporter.php:684`

To test:
- activate give-razorpay
- export donations
- choose a donation form
- select the custom field column razorpay_donation_response

### DIFF
--- a/includes/admin/tools/export/give-export-donations-exporter.php
+++ b/includes/admin/tools/export/give-export-donations-exporter.php
@@ -646,6 +646,7 @@ class Give_Export_Donations_CSV extends Give_Batch_Export {
 	 * @access public
 	 *
 	 * @since  2.1
+	 * @unreleased Use maybe_serialize() for column data (some data are arrays)
 	 *
 	 * @return string|false
 	 */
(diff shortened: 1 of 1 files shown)


jonwaldstein:
@glaubersilva since you are working on razorpay, could you review this please?

includes/admin/tools/export/give-export-donations-exporter.php
```diff
@@ -663,6 +663,7 @@ public function print_csv_rows() {
 				foreach ( $row as $col_id => $column ) {
 					// Make sure the column is valid
 					if ( array_key_exists( $col_id, $cols ) ) {
+                        $column = maybe_serialize( $column );
```
glaubersilva, line 666: We need to document this change. So, can you please add a new unreleased tag (describing this change) just above the `@since  2.1` tag?

Would be something like this:

```
@unreleased Use maybe_serialize() for column data
```

glaubersilva (commented):
@Genevieve-K Thanks for the fix! 

I wasn't able to reproduce this bug on PHP `7.4.3` but was able to reproduce it with PHP `8.1.23` and could confirm that your fix is working as expected. 

The code looks good to me, but I noticed a minor thing related to a missing `@unreleased` tag. I left a comment with more details about it. Let me know if you have any questions!

rickalday (approved):
Passed manual QA tests.

glaubersilva (approved):

PR #7452, opened 2024-07-17
Refactor: use v3 form for the onboarding wizard

JoshuaHungDinh:
Resolves [GIVE-997]

## Description
These changes address replacing the v2 wizard form from the onboarding wizard with a v3 form. To facilitate this several changes were made. 

### Creating the form
The form creation processes has been updated in the `FormRepository` class so that a v3 form is created. On creation the v3 form includes a header image, an alt for the image & the image position ('above').

### Adding/Removing selected features for the form
To update the form fields, the logic has changed from manipulating form meta to the updating the various blocks needed in the v3 form. While removing the blocks from a form there can be an error when removing a block that is not available on the form. To accommodate this, a conditional was included inside the BlockCollection `remove` method.

### Displaying the form
The form url used to src the iframe for display has been updated. Inconsistencies with providing the form url  immediately through the window prior to form creation lead to issues while displaying. Due to this the base form preview url is still passed through the window but the formId used is now managed on the frontend and received after form creation from the route.

## Affects
Onboarding Wizzard

## Testing Instructions
- Go through the onboarding wizard which can be found either by going to Donations-setup from the left sidebar then selecting "First Time Configuration" from the page or through a brand new give site.
- Try adding features from the wizard and verify they show up on the wizard form.
- Try navigating back and removing features and verify the removed features do NOT display.

## Pre-review Checklist
-   [ ] Includes unit tests

[GIVE-997]: https://stellarwp.atlassian.net/browse/GIVE-997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


assets/src/images/admin/onboarding/header-image.png
pauloiankoski, line 1: Could we use a JPEG instead so that we have a smaller file size?

pauloiankoski (changes requested):

assets/src/images/admin/onboarding/header-image.png
JoshuaHungDinh: Resolved: 6f504a0c15cc867e9bc7637056f793ca4fb3713d

src/Onboarding/FormRepository.php
```diff
@@ -66,17 +72,30 @@ protected function isFormAvailable($formID)
...
      */
     protected function makeAndPersist()
     {
-        $formID = $this->defaultFormFactory->make();
+        $form = new DonationForm([
+            'title'    => __('GiveWP Donation Form', 'give'),
+            'status'   => DonationFormStatus::PUBLISHED(),
+            'settings' => FormSettings::fromArray([
+                'designId' => 'multi-step',
+                'designSettingsImageUrl' => GIVE_PLUGIN_URL . '/assets/dist/images/admin/onboarding/header-image.png',
```
pauloiankoski, line 87: Remember to update it here too.

pauloiankoski (approved):
Left one last comment but then it's ready to merge

src/Onboarding/FormRepository.php
JoshuaHungDinh: Good catch, thanks! 2cf25936ee4ef5c9e6d3b6c41a4d009bce86de8c

PR #7463, opened 2024-07-30
Feature: Form Builder Tags, Categories

kjohnson:
Resolves [GIVE-927]

## Description

This PR adds a new Domain for Form Taxonomies, which includes form tags and form categories as custom taxonomies.

## Affects

Add the `$request` as an additional parameter to the `givewp_form_builder_updated` action hook.

## Testing Instructions

Enable form tags and/or form categories in the Give settings.

Create a new donation form and add/remove form tags and form categories.

## Visuals

| Tags & Categories |
|---|
|image |

[GIVE-927]: https://stellarwp.atlassian.net/browse/GIVE-927?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kjohnson:
I found a little more to do with the category selectors.

jonwaldstein:
@kjohnson please fill out the PR description with what this affects and testing instructions.  Also, in order for the jira ticket to be linked using resolves you need to add the brackets around it like this [GIVE-927]

[GIVE-927]: https://stellarwp.atlassian.net/browse/GIVE-927?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

kjohnson:
Build for QA: https://github.com/impress-org/givewp/actions/runs/10181644906

src/FormTaxonomies/Actions/EnqueueFormBuilderAssets.php
```diff
@@ -0,0 +1,89 @@
+<?php
+
+namespace Give\FormTaxonomies\Actions;
```
jonwaldstein, line 3: @kjohnson what's your reasoning behind making FormTaxonomies a top level domain in Give? I would think a lot of this should be within the DonationForms and/or FormBuilder domains.

kjohnson: Form taxonomies are an optional feature, so I think it makes sense to treat it almost like a built-in add-on.

Also, taxonomies are a WordPress concept and do not _belong_ to the donation form domain.

jonwaldstein: @kjohnson thanks for your insight.  Sounds like your viewing this as a separation of concerns, which I agree from a WordPress concept technically makes sense.  

I'm going to think out loud here for a minute and would appreciate your thoughts to help come to a conclusion 😄 

When it comes to domains, I always want to think about the top level business logic and context, this is in contrast to a structure of types.  What encompasses a donation form?  Well, we have things like form settings, form schema, front-end rendering, form request, server side validation, persistence, and even the form confirmation.  How about the form builder? well we have blocks, rendering, form request, persistence, etc.

Where do taxonomies fit in? Well, to your point, they are a separate WordPress concept that is used for content organization and relationships.  Taxonomies are also both a concept and a type at the same time (right?). However, our _usage_ of these taxonomies is very contextual in that they are used within donation forms.  So we are essentially utilizing a WordPress concept in a domain-contextual way (What's a form taxonomy without a donation form?).

Looking at our current structure and also our future, I think this comes down to a two-part questions in my mind:
If we were going with FormTaxonomies as a top level concept, would we potentially have more top level taxonomy domains like CampaignTaxonomies, EventTaxonomies, etc.?  Or would the contextual usage of those warrant them to live inside their respected domains?

kjohnson: > [o]ur usage of these taxonomies is very contextual in that they are used within donation forms.

These custom taxonomies are not used exclusively _within_ donation forms. Rather, they are managed at a global level in the WordPress admin, they are managed per-form in the Form Builder, and they are used as input for various shortcodes (such as the Form Grid) for querying donations forms in the database.

They aren't actually used by the donation form at all.

kjohnson: > If we were going with FormTaxonomies as a top level concept, would we potentially have more top level taxonomy domains like CampaignTaxonomies, EventTaxonomies, etc.? Or would the contextual usage of those warrant them to live inside their respected domains?

I could see those being additional top level domain.

I would say, "it depends".

pauloiankoski: Here’s my 2c: I don’t think a custom taxonomy should be a top-level domain. It seems more appropriate for it to be closely attached to the DonationForm domain, similar to its own post type. A taxonomy is rarely an independent entity, it is usually linked to a specific post type. In our case, it should be related to the donation form post type. Additionally, the interface for managing the new taxonomy in the VFB should live within the same area as the other form settings.

src/DonationForms/Properties/FormSettings.php
```diff
@@ -262,6 +262,20 @@ class FormSettings implements Arrayable, Jsonable
...
+     */
+    public $formTags;
+
+    /**
+     * Form Categories are a custom taxonomy. This setting is only used to pass the selected terms to the server.
+     * @unreleased
+     * @var array
+     */
+    public $formCategories;
+
```
jonwaldstein, line 278: Can you explain your comment here "This setting is only used to pass the selected terms to the server"? 

The form settings object stores all values into the database as json.  Is that an unwanted side effect?

src/FormTaxonomies/Actions/EnqueueFormBuilderAssets.php
```diff
@@ -0,0 +1,89 @@
...
+    public function __construct()
+    {
+        $this->settings = give_get_settings();
+    }
+
+    public function __invoke()
+    {
+        if($this->isFormTagsEnabled() || $this->isFormCategoriesEnabled()) {
+            wp_enqueue_script('givewp-builder-taxonomy-settings', GIVE_PLUGIN_URL . 'assets/dist/js/form-taxonomy-settings.js');
+            wp_add_inline_script('givewp-builder-taxonomy-settings','var giveTaxonomySettings =' . json_encode([
```
jonwaldstein, line 21: I think usage of `wp_localize_script` would be better suited here so you don't have to manually create a variable and json_encode the data as it does all the for you.

src/FormTaxonomies/Actions/EnqueueFormBuilderAssets.php
```diff
@@ -0,0 +1,89 @@
...
+
+        return array_map(function ($term) {
+            return [
+                'id' => $term->term_id,
+                'value' => $term->name,
+            ];
+        }, $terms) ?? [];
+    }
+
+    public function getFormCategories(): array
```
jonwaldstein, line 57: Can you please add doc blocks for all the methods here and include some unit tests?  I could see a view model being useful here with a method for exports that can be asserted

src/FormTaxonomies/Actions/UpdateFormTaxonomies.php
```diff
@@ -0,0 +1,25 @@
...
+     */
+    public function __invoke(DonationForm $form)
+    {
+        if($form->settings->formTags) {
+            wp_set_post_terms($form->id, array_column($form->settings->formTags, 'id'), 'give_forms_tag');
+        }
+
+        if($form->settings->formCategories) {
+            wp_set_post_terms($form->id, array_column($form->settings->formCategories, 'id'), 'give_forms_category');
+        }
```
jonwaldstein, line 23: If we decide to go with notion of taxonomies living outside the concept of donation forms and the form settings should not store their values then this would the alternative approach (this is how i'm currently recommending third parties create their own form settings so we are not storing their values in our form settings object.):

```
add_action('givewp_form_builder_updated', function($form) {
    if (isset($_POST['settings'])) {
        $settings = json_decode(give_clean($_POST['settings']), true);
        if (isset($settings['formTags'])) {
            $formTags = give_clean($settings['formTags']);

            wp_set_post_terms($form->id, array_column($formTags, 'id'), 'give_forms_tag');
        }
    }
});
```

src/FormTaxonomies/resources/form-builder/form-categories.tsx
```diff
@@ -0,0 +1,54 @@
+import {getAvailableFormCategories, getInitialFormCategories} from "./windowData";
+import {buildTermsTree} from "./utils/terms";
+import {CheckboxControl} from "@wordpress/components";
+import {decodeEntities} from "@wordpress/html-entities";
+
+
+const FormCategorySetting = ({settings, setSettings}) => {
+    const {
+        formCategories = getInitialFormCategories(),
+    } = settings;
```
jonwaldstein, line 10: Again, if we are bypassing our internal form settings and treating this as an extension than this would be the approach to take:

```suggestion
    const {
        formCategories = getInitialFormCategories(),
    } = = {...giveTaxonomySettings, ...settings};
```

src/FormTaxonomies/resources/form-builder/form-categories.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    const onChange = (categoryId ) => {
+        setSettings({formCategories: formCategories.includes( categoryId )
+                ? formCategories.filter( ( id ) => id !== categoryId )
+                : [ ...formCategories, categoryId ]
+        })
+    };
+
+    const renderTerms = (renderedTerms) => {
+        return renderedTerms.map((term) => {
```
jonwaldstein, line 22: Creating functions inside a react component means it will be re-created every re-render.  Some ways to avoid this would be to use [useCallback](https://react.dev/reference/react/useCallback), move the entire function outside of the component, or don't create a function and just return the map.

src/FormTaxonomies/Actions/EnqueueFormBuilderAssets.php
jonwaldstein: Thanks for the discussion @kjohnson and @pauloiankoski .  I'll leave the final decision up to you @kjohnson, just wanted to make sure we thought through this as we continue to build out our functionality.

If you decide to keep this separated domain than I left some suggestions on handling this as an extension.

src/FormTaxonomies/resources/form-builder/form-categories.tsx
```diff
@@ -0,0 +1,54 @@
...
+import {CheckboxControl} from "@wordpress/components";
+import {decodeEntities} from "@wordpress/html-entities";
+
+
+const FormCategorySetting = ({settings, setSettings}) => {
+    const {
+        formCategories = getInitialFormCategories(),
+    } = settings;
+
+    const categoryTree = buildTermsTree(getAvailableFormCategories());
```
jonwaldstein, line 12: Looks like you can move this outside of your component to avoid side effects of re-renders as it is working with data from the window that won't change.

src/FormTaxonomies/resources/form-builder/form-tags.tsx
```diff
@@ -0,0 +1,60 @@
+import {__} from "@wordpress/i18n";
+import {debounce} from "@wordpress/compose";
+import apiFetch from '@wordpress/api-fetch';
+import {FormTokenField} from "@wordpress/components";
+import {getInitialFormTags} from "./windowData";
+
+/**
+ * @note Not sure why it won't let me use a state hook here.
+ *       As a workaround, I'm overloading the form settings.
+ *       They will be parsed out by the server during publish.
```
jonwaldstein, line 10: Can you explain what you were trying to accomplish here?

jonwaldstein (changes requested):

jonwaldstein:
@kjohnson another thought here, have we already accounted for taxonomies in the v2 to v3 form migration process?  If not, can you please add a migration step?

src/FormTaxonomies/Actions/UpdateFormTaxonomies.php
kjohnson: Ok, that makes sense. I had considered reaching into the request directly, but it "felt hack-y", but we don't have an official way to pass "extra" data to the server on publish.

kjohnson:
@jonwaldstein thanks for the feedback on how to better separate this domain without using temporary form settings. Always good to get a second set of eyes with a "non-standard" approach. 

I'll get that updates.

Also, I'll look into the form migration - good catch.

src/FormTaxonomies/Actions/UpdateFormTaxonomies.php
jonwaldstein: Yeah I agree on the hacky lol, I could see adding a couple additional params to these form builder actions to pass the request along.  When it comes to 3rd parties adding their own settings, I really just want them to handle the storage themselves so we are not persisting unknown data.

src/FormTaxonomies/Actions/EnqueueFormBuilderAssets.php
kjohnson: My understanding is that `wp_add_inline_script` is the correct way to handle this, as `wp_localize_script` is intended for internationalization.

Also, this avoids some of the "gotchas" such as how localizing a script assumes top level values are strings. Using `wp_add_inline_script` gives more control as to how the data is added to the page.

jonwaldstein: That's fair, apparently WordPress recommends using `wp_add_inline_script` for sending "generic data" to Javascipt.  I've been partial to `wp_localize_script` because the syntax is cleaner but you do technically get more flexiblity with inline.

However, I do think instantiating your variable with `const` rather than `var` would make more sense so it can not be overwritten or declared more than once.

https://developer.wordpress.org/reference/functions/wp_localize_script/#more-information

>Though localization is the primary use, it was often used to pass generic data from PHP to JavaScript, because it was originally the only official way to do that. [wp_add_inline_script()](https://developer.wordpress.org/reference/functions/wp_add_inline_script/) was [introduced in WordPress Version 4.5](https://make.wordpress.org/core/2016/03/08/enhanced-script-loader-in-wordpress-4-5/), and is now the best practice for that use case. `[wp_localize_script()](https://developer.wordpress.org/reference/functions/wp_localize_script/) ` should only be used when you actually want to localize strings.

src/DonationForms/Properties/FormSettings.php
kjohnson: It was an imperfect solution, but I decided to go with your suggestion to pull the data from `$_POST`. See [558e723](https://github.com/impress-org/givewp/pull/7463/commits/558e723908fa1e8975fd3bc511b5b0dae76e2a47).

src/FormTaxonomies/Actions/EnqueueFormBuilderAssets.php
kjohnson: For this case it is important to remember that `wp_localize_script` forces string values and here I'm passing some booleans, which has lead to frustrating bugs in the past. The workaround is to nest non-string values within an array, but at that point might as well just use `wp_add_inline_script`.

Something like `wp_add_inline_data` would be nice, but that would just be a wrapper for `wp_add_inline_script` and `json_encode`.

And yeah, I can make that a `const`. I copy/pasted the structure, which is probably an older example.

kjohnson: Resolved, see [aad42d0](https://github.com/impress-org/givewp/pull/7463/commits/aad42d036621164ec5df0d73f78f8a432658519a).

src/FormTaxonomies/Actions/UpdateFormTaxonomies.php
kjohnson: Resolved, see [558e723](https://github.com/impress-org/givewp/pull/7463/commits/558e723908fa1e8975fd3bc511b5b0dae76e2a47).

kjohnson:
@jonwaldstein Re: Form Migration: I created a new task in Jira, see https://stellarwp.atlassian.net/browse/GIVE-1119.

src/FormTaxonomies/resources/form-builder/form-tags.tsx
kjohnson: I had to overload the form settings to persist the search results, because I'm getting errors when using React hooks, like `useState()`.

src/FormTaxonomies/Actions/EnqueueFormBuilderAssets.php
```diff
@@ -0,0 +1,97 @@
...
+
+    public function __construct()
+    {
+        $this->settings = give_get_settings();
+    }
+
+    public function __invoke()
+    {
+        if($this->isFormTagsEnabled() || $this->isFormCategoriesEnabled()) {
+            wp_enqueue_script('givewp-builder-taxonomy-settings', GIVE_PLUGIN_URL . 'assets/dist/js/form-taxonomy-settings.js');
```
jonwaldstein, line 20:
```suggestion
            ScriptAsset::get(GIVE_PLUGIN_DIR . 'assets/dist/js/formTaxonomySettings.asset.php');

            $scriptAsset = ScriptAsset::get(GIVE_PLUGIN_DIR . 'build/formTaxonomySettings.asset.php');

            wp_enqueue_script(
                'givewp-builder-taxonomy-settings',
                GIVE_PLUGIN_URL . 'build/formTaxonomySettings.js',
                $scriptAsset['dependencies'],
                $scriptAsset['version'],
                true
            );
```

webpack.mix.js
```diff
@@ -57,6 +57,7 @@ mix.setPublicPath('assets/dist')
     .ts('src/Promotions/InPluginUpsells/resources/js/donation-options.ts', 'js/donation-options.js')
     .ts('src/Promotions/InPluginUpsells/resources/js/payment-gateway.ts', 'js/payment-gateway.js')
     .ts('src/Promotions/WelcomeBanner/resources/js/index.tsx', 'js/welcome-banner.js')
+   .ts('src/FormTaxonomies/resources/form-builder/index.tsx', 'js/form-taxonomy-settings.js')
```
jonwaldstein, line 60: We use WordPress scripts for the form builder.  Technically this could be done with laravel mix but it's more complicated and we want to move away from this.  So, move this to `wordpress-scripts-webpack.config.js` like this:

```
formTaxonomySettings: srcPath('FormTaxonomies/resources/form-builder/index.tsx'),
```

jonwaldstein (changes requested):

webpack.mix.js
kjohnson: Resolved, see [e43fd0d](https://github.com/impress-org/givewp/pull/7463/commits/e43fd0de9790699d982ed8337e15b7cece75364e).

src/FormTaxonomies/Actions/EnqueueFormBuilderAssets.php
kjohnson: Good catch - thanks!

src/FormTaxonomies/resources/form-builder/form-tags.tsx
kjohnson: Resolving as moot.

src/FormTaxonomies/resources/form-builder/form-categories.tsx
kjohnson: The data doesn't seem to be available when called outside of the component, so I memoized it, see [0f00bb7](https://github.com/impress-org/givewp/pull/7463/commits/0f00bb708095850f4be64e4102ba2d0d1a728d32).

src/FormTaxonomies/resources/form-builder/form-categories.tsx
kjohnson: I was able to refactor the function so that it could be extracted (and is no longer coupled with local variables in the component context, see [53f4b0d](https://github.com/impress-org/givewp/pull/7463/commits/53f4b0d08d30f03d587b6922e22a7965b2c50f38).

jonwaldstein: Nice, thanks!

src/FormTaxonomies/resources/form-builder/form-categories.tsx
jonwaldstein: Okay that works too 👍

src/FormTaxonomies/Actions/EnqueueFormBuilderAssets.php
kjohnson: I extracted the methods to a view model ([56ea12e](https://github.com/impress-org/givewp/pull/7463/commits/56ea12e65678fb7b6732e2f1c54816b936897801)) and added some tests ([04bab95](https://github.com/impress-org/givewp/pull/7463/commits/04bab95ff10db7fc8f139d895960811375101cbd)).

src/FormTaxonomies/Actions/EnqueueFormBuilderAssets.php
```diff
@@ -0,0 +1,52 @@
...
+            $scriptAsset = ScriptAsset::get(GIVE_PLUGIN_DIR . 'build/formTaxonomySettings.asset.php');
+
+            wp_enqueue_script(
+                'givewp-builder-taxonomy-settings',
+                GIVE_PLUGIN_URL . 'build/formTaxonomySettings.js',
+                $scriptAsset['dependencies'],
+                $scriptAsset['version'],
+                true
+            );
+
```
jonwaldstein, line 42: Looks like we also need to enqueue the styles

```
wp_enqueue_style(
    'givewp-builder-taxonomy-settings',
    GIVE_PLUGIN_URL . 'build/style-formTaxonomySettings.css'
);
```
Screenshot 2024-08-16 at 8 20 26 AM

jonwaldstein (changes requested):

src/FormTaxonomies/ServiceProvider.php
```diff
@@ -0,0 +1,43 @@
...
+     */
+    public function register()
+    {
+        give()->bind(Actions\UpdateFormTaxonomies::class, function() {
+            $formBuilderSettingsPostData = isset($_POST['settings']) ? give_clean(@$_POST['settings']) : [];
+            /** @link https://github.com/impress-org/givewp/pull/7463#discussion_r1706988002 */
+            return new Actions\UpdateFormTaxonomies(
+                json_decode($formBuilderSettingsPostData, true)
+            );
+        });
```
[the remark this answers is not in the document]
kjohnson: I would be open to adding an additional parameter, but prefer not to lead with that idea during initial development.

If we agree that this warrants a change to the filter's signature then I can make that change in this PR.

src/FormTaxonomies/Actions/EnqueueFormBuilderAssets.php
kjohnson: Do the styles get stripped out and built separately? This was imported via JS, but if the build results in a separate asset then I'll enqueue that as well.

kjohnson: Resolved, see [d845a9c](https://github.com/impress-org/givewp/pull/7463/commits/d845a9ca565bb4b2b79952bbc50185d3d083a765).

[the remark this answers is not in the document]
kjohnson: Added, see [b09a69d](https://github.com/impress-org/givewp/pull/7463/commits/b09a69d396943c97653d8e436878157af9b49a5d).

src/FormTaxonomies/resources/form-builder/windowData.ts
```diff
@@ -0,0 +1,40 @@
+type FormTagToken = {
```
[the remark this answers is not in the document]
kjohnson: Resolved, see [cb8c20f](https://github.com/impress-org/givewp/pull/7463/commits/cb8c20fc63120659670456af4c3f98001508e797).

src/FormTaxonomies/resources/form-builder/utils/terms.ts
```diff
@@ -0,0 +1,85 @@
...
+import { decodeEntities } from '@wordpress/html-entities';
+
+/**
+ * Returns terms in a tree form.
+ *
+ * @param {Array} flatTerms Array of terms in flat format.
+ *
+ * @return {Array} Array of terms in tree format.
+ */
+export function buildTermsTree( flatTerms ) {
```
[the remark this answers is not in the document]
kjohnson: Resolved, see [cb8c20f](https://github.com/impress-org/givewp/pull/7463/commits/cb8c20fc63120659670456af4c3f98001508e797).

src/FormTaxonomies/resources/form-builder/taxonomy-settings.tsx
```diff
@@ -0,0 +1,52 @@
+import {SettingsSection} from "@givewp/form-builder-library";
+import {__} from "@wordpress/i18n";
+import {PanelRow} from "@wordpress/components";
+import FormTagSetting from "./form-tags";
+import FormCategorySetting from "./form-categories";
+import getWindowData, {isFormCategoriesEnabled, isFormTagsEnabled} from "./windowData";
+import {useState} from "react";
+
+const TaxonomySettings = ({settings, setSettings}) => {
```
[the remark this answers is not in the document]
kjohnson: Resolved, see [cb8c20f](https://github.com/impress-org/givewp/pull/7463/commits/cb8c20fc63120659670456af4c3f98001508e797).

src/FormTaxonomies/resources/form-builder/form-tags.tsx
```diff
@@ -0,0 +1,53 @@
+import {__} from "@wordpress/i18n";
+import {debounce} from "@wordpress/compose";
+import apiFetch from '@wordpress/api-fetch';
+import {FormTokenField} from "@wordpress/components";
+import {getInitialFormTags} from "./windowData";
+import {useState} from "react";
+
+const FormTagSetting = ({settings, setSettings}) => {
```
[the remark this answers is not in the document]
kjohnson: Resolved, see [cb8c20f](https://github.com/impress-org/givewp/pull/7463/commits/cb8c20fc63120659670456af4c3f98001508e797).

src/FormTaxonomies/resources/form-builder/form-categories.tsx
```diff
@@ -0,0 +1,54 @@
+import {getAvailableFormCategories, getInitialFormCategories} from "./windowData";
+import {buildTermsTree} from "./utils/terms";
+import {CheckboxControl} from "@wordpress/components";
+import {decodeEntities} from "@wordpress/html-entities";
+import {useMemo} from "react";
+
+const FormCategorySetting = ({settings, setSettings}) => {
```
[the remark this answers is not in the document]
kjohnson: Resolved, see [cb8c20f](https://github.com/impress-org/givewp/pull/7463/commits/cb8c20fc63120659670456af4c3f98001508e797).

src/FormTaxonomies/resources/form-builder/windowData.ts
```diff
@@ -0,0 +1,40 @@
...
+    id: number;
+    value: string;
+};
+
+interface TaxonomySettings {
+    formTagsEnabled: boolean;
+    formCategoriesEnabled: boolean;
+    formTagsSelected: FormTagToken[];
+    formCategoriesSelected: any[];
+    formCategoriesAvailable: any[];
```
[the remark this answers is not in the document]
kjohnson: Resolved, see [cb8c20f](https://github.com/impress-org/givewp/pull/7463/commits/cb8c20fc63120659670456af4c3f98001508e797).

tests/Unit/FormTaxonomies/ViewModels/FormTaxonomyViewModelTest.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+namespace Give\Tests\Unit\FormTaxonomies\ViewModels;
+
+use Give\FormTaxonomies\ViewModels\FormTaxonomyViewModel;
+use Give\Tests\TestCase;
+use Give\Tests\TestTraits\RefreshDatabase;
+use Give\Tests\Unit\DonationForms\TestTraits\LegacyDonationFormAdapter;
+
+class FormTaxonomyViewModelTest extends TestCase
```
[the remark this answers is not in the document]
kjohnson: Resolved, see [cd7b514](https://github.com/impress-org/givewp/pull/7463/commits/cd7b514479d84843ae779fda84924b9613c2f825).

src/FormTaxonomies/ServiceProvider.php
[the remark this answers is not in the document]
jonwaldstein: @kjohnson I like that, will make adding settings externally much easier 👍

src/FormTaxonomies/Actions/UpdateFormTaxonomies.php
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * @unreleased
+     */
+    public function __invoke(DonationForm $form)
+    {
+        if(isset($this->settings['formTags'])) {
+            wp_set_post_terms($form->id, array_column($this->settings['formTags'], 'id'), 'give_forms_tag');
+        }
+
+        if(isset($this->settings['formCategories'])) {
```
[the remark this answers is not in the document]
kjohnson: The IDs are already validated for hierarchical terms, but not flat terms.

The sure this up, I added a shared validation (since both are expected arrays of IDs) and bypassed `wp_set_post_terms` to call `wp_set_object_terms` directly with the validated terms.

See [25b5b0b](https://github.com/impress-org/givewp/pull/7463/commits/25b5b0b185546a2cb02681b5145a8f36fd247f0d).

kjohnson:
I may have found a bug in WordPress core - or I'm doing it wrong and don't see the issue.

This results in being unable to have a de-selected category removed from a form.

When I set the terms of a post for a hierarchical taxonomy, ie form category, the unlisted terms are not being removed.
image

This doesn't seem to be an issue with a flat taxonomy, ie form tag, the unlisted term is removed (as is expected).
image

Note: [wp_set_post_terms](https://developer.wordpress.org/reference/functions/wp_set_post_terms/) has a parameter $append, which is false by default, and when false should replace the terms.

src/FormTaxonomies/Actions/UpdateFormTaxonomies.php
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        if(isset($formBuilderSettings['formCategories'])) {
+            $formCategories = $this->validateTermIds($formBuilderSettings['formCategories']);
+            wp_set_object_terms($form->id, $formCategories, 'give_forms_category');
+        }
+    }
+
+    /**
+     * @param $termsIds
```
jonwaldstein, line 32:
```suggestion
     * @unreleased
```

src/FormTaxonomies/Actions/UpdateFormTaxonomies.php
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Give\FormTaxonomies\Actions;
+
+use Give\DonationForms\Models\DonationForm;
+use WP_REST_Request;
+
+/**
+ * @unreleased
+ */
+class UpdateFormTaxonomies
```
jonwaldstein, line 11: Can you write a test for this?

jonwaldstein (approved):
@kjohnson nice job! just one more request for a unit test, otherwise looks good & ready for QA 🌈

src/FormTaxonomies/Actions/UpdateFormTaxonomies.php
kjohnson: Resolved, see [03fde79](https://github.com/impress-org/givewp/pull/7463/commits/03fde7956e5c01ef2f1c19604bf38c7763e9fa88).

kjohnson:
Build for QA, see https://github.com/impress-org/givewp/actions/runs/10455929311.

src/FormTaxonomies/Actions/UpdateFormTaxonomies.php
jonwaldstein: Thanks!

rickalday (approved):
Passed manual QA tests.

PR #7477, opened 2024-08-05
Feature: Add additional gateway notice

kjohnson:
Resolves [GIVE-1092]

## Description

This PR adds a notice about additional payment gateways to the payment gateway block.

## Affects

Adds a notice in the inspector controls.

Dev note: the external link element cannot be added to the block card description directly, which only accepts a string. The implementation here is a workaround, which is documented.

## Visuals

| Inspector Controls for Payment Gateways block |
|---|
| image|
| Screenshot from 2024-08-05 13-27-22 |

## Testing Instructions

Create a v3 donation form.
Go to the Build tab/screen.
Click on the Payment Gateways block.
Dismiss the notice.
Reload the form builder.

## Pre-review Checklist

-   [x] Includes unit tests

[GIVE-1092]: https://stellarwp.atlassian.net/browse/GIVE-1092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kjohnson:
Build for QA, see https://github.com/impress-org/givewp/actions/runs/10253458388.

kjohnson:
Build for QA, see https://github.com/impress-org/givewp/actions/runs/10269404181.

tests/Unit/FormBuilder/ServiceProviderTest.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Give\Tests\Unit\FormBuilder;
+
+use Give\Tests\TestCase;
+use Give\Tests\TestTraits\RefreshDatabase;
+
+class ServiceProviderTest extends TestCase
```
glaubersilva, line 8: Can we add `@unreleased `tags for the class and the method?

glaubersilva (changes requested):
@kjohnson Looks good to me, I just left a comment related to a small change.  

Also, I noticed on the [Zeplin Design](https://app.zeplin.io/project/654141b179b50d20560405c8/screen/669e5605d289e0dade00bf00) that the offline instructions are displayed before the notice instead of after, which makes the implementation from this PR slightly different from the original design - check the attached screenshot for reference. 

Honestly, I don't know if it should be a concern. So, I would like to confirm with @jdghinson  if it's ok as it is currently or if we need to change it.

image

tests/Unit/FormBuilder/ServiceProviderTest.php
jdghinson: @glaubersilva Good catch! Yes, the notice should be below all the gateway options

kjohnson:
@glaubersilva thanks. I didn't have the Offline Donations showing when I added this. I'll take a look.

kjohnson:
@glaubersilva it seems that the Offline Donations settings are added by filtering the Block's Edit component, so there isn't a way to control this ordering from within the Payment Gateways block.

We _could_ inject this notice as a filter, which is what I'll try next.

kjohnson:
The order issue is actually due to the use of a Higher Order Component to append the Offline Donation inspector controls, where as the filter is only used to render the block itself.

kjohnson:
@glaubersilva Ok, I was able to re-position the notice using a filter and a higher order component, see [15ca140](https://github.com/impress-org/givewp/pull/7477/commits/15ca140220351c7253f54bc45268e88613c3b344).

image

tests/Unit/FormBuilder/ServiceProviderTest.php
kjohnson: I added the docblocks, see [1cbf0f9](https://github.com/impress-org/givewp/pull/7477/commits/1cbf0f9cf93fc0240862fac308ee0172a0b0661c).

glaubersilva:
@kjohnson Thanks for the changes, this seems more with the design now. However, to compile the files on my local machine I have to delete lines a couple of unused imports on the `FormBuilder/resources/js/form-builder/src/blocks/fields/payment-gateways/Edit.tsx` file.

Besides that, it's all good. So, I'll approve the PR to advance things, but remember to remove this not used imports before sending it to QA. =)

Nice work! 

image

kjohnson:
Thanks, @glaubersilva. I removed those unused imports, see [d1e3cde](https://github.com/impress-org/givewp/pull/7477/commits/d1e3cde84da52f5b31584f7ce922956c2e095d2b).

kjohnson:
Build for QA, see https://github.com/impress-org/givewp/actions/runs/10421224477.

rickalday (approved):
Passed manual QA tests.

PR #7483, opened 2024-08-08
Fix: sites with a lot of forms can cause slow queries on pages with form list views

glaubersilva:
Resolves [GIVE-1079]

## Description

This PR solves the performance issue introduced on the `Give 3.14` related to the long page load which was only happening on pages with lots of forms that had lots of donations and the progress bar enabled. 

Before we didn't face this problem due to the use of meta keys that were used to store the aggregated values for the form stats, since `Give 3.14` we started to calculate this data in real-time to keep the data accurate.

The problem with this new approach is that it can be time and resource-consuming in some sites with huge amounts of data.

So, to solve the problem a new approach to load the form stats data was implemented. Now, we load the admin form list views and the form grid view pages without loading the form stats - these data are loaded through async requests to the server only after the page gets loaded. 

Also, we just load the data of the forms visible on the screen, when the user scrolls down the form list we load the other stats by demand when it gets visible on the screen as well.

This performance issue doesn't happen in all sites. So, if the customer (or the support team) wishes to disable the async approach to the admin form list views or to the form grid is possible to do that by adding the following options on the `wp-config.php` file:

```php
/**
 * ADMIN FORM LIST VIEWS OPTIONS
 */

define('GIVE_IS_ALL_STATS_COLUMNS_ASYNC_ON_ADMIN_FORM_LIST_VIEWS', false);
//define('GIVE_IS_GOAL_COLUMN_ASYNC_ON_ADMIN_FORM_LIST_VIEWS', false);
//define('GIVE_IS_DONATIONS_COLUMN_ASYNC_ON_ADMIN_FORM_LIST_VIEWS', false);
//define('GIVE_IS_REVENUE_COLUMN_ASYNC_ON_ADMIN_FORM_LIST_VIEWS', false);


/**
 * FORM GRID OPTIONS
 */

define('GIVE_IS_ALL_PROGRESS_BAR_STATS_ASYNC_ON_FORM_GRID', false);
//define('GIVE_IS_PROGRESS_BAR_AMOUNT_RAISED_ASYNC_ON_FORM_GRID', false);
//define('GIVE_IS_PROGRESS_BAR_DONATIONS_COUNT_ASYNC_ON_FORM_GRID', false);
```

And if the customer (or support team) wishes to use the meta key aggregated values like it was before `Give 3.14`, it is possible to do that using the following snippet:

```php
<?php

/**
 * It will use the meta keys that store the aggregated values instead of the data retrieved in real-time from DB.
 */
add_filter('givewp_use_cached_form_stats_meta_keys_on_admin_form_list_views', '__return_true');
add_filter('givewp_use_cached_form_stats_meta_keys_on_form_grid', '__return_true');

```

Please note that some servers have rules to prevent too many requests at the same time for the same client as described here in this Cloudflare forum thread: https://community.cloudflare.com/t/error-429-message/625392

So, what can happen is that some servers can return a 429 error when some rules like that are enabled. It also will return a Retry-After header with the time the client needs to wait until trying to do a new request, usually 60 minutes. More details here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429#response_containing_retry-after_header

So, to try to avoid this error, a throttling technique was implemented to limit the requests made at once when the form list pages get loaded. By default, only one async request will be run per time.

But is possible to disable the throttling by adding the following in the `wp-config.php` file:

`define('GIVE_ASYNC_DATA_THROTTLING', false);`

However, it's recommended to let the throttling feature be enabled. Also, It's important to understand that even with this feature enabled the client can get 429 errors due to highly restrictive rate-limiting rules applied to the Server/Firewall - for these cases, the rules should be adjusted or removed.

## Affects

- The legacy admin form list view.
- The admin form list view.
- The form grid list view (shortcode and block).

## Visuals

**Demo video with more details and context:**

https://www.loom.com/share/e9b69a066260435293118b3ad19545a5?sid=4e1437bf-ec62-4d57-9333-4cab66222df7

**IMPORTANT:** A couple of constants and the snippet discussed in the 👆 video above have changed after the recording, so consider only the values presented in this PR's "description" section.

10 minutes to load the legacy form list view page **BEFORE** the changes implemented on this PR:

10-minutes-to-load-all-forms

3 seconds to load the legacy form list view **AFTER** the changes implemented on this PR:

image

## Testing Instructions

1. Setup a site with a lot of data that has the problem - you can find a site sample in the attached Jira task;
2. Setup the Give version of this branch on the problematic site;
3. Access the new and legacy form list views on the admin area;
4. Access a page with the form grid;
5. Make sure all pages from the previous steps got loaded in a reasonable time;

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-1079]: https://stellarwp.atlassian.net/browse/GIVE-1079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/DonationForms/AsyncData/resources/loadAsyncData.js
```diff
@@ -0,0 +1,243 @@
+document.addEventListener('DOMContentLoaded', () => {
```
jonwaldstein, line 1: This whole thing is impressive! can you add doc blocks with comments/descriptions to describe what's going on here?

includes/forms/functions.php
```diff
@@ -1285,9 +1281,18 @@ function give_admin_form_goal_stats( $form_id ) {
...
 		$html .= sprintf( '<span class="give-admin-goal-achieved"><span class="dashicons dashicons-star-filled"></span> %s</span>', __( 'Goal achieved', 'give' ) );
-	}
+	}*/
+
+    $opacity = $goal_stats['raw_actual'] >= $goal_stats['raw_goal'] ? 1 : 0;
+    $html .= sprintf(
+        '<span style="opacity:%s" class="give-admin-goal-achieved"><span class="dashicons dashicons-star-filled"></span> %s</span>',
+        apply_filters('give_admin_goal_progress_achieved_opacity', $opacity),
+        __('Goal achieved', 'give')
+    );
```
jonwaldstein, line 1295: I'm getting this error on legacy view:

Screenshot 2024-08-16 at 11 02 55 AM

jonwaldstein (changes requested):

src/DonationForms/Actions/getAsyncFormDataForListView.php
```diff
@@ -0,0 +1,92 @@
...
+
+namespace Give\DonationForms\Actions;
+
+use Give\DonationForms\DonationQuery;
+use Give\MultiFormGoals\ProgressBar\Model as ProgressBarModel;
+
+/**
+ * @unreleased
+ */
+class getAsyncFormDataForListView
```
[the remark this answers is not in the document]
glaubersilva: Resolved in 09aba60b019cfe9015f83a735188bf424d1c6dad

src/DonationForms/AsyncData/Actions/getAsyncFormDataForListView.php
```diff
@@ -0,0 +1,95 @@
...
+namespace Give\DonationForms\AsyncData\Actions;
+
+use Give\DonationForms\AsyncData\AdminFormListView\AdminFormListViewOptions;
+use Give\DonationForms\AsyncData\FormGrid\FormGridViewOptions;
+use Give\DonationForms\AsyncData\FormStats;
+
+/**
+ * @unreleased
+ */
+class getAsyncFormDataForListView
```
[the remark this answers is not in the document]
glaubersilva: Resolved in 09aba60b019cfe9015f83a735188bf424d1c6dad

src/DonationForms/ServiceProvider.php
```diff
@@ -80,6 +85,52 @@ public function boot()
...
+        $this->registerAsyncData();
+    }
+
+    /**
+     * @unreleased
+     */
+    private function registerAsyncData()
+    {
+        // Async ajax request
+        Hooks::addAction('wp_ajax_givewp_get_form_async_data_for_list_view', getAsyncFormDataForListView::class);
```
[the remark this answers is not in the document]
glaubersilva: Resolved in 09aba60b019cfe9015f83a735188bf424d1c6dad

src/DonationForms/ServiceProvider.php
```diff
@@ -80,6 +85,52 @@ public function boot()
...
+        Hooks::addAction('wp_ajax_nopriv_givewp_get_form_async_data_for_list_view', getAsyncFormDataForListView::class);
```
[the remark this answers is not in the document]
glaubersilva: Resolved in 09aba60b019cfe9015f83a735188bf424d1c6dad

includes/forms/functions.php
```diff
@@ -1285,9 +1281,18 @@ function give_admin_form_goal_stats( $form_id ) {
 		( 'donors' === $goal_stats['format'] ? __( 'donors', 'give' ) : ( 'donation' === $goal_stats['format'] ? __( 'donations', 'give' ) : '' ) )
 	);
 
-	if ( $goal_stats['raw_actual'] >= $goal_stats['raw_goal'] ) {
+    /*if (is_archive()){
+        $html .= sprintf( '<span style="opacity: 0" class="give-admin-goal-achieved"><span class="dashicons dashicons-star-filled"></span> %s</span>', __( 'Goal achieved', 'give' ) );
+    }elseif ( $goal_stats['raw_actual'] >= $goal_stats['raw_goal']) {
```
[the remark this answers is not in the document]
glaubersilva: Resolved in dc4db533885170da872d3ec67c762689f7368a72

assets/src/js/admin/admin.js
```diff
@@ -35,6 +35,8 @@ import './paypal-commerce';
 
 import '../../../../src/DonorDashboards/resources/js/admin';
 
+import '../../../../src/DonationForms/AsyncData/resources/loadAsyncData';
+
```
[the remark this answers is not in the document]
glaubersilva: Resolved in d4059de940b5a1eacf632eaa309cc812fd4c1226

assets/src/js/frontend/give.js
```diff
@@ -20,6 +20,8 @@ import iFrameResizer from '../plugins/form-template/iframe-content';
 import '../plugins/form-template/parent-page';
 import '../../../../src/Form/LegacyConsumer/resources/js/conditional-fields';
 
+import '../../../../src/DonationForms/AsyncData/resources/loadAsyncData';
+
```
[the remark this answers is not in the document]
glaubersilva: Resolved in d4059de940b5a1eacf632eaa309cc812fd4c1226

src/Framework/ListTable/ListTable.php
```diff
@@ -123,7 +123,10 @@ public function getItems(): array
     private function safelyGetCellValue(ModelColumn $column, Model $model, string $locale)
     {
         try {
-            $cellValue = $column->getCellValue($model, $locale);
+            /**
+             * @unreleased
+             */
+            $cellValue = apply_filters("givewp_list_table_cell_value_{$column::getId()}", $column->getCellValue($model, $locale));
```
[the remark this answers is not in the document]
glaubersilva: Resolved in d834bed0170307ac7ab74373d4cbb423aedf125c

includes/misc-functions.php
```diff
@@ -1925,13 +1925,23 @@ function give_get_nonce_field( $action, $name, $referer = false ) {
 	);
 }
 
+/**
+ * @unreleased
+ */
+function give_get_skeleton_placeholder_for_async_data($width = '100%', $height = '0.7rem'): string {
+
+    return '<span class="give-skeleton js-give-async-data" style="width: '. $width . '; height: '. $height . ';"></span>';
```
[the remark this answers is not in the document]
glaubersilva: Resolved in 345ae3a26e2cb7ff9b00b43de4cbb407e976914d

includes/forms/functions.php
glaubersilva: Resolved in bd5028b020901394eae3c50561ae3249f284f8ea

src/DonationForms/AsyncData/resources/loadAsyncData.js
glaubersilva: Resolved in b78516e3c5f33df00e0c6ff2be70dde8a51d8bf7

glaubersilva:
@jonwaldstein Thanks for the suggestions! This is ready for re-review.

includes/misc-functions.php
```diff
@@ -1971,16 +1978,23 @@ function give_goal_progress_stats( $form ) {
...
 		default:
-			/**
-			 * Filter the form income.
-			 *
-			 * @since 1.8.8
-			 */
-			$actual = apply_filters( 'give_goal_amount_raised_output', (new DonationQuery())->form($form->ID)->sumIntendedAmount(), $form->ID, $form );
+
+            if ($usePlaceholder) {
+                $actual = $placeholder;
```
jonwaldstein, line 1983: This part confuses me a bit because we're assigning an html element to a variable that is supposed to be a numeric value right?  

See comment below, this might not be necessary

includes/misc-functions.php
```diff
@@ -1997,17 +2011,17 @@ function give_goal_progress_stats( $form ) {
...
 
 		case 'amount' === $goal_format:
-			$actual     = give_currency_filter( give_format_amount( $actual ) );
+			$actual     = $usePlaceholder ? $placeholder : give_currency_filter( give_format_amount( $actual ) );
 			$total_goal = give_currency_filter( give_format_amount( $total_goal ) );
 			break;
 
 		default:
-			$actual     = give_format_amount( $actual, [ 'decimal' => false ] );
+			$actual     = $usePlaceholder ? $placeholder : give_format_amount( $actual, [ 'decimal' => false ] );
```
jonwaldstein, line 2024: It looks like all these conditionals could be simplified down the chain so they can remain the original way:

```php
$stats_array = array_merge(
		[
			'progress' => $progress,
			// here is the placeholder conditional
			'actual'   => $usePlaceholder ? $placeholder : $actual,
			'goal'     => $total_goal,
			'format'   => $goal_format,
		],
		$stats_array
	);
```

src/DonationForms/AsyncData/Actions/GetAsyncFormDataForListView.php
```diff
@@ -0,0 +1,95 @@
...
+            wp_send_json_error(['errorMsg' => __('Missing Form ID.', 'give')]);
+        }
+
+        if ( ! isset($options['nonce']) || ! check_ajax_referer('GiveDonationFormsAsyncDataAjaxNonce', 'nonce')) {
+            wp_send_json_error([
+                'errorMsg' => __('The current user does not have permission to execute this operation.', 'give'),
+            ]);
+        }
+
+        $formId = $options['formId'];
```
jonwaldstein, line 31: Should make sure this is an int

```suggestion
        $formId = absint($options['formId']);
```

src/DonationForms/AsyncData/Actions/GetAsyncFormDataForListView.php
```diff
@@ -0,0 +1,95 @@
...
+class GetAsyncFormDataForListView
+{
+    /**
+     * @unreleased
+     */
+    public function __invoke()
+    {
+        $options = give_clean($_GET);
+
+        if ( ! isset($options['formId'])) {
```
jonwaldstein, line 21: We should also validate that the `formId` is a DonationForm (and not just an arbitrary post ID or something else)

src/DonationForms/AsyncData/resources/loadAsyncData.js
```diff
@@ -0,0 +1,302 @@
...
+
+        // This class ensures that once the element has the fetch request triggered we'll not try to fetch it again.
+        itemElement.classList.add('give-async-data-fetch-triggered');
+
+        url =
+            window.GiveDonationFormsAsyncData.ajaxUrl +
+            '?action=givewp_get_form_async_data_for_list_view&formId=' +
+            formId +
+            '&nonce=' +
+            window.GiveDonationFormsAsyncData.ajaxNonce;
```
jonwaldstein, line 70: Simplify a little bit, you could also build this url largely in php as well

```suggestion
        const url = `${window.GiveDonationFormsAsyncData.ajaxUrl}?action=givewp_get_form_async_data_for_list_view&formId=${formId}&nonce=${window.GiveDonationFormsAsyncData.ajaxNonce}`;
```

src/DonationForms/AsyncData/resources/loadAsyncData.js
```diff
@@ -0,0 +1,302 @@
...
+
+        // It can be used to abort the async request when necessary.
+        const controller = new AbortController();
+        const signal = controller.signal;
```
jonwaldstein, line 74: Nice!

jonwaldstein (commented):
Getting close!!

I did run into this in the block editor where it was getting stuck loading, I also wonder what would happen if someone was using more than one form grid block 😬 

Screenshot 2024-08-16 at 4 58 43 PM

src/DonationForms/AsyncData/Actions/GetAsyncFormDataForListView.php
glaubersilva: Resolved in ea887b14a050fc16f2c4856cf7eebe8c49d022d7

src/DonationForms/AsyncData/Actions/GetAsyncFormDataForListView.php
glaubersilva: Resolved in ea887b14a050fc16f2c4856cf7eebe8c49d022d7

includes/misc-functions.php
glaubersilva: Resolved in 7d8059274d5a01e5df31058fd2a1fec04ad245d2

includes/misc-functions.php
glaubersilva: I did simplify it, but we still need to prevent the `give_goal_amount_raised_output` from being triggered when using the placeholder.

Resolved in 7d8059274d5a01e5df31058fd2a1fec04ad245d2

glaubersilva:
@jonwaldstein This is ready for re-review. I fixed the form grid on the WordPress block editor and now it should work like in the other places. Also, we can add multiple form grids on the same page because we are checking for the form grid items independently of how many blocks/shortcodes we have on the page.

includes/misc-functions.php
```diff
@@ -1971,12 +1977,14 @@ function give_goal_progress_stats( $form ) {
...
-			 */
-			$actual = apply_filters( 'give_goal_amount_raised_output', (new DonationQuery())->form($form->ID)->sumIntendedAmount(), $form->ID, $form );
+            /**
+             * Filter the form income.
+             *
+             * @unreleased Revert changes implemented on the 3.14.0 version and add placeholder condition
+             * @since 3.14.0 Replace "$form->earnings" with (new DonationQuery())->form($form->ID)->sumIntendedAmount()
+             * @since 1.8.8
+             */
+            $actual = $usePlaceholder ? 0 : apply_filters( 'give_goal_amount_raised_output', $form->earnings, $form->ID, $form );
```
jonwaldstein, line 1987: Now that you have simplified this file, I don't think we even need the `$usePlaceholder` conditionals and filter.  We can hook into the existing filters as they are doing something really simple.  We could make a little helper to be reused for using these filters, but the advantage would be that we are not modifying this function anymore.  Exampe:

```php
if (apply_filters('givewp_admin_list_form_view_goal_progress_stats_use_placeholder', false)){

  add_filter("give_goal_amount_raised_output", function($output) {
      return 0;
  });
  
  add_filter('give_goal_progress_stats', function($stats) {
      $stats['actual'] = AsyncDataHelpers::getSkeletonPlaceholder('1rem');
  
      return $stats;
  });
}
```

src/DonationForms/V2/ListTable/Columns/GoalColumn.php
```diff
@@ -49,9 +49,11 @@ public function getCellValue($model): string
             return __('No Goal Set', 'give');
         }
 
-        add_filter('give_get_form_earnings_stats', function ($earnings, $donationFormId) {
-            return (new DonationQuery())->form($donationFormId)->sumAmount();
-        }, 10, 2);
+        $usePlaceholder = apply_filters('givewp_admin_list_form_view_goal_progress_stats_use_placeholder', false);
```
jonwaldstein, line 52: You could drop an action here and hook in outside of the column so the placeholder logic is not so coupled:

```
do_action("givewp_list_table_cell_value_{$this::getId()}_before", $model);
```

(or we could even add to the ListTable framework)

jonwaldstein (commented):
@glaubersilva on initial load in the form builder the form grid appears to be loading but will fetch on scroll:

Screenshot 2024-08-19 at 11 25 59 AM

includes/misc-functions.php
glaubersilva: @jonwaldstein I think it will not work because passing 0 to the `give_goal_amount_raised_output` filter will not prevent other functions associated with this filter get triggered (including the one that calc the output retrieving real data from DB and slows down the page load) and we are trying to prevent exactly that. 

So, what the current logic does is prevent the `give_goal_amount_raised_output` filter from being triggered when using the placeholder and we need to do that to avoid triggering the function that slows the page load.

glaubersilva:
@jonwaldstein I'm not able to reproduce the problem you mentioned related to the Form Grid Block on the editor. Check out this 30 seconds video: https://www.loom.com/share/96f0c0475b9345f9b5c32418a810d5ed?sid=b5e7b10d-b4a0-4dcf-8ca3-e191ae3e9d07

src/DonationForms/V2/ListTable/Columns/GoalColumn.php
glaubersilva: Resolved in 149593253aa596f97009b26db3a96e69221de70c and 5992e441fd4870514cf99cb76116217ccdccf746

includes/misc-functions.php
glaubersilva: Ok, I think I found a way to make it work and was able to fully remove the filter to enable the placeholder. Now everything is enabled on the Service Provider. Resolved in 86784519beced7fd0ca1cd7d2248853091da4cd4 and 6a795e52fd1cc0a629c49b3a88701079debcf723 and 882da69f1cce0ad6562ab4d31aa409bd1ad2b85f and 687caa50e31ba3b9b0f60ead1b491bb6e1c713a7 and 567f8d0f9306da2fb6015ddc159ad91947affe26

src/DonationForms/ServiceProvider.php
```diff
@@ -80,6 +89,114 @@ public function boot()
...
+
+                    return $stats;
+                });
+            }
+        });
+        Hooks::addFilter('give_form_grid_progress_bar_amount_raised_value', FormGridView::class, 'maybeSetProgressBarAmountRaisedAsync',10,2);
+        Hooks::addFilter('give_form_grid_progress_bar_donations_count_value', FormGridView::class, 'maybeSetProgressBarDonationsCountAsync',10,2);
+
+        // Load assets on the admin list form pages and pages with the WordPress clock editor enable
+        $isAdminListFormPage = isset($_GET['post_type']) && 'give_forms' === $_GET['post_type'];
```
jonwaldstein, line 135: Looking at the methods in `Give\DonationForms\V2\DonationFormsAdminPage` like `::isShowing` we will need to account for this because our menu items have two links:
```
isset($_GET['page']) && $_GET['page'] === 'give-forms'
```

It looks like they both share the page === give-forms

jonwaldstein: Note: this is referring to clicking on the top level "Donations" item vs "All forms"

Screenshot 2024-08-21 at 1 54 10 PM

glaubersilva: I made a few changes to make sure the assets will be loaded only when necessary. Resolve in 74640b1f13df4db712f1a4786343c5de11f7d299 and 3cd5778205dc34ffb6b91d1c718e0e3cd96dd1d3

glaubersilva:
@jonwaldstein Thanks for all the great suggestions. This is ready for re-review.

src/DonationForms/ServiceProvider.php
```diff
@@ -80,6 +89,121 @@ public function boot()
...
+                    return $stats;
+                });
+            }
+        });
+        Hooks::addFilter('give_form_grid_progress_bar_amount_raised_value', FormGridView::class, 'maybeSetProgressBarAmountRaisedAsync',10,2);
+        Hooks::addFilter('give_form_grid_progress_bar_donations_count_value', FormGridView::class, 'maybeSetProgressBarDonationsCountAsync',10,2);
+
+        // Load assets on the admin list form pages
+        $isLegacyAdminFormListPage = isset($_GET['post_type']) && 'give_forms' === $_GET['post_type'] && !isset($_GET['page']);
+        $isAdminFormListPage = isset($_GET['post_type']) && 'give_forms' === $_GET['post_type'] && isset($_GET['page']) && 'give-forms' === $_GET['page'];
```
jonwaldstein, line 142: @glaubersilva sorry this still does not account for the initial page when clicking the "Donations" menu item:
This page does not have the `post_type` param, just the `page` arg:

```
isset($_GET['page']) && $_GET['page'] === 'give-forms'
```

Ref: https://github.com/impress-org/givewp/blob/f83c6c9bc01a77a4ddc76b8adb0b3816e97d4e14/src/DonationForms/V2/DonationFormsAdminPage.php#L268

Screenshot 2024-08-21 at 4 18 28 PM

glaubersilva: @jonwaldstein Oh ok, I had misunderstood it before. Resolve in 6cfe563590cc2ecf0c54542b5b113d6a5393ebef

jonwaldstein (approved):
@glaubersilva great job man! this is really working well and it was built in such a pluggable way 🔌 Let's battle test this with QA! 🌈

rickalday (approved):
Passed manual QA tests.

PR #7485, opened 2024-08-09
Fun: honeypot field

jonwaldstein:
Resolves [GIVE-1160]

## Description

This adds a simple honey pot field to GiveWP v3 forms.

This can be enabled with a filter:

```php
add_filter('givewp_donation_forms_honeypot_enabled', '__return_true')
```

Per form:

```php
add_filter('givewp_donation_forms_honeypot_enabled', function($enabled, $formId){
    if ($formId === 1) {
        $enabled = true;
    }

    return $enabled;
}, 10, 2);
```

The next step would be to create a UI setting in GiveWP to enable either per-form and/or globally.

## Affects

- v3 donation forms

## Visuals

Screenshot 2024-08-30 at 9 17 32 AM

Screenshot 2024-08-30 at 9 18 00 AM

## Testing Instructions

- Make a donation form request
- Add a value to the "donationBirthday" field
- You should get an error

## Pre-review Checklist

-   [x] Includes unit tests

[GIVE-1160]: https://stellarwp.atlassian.net/browse/GIVE-1160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/DonationForms/Actions/AddHoneyPotFieldToDonationForms.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * @unreleased
+     * @throws EmptyNameException
+     */
+    public function __invoke(DonationForm $form): void
+    {
+        $formNodes = $form->all();
+        $lastSection = $form->count() ? $formNodes[$form->count() - 1] : null;
+
+        if ($lastSection) {
```
kjohnson, line 24: Wouldn't there always be a last section, even if there is only one section?

tests/Unit/DonationForms/Rules/TestHoneyPotRule.php
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * @unreleased
+     * @dataProvider honeyPotProvider
+     */
+    public function testHoneyPotRule($value, bool $shouldBeValid): void
+    {
+        $rule = new HoneyPotRule();
+
+        if ($shouldBeValid) {
```
kjohnson, line 26: Would it be better to have a `validProvider` and a `invalidProvider` with separate tests, so that the test names are more precise (`testHoneyPotValidatesEmpty`, `testHoneypotInvalidatesSubmittedValues`) and the added logic within each test is reduced?

src/DonationForms/Actions/AddHoneyPotFieldToDonationForms.php
jonwaldstein: Ideally yes, however if for some reason there are no sections i'm preventing that with the check for `$form->count()`

tests/Unit/DonationForms/Rules/TestHoneyPotRule.php
kjohnson: Ah, ok. That makes sense. In that case, let's not fiddle with it at this time.

src/DonationForms/Actions/AddHoneyPotFieldToDonationForms.php
kjohnson: Ok. We would probably have other problems if there are _no sections_, but let's at least not make things worse here.

kjohnson:
@jonwaldstein Something to consider, as a SPAM prevention tactic, I think it might be _better_ to throw a higher level error which does not identify the SPAM detection field.

Specifically, as SPAM bots get more sophisticated, I think that adding an error to the honeypot field makes it easier to formulate a targeted attack.

image

For example, the Akismet integration hooks into `givewp_donate_form_data_validated` and throws an exception that is obfuscated by the donation process (but still logs with context).

jonwaldstein:
@kjohnson I like that idea, let me see if I can get that to work in the context of a form field.

jonwaldstein:
@kjohnson I updated the HoneyPot rule to throw an exception instead of failing which results in an unknown error in the response.  I also hid the UI error states for the field template to really confuse those bots 🤖.  My only hesitation with this approach is that throwing an exception inside the rule kind of breaks the rule convention of _failing_ itself but in this context I think it make sense.

This whole thing also let's folks add as many honeypot fields as they want to a form without having to deal with styling and such.

kjohnson (approved):
Looks good to me. I think the form level exception will work well here.

Good addition for Fun Day Friday projects entering production!

rickalday (approved):
Passed QA tests

PR #7491, opened 2024-08-16
Fun: UserPermissions facade

jonwaldstein:
Resolves #

## Description

The UserPermissions Facade is a class that allows you to check if a user has a specific capability or role but not necessarily in the _traditional_ way.  The DX focuses on the resource, not the capability, so we don't have to remember what specific caps or roles have access to the resource.   

It is a wrapper around the WordPress `current_user_can()` function but geared toward the roles in GiveWP.

Example:

```php
use Give\Framework\Permissions\Permissions;

if (!UserPermissions::donationForms()->can('edit')) {
    throw new Exception('You do not have permission to edit donation forms.');
}

if (!UserPermissions::donors()->can('delete')) {
    throw new Exception('You do not have permission to delete donors.');
}
```

## Affects

Nothing yet, but eventually would be used for validating permissions

## Visuals

N/A

## Testing Instructions

- Can test programmatically

## Pre-review Checklist

-   [ ] Includes unit tests


jonwaldstein:
This was a good spike/exploration but i'm not sure I like the api anymore 😄 .  I'm going to close for now and potentially explore a different, simpler direction that doesn't try to abstract away too much.

src/Framework/Permissions/Facades/UserPermissions.php
```diff
@@ -0,0 +1,23 @@
...
+use Give\Framework\Permissions\DonationPermissions;
+use Give\Framework\Permissions\DonorPermissions;
+use Give\Framework\Support\Facades\Facade;
+
+/**
+ * @unreleased
+ *
+ * @method static DonationFormPermissions donationForms()
+ * @method static DonationPermissions donations()
+ * @method static DonorPermissions donors()
```
pauloiankoski, line 21: The `@method` annotations in the facade class seem to be incomplete.

```suggestion
 * @method static DonationFormPermissions donationForms()
 * @method static DonationPermissions donations()
 * @method static DonorPermissions donors()
 * @method static CampaignPermissions campaigns()
 * @method static ReportsPermissions reports()
 * @method static SensitiveDataPermissions sensitiveData()
 * @method static SettingsPermissions settings()
```

src/Framework/Permissions/Traits/WithPrivatePermissions.php
```diff
@@ -0,0 +1,24 @@
...
+trait WithPrivatePermissions
+{
+    /**
+     * Check if user can view items with restricted visibility.
+     *
+     * @unreleased
+     */
+    public function canViewPrivate(): bool
+    {
+        return $this->currentUserCan('read_private');
```
pauloiankoski: Since this trait relies on a method that may not exist in the context where it’s used, do you think we could add a check like `method_exists($this, 'currentUserCan')` before calling it?

pauloiankoski (changes requested):
Left comments about some small details. One last question: do you think it would be worth registering `UserPermissionsFacade::class` as a singleton in a service provider? The idea would be to prevent multiple instances of that class from being created.

src/Framework/Permissions/Facades/UserPermissions.php
jonwaldstein: oh yeah good catch!

src/Framework/Permissions/Traits/WithPrivatePermissions.php
jonwaldstein: Yeah you're right, I ended up removing it as we can tackle more specifics needed later during implementation de14bd66

jonwaldstein:
@pauloiankoski I made some updates and I have some more ideas but will wait until implementation before overcomplicating this 😄 

Regarding your question about registering this as a singleton - the Facade class actually already[ does that for us](https://github.com/impress-org/givewp/blob/42b82ca57a18193c46fd320c6a26132ec4e9529d/src/Framework/Support/Facades/Facade.php#L35).  The only difference is it's done at runtime (which I think make sense for permissions) so the first time you use the Facade it will become a singleton and subsequent usage will use the same instance.

pauloiankoski (approved):
Looks really good. This will make working with permissions much easier.

PR #7493, opened 2024-08-20
Fix: exclude Tests folder from strauss

jonwaldstein:
Resolves [GIVE-1136]

## Description

While using Strauss for the `"symfony/http-foundation"`, it was copying over its Tests folder.  This adds the Tests folder to the list of files to exclude during copy.

## Affects

- The composer build process
- This should not have an affect on production as we are only removing a dev testing folder

## Visuals

Screenshot 2024-08-20 at 9 42 17 AM

## Testing Instructions

Build: https://github.com/impress-org/givewp/actions/runs/10472792166

Run composer install (delete vendor-prefixed if already exists) and make sure this `Tests` folder is not present:

Screenshot 2024-08-20 at 9 06 08 AM

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-1136]: https://stellarwp.atlassian.net/browse/GIVE-1136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


composer.json
```diff
@@ -104,6 +104,9 @@
             "exclude_from_copy": {
                 "packages": [
                     "symfony/deprecation-contracts"
+                ],
+                "file_patterns": [
+                    "/Tests/"
```
JasonTheAdams, line 109: @jonwaldstein Is this case-sensitive? Do we want to add a `"/tests/"` line as well?

JasonTheAdams: Fair enough. I just wasn't sure if adding that would protected further into the future, when new packages are added.

jonwaldstein: Yeah I thought about adding additional checks but it appears this is not common for other packages to include their tests folder, this one just seems to be the outlier - maybe it's not respecting the --no-dev flag or something

JasonTheAdams: Ok! Since you looked into it I'm good!

JasonTheAdams (approved):

rickalday (approved):

PR #7500, opened 2024-08-21
Fix: Text domain in index.tsx

DAnn2012:
Fix: Text domain in index.tsx


jonwaldstein (approved):
@DAnn2012 thanks for your contribution!

PR #7501, opened 2024-08-21
Fix: Made string translatable in class-settings-advanced.php

DAnn2012:
Fix: Made string translatable in class-settings-advanced.php


jonwaldstein (approved):
@DAnn2012 thanks for your contribution!

PR #7515, opened 2024-08-23
Fix: prevent Stripe Payment Element from rendering when amount is zero

jonwaldstein:
Resolves [GIVE-509]

## Description

This prevents the Stripe Payment Element from rendering when the donation amount is zero by adding a conditional before loading the fields.

## Affects

Stripe payment element gateways

## Visuals

Before:

Screenshot 2024-08-23 at 9 29 07 AM

After

Screenshot 2024-08-23 at 10 10 35 AM

## Testing Instructions

- create a v3 form
- specify 0 for the amount
- click on stripe payment element

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-509]: https://stellarwp.atlassian.net/browse/GIVE-509?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jonwaldstein:
Closing for now:

> I started working on this for Stripe but quickly realized the nuances become a weird experience.  When 0 is entered we can display a message within the gateway but you can still technically submit the form and receive an error. Furthermore, Stripe uses the amount value, while events are not included in this value.

src/PaymentGateways/Gateways/Stripe/StripePaymentElementGateway/stripePaymentElementGateway.tsx
```diff
@@ -212,6 +213,10 @@ const stripePaymentElementGateway: StripeGateway = {
             appearance: appearanceOptions,
         };
 
+        if (donationAmount <= 0) {
+            return <>{__('Donation amount must be greater than zero to proceed.', 'give')}</>;
+        }
```
JasonTheAdams, line 218: @jonwaldstein Should this be resolved upstream? Might this affect other gateways as well?

jonwaldstein: It's a good question and something I have considered [and even spiked](https://github.com/impress-org/givewp-zero-dollar-donation/blob/main/resources/js/GatewaysField.jsx)

However, In order to accomplish this upstream the options I see are:
1. Assume all gateways will not render fields when amount is 0
2. The gateway API has a concept of accepts $0 and each gateway implements (test gateway, offline, etc would potentially return true)

These options further solidify our rejection of $0 donations which we're still working out in our product.  So at this point, I think it's okay for the gateways to handle this themselves (which most are doing) - at least until we come to a solid conclusion and direction for our product around the subject.

JasonTheAdams: I'm not sure if this helps, but I know that we're planning _at some point_ to allow $0 donations for things like events. I think, at that time, we're actually more likely to do something like hide the gateways. In the meantime, I think we can simply enforce the current assumption that donations _must_ be at least $1 — or whatever the actual minimum donation value is.

I think it's fine to go this route for now, knowing we may change it later.

jonwaldstein: Yeah hiding the gateways was an option on my $0 spike. That was something I built for the marketing team to allow $0 donations.  I'd like to pull out some of what I did over there for core.

I do also like the idea of using the minimum amount value too 🤔

However, the scope of this PR is so small and would solve the immediate weirdness.

I'd be happy to open a separate one that ports over the gateway filtering to play around with _all gateways_ which of course will increase the scope.

jonwaldstein: Started scaffolding the other PR https://github.com/impress-org/givewp/pull/7599

JasonTheAdams: Up to you whether you want to merge this PR to fix the immediate issue and proceed or not. This would ultimately become a tiny amount of technical debt.

jonwaldstein: Yeah, I'd like to merge this for the immediate issue, as keeping it won't hurt anything.  There are some nuances to the _all gateways_ approach that i'd like to spend a little more time testing.  After that I can choose to remove this piece.

rickalday (approved):
Passed manual QA tests.

PR #7519, opened 2024-08-26
Create campaign tables

alaca:
## Description

This PR is reusing the `give_campaigns` table created by P2P.  Also, it adds a new column to the `give_campaigns` table - `campaign_type`. The `give_core_campaign_forms` table is a lookup table that exists solely to avoid possible performance issues when using the post meta table.

## Pre-review Checklist


JasonTheAdams:
@alaca The intent is to re-use the existing P2P `give_campaigns` table, right? Ultimately, we want the two merged into a single table with a `type` that differentiates the various campaign types.

alaca:
@JasonTheAdams we talked about it and the plan was to create a new table, at least I understood it like that. The campaigns table created from P2P seems too oriented on P2P (it has columns like `form_id`, `campaign_url`, etc. ) which is something we don't need here. But yeah, we can definitely make use of the `give_campaigns` table.

JasonTheAdams:
@alaca I see, however I disagree with that conclusion. Ultimately, we want to be able to query all the campaigns at once. It sounds more like we need to offload some of the columns from the existing table to another P2P-specific table. It was intended to be generic when we built it, seems like some specific columns snuck their way in when we built it 3 years ago. 🤷‍♂️

JasonTheAdams:
@kjohnson Yeah, so we'll want to make two migrations:
1. A conditional one in core that creates the table if it doesn't already exist
2. A P2P migration to move some of the P2P-specific columns into the `give_p2P_campaigns` table

alaca:
@JasonTheAdams @kjohnson okay, then we will use the existing `give_campaigns` table. We will just have to add some conditional stuff.

JasonTheAdams:
Thanks @alaca! You're the best! 🙌

JasonTheAdams:
As a note, @alaca, this is a good time to assess the existing columns — names, types, etc. — and make sure they're what we want. Since we're making migrations to move and adjust things now, best to give it some thought and not have to make more later. I think it's worth some discussion.

alaca:
@JasonTheAdams @kjohnson I updated the table name and the columns list to be the same as the one created from P2P. 
And the awesome thing here is that we don't have to add any checks because we use `dbDelta`. `dbDelta` will check if that table exists, and if not, it will create one.  Also, it will check the columns and update the table if needed. So the only extra column I have to add here is `campaign_type`.

src/Campaigns/Migrations/Tables/CreateCampaignFormsTable.php
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * @inheritDoc
+     * @throws DatabaseMigrationException
+     */
+    public function run(): void
+    {
+        global $wpdb;
+
+        $table = $wpdb->prefix . 'give_core_campaign_forms';
```
kjohnson, line 48: I think we can drop the references to `core` in these migrations (title, tables, etc).

alaca: I agree!

src/Campaigns/Migrations/Tables/CreateCampaignsTable.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        global $wpdb;
+
+        $table = $wpdb->prefix . 'give_campaigns';
+        $charset = DB::get_charset_collate();
+
+        $sql = "CREATE TABLE $table (
+			id INT UNSIGNED NOT NULL AUTO_INCREMENT,
+			form_id INT NOT NULL,
+			campaign_type VARCHAR(12) NOT NULL DEFAULT '',
```
kjohnson, line 54: @alaca since this has previously existed (in P2P), could you please document any changes using a `@since/@unreleased` tag in the doc block. I think we need to treat this as a change and not necessarily as something "new".

kjohnson, line 54: We'll need to add a one-time migration that sets the `campaign_type` for preexisting P2P campaigns.

src/Campaigns/Migrations/Tables/CreateCampaignsTable.php
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * @inheritDoc
+     * @throws DatabaseMigrationException
+     */
+    public function run()
+    {
+        global $wpdb;
+
+        $table = $wpdb->prefix . 'give_campaigns';
```
kjohnson, line 48: In P2P we registered the table name in the Service Provider to avoid the need for prefixing elsewhere.

```php
public function register() {
  global $wpdb;
  $wpdb->give_campaigns = "{$wpdb->prefix}give_campaigns";
}
```

src/Campaigns/Migrations/Tables/CreateCampaignsTable.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        return 'Create give_campaigns table from core';
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public static function timestamp(): string
+    {
+        return strtotime('2024-08-26 00:00:00');
```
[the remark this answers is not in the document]
alaca: I think it doesn't matter, but yeah, we can do that.

src/Campaigns/Migrations/Tables/CreateCampaignsTable.php
alaca: Yes, but that will work only if P2P is installed on the site.

src/Campaigns/Migrations/Tables/CreateCampaignsTable.php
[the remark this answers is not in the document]
kjohnson: The timestamp determines run order, if I understand correctly.

It may not be an issue in this particular instance, but I think it would be prudent to maintain the timestamp as a general rule (not that we have ever moved a migration like this).

src/Campaigns/Migrations/Tables/CreateCampaignsTable.php
kjohnson: Ah, yes, I mean that we should move over the table name registration to core also.

kjohnson: I commented on the migration, but the definition would need to be added to the Service Provider.

src/Campaigns/Migrations/Tables/CreateCampaignsTable.php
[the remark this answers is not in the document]
alaca: > The timestamp determines run order, if I understand correctly.

Yes, and we want this migration to run later than the migration from P2P (if P2P is already installed). This one will add the `campaign_type` column if the give_campaign table exists. Also, we will have to add a check (give version probably) in P2P to conditionally run P2P migration for creating the campaign table.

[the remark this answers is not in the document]
kjohnson: > This one will add the campaign_type column if the give_campaign table exists.

Good point. Should we introduce these changes as an additional migration?

You mentioned previously that `dbDelta` can handle this automatically, but we haven't used it that way. Rather, we've used separation migrations that run synchronously. Perhaps we should port the "add campaign" migration and in addition introduce a new migration that modifies the original table.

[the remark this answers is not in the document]
kjohnson: > we will have to add a check (give version probably) in P2P to conditionally run P2P migration for creating the campaign table.

We won't necessarily have to add a logic check, if we leverage the migration and timestamps.

If the previous migration has run via P2P, then porting the migration and the timestamp should mean that it would inherently not run again.

src/Campaigns/Migrations/Tables/CreateCampaignsTable.php
alaca: Resolved in another PR: https://github.com/impress-org/givewp/pull/7520/commits/d52f7697bd6b1383bf98eaf2b0aa3e6ca1272f6a

src/Campaigns/Migrations/Tables/CreateCampaignsTable.php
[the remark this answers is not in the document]
alaca: > Good point. Should we introduce these changes as an additional migration? You mentioned previously that dbDelta can handle this automatically, but we haven't used it that way. Rather, we've used separation migrations that run synchronously. Perhaps we should port the "add campaign" migration and in addition introduce a new migration that modifies the original table.

I think I know how you feel about using `dbDelta`, I'm a little anxious about using that as well 😅  
But `dbDelta` does the job. I'm perfectly fine going with the "standard" way of doing things as it adds a bit more clarity to what is going on. So yeah, let's add column type as a separate migration.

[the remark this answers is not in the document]
alaca: Resolved: https://github.com/impress-org/givewp/pull/7519/commits/986cfe14895a177431734823a8b4d121549cb676

kjohnson (commented):
Approved (forgot to click this button before merging into the epic).

PR #7524, opened 2024-08-27
Feature: add list table for campaigns

glaubersilva:
Resolves [GIVE-1151]

## Description

This PR implements the campaigns list table and a few API routes to implement CRUD operations.

The new UI allows users to view, create, update, and delete campaigns. 

We still need to work on the design, form inputs, and the columns that we want to display in the list table - it should be done in subsequent PRs.

## Affects

The campaigns admin page.

## Visuals

image

image

image

image

## Testing Instructions

1. Go to `Give > Campaigns` page;
2. Create a few campaigns;
3. Use the Search box to filter a specific campaign;
4. Use the Delete action to remove a specific campaign;
5. Make sure everything worked as expected.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-1151]: https://stellarwp.atlassian.net/browse/GIVE-1151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Campaigns/Actions/LoadCampaignsListTableAssets.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Give\Campaigns\Actions;
+
+use Give\Campaigns\ListTable\CampaignsListTable;
+
+/**
+ * @unreleased
+ */
+class LoadCampaignsListTableAssets
```
alaca, line 10: Why do we need these static functions that are only used here? Wouldn't it be better to have all this in the `__invoke` method, and add just hook up to an event? Of course, we will have to add a check for the `post_type` so that assets are loaded only on the campaigns page.

src/Campaigns/ListTable/Columns/StatusColumn.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    /**
+     * @unreleased
+     *
+     * @param Campaign $model
+     */
+    public function getCellValue($model)
+    {
+        return $model->status->getValue();
```
alaca, line 36: We should return a translatable string here, not the status value. 
Something like:
```php
switch($model->status->getValue()){
     case 'active':
          return __('Active', 'give');
     ...
}
```

src/Campaigns/Routes/GetCampaignsListTable.php
```diff
@@ -0,0 +1,185 @@
...
+    public function handleRequest(WP_REST_Request $request): WP_REST_Response
+    {
+        $this->request = $request;
+        $this->listTable = give(CampaignsListTable::class);
+
+        $campaigns = $this->getCampaigns();
+        $campaignsCount = $this->getTotalCampaignsCount();
+        $pageCount = (int)ceil($campaignsCount / $request->get_param('perPage'));
+
+        if ('model' === $this->request->get_param('return')) {
```
alaca, line 97: Not sure why we have this.

src/Campaigns/resources/admin/components/CampaignsListTable/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+import ListTableApi from '@givewp/components/ListTable/api';
+import {BulkActionsConfig, FilterConfig} from '@givewp/components/ListTable/ListTablePage';
+import {IdBadge} from '@givewp/components/ListTable/TableCell';
+import {Interweave} from 'interweave';
+import {CampaignsRowActions} from './CampaignsRowActions';
+import styles from './CampaignsListTable.module.scss';
+import {GiveCampaignsListTable} from './types';
+import CreateCampaignModal from '../CreateCampaignModal';
+
+declare global {
```
alaca, line 12: Let's use a bit different approach here. 
```ts
declare const window: {
    GiveCampaignsListTable
} & Window;
```

The problem with `declare global` is that all projects opened in the IDE will reference this. Let's keep this local.

src/Campaigns/resources/admin/components/CampaignFormModal/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+                aria-disabled={!isDirty}
+                disabled={!isDirty}
+            >
+                {campaign?.id ? __('Save changes', 'give') : __('Save campaign', 'give')}
+            </button>
+        </FormModal>
+    );
+}
+
+type Campaign = {
```
alaca, line 128: Maybe this is nitpicking, but I would love to see interfaces and types defined at the top of the file.

alaca (commented):
Hey @glaubersilva, good job man. 
There are some minor issues, but nothing special. I left a few comments, and questions.

src/Campaigns/Actions/LoadCampaignsListTableAssets.php
glaubersilva: Resolved in 274266fc4411b0e9904b517a22bcb7ec3e45526b

src/Campaigns/ListTable/Columns/StatusColumn.php
glaubersilva: Resolved in 3e1af62cccdf32ef86d02f13e54f19d9a599598f

src/Campaigns/Routes/GetCampaignsListTable.php
glaubersilva: To be honest, I also don't know, but I believe it can be used somehow for our List Table component. I didn't check it deeper, but if you run a search in the codebase looking for `'model' === $this->request->get_param('return')` you'll see that it's used in all ListTables implemented.

src/Campaigns/resources/admin/components/CampaignsListTable/index.tsx
glaubersilva: Resolved in c09555969c69e626938f96c89b83ecaee94a4bd3 and ed96921e53a00205882df49881b9c6aac5f5995d

src/Campaigns/resources/admin/components/CampaignFormModal/index.tsx
glaubersilva: Resolved in 505c618cb7261de08d622eee69c5ff24137b1728

glaubersilva:
@alaca Thanks for the suggestions, this is ready for re-review.

src/Campaigns/Routes/GetCampaignsListTable.php
alaca: I'm unsure why this exists in other places, but it's wrong. PHP does not have a `__toArray` magic method to convert objects to an array. Outputting an array of objects like that will not do anything. So yeah, go ahead and remove this. 
Maybe I'm missing something so I will ask for clarification in Slack from others on why we have this in the first place.

glaubersilva: I agree with you, and thinking about what you said about `__toArray` I searched on the tests and I could see it being used there as you can check here: https://github.com/impress-org/givewp/blob/develop/tests/Unit/DonationForms/Repositories/DonationFormsRepositoryTest.php#L43

So seems that the only reason to have this is to make the unit tests easier. Not sure if is useful for anything else, let's check with the team.

glaubersilva: @alaca Thank you! Resolve in 5d1b45f7538c90e27376b6941762543c7e232a9b

PR #7525, opened 2024-08-28
Add campaign id to Donation Form model

alaca:
## Description

This PR updates the Donation Form model by adding a new property - `campaignId`

## Pre-review Checklist


kjohnson:
@alaca I don't think we need to add this property directly to the Donation Form, since it is a one-of-many with the Campaign.

If anything, wouldn't we populate this information from the lookup table, instead of storing as post meta?

alaca:
@kjohnson looking at it now, I think you are right. We don't need this. I will close this PR. Thanks!

PR #7526, opened 2024-08-28
Feature: Relate Campaigns to Donation Forms with one-to-many relationship

kjohnson:
## Description

This PR adds the `forms()` relationship method to the `Campaign` model.

Introduces query builder logic for relating Donation Forms to Campaigns by leveraging the `give_campaign_forms` lookup table.

## Affects

Adds an alias (`{prefix}_posts` as `forms`) in the `DonationForm` prepared query.

Updates the `PRIMARY` key from the `give_campaign_forms` table to support many forms per campaign.

## Testing Instructions

```php
use Give\Campaigns\Models\Campaign;

/** @return DonationForm[] */
Campaign::find(17)
  ->forms()
  ->getAll();
```

## Pre-review Checklist

-   [x] Includes unit tests


src/Campaigns/Migrations/Tables/CreateCampaignFormsTable.php
```diff
@@ -51,8 +51,7 @@ public function run(): void
         $sql = "CREATE TABLE $table (
             campaign_id INT UNSIGNED NOT NULL,
             form_id INT UNSIGNED NOT NULL,
-            PRIMARY KEY  (campaign_id),
-            KEY form_id (form_id)
+            PRIMARY KEY (campaign_id, form_id),
```
kjohnson, line 54:
```suggestion
            PRIMARY KEY (campaign_id, form_id)
```

src/Campaigns/Migrations/Tables/CreateCampaignFormsTable.php
```diff
@@ -51,8 +51,7 @@ public function run(): void
         $sql = "CREATE TABLE $table (
             campaign_id INT UNSIGNED NOT NULL,
             form_id INT UNSIGNED NOT NULL,
-            PRIMARY KEY  (campaign_id),
-            KEY form_id (form_id)
+            PRIMARY KEY (campaign_id, form_id)
```
alaca, line 54: I'm pretty sure we don't need a composite key here. Form ID is unique for each row, so adding an index on that should be sufficient - which we already have. It will improve performance in case we want to query the table by using the Form ID only. 

Composite keys are great when you query the table on multiple criteria, and you have a bunch of columns that are not unique, but that is not the case here. 
For example, we will probably most of the time query this table by using the Campaign ID only which is already a primary key, and not by Campaign ID and Form ID at the same time (where having the composite key will make sense).  

Also, if for some reason we want to create a new table that has a foreign key constraint on the campaign_id from this table, we will have to reference both campaign_id and form_id as foreign keys - an unlikely scenario, but just something to keep in mind.

alaca (commented):
This is great. I only have one question about using composite keys. I tried to elaborate in the comment on why I'm against using them here.

src/Campaigns/Migrations/Tables/CreateCampaignFormsTable.php
kjohnson: I added the composite key as primary to enforce a unique combination, as I don't want duplicate rows (which would result in duplicate forms when fetching the one-to-many relationship).

kjohnson: We could also have a `KEY` on both the `campaign_id` and `form_id` for query performance.
Although, at some point, given how simple this table is, does adding multiple keys potentially reduce the performance? Or is the only downside table size increase from the additional keys?

kjohnson: I added index keys for both `campaign_id` and `form_id`, see [c1e03c8](https://github.com/impress-org/givewp/pull/7526/commits/c1e03c84d66c2d7b8b0b3f0f5590ef7680765403).

alaca: Awesome, thanks!

alaca (approved):
Great job, @kjohnson

PR #7532, opened 2024-09-04
Refactor: Update Forms to Camapigns migration to be closer to the data layer

### LINKED ISSUE #1178: a11y: Donation form required fields
## Issue Overview

We need to use `aria-required="true"` on required fields.

required
## Current Behavior

If the field is required, the HTML is `<input required>`.
## Expected Behavior

The HTML should be`<input required aria-required="true">` for screen-readers.
## Addons

We should also update all our addons - Mainly the "Form Field Manager". To confirm they add ARIA tags in the front-end.

kjohnson:
Resolves #1178

## Pre-review Checklist

-   [ ] Includes unit tests


src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
```diff
@@ -38,57 +38,99 @@ public static function timestamp(): int
...
-        ]);
+        $formId = $formData->id;
+        $formTitle = $formData->title;
+        $formStatus = $formData->status;
+        $formCreatedAt = $formData->createdAt;
+        $formSettings = json_decode($formData->settings);
+
+        $campaignId = DB::table('give_campaigns')
+            ->insert([
+                'campaign_type' => CampaignType::CORE()->getValue(),
```
JasonTheAdams, line 89: @kjohnson Even this, I would put in the actual value and not reference the VO, as we'd potentially have a migration later that changes this value, moves the VO, etc.

glaubersilva (approved):
@kjohnson Looks good to me! I just left a comment related to a minor change, but I'll approve it to advance things.

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
```diff
@@ -38,57 +38,99 @@ public static function timestamp(): int
      */
     public function run()
```
glaubersilva, line 39: Let's add a `@throws Exception` to the doc block.

glaubersilva (approved):

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
kjohnson: Resolved, see [bde30e0](https://github.com/impress-org/givewp/pull/7532/commits/bde30e049132c1d485ff58253ffd30d2b2044ff8).

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
```diff
@@ -29,66 +26,118 @@ public static function id(): string
...
+                return 'draft';
+
+            case 'trash':
+                return 'inactive';
 
-            case DonationFormStatus::DRAFT():
-                return CampaignStatus::DRAFT();
+            case 'publish':
+            case 'private':
+            case 'upgraded':
```
kjohnson, line 136: On second thought, I think the `upgraded` status only applied after a v2 form has been migrated to a v3 (new) form and the donations have been transferred, which means that forms with the `upgraded` status should probably be excluded from this migration.

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
kjohnson: Resolved, see [7ec7b59](https://github.com/impress-org/givewp/pull/7532/commits/7ec7b59f4808b86f05b30fb08e98f64e1ecf8cb6).

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
kjohnson: See [fe3f605](https://github.com/impress-org/givewp/pull/7532/commits/fe3f605c7c5424627bc9d9dec1156c5d128d47d2)

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
```diff
@@ -29,66 +26,153 @@ public static function id(): string
...
+        return DB::table('posts', 'forms')
+            ->select(['forms.ID', 'formId'], ['campaign_forms.campaign_id', 'campaignId'])
+            ->attachMeta('give_formmeta', 'ID', 'form_id', 'migratedFormId')
+            ->join(function (JoinQueryBuilder $builder) {
+                $builder
+                    ->rightJoin('give_campaign_forms', 'campaign_forms')
+                    ->on('campaign_forms.form_id', 'forms.ID');
+            })
+            ->where('forms.post_type', 'give_forms')
+            ->where('forms.post_status', 'trash', '!=')
```
glaubersilva, line 101: Not sure if we should exclude forms with the trash status. I know that our list tables don't provide a way to restore items with this status, but in the legacy list view, this is possible. So, thinking about the feature flag approach that we discussed previously, maybe some customers still have access to that legacy list view table which will allow them to restore these forms that will not be associated with any campaign in that scenario.

image

glaubersilva (commented):
@kjohnson I left a comment about the `trash` status, but I am not sure if my point should be a concern. So let me know your thoughts on it.

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
```diff
@@ -29,66 +26,153 @@ public static function id(): string
...
      */
     public function run()
     {
-        foreach(DonationForm::query()->getAll() ?? [] as $form) {
-            $this->createParentCampaignForDonationForm($form);
-        }
+        DB::transaction(function() {
+            try {
+                array_map([$this, 'createCampaignForForm'], $this->getFormData());
+                array_map([$this, 'addUpgradedFormToCampaign'], $this->getUpgradedFormData());
```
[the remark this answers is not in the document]
JasonTheAdams: As a note, inserting multiple rows at once has to be batched. I think the limit is somewhere around 10k at once. We are using a transaction here, so that will help quite a bit.

This may be entirely fine as it is, to be clear, I'm just sharing thoughts. My typical preference for these sorts of migrations is to use INSERT INTO with a subquery — and let the DB engine handle the whole thing.

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
kjohnson: Good point, @glaubersilva. This also brings up the direction that "every form is part of a campaign", even a deleted for would need a campaign.

kjohnson: Resolved, see [6146be5](https://github.com/impress-org/givewp/pull/7532/commits/6146be5b853c6048987b4147e1e690e7932ff502).

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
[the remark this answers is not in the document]
kjohnson: Good point, @JasonTheAdams. This could certainly be improved by inserting multiple rows in a single query, particularly for the upgraded forms.

Our `DB::insert` method maps to `wpdb::insert`, which only supports inserting a single row. I'm not seeing any `wpdb` functions for inserting multiple rows.

I'll look into using a sub-query.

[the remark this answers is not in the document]
kjohnson: @JasonTheAdams I was looking into manually adding the INSERT INTO with a subquery for the migration as a step towards updating the query builder, but I think it would be better to just go ahead and update the query builder and refactor this work accordingly in a subsequent PR.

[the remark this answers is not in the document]
JasonTheAdams: I'm totally good with that. 👍

glaubersilva (approved):
@kjohnson Thanks for the feedback and update. Great work!

JasonTheAdams (approved):
Great work! This is well done and future-proof. 💪

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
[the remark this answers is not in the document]
kjohnson: @JasonTheAdams for follow up see https://github.com/impress-org/givewp/pull/7538.

PR #7536, opened 2024-09-12
Feature: add campaign details page header with tabs, save buttons, and form with sample input

glaubersilva:
Resolves [GIVE-1218]

## Description

This PR implements the campaign details page with the following items:

- Breadcrumbs
- Title
- Update/Publish campaign buttons
- Tabs (easily extensible) that can be accessed through links and also keep the history of the "back" and "forward" buttons from the browser - without using any external library for that.
- Form using `react-hook-form` to keep the context through the different tabs

This will require a subsequent PR for design review, but the main functionality is already in place. The idea here is to move forward with the project and not block the rest of the team that is depending on this screen to keep going on their tasks.

## Affects

The campaign details page.

## Visuals

https://www.loom.com/share/f690359235444dc38bad921c4616e2b8?sid=c2137137-c1e0-41e6-9b06-37056dc87eba

## Testing Instructions

1. Create a new campaign;
2. Go to the settings tab and change the title, save as a draft;
3. Publish the campaign with the new name.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-1218]: https://stellarwp.atlassian.net/browse/GIVE-1218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JasonTheAdams:
Great progress, @glaubersilva!

@alaca @kjohnson @jonwaldstein This is a good example of why I think it would be a good idea to set up the Data API as soon as possible. All of the state described here would work well within the store. It also wouldn't have the same rendering issues as the Context API.

As a note, we'd also be able to use the Redux dev tools to start to do things like watch, edit, and rewrite state changes for debugging.

kjohnson:
On a related note, we should discuss a strategy for handling how Admin Notices appear within our admin pages.

Probably a larger discussion, but noting it here.

image

src/Campaigns/Routes/PublishCampaign.php
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * @unreleased
+ */
+class PublishCampaign implements RestRoute
+{
+    use RestResponses;
+
+    /** @var string */
+    protected $endpoint = 'campaigns/publish/(?P<id>[0-9]+)';
```
kjohnson, line 21: This endpoint should probably be `campaigns/{id}/publish`, which would more closely match a RESTful resource.

src/Campaigns/Routes/Traits/RestResponses.php
```diff
@@ -0,0 +1,51 @@
...
+    public function badRequestResponse(Exception $e, $response = 'rest_bad_request'): WP_REST_Response
+    {
+        return new WP_REST_Response(
+            [
+                'status' => 400,
+                'response' => $response,
+                'body_response' => [
+                    'message' => html_entity_decode(
+                        sprintf(
+                            esc_html__('%s. Search the logs at Donations > Tools > Logs for a more specific cause of the problem.',
```
kjohnson, line 43: Should this messaging be moved to the client (instead of being part of the route)?

src/Campaigns/Routes/PublishCampaign.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * @unreleased
+     */
+    public function handleRequest(WP_REST_Request $request): WP_Rest_Response
+    {
+        $campaignId = $request->get_param('id');
+        $campaign = Campaign::find($campaignId);
+
+        if ( ! $campaign) {
```
kjohnson, line 58: We might consider moving this validation to the route arguments, which has a `validate_callback` in addition to the `sanitize_callback`.

kjohnson: Using `validate_callback` would also handle the `404` response if the validation fails.

src/Campaigns/Routes/PublishCampaign.php
```diff
@@ -0,0 +1,73 @@
...
+            return $this->notFoundResponse(esc_html__(sprintf('Campaign %s not found.', $campaignId), 'give'));
+        }
+
+        try {
+            $campaign->status = CampaignStatus::ACTIVE();
+            $campaign->save();
+
+            $response = json_encode($campaign->toArray());
+        } catch (Exception $e) {
+            return $this->badRequestResponse($e);
```
kjohnson, line 68: We may not have to catch the `Exception` to handle the response type. If I remember correctly, the WP_REST_API will convert an `Exception` to a `500` error response.

src/Campaigns/Routes/UpdateCampaign.php
```diff
@@ -0,0 +1,76 @@
...
+        $campaignId = $request->get_param('id');
+        $campaign = Campaign::find($campaignId);
+
+        if ( ! $campaign) {
+            return $this->notFoundResponse(esc_html__(sprintf('Campaign %s not found.', $campaignId), 'give'));
+        }
+
+        try {
+            $title = $request->get_param('title');
+            $campaign->title = $title ? $request->get_param('title') : $campaign->title;
```
kjohnson, line 65: We can add a check on the `title` parameter first, to avoid an unnecessary call to the database.

kjohnson: On a side note, this ternary can be shortened to `$request->get_param('title') ?: $campaign->title`.

src/Campaigns/Routes/PublishCampaign.php
glaubersilva: Resolved in d362f6f69afe0305a59fe6da75ee94c004dfda86 and 849de72a133e094686bc547f9a33201cb4eb91a9

src/Campaigns/Routes/Traits/RestResponses.php
glaubersilva: Resolved in 7f64906a2d650e4af1d6f95a6ba444ff62404ec4 and ea7a3d193d985b101ce808c5311ba7e3cd16e0ee

src/Campaigns/Routes/PublishCampaign.php
glaubersilva: Resolved in 7f6062c3bcc0f7f47453fe7a1cc851f14cace14d and 7f64906a2d650e4af1d6f95a6ba444ff62404ec4 and ea7a3d193d985b101ce808c5311ba7e3cd16e0ee

src/Campaigns/Routes/PublishCampaign.php
glaubersilva: Resolved in 7f6062c3bcc0f7f47453fe7a1cc851f14cace14d

src/Campaigns/Routes/UpdateCampaign.php
glaubersilva: Resolved in d362f6f69afe0305a59fe6da75ee94c004dfda86

glaubersilva:
@kjohnson Thanks for the suggestions. I already implemented them. Let me know if there is something else.

About the admin notices, those are displayed after the `<div className={cx('wp-header-end', 'hidden')} />` element, and in that case, I choose put it after the tabs navigation because it has the same color as the header 

I think it would be a bit weird to display the notices before the tabs because the tabs and the header seem to be one single thing, different from the event details page where the tabs navigation and the head have different colors - check the attached screenshots for reference.

Event Details Page - Notices **before** the tabs navegation:

image

Campaign Details Page - Notices **before** the tabs navegation:

image

Campaign Details Page - Notices **after** the tabs navegation:

image

kjohnson (approved):

PR #7538, opened 2024-09-12
Feature: Add new Insert Into method to the Query Builder

kjohnson:
## Description

This PR adds support for inserting the results on a query into another table using `INSERT INTO`.

## Affects

Adds a new `insertInto()` method to the `CRUD` trait of the `QueryBuilder`.

## Visuals

```php
DB::table('postmeta')
    ->insertInto(
        ['post_id', 'meta_key', 'meta_value'],
        DB::table('posts')
            ->select(['ID', 'post_id'])
            ->selectRaw('(SELECT "postTitle") as meta_key')
            ->select(['post_title', 'meta_value'])
    );
```

```

```sql
INSERT INTO wp_postmeta (post_id, meta_key, meta_value)
    SELECT
        ID AS post_id,
        (SELECT "postTitle") as meta_key,
        post_title AS meta_value
    FROM wp_posts;
```

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Framework/QueryBuilder/Concerns/CRUD.php
+++ b/src/Framework/QueryBuilder/Concerns/CRUD.php
@@ -3,6 +3,7 @@
 namespace Give\Framework\QueryBuilder\Concerns;
 
 use Give\Framework\Database\DB;
+use Give\Framework\QueryBuilder\QueryBuilder;
 
 /**
  * @since 2.19.0
@@ -27,6 +28,20 @@ trait CRUD
             $data,
             $format
         );
+    }
+
+    /**
+     * @unreleased
+     *
+     * @param array        $columns
+     * @param QueryBuilder $query
+     *
+     * @return bool|int
+     */
+    public function insertInto(array $columns, QueryBuilder $query)
+    {
+        $columns = implode(', ', $columns);
+        return DB::query("INSERT INTO {$this->getTable()} ({$columns}) {$query->getSQL()}");
     }
 
     /**
--- a/tests/Unit/Framework/QueryBuilder/CRUDTest.php
+++ b/tests/Unit/Framework/QueryBuilder/CRUDTest.php
@@ -58,6 +58,30 @@ final class CRUDTest extends TestCase
     }
 
     /**
+     * @unreleased
+     */
+    public function testInsertIntoShouldAddMultipleRowsToTheDatabase()
+    {
+        DB::table('posts')->insert(['post_title' => 'Aye Post']);
+        $postIdAye = DB::last_insert_id();
+
+        DB::table('posts')->insert(['post_title' => 'Bee Post']);
+        $postIdBee = DB::last_insert_id();
+
+        DB::table('postmeta')
+            ->insertInto(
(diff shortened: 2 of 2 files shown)


src/Framework/QueryBuilder/Concerns/CRUD.php
```diff
@@ -29,6 +30,20 @@ public function insert($data, $format = null)
...
 
+    /**
+     * @unreleased
+     *
+     * @param array        $columns
+     * @param QueryBuilder $query
+     *
+     * @return bool|int
+     */
+    public function insertInto(array $columns, QueryBuilder $query)
```
alaca, line 41: This is nice, but very limited. What if I want to pass values instead of the QB instance?

alaca: This will be a great addition to QB, but let's leverage as many MySQL features as we can. Providing just a QB instance means that we can only use subqueries to select the data we want to insert, but I would like to do more than that. As I mentioned above, I would also like to have something like this:

```php
->insertInto(['name', 'last_name'], ['ante', 'laca']);
```

Or even something like this:

```php
->insertInto(['name', 'last_name'], [['ante', 'laca'], ['kyle', 'johnson']]);
```

JasonTheAdams:
Throwing in my 0.02, I was imagining a syntax such as:
```php
DB::table('table1'
  ->select('a', 'b', 'c')
  ->where('a', '=', 'Kyle')
  ->insertColumns('col1', 'col2', 'col3')
  ->insertInto('talble2')
```

src/Framework/QueryBuilder/Concerns/CRUD.php
alaca: The biggest _problem_ with QB is that we are using `wpdb::insert`, `wpdb::update`, etc. under the hood in all CRUD methods.  Switching to _pure_ SQL will be much better, but that's something we can discuss some other time.

kjohnson: I was looking into a dynamic second parameter, but union types require PHP 8.0, so I considered multiple methods (`insertDataInto`, `insertQueryInto`) and didn't like that. Still considering.

kjohnson:
@JasonTheAdams I see what you were going for. When I did this first pass I was transcribing the `INSERT INTO` syntax to the Query Builder, but I'm reminded that a big benefit of the Query Builder is creating a syntax that better fits the application logic.

Ironic given my recent Slack post about this very benefit!

Goes to show how perspective can be a blind spot. As a writer I connected the path of least resistance.

src/Framework/QueryBuilder/Concerns/CRUD.php
kjohnson: @alaca I added a note to the PR explaining why I'm not including support for inserting multiple rows of known data as a part of this feature.

JasonTheAdams:
I'm going to close this as we'll re-open this in the https://github.com/stellarwp/db library after some syntax brainstorming. Let's link back to this, though, for reference!

PR #7543, opened 2024-09-17
Feature: add create new campaign flow

glaubersilva:
Resolves [GIVE-1217]

## Description

This PR split the form to create a new campaign in two steps and add all necessary fields and styles to make it work as expected.  The main goal here is to implement the functionality that allows us to run some internal tests, but we'll need to tweak some things later related to design adjustments and other improvements.

**Important:** Please note that I'm planning to open a subsequent PR when PR #7547 gets merged to the Epic, so I'll be able to refactor this form to use the new `@wordpress/core-data approach` used there and also will replace the Goal input to use a CurrencyInput (which is being introduced in the mentioned PR) when the "Amount raised" option is selected on the Goal Type options.

## Affects

The "Create New Campaign" form.

## Visuals

image

image

## Testing Instructions

1. Go to `Give > Campaigns` page;
2. Click on the `Create campaign` button;
3. Follow the steps to create a new campaign;
4. Make sure everything worked as expected.

## Pre-review Checklist

-   [ ] Acceptance criteria satisfied and marked in related issue
-   [ ] Relevant `@unreleased` tags included in DocBlocks
-   [ ] Includes unit tests
-   [ ] Reviewed by the designer (if follows a design)
-   [ ] [Self Review](https://give.gitbook.io/development-manual/devops/github/code-reviews#self-review) of code and UX completed

[GIVE-1217]: https://stellarwp.atlassian.net/browse/GIVE-1217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7547, opened 2024-09-23
Feature: Campaign Settings tab

alaca:
## Description

This PR scaffolds the structure for Campaign settings. That includes registering a campaign entity and adding new and refactoring existing campaign routes. The Campaign components are also simplified and tabs logic is moved into separate files. Also, a new schema validation library ([ajv](https://ajv.js.org/)) is installed and it's used to validate form inputs. The great thing about using `ajv` is that it uses [JSON schema](https://json-schema.org/), which means we can reuse the schema from the endpoint.  

Note: there is one field missing (campaign_image) and that will be added after [this](https://github.com/impress-org/givewp/pull/7543) PR is merged. 
## Affects

Campaign details page.

## Pre-review Checklist


src/Campaigns/Routes/GetCampaigns.php
```diff
@@ -0,0 +1,77 @@
...
+                    'page' => [
+                        'type' => 'integer',
+                        'default' => 1,
+                        'minimum' => 1,
+                    ],
+                    'per_page' => [
+                        'type' => 'integer',
+                        'default' => 30,
+                        'minimum' => 1,
+                    ],
```
TimothyBJacobs, line 45: I'd reocmmend attaching a maximum to this. WordPress Core set's `100`.

src/Campaigns/Routes/GetCampaigns.php
```diff
@@ -0,0 +1,77 @@
...
+            ->limit($perPage)
+            ->offset(($page - 1) * $perPage);
+
+        $campaigns = $query->getAll() ?? [];
+        $totalCampaigns = empty($campaigns) ? 0 : Campaign::query()->count();
+        $totalPages = (int)ceil($totalCampaigns / $perPage);
+
+        $response = rest_ensure_response($campaigns);
+        $response->header('X-WP-Total', $totalCampaigns);
+        $response->header('X-WP-TotalPages', $totalPages);
```
TimothyBJacobs, line 73: I'd recommend setting response header links for pages. That way it works with things like the `@wordpress/api-fetch` fetch all middleware.

Eg: https://github.com/ithemes/ithemes-security-pro/blob/73caf6c88c69a4a3ecbf78f4565423ebf36be7ed/core/lib/class-itsec-lib-rest.php#L334

src/Campaigns/resources/admin/components/CampaignDetailsPage/campaignSchema.ts
```diff
@@ -0,0 +1,44 @@
+import {__} from '@wordpress/i18n';
+import {JSONSchemaType} from 'ajv';
+import {CampaignInputFields} from './types';
+
+const campaignSchema: JSONSchemaType<CampaignInputFields> = {
```
TimothyBJacobs, line 5: I'd recommend pulling this from the REST API. That way you are actually doing the validation on the server as well, and it can be updated by things like extensions. You can grab it by making an `OPTIONS` request to the endpoint.

src/Campaigns/Routes/UpdateCampaign.php
```diff
@@ -60,18 +61,34 @@ public function registerRoute()
...
-                $this->campaign->setAttribute($key, $value);
+            switch ($key) {
+                case 'id':
+                    break;
+                case 'status':
+                    $status = array_key_exists($value, $statusMap)
+                        ? $statusMap[$value]
+                        : CampaignStatus::DRAFT();
+
+                    $this->campaign->setAttribute('status', $status);
```
TimothyBJacobs, line 78: The `validate_callback` must be a pure function. Having side effects like setting a property on the controller is not safe. This may already be broken by things like the Batch API, and could break further in the future.

This isn't the exact line that this happens on, but I can't commend on the `validate_callback` declaration.

TimothyBJacobs (commented):
👋 Just giving a quick review on a couple of things I noticed.

src/Campaigns/Routes/GetCampaigns.php
alaca: I don't have access to this repo.

src/Campaigns/resources/admin/components/CampaignDetailsPage/campaignSchema.ts
alaca: Cool idea, thanks!

src/Campaigns/Routes/UpdateCampaign.php
alaca: 👆  @kjohnson @glaubersilva

src/Campaigns/Routes/GetCampaigns.php
TimothyBJacobs:
```

public static function paginate( WP_REST_Request $request, WP_REST_Response $response, int $count, string $path ) {
		$max_pages = ceil( $count / $request['per_page'] );
		$response->header( 'X-WP-Total', $count );
		$response->header( 'X-WP-TotalPages', $max_pages );

		$request_params = $request->get_query_params();
		$base           = add_query_arg(
			map_deep( $request_params, function ( $value ) {
				if ( is_bool( $value ) ) {
					$value = $value ? 'true' : 'false';
				}

				return urlencode( $value );
			} ),
			rest_url( $path )
		);

		if ( $request['page'] > 1 ) {
			$prev_page = $request['page'] - 1;

			if ( $prev_page > $max_pages ) {
				$prev_page = $max_pages;
			}

			$prev_link = add_query_arg( 'page', $prev_page, $base );
			$response->link_header( 'prev', $prev_link );
		}

		if ( $max_pages > $request['page'] ) {
			$next_page = $request['page'] + 1;
			$next_link = add_query_arg( 'page', $next_page, $base );

			$response->link_header( 'next', $next_link );
		}
	}

```

src/Campaigns/Routes/UpdateCampaign.php
TimothyBJacobs: You shouldn't need the callback at all if the type is set to `integer`. But yeah you'd fetch the model in the controller callback.

src/Campaigns/resources/admin/components/CampaignDetailsPage/index.tsx
```diff
@@ -1,214 +1,139 @@
...
-        };
-
-        // Updates state based on URL when user navigates with "Back" or "Forward" buttons
-        window.addEventListener('popstate', handlePopState);
-
-        // Cleanup listener on unmount
-        return () => {
-            window.removeEventListener('popstate', handlePopState);
-        };
+        fetch(`/wp-json/give-api/v2/campaigns/${campaignId}`, {method: 'OPTIONS'})
```
TimothyBJacobs, line 22: You'll want to use `@wordpress/api-fetch`. It's preconfigured to talk to the REST API. Then you'll drop the `/wp-json/` prefix (it's configurable via a filter, and may not work at all since pretty permalinks aren't guaranteed), since WP will speak to the right URL by default.

src/Campaigns/Routes/Campaign.php
```diff
@@ -0,0 +1,190 @@
...
+                    'permission_callback' => function () {
+                        return current_user_can('manage_options');
+                    },
+                ],
+                'args' => [
+                    'id' => [
+                        'type' => 'integer',
+                        'required' => true,
+                        'validate_callback' => function ($id) {
+                            return filter_var($id, FILTER_VALIDATE_INT);
```
TimothyBJacobs, line 44: You don't need a callback here. The rest api takes care of this by default. 

Check out: https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/

src/Campaigns/Routes/Campaign.php
```diff
@@ -0,0 +1,190 @@
...
+     * @unreleased
+     *
+     * @throws Exception
+     */
+    public function handleUpdateRequest(WP_REST_Request $request): WP_REST_Response
+    {
+        $campaign = CampaignModel::find($request->get_param('id'));
+
+        if ( ! $campaign) {
+            return new WP_REST_Response([
```
TimothyBJacobs, line 111: You'll want to return a WP_Error to properly format errors.

src/Campaigns/Routes/GetCampaigns.php
alaca: Resolved:  https://github.com/impress-org/givewp/pull/7547/commits/7c3e40fb01c5b7071eb5eedfef4909fa168f59f6

src/Campaigns/Routes/UpdateCampaign.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/7547/commits/e9253fbcda04ea19489be095af0493448c6d8c2c

src/Campaigns/Routes/Campaign.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/7547/commits/d57d213d7df34156fb89fd8cd904c3838b5e2de5

src/Campaigns/resources/admin/components/CampaignDetailsPage/index.tsx
alaca: Resolved: https://github.com/impress-org/givewp/pull/7547/commits/31a91bef3143fce84170bbb63a84e8184b635869

src/Campaigns/resources/admin/components/CampaignDetailsPage/index.tsx
```diff
@@ -1,214 +1,144 @@
...
+                    </div>
+                </div>
+            </div>
+        );
+    }
+
     return (
         <FormProvider {...methods}>
             <form onSubmit={handleSubmit(onSubmit)}>
                 <article className={styles.page}>
```
glaubersilva, line 87: Currently, the `ClassicEditor` component is throwing some errors in the browser console because it was designed to be used inside the Gutenberg editor at the first moment. 

So it's looking for an element in the DOM with the `.interface-interface-skeleton__content` class necessary to make [this workaround](https://github.com/impress-org/form-builder-library/blob/master/src/ClassicEditor/index.tsx#L109-L127) work. 

To prevent these errors we can change this line to something like this:

```
<article className={`interface-interface-skeleton__content ${styles.page}`}>
```

This way, we can use the `ClassicEditor` component anywhere on the campaign details page without throwing errors in the console.

alaca: @glaubersilva it works, awesome. Thanks, man!

src/Campaigns/Routes/Campaign.php
```diff
@@ -77,16 +78,16 @@ public function registerRoute()
...
-    public function handleGetRequest(WP_REST_Request $request): WP_Rest_Response
+    public function handleGetRequest(WP_REST_Request $request)
     {
         $campaign = CampaignModel::find($request->get_param('id'));
 
         if ( ! $campaign) {
-            return new WP_REST_Response([
-                'message' => __('Campaign not found', 'give'),
-            ], 400);
+            return new WP_Error(400,  __('Campaign not found', 'give'));
```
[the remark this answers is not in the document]
alaca: Yeah, I refactored it see https://github.com/impress-org/givewp/pull/7547/commits/d57d213d7df34156fb89fd8cd904c3838b5e2de5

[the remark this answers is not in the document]
TimothyBJacobs: cool, sorry I didn't notice that!

src/Campaigns/Routes/GetCampaigns.php
JasonTheAdams: @alaca You have access to that repo now, FYI.

src/Campaigns/Routes/Campaign.php
```diff
@@ -0,0 +1,249 @@
...
+    public function registerRoute(): void
+    {
+        // Get campaign
+        register_rest_route(
+            'give-api/v2',
+            $this->endpoint,
+            [
+                [
+                    'methods' => WP_REST_Server::READABLE,
+                    'callback' => [$this, 'handleGetRequest'],
```
jonwaldstein, line 35: @alaca I recommend creating a controller to handle the request logic.  This gives a nice separation of the route and request, while giving you a controller to write unit tests for.

You can use the form builder as an example:
https://github.com/impress-org/givewp/blob/8adc13337a1289e1a30f52dc9ec632c9b543d0d9/src/FormBuilder/Routes/RegisterFormBuilderRestRoutes.php#L49

https://github.com/impress-org/givewp/blob/304c495bd0a727b0d8e1149c54e7fb1bfecdbcb0/src/FormBuilder/Controllers/FormBuilderResourceController.php

alaca: Resolved: https://github.com/impress-org/givewp/pull/7547/commits/309cd0f8e96323c082b2d5eb152a9d7617f76a3c

alaca:
Hey @TimothyBJacobs, I have a question regarding schema validation. 
It looks like the schema validation on the server is not working in my case.

According to docs https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#api

> If your endpoint does not follow the controller class pattern, args returned from WP_REST_Controller::get_collection_params(), or any other instance where callbacks are not specified, the WP_REST_Request object will apply sanitization and validation using the rest_parse_request_arg function. 
Importantly, this is only applied when the sanitize_callback is not defined. As such, if you specify a custom sanitize_callback for your argument definition, the built-in JSON Schema validation will not apply. If you need this validation, you should manually specify rest_validate_request_arg as the validate_callback in your argument definition.

So I tried to remove `sanitize_callback` and `validate_callback` - nothing happened. I'm able to pass whatever for the `title` param.

I also tried this, but it doesn't work either. 
```php
'title' => [
    'type' => 'string',
    'sanitize_callback' => 'rest_sanitize_request_arg',
    'validate_callback' => 'rest_validate_request_arg'
]
```

I'm probably doing something wrong here, can you please take a look? Thanks!

src/Campaigns/Routes/Campaign.php
jonwaldstein: Thanks!

Now that you have a controller, you can even write tests for it! 😄 

Example:
https://github.com/impress-org/givewp/blob/c9b8d7396118eee0cfee485f644b7a26b975214b/tests/Feature/Controllers/FormBuilderResourceControllerTest.php#L21

src/Campaigns/resources/admin/components/CampaignDetailsPage/style.module.scss
```diff
@@ -199,14 +199,115 @@
             margin: 0 0 var(--givewp-spacing-2);
         }
     }
+
+    .sections {
```
glaubersilva, line 203: We need to rename this file and the places that are using it to `CampaignDetailsPage.module.scss` because the way it's being used right now is preventing the CSS files from the Form List table from being loaded - check the attached screenshot.

image

src/Campaigns/Actions/LoadCampaignDetailsAssets.php
```diff
@@ -14,30 +12,33 @@ class LoadCampaignDetailsAssets
...
-                'campaign' => (new CampaignDetailsPage($campaign))->exports(),
+                'currency' => give_get_currency()
             ]
         );
 
         wp_enqueue_script($handleName);
         wp_enqueue_style('givewp-design-system-foundation');
+        wp_enqueue_style(
+            $handleName,
+            GIVE_PLUGIN_URL . 'build/style-campaignDetails.css'
```
glaubersilva, line 41: We need to change this line to this `GIVE_PLUGIN_URL . 'build/campaignDetails.css'` so we can properly load the CSS files for the Form List Table in the Forms tab.

But it will introduce a new problem that prevents the details page styles from being loaded as you can see in this screenshot:

image

So to fix this one you need to implement the change that I mentioned in this comment: https://github.com/impress-org/givewp/pull/7547/files#r1777705915

After that, we'll be able to have all styles being loaded like you can check on this screenshot: 

image

src/Campaigns/resources/admin/components/CampaignDetailsPage/style.module.scss
alaca: Resolved: https://github.com/impress-org/givewp/pull/7547/commits/8068f9b8474f1df1e36e0eb9a6fac7dad23876d1

src/Campaigns/Routes/Campaign.php
alaca: 😆 
image

Resolved: https://github.com/impress-org/givewp/pull/7547/commits/18ee165f25c9cc5722e49da8c5e194387ada5c4e

TimothyBJacobs:
> Hey @TimothyBJacobs, I have a question regarding schema validation. It looks like the schema validation on the server is not working in my case.
> I'm probably doing something wrong here, can you please take a look? Thanks!

The `args` list in `register_rest_route` should be at the same level as your `method`, `callback`, etc... You may want to consider using the `rest_get_endpoint_args_for_schema` function as well to help build this list.

src/Campaigns/Routes/Campaign.php
jonwaldstein: haha Yay!!!

src/Campaigns/Routes/RegisterCampaignRoutes.php
```diff
@@ -0,0 +1,305 @@
...
+                    'title' => [
+                        'type' => 'string',
+                        'required' => true,
+                        'sanitize_callback' => 'sanitize_text_field',
+                    ],
+                    'description' => [
+                        'type' => 'string',
+                        'required' => false,
+                        'sanitize_callback' => 'sanitize_text_field',
+                    ],
```
glaubersilva, line 173: Let's add here these args:

```
 'image' => [
    'type' => 'string',
    'required' => false,
    'sanitize_callback' => 'sanitize_url',
],
'goalType' => [
    'type' => 'string',
    'required' => true,
    'sanitize_callback' => 'sanitize_text_field',
],
'goal' => [
    'type' => 'integer',
    'required' => true,
    'sanitize_callback' => 'absint',
],
```

This way, we can keep the changes that I did on `CreateCampaign.php` file: https://github.com/impress-org/givewp/pull/7543/files#diff-2adc6a525044d3090b82717de88d883fd6156a30099c8c238698db2f421900a3

src/Campaigns/Controllers/CampaignRequestController.php
```diff
@@ -0,0 +1,159 @@
...
+     * @unreleased
+     *
+     * @throws Exception
+     */
+    public function createCampaign(WP_REST_Request $request): WP_REST_Response
+    {
+        $campaign = Campaign::create([
+            'type' => CampaignType::CORE(),
+            'title' => $request->get_param('title'),
+            'shortDescription' => $request->get_param('shortDescription'),
```
glaubersilva, line 144: As this field is not required on the UI, I think we can replace it by this:  `$request->get_param('shortDescription') ?? ''`

src/Campaigns/Controllers/CampaignRequestController.php
```diff
@@ -0,0 +1,159 @@
...
+            'longDescription' => '',
+            'logo' => '',
+            'image' => '',
+            'primaryColor' => '',
+            'secondaryColor' => '',
+            'goal' => 1000,
+            'goalType' => 'amount',
+            'status' => CampaignStatus::DRAFT(),
+            'startDate' => $request->get_param('startDateTime'),
+            'endDate' => $request->get_param('endDateTime'),
```
glaubersilva, line 154: As these fields are not required on the UI, I think we can replace it by this:

```
'startDate' => $request->get_param('startDateTime') ?? null,
'endDate' => $request->get_param('endDateTime') ?? null,
```

src/Campaigns/Controllers/CampaignRequestController.php
```diff
@@ -0,0 +1,159 @@
...
+     */
+    public function createCampaign(WP_REST_Request $request): WP_REST_Response
+    {
+        $campaign = Campaign::create([
+            'type' => CampaignType::CORE(),
+            'title' => $request->get_param('title'),
+            'shortDescription' => $request->get_param('shortDescription'),
+            'longDescription' => '',
+            'logo' => '',
+            'image' => '',
```
glaubersilva, line 147: This also is not required on the UI, so let's replace this line with this one: `'image' => $request->get_param('image') ?? '',`

src/Campaigns/Routes/RegisterCampaignRoutes.php
```diff
@@ -0,0 +1,305 @@
...
+                        return current_user_can('manage_options');
+                    },
+                ],
+                'args' => [
+                    'title' => [
+                        'type' => 'string',
+                        'required' => true,
+                        'sanitize_callback' => 'sanitize_text_field',
+                    ],
+                    'description' => [
```
glaubersilva, line 169: Let's rename it to `shortDescription` to keep the same name used in the controller and model.

src/Campaigns/Routes/RegisterCampaignRoutes.php
```diff
@@ -0,0 +1,305 @@
...
+                        'type' => 'string',
+                        'required' => false,
+                        'sanitize_callback' => 'sanitize_text_field',
+                    ],
+                    'startDateTime' => [
+                        'type' => 'string',
+                        'format' => 'date-time', // @link https://datatracker.ietf.org/doc/html/rfc3339#section-5.8
+                        'required' => true,
```
glaubersilva, line 177: As this is not required on the UI, let's replace this line with this one: `'required' => false,`

src/Campaigns/Actions/LoadCampaignDetailsAssets.php
```diff
@@ -14,30 +12,33 @@ class LoadCampaignDetailsAssets
     /**
      * @unreleased
      */
-    public function __invoke(Campaign $campaign)
+    public function __invoke()
```
glaubersilva, line 15: The `CampaignsAdminPage.php` file has a call for this file where the `$campaign` is passed as a parameter, make sure to remove it from there as it's no longer necessary.

glaubersilva (changes requested):
@alaca Awesome job man! I just left some minor comments related to minor changes to make sure your PR will not conflict with my one that implements a way to add new campaigns with all necessary inputs.

src/Campaigns/Actions/LoadCampaignDetailsAssets.php
```diff
@@ -14,30 +12,33 @@ class LoadCampaignDetailsAssets
...
+                'currency' => give_get_currency()
             ]
         );
 
         wp_enqueue_script($handleName);
         wp_enqueue_style('givewp-design-system-foundation');
+        wp_enqueue_style(
+            $handleName,
+            GIVE_PLUGIN_URL . 'build/campaignDetails.css'
+        );
```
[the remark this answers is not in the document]
alaca: Resolved; https://github.com/impress-org/givewp/pull/7547/commits/a26ece8bdcb227f397a19fa338fe3fb490f724b1

src/Campaigns/Controllers/CampaignRequestController.php
alaca: If any of these parameters are not set, `get_param` will return null anyway because there is no default value set.

src/Campaigns/Routes/RegisterCampaignRoutes.php
alaca: The schema should handle this automatically, and I don't want to duplicate code unnecessarily. Currently, the schema doesn't work for some reason. I'm working on it to figure out why.

src/Campaigns/resources/admin/components/CampaignDetailsPage/index.tsx
```diff
@@ -1,214 +1,165 @@
...
                                     )}
                                 >
-                                    {campaign.properties.status}
+                                    {campaign.status}
                                 </span>
                             </div>
 
                             <div className={styles.flexRow}>
-                                {campaign.properties.status === 'draft' && (
+                                {campaign.status === 'active' && (
```
glaubersilva, line 119: The "Save as draft" button is not being displayed when the campaign is draft, only when it's active. So, if I want to edit the settings of the campaign without publishing it, it isn't possible as you can see on the attached screenshot. I think we need to display this button only when the campaign has the draft status and maybe a _"Convert to draft"_ button when it's active.

image

glaubersilva (changes requested):
@alaca We are almost there, I just left one more comment related to the "Save as draft" button.

src/Campaigns/resources/admin/components/CampaignDetailsPage/index.tsx
alaca: Good catch! Resolved: https://github.com/impress-org/givewp/pull/7547/commits/b41e8e9012f2079b632dd98cd516eba4e30817b6

glaubersilva (approved):
@alaca Thanks for the changes. Great work man!

src/Campaigns/resources/entity.ts
```diff
@@ -0,0 +1,18 @@
+import {__} from '@wordpress/i18n';
+import {dispatch} from '@wordpress/data';
+import {store as coreStore} from '@wordpress/core-data';
+
+//@ts-ignore
+dispatch(coreStore).addEntities([
```
jonwaldstein, line 6: 🎉

src/Campaigns/Models/Campaign.php
```diff
@@ -53,6 +54,7 @@ class Campaign extends Model implements ModelCrud, ModelHasFactory
         'primaryColor' => 'string',
         'secondaryColor' => 'string',
         'goal' => 'int',
+        'goalType' => 'string',
```
jonwaldstein, line 57: We should make an enum for this (doesn't have to be now), similar to the donation form [GoalType](https://github.com/impress-org/givewp/blob/162cdcab2d3bbac090efddd1b059085404013091/src/DonationForms/ValueObjects/GoalType.php#L24)

src/Campaigns/Controllers/CampaignRequestController.php
```diff
@@ -0,0 +1,159 @@
...
+        foreach ($request->get_params() as $key => $value) {
+            switch ($key) {
+                case 'id':
+                    break;
+                case 'status':
+                    $status = array_key_exists($value, $statusMap)
+                        ? $statusMap[$value]
+                        : CampaignStatus::DRAFT();
+
+                    $campaign->setAttribute('status', $status);
```
jonwaldstein, line 116: We should be using the direct model setter. `$model->setAttribute` is really intended for abstraction of the model setter

```suggestion
                    $campaign->status = $status;
```

jonwaldstein (approved):
Looks great! I'm especially excited about the server and front-end validation using JSON schema 🌮 Just added a couple notes.

src/Campaigns/Models/Campaign.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/7547/commits/dae29349803acb504ed17ba09778e647e96bb791

src/Campaigns/Controllers/CampaignRequestController.php
alaca: Resolved:  https://github.com/impress-org/givewp/pull/7547/commits/f172b8c2b406c671434a71200c8b55dfbfa04eb2

PR #7555, opened 2024-09-26
Feature: add Forms tab to the campaign details page

glaubersilva:
Resolves [GIVE-1219]

## Description

This PR implements the Forms tab on the campaign details page where is possible to see the forms associated with the campaign and create new ones.

Please note that the list table here does not fully match the prototype design. As we are using the same list table component used in other places of the codebase, I think the best approach will be to change the list table design in a subsequent PR implementing the necessary changes and checking how it reflects in other views - like the subscriptions list.  

Also, the prototype design does not consider the pagination of the list table which can be a problem for campaigns with large amounts of forms, so probably a few discussions about it will be required.

## Affects

The "Forms" tab of the campaign details page.

## Visuals

image

## Testing Instructions

1. Go to `Give > Campaigns` page;
2. Create a new Campaign;
3. Go to the `Forms` tab and make sure a default form was created;
4. Create a new campaign form and make sure it gets visible in the `Forms` tab as well.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-1219]: https://stellarwp.atlassian.net/browse/GIVE-1219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


glaubersilva:
@alaca Yep! I'm using these same code standards.

src/Campaigns/Models/Campaign.php
```diff
@@ -138,6 +138,24 @@ public function save(): void
...
+    public function addForm(DonationForm $donationForm, $updateDefaultDonationForm = false)
+    {
+        if ( ! $this->id) {
+            give(CampaignRepository::class)->insert($this, $donationForm);
+        } else {
+            if ( ! $this->defaultForm() && ! $updateDefaultDonationForm) {
+                $updateDefaultDonationForm = true;
+            }
+
+            give(CampaignRepository::class)->update($this, $donationForm, $updateDefaultDonationForm);
```
alaca, line 155: So instead of doing all these and modifying Campaign model `insert` and `update` methods, I will add a new method to `CampaignRepository` - `addCampaignForm` which will take three parameters (Campaign, $donationFormId, $isDefault = false), and then handle adding form there.

src/Campaigns/Repositories/CampaignRepository.php
```diff
@@ -42,11 +44,18 @@ public function getById(int $id)
      *
      * @throws Exception|InvalidArgumentException
      */
-    public function insert(Campaign $campaign): void
+    public function insert(Campaign $campaign, DonationForm $defaultDonationForm = null): void
```
alaca, line 47: Why do we need this? Creating a new campaign should automatically create the default campaign form. We are not creating campaigns when creating new forms, you won't even be able to create a new form without creating a campaign first.

alaca (changes requested):
Hey @glaubersilva I reviewed only the backend part for now. Once that's resolved, I will take a look at the frontend stuff.

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
```diff
@@ -131,24 +131,35 @@ public function createCampaignForForm($formData): void
...
-            ]);
+        $this->addCampaignFormRelationship($data->migratedFormId, $data->campaignId, false);
+    }
+
+    /**
+     * @unreleased
+     */
+    protected function addCampaignFormRelationship($formId, $campaignId, $isDefault)
+    {
+        $relationshipExists = (bool)DB::table('give_campaign_forms')
```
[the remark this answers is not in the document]
glaubersilva: Resolved in 8bca5caab8c88fdc00961d50e6816b24c4a15cfb

src/Campaigns/Routes/RegisterCampaignRoutes.php
```diff
@@ -220,7 +220,7 @@ public function getSchema(): array
                     'errorMessage' => esc_html__('Must be a number', 'give'),
                 ],
                 'goalType' => [
-                    'enum' => ['amount', 'donation', 'donors'],
+                    'enum' => ['amount', 'donations', 'donors'],
```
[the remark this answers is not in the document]
glaubersilva: Resolved in 42c35d96fcaf43ef4f1644ef0ab7bde82dd7b12c

src/Campaigns/ValueObjects/CampaignGoalType.php
```diff
@@ -8,15 +8,15 @@
...
  * @method bool isAmount()
- * @method bool isDonation()
+ * @method bool isDonations()
  * @method bool isDonors()
  */
 class CampaignGoalType extends Enum
 {
     const AMOUNT = 'amount';
-    const DONATION = 'donation';
+    const DONATIONS = 'donations';
```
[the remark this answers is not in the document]
glaubersilva: Resolved in 42c35d96fcaf43ef4f1644ef0ab7bde82dd7b12c

src/Campaigns/resources/admin/components/CampaignDetailsPage/Tabs/Forms.tsx
```diff
@@ -1,17 +1,8 @@
-import {__} from '@wordpress/i18n';
-
-import styles from '../CampaignDetailsPage.module.scss';
+import DonationFormsListTable from '../../../../../../DonationForms/V2/resources/components/DonationFormsListTable';
```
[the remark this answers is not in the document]
glaubersilva: We don't have an alias for this path.

src/Campaigns/Repositories/CampaignRepository.php
```diff
@@ -88,22 +104,26 @@ public function insert(Campaign $campaign): void
...
+        Hooks::doAction('givewp_campaign_created', $campaign, $defaultDonationForm);
     }
 
     /**
      * @unreleased
      *
      * @throws Exception|InvalidArgumentException
      */
-    public function update(Campaign $campaign): void
+    public function update(
```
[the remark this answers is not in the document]
glaubersilva: Resolved in 0714b47594199a3bb112162fbf5e9bfcb0e8d6f4

src/Campaigns/Models/Campaign.php
```diff
@@ -138,6 +138,24 @@ public function save(): void
...
     }
 
+    /**
+     * @unreleased
+     *
+     * @throws Exception
+     */
+    public function addForm(DonationForm $donationForm, $updateDefaultDonationForm = false)
+    {
+        if ( ! $this->id) {
```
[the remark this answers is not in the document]
glaubersilva: Resolved in 0714b47594199a3bb112162fbf5e9bfcb0e8d6f4

src/Campaigns/Models/Campaign.php
glaubersilva: Resolved in 0714b47594199a3bb112162fbf5e9bfcb0e8d6f4

src/Campaigns/Repositories/CampaignRepository.php
glaubersilva: Resolved in 0714b47594199a3bb112162fbf5e9bfcb0e8d6f4

src/Campaigns/Repositories/CampaignRepository.php
```diff
@@ -125,6 +145,30 @@ public function update(Campaign $campaign): void
                     'start_date' => $startDateFormatted,
                     'end_date' => $endDateFormatted,
                 ]);
+
+            if ( ! is_null($donationForm)) {
```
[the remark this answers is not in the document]
glaubersilva: The query you mentioned only updates the values, but we can't insert DB entries with that....

I was able the replace the queries for insert/delete with a single one that can insert or update value in case of duplication. 

But I still need an additional query for cases where the $isDefault is true to make sure that we'll keep only one form as default on DB.

Resolved in af83cc387557764bd1b8c1e60417b72f58178ced

glaubersilva:
@alaca Thanks for the suggestions, this is ready for re-review.

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
```diff
@@ -131,24 +131,35 @@ public function createCampaignForForm($formData): void
...
      * @param $data
      */
     protected function addUpgradedFormToCampaign($data): void
     {
-        DB::table('give_campaign_forms')
-            ->insert([
-                'form_id' => $data->migratedFormId,
-                'campaign_id' => $data->campaignId,
-            ]);
+        $this->addCampaignFormRelationship($data->migratedFormId, $data->campaignId, false);
```
alaca, line 142: At this point, there is only one form per campaign. You don't have to check if a campaign-form relationship exists. I'd suggest rolling back this commit.

src/Campaigns/Repositories/CampaignRepository.php
```diff
@@ -138,6 +140,50 @@ public function update(Campaign $campaign): void
...
+            // Make sure we'll have only one default form
+            if ($isDefault) {
+                DB::table('give_campaign_forms')
+                    ->where('campaign_id', $campaign->id)
+                    ->update([
+                        'is_default' => false,
+                    ]);
+            }
+
+            $table = DB::prefix('give_campaign_forms');
```
alaca, line 164: My example was more of an example of how to set the default form from one query without the need to do three DB actions. Let's have a simple insert here without fancy things like `ON DUPLICATE KEY UPDATE` because this method should do only one thing, and that's creating a relationship between the campaign and the form. Also, this method is used only when a new form is created which means there shouldn't be duplicate entries. 

BTW. `%s` is used for strings, and all parameters here are integers so it would make sense to use `%d` as a placeholder.

src/Campaigns/Repositories/CampaignRepository.php
```diff
@@ -138,6 +140,50 @@ public function update(Campaign $campaign): void
         Hooks::doAction('givewp_campaign_updated', $campaign);
     }
 
+    /**
+     * @unreleased
+     *
+     * @throws Exception
+     */
+    public function addCampaignForm(Campaign $campaign, DonationForm $donationForm, bool $isDefault = false)
```
alaca, line 148: Do we need a Donation form object here or do we need just the form ID? I don't see the benefit of passing the entire object where just an ID will be sufficient.

src/Campaigns/Repositories/CampaignRepository.php
```diff
@@ -138,6 +140,50 @@ public function update(Campaign $campaign): void
...
+                        $isDefault,
+                        $isDefault,
+                    ])
+            );
+        } catch (Exception $exception) {
+            DB::query('ROLLBACK');
+
+            Log::error('Failed updating a campaign form', compact('campaign'));
+
+            throw new $exception('Failed updating a campaign form');
```
alaca, line 179: The error message doesn't reflect the action we are doing here.

src/DonationForms/V2/DonationFormsAdminPage.php
```diff
@@ -171,8 +172,13 @@ private function preloadDonationForms()
         $queryParameters = [
             'page' => 1,
             'perPage' => 30,
+
         ];
 
+        if (isset($_GET['id']) && isset($_GET['page']) && 'give-campaigns' == isset($_GET['page'])) {
+            $queryParameters['campaignId'] = absint($_GET['id']);
```
alaca, line 179: Campaign ID is the same as the Form ID?

src/DonationForms/V2/resources/components/DonationFormsListTable.tsx
```diff
@@ -63,6 +63,13 @@ const donationStatus = [
     },
 ];
 
+const urlParams = new URLSearchParams(window.location.search);
+urlParams.get('id');
```
alaca, line 67: I guess this is a leftover?

alaca (changes requested):
There are still some minor issues we have to resolve, but overall good progress  👍

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
glaubersilva: Without this checking, I getting the errors described here: https://lw.slack.com/archives/C04SLRDD9CK/p1727723671542809

src/Campaigns/Repositories/CampaignRepository.php
glaubersilva: About it:

> this method is used only when a new form is created

In the current codebase it's true, but we can use this same method to add a form to the campaign at any time like on [this unit test](https://github.com/impress-org/givewp/blob/feature/forms-list-table-GIVE-1219/tests/Unit/Campaigns/Models/CampaignModelTest.php#L58), so we need a way to make sure that we'll not get a `WordPress database error Duplicate entry...` message when someone tries to insert a form already added before - because of that we need the `ON DUPLICATE KEY UPDATE` statement.  

Besides that, we also need a way to make sure the campaign will have only one default form when someone calls this method - because of that we need the additional query at the top that only gets triggered when `$isDefault` is true.

The placeholder was replaced here: 7fa4f07e752ecc8624916b3707b7cb0c2d7ab741

src/DonationForms/V2/DonationFormsAdminPage.php
glaubersilva: Nope! The condition is checking if we are on the campaign details page. To make it clear I created a helper function here: 56abe5317c47add063113bb0427964136ef4dd98

src/DonationForms/V2/resources/components/DonationFormsListTable.tsx
glaubersilva: Yep! Resolve in 93621fbedf84ce0f43ba9280373e0ac12c0f9881

src/Campaigns/Repositories/CampaignRepository.php
glaubersilva: Resolved in 99a801dbdec928b92d8c4bd9fe1bd086c6b66148

src/Campaigns/Repositories/CampaignRepository.php
glaubersilva: This way we don't need to use the `Campaign::find()` method to check if the campaign exists. If we already have the object, we know that the campaign exists.

glaubersilva:
@alaca This is ready for a new review.

src/Campaigns/Repositories/CampaignRepository.php
alaca: > Do we need a Donation form object here or do we need just the form ID?

I was talking about `DonationForm`.

src/Campaigns/Actions/AddCampaignFormFromRequest.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+class AddCampaignFormFromRequest
+{
+    /**
+     * @throws Exception
+     */
+    public function __invoke(DonationForm $donationForm)
+    {
+        if (isset($_GET['campaignId']) && $campaign = Campaign::find(absint($_GET['campaignId']))) {
+            $isDefault = isset($_GET['isDefault']) && $_GET['isDefault'];
```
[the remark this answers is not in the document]
alaca: Do we have that feature request? If not, then we don't need it.

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
alaca: So there can be a problem with data integrity. Okay, we can use `distinct` on query when fetching meta table data.  This will be much more performant than always checking if the form exist in table, which is basically an N+1 problem.

src/Campaigns/Repositories/CampaignRepository.php
alaca: > In the current codebase it's true, but we can use this same method to add a form to the campaign at any time like on this unit test

Unit test should create a new form and not use an existing one. 
Basically, you added all this because there is a problem with data (duplicate meta for formBuilderSettings) on your end. See my previous comment about how we can fix that. 

> because of that we need the additional update action at the top that only gets triggered when $isDefault is true.

I don't have an issue with this part.

alaca (changes requested):
Almost there

src/Campaigns/Repositories/CampaignRepository.php
glaubersilva: > Unit test should create a new form and not use an existing one.

Yeah! In the mentioned unit test we are doing exactly that on this line: `$newDefaultForm = DonationForm::factory()->create();`

And then updating the campaign default form using this other one: `give(CampaignRepository::class)->addCampaignForm($campaign, $newDefaultForm, true);`

Now, let's suppose that we call the `addCampaignForm()` method informing the `$newDefaultForm` form again as the parameter in some other place on the code. 

Without the `ON DUPLICATE KEY` statement we'll get `Duplicate entry` error because this form is already on DB. 

> Basically, you added all this because there is a problem with data (duplicate meta for formBuilderSettings) on your end. See my previous comment about how we can fix that.


Nope! I added it thinking about a defensive approach because we can use the `addCampaignForm()` method with a form already added before by mistake and throw an error.

Maybe I'm missing something here, but I can't understand how the `ON DUPLICATE KEY` statement can be a problem. Could you please elaborate more on that?

alaca: Okay, if you think we should do it like this, then it's okay. I don't want to hold this important PR.

alaca: > Maybe I'm missing something here, but I can't understand how the ON DUPLICATE KEY statement can be a problem. Could you please elaborate more on that?

It is not a problem, it means that we are doing something wrong. There shouldn't be duplicate forms in this table, and if someone tries to insert a form that already exists, he will get an error from DB because we are using capaign_id and form_id to create the composite key - which is totally fine, and even expected.  You can't do that from UI, only from code. So we are already defensive about it without doing anything extra.

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
glaubersilva: The `distinct` works fine with the query generated by the `getFormData()` method, but does not for the query generated by the `getUpgradedFormData()` method -  here is a version a bit different (only added the status column) of this last query:

```
SELECT forms.post_status, forms.ID AS formId, campaign_forms.campaign_id AS campaignId, give_formmeta_attach_meta_migratedFormId.meta_value AS migratedFormId 
FROM wp_posts AS forms 
LEFT JOIN wp_give_formmeta give_formmeta_attach_meta_migratedFormId ON ID = give_formmeta_attach_meta_migratedFormId.form_id AND give_formmeta_attach_meta_migratedFormId.meta_key = 'migratedFormId' 
RIGHT JOIN wp_give_campaign_forms campaign_forms ON campaign_forms.form_id = forms.ID 
WHERE forms.post_type = 'give_forms' AND give_formmeta_attach_meta_migratedFormId.meta_value IS NOT NULL
```

As you can see on the attached screenshot, this query is returning the same `migratedFormId` for different forms. So, we are trying to insert this same `migratedFormId` on the DB multiple times and it's throwing the duplicated entry errors that prevent the migration from getting finished.

Honestly, I don't know the steps to replicate the problem and make you get a similar result on your end. But it makes me think that if it's happening on my end, could happen for customers as well.

The problem only happens in forms with the `trash` status, so we could remove those for the query, but I don't think it's a good idea since the form legacy view has the restore option available - check the secound attached screenshot.

For now, I'll just left the distinct that solves the problem partially and will allow us move forward with this PR, but later I'll keep checking this problem to see if there is something we can do here.

Partially resolved in: da3cd569bb101bb564b97ee4e286af18036e5f9b

image

image

src/Campaigns/Repositories/CampaignRepository.php
glaubersilva: We can apply the same logic as we don't want to add a form id that doesn't exist on the DB.

src/Campaigns/Repositories/CampaignRepository.php
glaubersilva: Resolved in 3ed2a0713bb6462619b9dbe9187058460cb11b17

src/Campaigns/Actions/AddCampaignFormFromRequest.php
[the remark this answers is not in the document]
glaubersilva: Resolved in f5cff1ea9dc2c7b7027d5e1bfe12717b7b5e16f2

glaubersilva:
@alaca Thanks for the patience and explanations, this PR is ready for a new review.

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
alaca: Ok, so `getUpgradedFormData` is used to get upgraded form data, right? Then we are missing something really important here, and that's where condition with post status value. 

Currently, this query will return all forms regardless of the status of the form. So the easy solution here would be to add another `WHERE` condition where we want forms only with the status `upgraded`. 
```php
->where('forms.post_status', 'upgraded')
```

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
```diff
@@ -89,7 +89,7 @@ protected function getFormData(): array
      */
     protected function getUpgradedFormData(): array
     {
-        return DB::table('posts', 'forms')
+        return DB::table('posts', 'forms')->distinct()
```
alaca, line 92: We can remove this `distinct`

src/Campaigns/Repositories/CampaignRepository.php
alaca: Having a `DonationForm` object here doesn't necessarily mean that form exists, but okay.

alaca (commented):
Thank you, Glauber. 
There is one small thing I want you to add and test. If that works, we are good to go. 
Great job, man!

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
glaubersilva: Oh yeah! It makes much more sense. 😅

Resolved in 9ac7b978746a1b487ffb4ba47c472147d538d6a1

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
glaubersilva: Resolved in 9ac7b978746a1b487ffb4ba47c472147d538d6a1

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
glaubersilva: Ops, I just noticed the `MigrateFormsToCampaignFormsTest::testUpgradedFormsAreNotDefault` is failing after this change.

glaubersilva: In fact, the proper where is this one:

`->where('forms.post_status', 'publish')`

Because this query is getting the forms that already are using the Visual Form Builder and the `migratedFormId` meta key value is reference for they source, the legacy form where they came from - this source that have the `upgraded` status.

alaca: Yeah, makes sense.

alaca (approved):
Good job, man! 🚀

PR #7559, opened 2024-10-02
Feature: Add campaign overview statistics

kjohnson:
Resolves [GIVE-1222]

## Description

This PR adds widgets to the Campaign Overview page, including Donation Amount, Donation Count, Donor Count, and a Donation Goal Progress widget.

## Visuals

image

## Testing Instructions

View a Campaign's detail page and toggle the date range filters

## Pre-review Checklist

-   [x] Includes unit tests

[GIVE-1222]: https://stellarwp.atlassian.net/browse/GIVE-1222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alaca (approved):
Wow, this looks awesome! Great job, man! 
There are a few minor styling issues, but we can address them later.

src/Campaigns/Actions/LoadCampaignDetailsAssets.php
```diff
@@ -31,6 +31,8 @@ public function __invoke()
             [
                 'adminUrl' => admin_url(),
                 'currency' => give_get_currency(),
+                'apiRoot' => esc_url_raw(rest_url('give-api/v2/campaigns')),
+                'apiNonce' => wp_create_nonce('wp_rest'),
```
JasonTheAdams, line 35: If we're using https://developer.wordpress.org/block-editor/reference-guides/packages/packages-api-fetch/ then we shouldn't need to handle the nonce as it does for us.

Also, using entities handles this, too.

kjohnson: @JasonTheAdams the entity didn't seem to work in this case. I discussed the setup with @alaca, but during implementation it quickly resulted in work around. These stats aren't entities in and of themselves and are not fetched by an record Id.

kjohnson: @JasonTheAdams I'm leveraging the existing `CampaignsApi` class, which is configured with an `apiRoot` and an `apiNonce`. It seems to use `fetch` directly.

Refactoring to no longer need to localize a Nonce would be very cool, though. I'll look into the API Fetch package and see if I can submit a subsequent PR before my time off.

kjohnson: This looks like an easy swap - giving it a quick go.

kjohnson: @JasonTheAdams resolved, see https://github.com/impress-org/givewp/pull/7559/commits/70cb37ec97df275091bf7afb5856c3d8e4e95f12.

PR #7564, opened 2024-10-04
Refactor: update Honorific block to only accept labels and add field validation

jonwaldstein:
Resolves [GIVE-1292]

## Description

I noticed the honorific block was offering to "show values", except it didn't actually do anything with the values.  Since honorifics are both a value and label, also used in global settings I decided to remove the option for showing values.

Furthermore, extra field validation was added to only accept the available select options. 

## Affects

The honorific block & field

## Visuals

**The problem discovery**

https://www.loom.com/share/1ae99a62f60a4e96b47d00f394de0681?sid=79a84b4c-427e-4b0d-9f99-44d684b6f0bb

**The fix**

https://www.loom.com/share/0bc214e053b64e458a4bea57a5454d0f?sid=28e23e37-1510-4f56-9edf-248cb884f2bf

## Testing Instructions

- Interact with the honorific block & field
- Try submitting a value that is not in the listed options

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-1292]: https://stellarwp.atlassian.net/browse/GIVE-1292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
+++ b/src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
@@ -295,7 +295,8 @@ class ConvertDonationFormBlocksToFieldsApi
 
 
             if ($block->hasAttribute('showHonorific') && $block->getAttribute('showHonorific') === true) {
-                $options = array_values((array)$block->getAttribute('honorifics'));
+                $options = array_filter(array_values((array)$block->getAttribute('honorifics')));
+
                 if (!empty($options)){
                     $group->getNodeByName('honorific')
                         ->label(__('Title', 'give'))
--- a/src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
+++ b/src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
@@ -261,7 +261,7 @@ class ConvertDonationFormBlocksToFieldsApi
     }
 
     /**
-     * @unreleased updated honorific field with validation
+     * @unreleased updated honorific field with validation, global options, and user defaults
      *
      * @since 3.0.0
      */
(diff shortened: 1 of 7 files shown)


JoshuaHungDinh:
@jonwaldstein I noticed the build settings are not persisting properly after you refresh the page. It looks like the options state is being updated from a `useEffect` in givewp/src/FormBuilder/resources/js/form-builder/src/blocks/fields/donor-name/Edit.tsx. 

I confirmed this is currently an issue in develop as well but could be fixed with a conditional if you would like to include that in this PR since its relative to this setting

Heres a short video of the bug: .

https://github.com/user-attachments/assets/7b28c49f-e294-495b-ad78-e92194f026dd

src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
```diff
@@ -293,9 +295,13 @@ protected function createNodeFromDonorNameBlock(BlockModel $block): Node
 
 
             if ($block->hasAttribute('showHonorific') && $block->getAttribute('showHonorific') === true) {
-                $group->getNodeByName('honorific')
-                    ->label('Title')
-                    ->options(...array_values($block->getAttribute('honorifics')));
+                $options = array_values((array)$block->getAttribute('honorifics'));
```
JoshuaHungDinh, line 298: Should we check for empty strings when passing the options? I noticed you can add an empty option and it will display. Although maybe it is something we can look into another time.

JoshuaHungDinh (changes requested):

src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
jonwaldstein: nice catch! resolved f3eed612

jonwaldstein:
@JoshuaHungDinh this is ready for re-review.  Also fixed those options not persisting 74de0094

I think there are further improvements we can make here but this should be good for now.

JoshuaHungDinh (approved):
Awesome, this should be good to go!

rickalday (approved):
Pass manual QA tests.

PR #7565, opened 2024-10-07
Feature: Archive campaigns

alaca:
## Description

WIP

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Campaigns/ValueObjects/CampaignStatus.php
+++ b/src/Campaigns/ValueObjects/CampaignStatus.php
@@ -9,12 +9,14 @@ use Give\Framework\Support\ValueObjects\Enum;
  *
  * Statuses copied from https://github.com/impress-org/give-peer-to-peer/blob/develop/src/P2P/ValueObjects/Status.php
  *
+ * @method static CampaignStatus ARCHIVE()
  * @method static CampaignStatus ACTIVE()
  * @method static CampaignStatus INACTIVE()
  * @method static CampaignStatus DRAFT()
  * @method static CampaignStatus PENDING()
  * @method static CampaignStatus PROCESSING()
  * @method static CampaignStatus FAILED()
+ * @method bool isArchive()
  * @method bool isActive()
  * @method bool isInactive()
  * @method bool isDraft()
@@ -24,6 +26,7 @@ use Give\Framework\Support\ValueObjects\Enum;
  */
 class CampaignStatus extends Enum
 {
+    const ARCHIVE  = 'archive';
     const ACTIVE  = 'active';
     const INACTIVE  = 'inactive';
     const DRAFT = 'draft';
--- a/src/Campaigns/resources/admin/components/CampaignDetailsPage/CampaignDetailsPage.module.scss
+++ b/src/Campaigns/resources/admin/components/CampaignDetailsPage/CampaignDetailsPage.module.scss
@@ -73,6 +73,7 @@
 }
 
 .flexRow {
+    position: relative;
     display: flex;
     flex-direction: row;
     align-items: center;
@@ -272,7 +273,8 @@ select[name="campaignId"] {
             }
 
             input,
-            select {
+            select,
+            textarea {
                 font-size: 16px;
                 line-height: 2;
                 display: block;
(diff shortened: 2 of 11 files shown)


src/Campaigns/resources/admin/components/CampaignDetailsPage/index.tsx
```diff
@@ -120,29 +199,17 @@ export default function CampaignsDetailsPage({campaignId}) {
...
                             <div className={styles.flexRow}>
                                 <h1 className={styles.pageTitle}>{campaign.title}</h1>
-                                <StatusBadge />
+                                <span
+                                    className={cx(
+                                        styles.status,
+                                        styles[`${campaign.status}Status`],
+                                    )}
+                                >
+                                    {getStatus(campaign.status)}
```
glaubersilva, line 208: Is there some specific reason to replace the `<StatusBadge />` component with it? 

The main advantage of using the previous component is that it uses the same status styles from the list tables that live on this file: https://github.com/impress-org/givewp/blob/epic/campaigns/src/Views/Components/ListTable/InterweaveSSR/styles.scss

src/Campaigns/resources/admin/components/CampaignDetailsPage/CampaignDetailsPage.module.scss
```diff
@@ -118,6 +119,34 @@
...
+        background: #CEF2CF;
+        color: #022603;
+    }
+
+    &.archiveStatus {
+        background: #dadada;
+        color: #0e0e0e;
+    }
+}
+
```
glaubersilva, line 149: If we use the `<StatusBadge />` component as mentioned before, we don't need these styles for status.

src/Campaigns/ValueObjects/CampaignStatus.php
```diff
@@ -24,6 +26,7 @@
  */
 class CampaignStatus extends Enum
 {
+    const ARCHIVE  = 'archive';
```
glaubersilva, line 29: I would prefer to rename it `archived` to keep the same name displayed on the screen. 

Also, the CSS class is already using this naming here:  https://github.com/impress-org/givewp/blob/epic/campaigns/src/Views/Components/ListTable/InterweaveSSR/styles.scss#L41

Another thing to notice is that currently this status is not being displayed correctly in the campaigns list table, it's being displayed as "Draft" and with the wrong style because of this difference between the `enum` and the `scss` file.

After renaming it, it's also necessary to refactor the `src/Campaigns/ListTable/Columns/StatusColumn.php` file and add something like it there:

```
case 'archived':
    $statusLabel = __('Archived', 'give');
    break;
```

**Campaign 6881 on the list table page:**
image

**Campaign 6881 on the details page:**
image

glaubersilva (changes requested):
@alaca  This looks great man, good job! I just left a few comments related to the naming and styles for status. I would like to use the same pattern used in the campaigns list table which already was approved by Jeffrey previously.

src/Campaigns/resources/admin/components/CampaignDetailsPage/index.tsx
alaca: That was a result of resolving the merge conflict 😅 
Resolved: https://github.com/impress-org/givewp/pull/7565/commits/38f5fe7165024960ff6f09c973aff2fb4b68f406

src/Campaigns/ValueObjects/CampaignStatus.php
alaca: The problem with this approach is that it is not translatable. Anyway, I refactored the campaign status to `archived`, but I added support for translations as well https://github.com/impress-org/givewp/pull/7565/commits/1d32e125ad84371016ced11454a8b623534c3251

src/Campaigns/resources/admin/components/CampaignDetailsPage/CampaignDetailsPage.module.scss
```diff
@@ -118,6 +119,34 @@
     font-weight: 600;
 }
 
+.status {
```
glaubersilva, line 122: Now that we are using the <StatusBadge /> component again, we can remove this CSS class.

glaubersilva (approved):
@alaca Thanks for the changes. It looks better now. I just left a comment asking to remove an unused CSS class but I will approve it right now just to advance things. So please remember to remove it before merge, great work!

src/Campaigns/resources/admin/components/CampaignDetailsPage/CampaignDetailsPage.module.scss
alaca: Resolved: https://github.com/impress-org/givewp/pull/7565/commits/6ef588dac2c50f5fcf2c5663ecbc8b3cb9979326

PR #7572, opened 2024-10-15
Hot Fix: prevent PHP 8+ fatal errors when Tributes add-on is enabled

glaubersilva:
Resolves [GIVE-1306]

## Description

The problem was happening on PHP 8+ because the `ltrim()` method on those versions is more restrictive about which type of parameter is accepted. In previous versions, not-string values only threw wanings but in PHP 8+ version it threw a fatal error that prevented the donation workflow. 

So, the solution was to prevent the use of values different from a string on the `ltrim()` method.

## Affects

Option-based forms with the Tributes add-on enabled.

## Testing Instructions

1. Create a v2 form and enable Tributes;
2. Make a donation but do not opt-in for Tributes;
3. The donation should be finished without fatal errors.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-1306]: https://stellarwp.atlassian.net/browse/GIVE-1306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/includes/process-donation.php
+++ b/includes/process-donation.php
@@ -418,6 +418,7 @@ function give_donation_form_validate_fields() {
 /**
  * Detect serialized fields.
  *
+ * @unreleased Make sure only string parameters are used with the ltrim() method to prevent PHP 8+ fatal errors
  * @since 3.16.4 updated to check all values for serialized fields
  * @since 3.16.2 added additional check for stripslashes_deep
  * @since 3.14.2 add give-form-title, give_title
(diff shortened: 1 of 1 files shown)


jonwaldstein (approved):
Looks good to me, thanks!

rickalday (approved):
Passed manual QA tests

PR #7574, opened 2024-10-16
Chore: Replace test data from Campaign Stats

kjohnson:
## Description

This PR replaces the test data used in the Campaign Stats with real data returned by the API.

## Visuals

image


alaca (approved):
👍

PR #7578, opened 2024-10-21
Fix: Resolve issue with "Update Subscription" button being always disabled for Stripe

pauloiankoski:
Resolves [GIVE-1326]

## Description

This PR fixes an issue where the “Update Subscription” button was always disabled, preventing users from updating subscription amounts. The issue affected both the Option Based Form Editor (V2) and the Visual Form Builder (VFB) forms.

The problem stemmed from a recently introduced `id` property in the subscription status object, which will only be available in the next release of Give Recurring. To ensure backwards compatibility, this fix adds a fallback to use a lowercased version of the `label` property when `id` is not present.

With this adjustment, the “Update Subscription” button will be correctly enabled, allowing users to update subscription amounts as expected, regardless of whether their data includes the new id property.

## Affects
Manage Subscription on the Donor Dashboard

## Testing Instructions
1.	Create a subscription.
2.	Go to the Donor Dashboard.
3.	Click the “Manage Subscription” link.
4.	The “Update Subscription” button should now be enabled (the subscription must have an active status).

## Pre-review Checklist

-   [ ] Includes unit tests


JoshuaHungDinh (approved):
Good to go!

rickalday (approved):
Passed manual QA tests.

PR #7582, opened 2024-10-23
Fix: When using the Donor Wall shortcode or block, filtering by only_comments returns no donations

alaca:
## Description
When using the Donor Wall shortcode or block, filtering by `only_comments` returns no donations. 

The problem was that the donor comment was stored in the donation meta table, but we tried to fetch it from the `give_comments` table which resulted in no donations being returned. The issue is fixed by querying the correct table. 

## Affects

Donor Wall block/shortcode

## Visuals

https://github.com/user-attachments/assets/0f124b72-76ee-4af5-8fa9-61ebc0c546c4

## Testing Instructions

-  enable comments either on a per-form basis or at Donations >> Settings >> Default Options
- using GiveWP, create donations both with and without comments.
- use this shortcode [give_donor_wall only_comments="true"]
- or use the Donor Wall block and turn on the Only Comments at Display Elements >> Wall Attributes

## Pre-review Checklist

### DIFF
--- a/includes/donors/class-give-donor-wall.php
+++ b/includes/donors/class-give-donor-wall.php
@@ -479,8 +479,8 @@ class Give_Donor_Wall {
 
 		// exclude donations which does not has donor comment.
 		if ( $query_params['only_comments'] ) {
-			$sql   .= " INNER JOIN {$wpdb->give_comments} as gc1 ON (p1.ID = gc1.comment_parent)";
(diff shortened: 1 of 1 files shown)


kjohnson (commented):
Are we able to  cover this with a unit test?

rickalday (approved):
Passed manual QA tests

PR #7584, opened 2024-10-23
Feature: Add campaign default form widget

kjohnson:
Resolves [GIVE-1347]

## Description

This PR adds the "Default campaign form" widget to the Campaigns Overview page.

Additionally, it extracts the `<HeaderText/>` and `<HeaderSubText />` components for re-use in standalone components/widgets.

## Visuals

Screen Shot 2024-10-31 at 12 17 14

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-1347]: https://stellarwp.atlassian.net/browse/GIVE-1347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Campaigns/Actions/LoadCampaignDetailsAssets.php
+++ b/src/Campaigns/Actions/LoadCampaignDetailsAssets.php
@@ -32,17 +32,13 @@ class LoadCampaignDetailsAssets
             true
         );
 
+        $defaultForm = $campaign->defaultForm();
         wp_localize_script($handleName, 'GiveCampaignDetails',
             [
                 'adminUrl' => admin_url(),
                 'currency' => give_get_currency(),
                 'isRecurringEnabled' => defined('GIVE_RECURRING_VERSION') ? GIVE_RECURRING_VERSION : null,
-                'campaignForms' => array_map(function ($form) {
-                    return [
-                        'id' => $form->id,
-                        'title' => $form->title
-                    ];
-                }, $campaign->forms()->getAll()),
+                'defaultForm' => $defaultForm ? $defaultForm->settings->formTitle : null,
             ]
         );
 
--- a/src/Campaigns/Actions/LoadCampaignDetailsAssets.php
+++ b/src/Campaigns/Actions/LoadCampaignDetailsAssets.php
@@ -2,6 +2,7 @@
 
 namespace Give\Campaigns\Actions;
 
+use Give\Campaigns\Models\Campaign;
 use Give\Framework\Support\Facades\Scripts\ScriptAsset;
 
 /**
@@ -14,6 +15,10 @@ class LoadCampaignDetailsAssets
      */
     public function __invoke()
     {
+        $campaign = Campaign::find(
+            absint($_GET['id'])
+        );
+
         $handleName = 'givewp-admin-campaign-details';
         $scriptAsset = ScriptAsset::get(GIVE_PLUGIN_DIR . 'build/campaignDetails.asset.php');
 
@@ -32,6 +37,12 @@ class LoadCampaignDetailsAssets
                 'adminUrl' => admin_url(),
                 'currency' => give_get_currency(),
                 'isRecurringEnabled' => defined('GIVE_RECURRING_VERSION') ? GIVE_RECURRING_VERSION : null,
+                'campaignForms' => array_map(function ($form) {
+                    return [
+                        'id' => $form->id,
+                        'title' => $form->title
+                    ];
+                }, $campaign->forms()->getAll()),
             ]
         );
 
--- a/src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats.tsx
+++ b/src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats.tsx
@@ -65,7 +65,7 @@ const CampaignStats = () => {
                 <Column flex={1}>
                     <GoalProgressWidget />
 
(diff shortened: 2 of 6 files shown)


src/Campaigns/Actions/LoadCampaignDetailsAssets.php
```diff
@@ -32,6 +37,12 @@ public function __invoke()
                 'adminUrl' => admin_url(),
                 'currency' => give_get_currency(),
                 'isRecurringEnabled' => defined('GIVE_RECURRING_VERSION') ? GIVE_RECURRING_VERSION : null,
+                'campaignForms' => array_map(function ($form) {
```
kjohnson, line 40: This could probably be replaced with an entity, but I need to figure out how to fetch the correct endpoint for a nested resource, ie `campaigns/{id}/forms`.

glaubersilva (commented):
@kjohnson The code looks good to me, but I'm a bit confused about the goal of this widget. It shows a list of all available forms of the campaign in the drop-down with the first one being the default form, but how is the point? 

image


I thought it would have a button to update the default form which would use the "update campaign form" method implemented on PR #7585

Are we planning to change the default form automatically when selecting another form from the drop-down without clicking on any button?

kjohnson:
@glaubersilva I was expecting to address persistence when the logic for updating the default form was merged.

My understanding is that this will need to persist the change when the user clicks the "Update Campaign" button.

glaubersilva:
@kjohnson Hmm, this seems weird/confusing to me because when I think of a widget I think of a "stand-alone" component that has the whole functionality embedded inside it without the need to interact with external elements. 

The way it's working right now makes it feel more like a field that should be on the settings page than a widget. So, would like to double-check with @jdghinson if this is the expected behavior.

jdghinson:
@glaubersilva Good point! and I get where you are coming from, however, either we have at it as a setting or widget, you would still have to click on "update campaign" to reflect the changes.

@kjohnson If Glauber is finding this interaction confusing, I'm sure we will have similar issues with some of our customers, what do you think about having it as a setting like the image below?

image

glaubersilva:
@jdghinson Thanks for the insights! It also makes me wonder if we need this field on the settings page since we'll have the same functionality in the forms tab. Having it in both places seems a bit redundant to me. I would love to hear your feedback about it.

image

jdghinson:
@glaubersilva The setting option still help in cases where you have a lot of forms and you want to quickly change the default form instead of going through all the forms in the list table. Plus its in the settings page, where most people will likely go to when making changes to their campaign.

glaubersilva:
@jdghinson About it:

> The setting option still help in cases where you have a lot of forms and you want to quickly change the default form instead of going through all the forms in the list table. 

I think it is the opposite, the list table is easier because users can filter and search forms by name, but in the dropdown, find the proper form can be a bit more complicated in cases where there are lots of forms.

And about it:

> Plus its in the settings page, where most people will likely go to when making changes to their campaign.

As we have a tab with the name "forms", I think the logical path to make changes related to form is to go directly to this tab.

So, I agree with @kjohnson that keeping the feature only in the list table is enough.

jdghinson:
@glaubersilva The only intro we have with this feature is by the use of the tooltip which when dismissed you won't see again. There's a chance, users will dismiss it without reading, so to rely on the quick actions for discoverability won't be enough that's why I added the widget to show which form is currently the default, since the overview page is the first page you see when you open a campaign.

Cc: @kjohnson

glaubersilva:
@jdghinson Ok, I got your point about discoverability. So in that case, I think makes more sense to keep it as a widget, even if the behavior to update it is a bit weird since the main goal is to provide a quick overview for users.

kjohnson:
@jdghinson @glaubersilva what if we update the widget design to highlight the default form and link to the forms/table or the setting (as opposed to the widget being responsible for updating the campaign)?

We could double-down on this widget being informational and educational.

jdghinson:
@kjohnson I like this direction better, where the widget highlights the default form and there's an "edit default form" button in the widget which leads to the settings page. I will work on a design asap and sent it over for review.

Cc: @glaubersilva

jdghinson:
@glaubersilva It will be a confusing experience when we send them to the form list 'cos they clicked on edit and it should match their expectations.

kjohnson:
@jdghinson I've update the widget.

Screen Shot 2024-10-31 at 12 17 14

kjohnson:
@glaubersilva we are going to need a way to change the tab navigation from the default form widget. What do you think about reviewing this PR as introducing the widget and then introduce the navigation change in a separate PR?

glaubersilva:
@kjohnson No problem! For me sounds good. But if want to implement it here, all you need to do is set the ID of the tab in the tab parameter from the URL, it should arrive at the target tab. 

image


However, if the goal is to add an anchor to show the exact field that can be used to edit the form, then it would require a new implementation to create these anchors in all fields from the settings page since we can have similar situations like this one in the future.

glaubersilva:
@kjohnson Also, I think we need to keep the browser history for the navigation, so probably we can reuse the functions that handle the tabs navigation in this scenario which already handle the browser history.

glaubersilva (approved):
@kjohnson Nice work! Looks good to me. Just would adjust the margin of the "Edit" button because in large screens the text description is located too close to it - check the attached screenshot for reference.

image

PR #7597, opened 2024-10-30
Feature: add merge campaigns methods

glaubersilva:
Resolves [GIVE-1404]

## Description

This PR implements the `Campaign::merge()` and `CampaignRepository::mergeCampaigns()` methods in the model and repository classes which allow us to inform the `$campaingsToMerge` and `$destinationCampaign` as parameters.

Also, a few unit tests to check different outcomes generated by the `CampaignRepository::mergeCampaigns()` method were implemented.

## Affects

The campaign model and repository classes.

## Visuals

image

## Testing Instructions

1. Run the `composer test -- --filter CampaignRepositoryTest` command in your terminal.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-1404]: https://stellarwp.atlassian.net/browse/GIVE-1404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Campaigns/Repositories/CampaignRepository.php
+++ b/src/Campaigns/Repositories/CampaignRepository.php
@@ -265,53 +265,58 @@ class CampaignRepository
     /**
      * @unreleased
      *
-     * @param Campaign[] $campaigns
+     * @param Campaign[] $campaignsToMerge
      *
      * @throws Exception
      */
-    public function mergeCampaigns(array $campaigns, Campaign $destinationCampaign): bool
+    public function mergeCampaigns(array $campaignsToMerge, Campaign $destinationCampaign): bool
     {
-        Hooks::doAction('givewp_campaigns_merging', $campaigns, $destinationCampaign);
+        Hooks::doAction('givewp_campaigns_merging', $campaignsToMerge, $destinationCampaign);
 
         DB::query('START TRANSACTION');
 
+        // Make sure the destination campaign ID will not be included into $campaignsToMergeIds
+        $campaignsToMergeIds = array_column($campaignsToMerge, 'id');
+        if ($key = array_search($destinationCampaign->id, $campaignsToMergeIds)) {
+            unset($campaignsToMergeIds[$key]);
+        }
+
+        // Convert $campaignsToMergeIds to string to use it in the queries
+        $campaignsToMergeIdsString = implode(', ', $campaignsToMergeIds);
+
         try {
-            foreach ($campaigns as $campaign) {
-                if ($campaign->id === $destinationCampaign->id) {
-                    continue;
-                }
+            // Migrate forms from campaigns to merge to the destination campaign
+            DB::query(
+                DB::prepare("UPDATE " . DB::prefix('give_campaign_forms') . " SET campaign_id = %d WHERE campaign_id IN ($campaignsToMergeIdsString)",
+                    [
+                        $destinationCampaign->id,
+                    ])
+            );
 
-                DB::query(
-                    DB::prepare('UPDATE ' . DB::prefix('give_campaign_forms') . ' SET campaign_id = %d WHERE campaign_id = %d',
-                        [
-                            $destinationCampaign->id,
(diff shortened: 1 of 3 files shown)


src/Campaigns/Repositories/CampaignRepository.php
```diff
@@ -262,6 +262,65 @@ public function delete(Campaign $campaign): bool
...
     }
 
+    /**
+     * @unreleased
+     *
+     * @param Campaign[] $campaignsToMerge
+     *
+     * @throws Exception
+     */
+    public function mergeCampaigns(array $campaignsToMerge, Campaign $destinationCampaign): bool
```
JasonTheAdams, line 272: @glaubersilva Minor suggestion, I'd use a variadic parameter here:
```php
public function mergeCampaigns(Campaign $destinationCampaign, Campaign ...$campaignsToMerge)
```

Note that I am swapping the parameter order.

glaubersilva: @JasonTheAdams Thanks for the suggestion! Resolved in c3db75b53b83d2449c7b0cae4e2a003c7f310a32

src/Campaigns/Repositories/CampaignRepository.php
```diff
@@ -262,6 +262,63 @@ public function delete(Campaign $campaign): bool
...
+            // Migrate revenue entries from campaigns to merge to the destination campaign
+            DB::query(
+                DB::prepare("UPDATE " . DB::prefix('give_revenue') . " SET campaign_id = %d WHERE campaign_id IN ($campaignsToMergeIdsString)",
+                    [
+                        $destinationCampaign->id,
+                    ])
+            );
+
+            // Delete campaigns to merge now that we already migrated the necessary data to the destination campaign
+            DB::query("DELETE FROM " . DB::prefix('give_campaigns') . " WHERE id IN ($campaignsToMergeIdsString)");
```
alaca, line 303: @glaubersilva since every campaign has a default form, we should also delete these references in the `give_campaign_forms` table. But that begs the question, what happens with the forms from other campaigns after the merge?

glaubersilva: Nice catch! I think we should keep the default form from the destination campaign because in the UI we have a text which says that all settings from the destination campaign will be kept - as a note, we'll have a new field in the settings tab to set the default form. Resolved in 446f16fbb60712c59312a1b816cf2e5f9ba8ef72

But your question makes me think that we probably should refactor the `CampaignRepository::delete()` method to also delete the entries (that are associated with the campaign being deleted) on the `give_revenue` and `give_campaign_forms` tables.

glaubersilva: @alaca @kjohnson Let me know what you guys think about the change related to the `CampaignRepository::delete()` method that I mentioned in my previous comment. I think we probably should handle it in a subsequent PR, but I would like to know your thoughts about it.

alaca: @glaubersilva we shouldn't delete anything from the `give_revenue` table. 

If we keep relationships in `give_campaign_forms`, those forms will reference the campaign that no longer exists. Since forms _can't exist_ outside the context of a campaign, it will make sense to delete them. 

If we want to keep the settings data, then we can't delete anything and we need something else.

JasonTheAdams: To confirm, yes: Only the merged campaigns themselves should be deleted; all associated data (donations, revenue, etc.) should be re-associated with the final campaign being merged into.

alaca: Go ahead and merge it 👍 

Good job, @glaubersilva

alaca (approved):
👍

PR #7606, opened 2024-11-05
Feature: add "Make as default" action to the forms list view

glaubersilva:
Resolves [GIVE-1333]

## Description

This PR adds a "Make as default" action in the forms list table which lives on the forms tab from the campaigns details page.

## Affects

The forms list table.

## Visuals

image

## Testing Instructions

1. Go to the "Forms" tab from the Campaigns details page;
2. Make sure to have more than one form;
3. Use the "Make as default" action and make sure everything works as expected.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-1333]: https://stellarwp.atlassian.net/browse/GIVE-1333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Campaigns/Routes/UpdateDefaultCampaignForm.php
```diff
@@ -0,0 +1,59 @@
...
+use Give\Campaigns\Models\Campaign;
+use Give\Campaigns\Repositories\CampaignRepository;
+use Give\Campaigns\ValueObjects\CampaignRoute;
+use WP_REST_Response;
+use WP_REST_Server;
+
+/**
+ * @unreleased
+ */
+class UpdateDefaultCampaignForm
```
alaca, line 15: @glaubersilva I think it will make sense to use the existing `RegisterCampaignRoutes.php` and add this logic there.

alaca (approved):

alaca (commented):
I accidentally approved this PR. I mean, it looks good, but I would like to see this a bit more organized. See my previous comment.

alaca (changes requested):
@glaubersilva after testing this, it looks like setting the default form doesn't work. After clicking on the link, I get the option to set the default form for all forms in the list table.

src/Campaigns/Routes/UpdateDefaultCampaignForm.php
glaubersilva: Resolved in 85524863d9e5e1c925a8b4fda4b6a9c03089d036

glaubersilva:
@alaca Make sure to run the `npm run dev` instead of the `npm run dev:v3` and it should work as expected. Here is a demo video showing the feature in action: https://www.loom.com/share/6be837ce4af7412184a1b167680aca65?sid=6ac3df5d-f122-427c-b5bf-31b16e59e93f

Let me know if you were able to make it work and if there is anything else to change.

alaca:
@glaubersilva 

https://github.com/user-attachments/assets/55215633-1d2c-4659-b8e9-84fa5d568c4c

src/Campaigns/Actions/LoadCampaignDetailsAssets.php
```diff
@@ -33,12 +34,13 @@ public function __invoke()
...
 
         $defaultForm = $campaign->defaultForm();
+        $defaultFormTitle = Utils::isV3Form($defaultForm->id) ? $defaultForm->settings->formTitle : $defaultForm->title;
         wp_localize_script($handleName, 'GiveCampaignDetails',
             [
                 'adminUrl' => admin_url(),
                 'currency' => give_get_currency(),
                 'isRecurringEnabled' => defined('GIVE_RECURRING_VERSION') ? GIVE_RECURRING_VERSION : null,
-                'defaultForm' => $defaultForm ? $defaultForm->settings->formTitle : null,
+                'defaultForm' => $defaultFormTitle,
```
alaca, line 43: I don't see this being used anywhere. Also, this type of information can change without page reloading, so if you need this type of information it will be better to get it from the frontend.

src/DonationForms/V2/Endpoints/ListDonationForms.php
```diff
@@ -141,6 +142,7 @@ public function handleRequest(WP_REST_Request $request): WP_REST_Response
                 $item['permalink'] = get_permalink($item['id']);
                 $item['v3form'] = (bool)give_get_meta($item['id'], 'formBuilderSettings');
                 $item['status_raw'] = $forms[$i]->status->getValue();
+                $item['isDefaultCampaignForm'] = ($campaign = Campaign::findByFormId($item['id'])) && $item['id'] === $campaign->defaultForm()->id;
```
alaca, line 145: This shouldn't be in the loop because it's creating n+1, and we want to avoid that. You can get the default form ID outside the loop and then just compare the item ID against the default form ID.

src/Campaigns/Models/Campaign.php
```diff
@@ -84,6 +95,18 @@ public function forms(): ModelQueryBuilder
             })->where('campaign_forms.campaign_id', $this->id);
     }
 
+    /**
+     * @unreleased
+     */
+    public function legacyForms(): ModelQueryBuilder
```
alaca, line 101: Honestly, I'm not a big fan of this. The `forms` method should return all forms associated with the campaign regardless of whether it's V2 or V3. I understand that there are fundamental differences between the two, and I also know this would require extra effort to accommodate what I'm suggesting. @kjohnson any ideas?

Maybe the best approach to get what I want here is to refactor the  `defaultForm` method. Using the `give_campaign_forms` table we get the default form ID, then we check the type of the form (we can do all this in one query by joining `wp_posts`, `wp_postmeta` and `give_campaign_forms` and including `formBuilderSettings` meta key to the query - if the returned value is null it means that we have V2 form ), and then we return the correct form model.

alaca (commented):
Hey @glaubersilva nice job, man. There are some things that I would like to do differently and would like to hear what you think.

src/Campaigns/Models/Campaign.php
kjohnson: I just came across this while working on the Default Form select setting.

When I wrote the `forms()` method the assumption, if I remember correctly, was that these would be v3 forms. We'll need to update this to return both v2 and v3 forms.

kjohnson: This would be difficult to do as a single query with support for hydrating models for multiple types. We could extend the `ModelQueryBuilder` class and do some switching logic in the `getRowAsModel()` method.

src/Campaigns/Actions/LoadCampaignDetailsAssets.php
kjohnson: I had added this as part of the Default Form Widget work, but it is work-in-progress (and should probably be replaced with better state management).

src/DonationForms/V2/resources/components/DonationFormsRowActions.tsx
```diff
@@ -51,11 +53,35 @@ export function DonationFormsRowActions({data, item, removeRow, addRow, setUpdat
...
-            UpgradeModalContent,
+            __('Make as default', 'give'),
+            (selected) => (
+                <p>
+                    {__('Really make the following campaign form default?', 'give')}
+                    <br />
+                    <Interweave content={item?.title} />
+                </p>
+            ),
             async (selected) => {
```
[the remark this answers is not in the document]
glaubersilva: Resolved in 204f5575db7be9b0b9c22ec28c67cb062eb7ecf9

src/Campaigns/Actions/LoadCampaignDetailsAssets.php
glaubersilva: @alaca As @kjohnson mentioned, this is out of the scope of this PR, I only changed this line to make it work alongside the changes implemented on the `Campaign::defaultForm()` method as it now can retrieve both `V2` and `V3` forms.

src/DonationForms/V2/Endpoints/ListDonationForms.php
glaubersilva: Not sure if I understand it. How can I use the `Campaign::findByFormId($item['id'])` outside of the loop if the `$item` var is available only on the loop?

src/Campaigns/Models/Campaign.php
glaubersilva: @alaca I agree that this is not the best approach, but as we already have a [**task**](https://stellarwp.atlassian.net/browse/GIVE-1949) to update how the default form ID will be stored and retrieved from DB, I think we can leave it as it is right now to move forward and then revisit it later, even because a query written now well need get changed when we change the place to store the default form ID.

src/DonationForms/V2/Endpoints/ListDonationForms.php
alaca: All these forms are associated with one campaign, right? And we already have the campaign ID.

So instead of getting the campaign by form ID in a loop, we should get the campaign by campaign ID outside the loop. 

So instead of doing this:

```php            
foreach ($items as $i => &$item) {
      $item['isDefaultCampaignForm'] = ($campaign = Campaign::findByFormId($item['id'])) && $item['id'] === $campaign->defaultForm()->id;
} 
```

We should do something like this:

```php            
$defaultForm = Campaign::find($this->request->get_param('campaignId'))->defaultForm();

foreach ($items as $i => &$item) {
      $item['isDefaultCampaignForm'] = $item['id'] === defaultForm->id;
} 
```

The difference here is that we only make one request. What we have now is making requests for each form. So imagine someone has 100 forms, we'll be doing 100 unnecessary requests.

src/Campaigns/Models/Campaign.php
alaca: Hmm okay, that makes sense. I'm working on that task so I will refactor this anyway.

src/DonationForms/V2/Endpoints/ListDonationForms.php
glaubersilva: Oh man, I totally ignored the `campaignId` parameter, now makes sense, thanks for noticing that! Resolved in e27a1c4d729ebe225064a9a5c69578743fcf1834

alaca: Thanks, man!

alaca (approved):
Nice job, man! 👍 
I'll do some refactoring on the `forms` and `defaultForm` methods in separate PR.

Just resolve the failing tests before merging.

alaca:
@glaubersilva hey man, where is the route that makes form default? 😅

PR #7614, opened 2024-11-07
Feature: Add campaign setting for default form

kjohnson:
Resolves [GIVE-1493]

## Description

This PR adds the Default Form setting in the Campaign Details Page, on the Settings tab.

## Affects

- Adds `CampaignDetailsData` DTO class
- Updates response data to use DTO in `CampaignRequestController`

## Visuals

image

## Testing Instructions

Go to the Campaign Details Page
Go to the Settings tab
Select a default form
Publish the campaign
Check that the default form was updated on the Default Form widget

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-1493]: https://stellarwp.atlassian.net/browse/GIVE-1493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Campaigns/Actions/LoadCampaignDetailsAssets.php
+++ b/src/Campaigns/Actions/LoadCampaignDetailsAssets.php
@@ -5,6 +5,7 @@ namespace Give\Campaigns\Actions;
 use Give\Campaigns\Models\Campaign;
 use Give\DonationForms\Models\DonationForm;
 use Give\Framework\Support\Facades\Scripts\ScriptAsset;
+use Give\Helpers\Form\Utils;
 
 /**
  * @unreleased
@@ -38,7 +39,6 @@ class LoadCampaignDetailsAssets
                 'adminUrl' => admin_url(),
                 'currency' => give_get_currency(),
                 'isRecurringEnabled' => defined('GIVE_RECURRING_VERSION') ? GIVE_RECURRING_VERSION : null,
-                'defaultFormId' => $campaign->defaultForm()->id,
                 'donationForms' => array_map(function(DonationForm $form) {
                     return [
                         'id' => $form->id,
--- a/src/Campaigns/Actions/LoadCampaignDetailsAssets.php
+++ b/src/Campaigns/Actions/LoadCampaignDetailsAssets.php
@@ -3,6 +3,7 @@
 namespace Give\Campaigns\Actions;
 
 use Give\Campaigns\Models\Campaign;
+use Give\DonationForms\Models\DonationForm;
 use Give\Framework\Support\Facades\Scripts\ScriptAsset;
 
 /**
@@ -39,6 +40,12 @@ class LoadCampaignDetailsAssets
                 'currency' => give_get_currency(),
                 'isRecurringEnabled' => defined('GIVE_RECURRING_VERSION') ? GIVE_RECURRING_VERSION : null,
                 'defaultForm' => $defaultForm ? $defaultForm->settings->formTitle : null,
+                'donationForms' => array_map(function(DonationForm $form) {
+                    return [
+                        'id' => $form->id,
+                        'title' => $form->settings->formTitle,
+                    ];
+                }, $campaign->forms()->getAll()),
             ]
         );
 
--- a/src/Campaigns/Actions/LoadCampaignDetailsAssets.php
+++ b/src/Campaigns/Actions/LoadCampaignDetailsAssets.php
@@ -33,17 +33,16 @@ class LoadCampaignDetailsAssets
             true
         );
 
-        $defaultForm = $campaign->defaultForm();
         wp_localize_script($handleName, 'GiveCampaignDetails',
             [
                 'adminUrl' => admin_url(),
                 'currency' => give_get_currency(),
                 'isRecurringEnabled' => defined('GIVE_RECURRING_VERSION') ? GIVE_RECURRING_VERSION : null,
-                'defaultForm' => $defaultForm ? $defaultForm->settings->formTitle : null,
+                'defaultFormId' => $campaign->defaultForm()->id,
                 'donationForms' => array_map(function(DonationForm $form) {
                     return [
                         'id' => $form->id,
-                        'title' => $form->settings->formTitle,
+                        'title' => $form->hasProperty('settings') ? $form->settings->formTitle : $form->title,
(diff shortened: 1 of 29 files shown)


src/Campaigns/Routes/RegisterCampaignRoutes.php
```diff
@@ -231,7 +231,7 @@ public function getSchema(): array
                     'description' => esc_html__('Campaign goal type', 'give'),
                 ],
                 'defaultFormId' => [
-                    'type' => 'integer',
+                    'type' => 'number',
```
glaubersilva, line 234: There is any specific reason to use `number` instead of `integer` here?

src/Campaigns/DataTransferObjects/CampaignDetailsData.php
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * @unreleased
+     */
+    public function toArray(): array
+    {
+        return array_merge(
+            $this->campaign->toArray(),
+            [
+                'defaultFormId' => $this->campaign->defaultForm()->id,
```
glaubersilva, line 34: I think we don't need this DTO since the campaign model will have the default form ID available when this PR is merged: https://github.com/impress-org/givewp/pull/7616

src/Campaigns/Actions/LoadCampaignDetailsAssets.php
```diff
@@ -33,14 +34,17 @@ public function __invoke()
...
                 'adminUrl' => admin_url(),
                 'currency' => give_get_currency(),
                 'isRecurringEnabled' => defined('GIVE_RECURRING_VERSION') ? GIVE_RECURRING_VERSION : null,
-                'defaultForm' => $defaultFormTitle,
+                'donationForms' => array_map(function(DonationForm $form) {
+                    return [
+                        'id' => $form->id,
+                        'title' => $form->hasProperty('settings') ? $form->settings->formTitle : $form->title,
+                    ];
+                }, $campaign->forms()->getAll()),
```
glaubersilva, line 47: We need a way to load this form list beyond the first page load because it can get inaccurate easily depending of how user add new forms to the campaign - check my attached video for more details.

glaubersilva (changes requested):
@kjohnson I found a few problems while testing it and recorded a video to demonstrate that, here is the link: https://www.loom.com/share/f77ce3b6d30b4932ac81edb8d308972f?sid=48028945-a72f-469b-9eaa-1e67190af3e7

src/Campaigns/Routes/RegisterCampaignRoutes.php
kjohnson: My understanding is that `reactHookForms` is validating as a `number`. When I used `integer` it would throw an error. Likewise, the `goal` property of the schema uses `number`. The `id` property uses `integer`, but the `id` isn't being validated with `reactHookForms`.

src/Campaigns/DataTransferObjects/CampaignDetailsData.php
kjohnson: Good point. I'm looking at #7616 now.

src/Campaigns/Actions/LoadCampaignDetailsAssets.php
kjohnson: Is a list of forms available for re-use from the Campaign Forms List Table?

src/Campaigns/Routes/RegisterCampaignRoutes.php
glaubersilva: @kjohnson In this case, I think the best approach is to create a `react-hook-form` custom rule in the fronted to validate integer numbers because changing this parameter here to `number` will not reflect what exactly this endpoint accepts as a parameter. If someone using our API uses the `OPTIONS` to retrieve the parameters schema for this endpoint, they will see that the `id` can be a `number` which is wrong since it can only be an `integer` value. Also, if we decide to provide some documentation for our API using a plugin like the `wp-api-swaggerui` the docs generated by it will display the wrong type for this parameter as well.

src/Campaigns/Actions/LoadCampaignDetailsAssets.php
glaubersilva: @kjohnson  I don't know if our List Table component provides some way to access the list of available items through a JS object or something like that, but I think not, even because the list table can have pagination wich makes me think again that this field to update the default form trough the settings tab is not a good idea because in situations where a campaign has a massive amount of forms in a single campaign, populate this dropdown can become a problem.

src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/DefaultForm.tsx
```diff
@@ -1,12 +1,25 @@
...
 
 /**
  * @unreleased
  */
-const DefaultFormWidget = ({defaultForm}: {defaultForm: string}) => {
+const DefaultFormWidget = () => {
+
+    const {watch} = useFormContext();
+
+    const [defaultFormId] = watch(['defaultFormId']);
```
jonwaldstein, line 18: Any reason to not use the `useWatch` hook instead? I typically use the hook for the performance benefits unless I need to watch many values at the root level:

From the docs about [watch](https://react-hook-form.com/docs/useform/watch)
>This API will trigger re-render at the root of your app or form, consider using a callback or the [useWatch](https://react-hook-form.com/docs/usewatch) api if you are experiencing performance issues.

kjohnson: That would probably work. I had copy/pasted from another part of the codebase, but I'll refactor and try it.

alaca (commented):
Hey Kyle, nice job man. 
What would be really cool is that we use the `useEntityRecord` hook to get the campaign data, we don't have to localize scripts then. The `defaultFormId` should be a property of the campaign entity so we don't have to use the `react-hook-forms` watch method to get the form id because the entity will reflect that once it's updated.

kjohnson:
Closing as stale due to refactors and de-prioritization.

PR #7625, opened 2024-11-15
Fix: disable currency abbreviations in custom amounts

### LINKED ISSUE #7537: [Broken functionality] Entering currency amount in cs_CZ locale does not behave correctly
## User Story

I am using GiveWP plugin in the Czech Republic (cs_CZ locale) and currency is CZK (Czech Koruna) which is placed after the amount value. I'd like to type in the desired amount normally (digit by digit) without any issues.

## Details

Currently, when I type in first digit, then the cursor immediately moves to the end of the input field string. That is after the "Kč" currency label. Then next digit pressed immediately generates NaN error.
Other weird issue is, that when I enter the currency field with some value already there (like the predefined amounts in the Builder), and then just leave the field by clicking elsewhere, the field value is updated with extra "000" ... so from "1 000 Kč" string is suddenly "1 000 000 Kč". The same happens when pressing Delete button. 
I've seen this behavior so far in the Builder under the Donation Options and also in the Custom Amount field
 
## Expected Behavior

The currency input fields should work correctly in all locales no matter if the currency name is before or after the value.

## Steps to Reproduce

Use the latest Give WP plugin with the locale where its currency name is after the value. In my case it is czech locale (cs_CZ).

## Visuals

https://github.com/user-attachments/assets/f9e7d2ba-e11c-4fa8-910f-12fbd5e99700

## System Information

Give WP 3.16.1

</details>

JoshuaHungDinh:
Resolves #7537
Resolves [GIVE-1985]

## Description
References:
https://github.com/cchanxzy/react-currency-input-field/issues/255
https://github.com/cchanxzy/react-currency-input-field#abbreviations

Some languages in certain browsers update the `CustomAmount` on forms with abbreviated letters for the value. This results in an issue where the input can render `NaN`  & adds "000" for every additional 0 included.

This PR addresses the issue by disabling the abbreviations.

## Affects
Donation Forms

### Disable Abbreviations
https://github.com/user-attachments/assets/66d4cf7f-336f-44d0-b83b-2633e25fdf84

### With calculated separators
https://github.com/user-attachments/assets/6fbe9d9f-7c66-416d-84c5-5eb8d07d2ff3

## Testing Instructions
- Create a form.
- Using Chrome browser change the preferred language to Swedish.
- Update currency from settings to Swedish krona/SEK.
- View Form and enter amounts in the custom field, were testing that the correct number of "0"s are added each time and that `NaN` errors do not occur.

## Pre-review Checklist

[GIVE-1985]: https://stellarwp.atlassian.net/browse/GIVE-1985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kjohnson:
@JoshuaHungDinh the screen recording in the PR description suggests that the abbreviation is still used. Is that correct?

kjohnson:
Oh, ok, so the underlying issue was that it was interpreting the `k` as an abbreviations for `thousand`, which was being converted from `1k` to `1,000`. Interesting. That would explain why it was intermittent based on the currency.

> ## Abbreviations
> 
> It can parse values with abbreviations k, m and b
> 
> Examples:
> 
>     1k = 1,000
>     2.5m = 2,500,000
>     3.456B = 3,456,000,000
> 
> This can be turned off by passing in disableAbbreviations.

kjohnson:
> There was another suggestion to calculate the group and decimal separators outside of react-currency-input & then provide them to the component as props.

@JoshuaHungDinh is this a second, related issue being addressed?

src/DonationForms/resources/registrars/templates/fields/Amount/useCurrencySeparators.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Custom hook to determine the group and decimal separators based on locale.
+ */
+export default function useCurrencySeparator(){
+    const [groupSeparator, setGroupSeparator] = useState(',');
+    const [decimalSeparator, setDecimalSeparator] = useState('.');
+
+    useEffect(() => {
+        const formatter = new Intl.NumberFormat();
```
kjohnson, line 11: This will override the browser's locale and assume a US locale for separators.

While is does prevent the `NaN` issue, we shouldn't do so by locking the local config.

kjohnson (changes requested):

src/DonationForms/resources/registrars/templates/fields/Amount/useCurrencySeparators.tsx
kjohnson: Also, I don't think we need a hook for this, since the value shouldn't change after page load. Instead, we could initialize these variables at the top of the component file,

```tsx
import ...

const formatter = new Intl.NumberFormat();
// ...

const Component = () => {
    // ...
}

export default Component;
```

kjohnson: @JoshuaHungDinh I updated this to use the browser locale configuration and replaced the non-breaking space on the group separator to avoid any conflicts with the suffix separator used by the `<CurrencyInput />` component.

```tsx
const formatter = new Intl.NumberFormat(navigator.language);
const groupSeparator = formatter.format(1000).replace(/[0-9]/g, '');
const decimalSeparator = formatter.format(1.1).replace(/[0-9]/g, '');

  //
  groupSeparator={
    /**
     * Replace non-breaking space to avoid conflict with the suffix separator.
     * @link https://github.com/cchanxzy/react-currency-input-field/issues/266
     */
    groupSeparator.replace(/\u00A0/g, ' ')
  }
  //
```

kjohnson (approved):
@JoshuaHungDinh good job researching the underlying conflict!

kjohnson:
Build for QA, see https://github.com/impress-org/givewp/actions/runs/12205301456.

src/DonationForms/resources/registrars/templates/fields/Amount/useCurrencySeparators.tsx
JoshuaHungDinh: Awesome, that makes sense.

Thanks for helping me move this forward. I'll send it to QA.

PR #7626, opened 2024-11-15
Feature: Add deferred resolution to model factory definitions

kjohnson:
## Description

This PR adds deferred resolution to model factory definitions, such as nested model factories, so that defaults can be overridden without instantiating unused model instances.

## Affects

Updates `Give\Framework\Models\Factories\ModelFactory::makeInstance()` to support resolving attributes that are `callable`.

## Visuals

```php
public function definition(): array
{
    return [
        'title' => __('GiveWP Campaign', 'give'),
        'formId' => DonationForm::factory()->createAndResolveTo('id'),
    ];
}
```

## Testing Instructions

This feature is covered by unit tests and does not have a customer facing application.

## Pre-review Checklist

-   [x] Includes unit tests

### DIFF
--- a/src/Framework/Models/Factories/ModelFactory.php
+++ b/src/Framework/Models/Factories/ModelFactory.php
@@ -101,7 +101,7 @@ abstract class ModelFactory
     /**
      * @unreleased
      */
-    public function createAndResolveTo($property)
+    public function createAndResolveTo($property): Closure
     {
         return function() use ($property) {
             return is_array($results = $this->create())
--- a/src/Framework/Models/Factories/ModelFactory.php
+++ b/src/Framework/Models/Factories/ModelFactory.php
@@ -113,7 +113,7 @@ abstract class ModelFactory
     /**
      * Creates an instance of the model from the attributes and definition.
      *
-     * @unreleased Add support for resolving callables.
+     * @unreleased Add support for resolving Closures.
      * @since 2.23.0
      *
      * @return M
--- a/src/Framework/Models/Factories/ModelFactory.php
+++ b/src/Framework/Models/Factories/ModelFactory.php
@@ -121,7 +121,7 @@ abstract class ModelFactory
     protected function makeInstance(array $attributes)
     {
         return new $this->model(array_map(function($attribute) {
-            return is_callable($attribute) ? $attribute() : $attribute;
+            return $attribute instanceof Closure ? $attribute() : $attribute;
         }, array_merge($this->definition(), $attributes)));
     }
 
--- a/src/Framework/Models/Factories/ModelFactory.php
+++ b/src/Framework/Models/Factories/ModelFactory.php
@@ -113,6 +113,7 @@ abstract class ModelFactory
     /**
      * Creates an instance of the model from the attributes and definition.
      *
+     * @unreleased Add support for resolving callables.
      * @since 2.23.0
      *
      * @return M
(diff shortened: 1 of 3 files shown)


src/Framework/Models/Factories/ModelFactory.php
```diff
@@ -85,16 +98,31 @@ public function create(array $attributes = [])
         return $this->count === 1 ? $instances[0] : $instances;
     }
 
+    /**
+     * @unreleased
+     */
+    public function createAndResolveTo($property)
```
JasonTheAdams, line 104: Looks like we can add the `Closure` return type here, too!

src/Framework/Models/Factories/ModelFactory.php
```diff
@@ -85,16 +98,31 @@ public function create(array $attributes = [])
...
      * @since 2.23.0
      *
      * @return M
      */
     protected function makeInstance(array $attributes)
     {
-        return new $this->model(array_merge($this->definition(), $attributes));
+        return new $this->model(array_map(function($attribute) {
+            return is_callable($attribute) ? $attribute() : $attribute;
+        }, array_merge($this->definition(), $attributes)));
```
JasonTheAdams, line 125: I wonder if the `is_callable()` is too generic? For example, model properties can be a class (say a VO), and if that class instance is invokable, then that would pass as `is_callable()` in which case it would be called here unintentionally.

I think it's safer to use `$attribute instanceof Closure` instead. I can't think of a time wherein a model property would be a closure, as that's not something persistable.

JasonTheAdams (changes requested):
Great work, @kjohnson! A couple little suggestions! 😄

src/Framework/Models/Factories/ModelFactory.php
kjohnson: Good point. I added a test to confirm and the test failed. Updating the resolver to check for a `Closure` type passed the test.

```php
    public function testDoesNotResolveInvokableClasses()
    {
        $factory = new class(MockModelWithInvokableProperty::class) extends ModelFactory {
            public function definition(): array {
                return [
                    'invokable' => new class extends MockInvokableClass {
                        public function __invoke() {
                            throw new \Exception('Invokable classes should not be resolved by factories.');
                        }
                    },
                ];
            }
        };

        $object = $factory->make();

        $this->assertInstanceOf(MockInvokableClass::class, $object->invokable);
    }
```

kjohnson: Resolved, see [dd89630](https://github.com/impress-org/givewp/pull/7626/commits/dd896308a814d43cec9a6098301b95cc5c5a5656).

src/Framework/Models/Factories/ModelFactory.php
kjohnson: Good catch. I added the `Closure` to this method as well, see [2aa34ce](https://github.com/impress-org/givewp/pull/7626/commits/2aa34cedeecb683ad5d3cb936b38c05ed8268280).

JasonTheAdams (approved):
Nice work, @kjohnson! 👍 

Would you be willing, on your next Fun Friday, to add this to the StellarWP library? We _really_ need to move over to that library as to avoid drifting apart.

kjohnson:
> Would you be willing, on your next Fun Friday, to add this to the StellarWP library?

Yeah, that sounds like a good next step.

PR #7630, opened 2024-11-21
Feature: Update Campaign Donation Query

alaca:
## Description

This PR touches several places to update the `CampaignDonationQuery` class so that we are no longer using the `give_campaign_forms` lookup table to get donations. Since we don't have a donations table, the only fairly reasonable option was to add another meta field that will reference the campaign ID. The donation model is updated so that it is now possible to get a campaign from a donation object (`$donation->campaign->id`). The donation repository has also been updated to store the campaign ID as a donation meta. Also, Revenue DonationHandler is updated to insert the campaign ID on each donation, although I'm not entirely sure we need the campaign ID column in this table. It might be handy for reports. 

## Affects
Donation model
Donations repository
Donation meta
Give container
Revenue Donation Handler

## Pre-review Checklist


src/Campaigns/Migrations/Donations/AddCampaignId.php
```diff
@@ -0,0 +1,88 @@
...
+
+            foreach($donations as $donation) {
+                foreach ($relationships as $campaignId => $formIds) {
+                    if (in_array($donation->formId, $formIds)) {
+                        DB::table('give_donationmeta')
+                            ->insert([
+                                'donation_id' => $donation->ID,
+                                'meta_key' => DonationMetaKeys::CAMPAIGN_ID,
+                                'meta_value' => $campaignId,
+                            ]);
```
JasonTheAdams, line 77: Seems like this could be accomplished as a single insert of many rows?

src/Campaigns/Migrations/Donations/AddCampaignId.php
```diff
@@ -0,0 +1,88 @@
...
+    public static function timestamp(): string
+    {
+        return strtotime('2024-11-22 00:00:00');
+    }
+
+    /**
+     * @inheritDoc
+     * @throws DatabaseMigrationException
+     */
+    public function run()
```
JasonTheAdams, line 44: I can't remember. Is this automatically run inside of a transaction? Or should we apply one here?

src/Campaigns/Migrations/Donations/AddCampaignId.php
alaca: Yes, with SQL. With WPDB, not really. I would love to have that feature. Maybe we should consider updating the `insert` method to support that.

src/Campaigns/Migrations/Donations/AddCampaignId.php
alaca: Yep, see https://github.com/impress-org/givewp/blob/d056e04cbb1b43664b2b13c1df6fb7328cc0db11/src/Framework/Migrations/MigrationsRunner.php#L107

src/Campaigns/Migrations/Donations/AddCampaignId.php
JasonTheAdams: Seems like a shame to knowingly have a nasty N+1 issue here that we could definitely circumvent just because of a lack of a WPDB feature. I think it's worth taking the moment to build the insert many feature for this.

src/Revenue/DonationHandler.php
```diff
@@ -17,40 +17,26 @@ class DonationHandler
...
+        $campaign = give()->campaigns->getByFormId($formId);
 
-        return [
+        $data = [
             'donation_id' => $donationId,
-            'form_id' => $formId,
-            'amount' => $money->getMinorAmount(),
+            'form_id' =>  $formId,
+            'amount' => Money::of($amount, $currency)->getMinorAmount(),
+            'campaign_id' => $campaign ? $campaign->id : 0,
```
JasonTheAdams, line 37: Why is this 0 instead of null?

src/Campaigns/Migrations/Donations/AddCampaignId.php
alaca: @JasonTheAdams, I have been thinking about this in the past. Since data for inserting will probably be collected inside the loop, I thought having something like this might be handy: 

```php

        $qb = DB::table('give_donationmeta');

        foreach($data as $row) {
            $qb->set([
                'donation_id' => $row->id,
                'meta_key' => $row->key,
                'meta_value' => $row->value,
            ]);
        }

        $qb->insert();
```

This approach will require altering the `insert` method to be something like this:

```php
    public function insert($data = [], $format = null)
    {
        // collection represents collected data for insert
        if ($this->collection instanceof Collection) {
            $this->insertMany($this->collection);
        }
        
        if (empty($data)) {
            throw new InvalidArgumentException('data cannot be empty');
        }
        
        return DB::insert(
            $this->getTable(),
            $data,
            $format
        );
    }
```

src/Framework/QueryBuilder/Concerns/CRUD.php
```diff
@@ -22,6 +23,12 @@ trait CRUD
      */
     public function insert($data, $format = null)
     {
+        if (array_is_list($data)) {
```
JasonTheAdams, line 26: @alaca Do we have a polyfill for this somewhere? This is a PHP 8.1 function.

alaca: @JasonTheAdams Yeah. I installed and am using the stellarwp/arrays package now.

kjohnson (approved):

PR #7632, opened 2024-11-25
Feature: Add support for integrating Blink Payment with the Donor Dashboard

pauloiankoski:
Needed for https://github.com/impress-org/give-recurring/pull/1212 and https://github.com/impress-org/give-blink/pull/10/files

## Description
Blink Payments does not support updating a subscription amount, only the credit card number. For this reason, this Pull Request adds several WordPress filters that allow us to hide the Amount field on the Subscription Manager page. Additionally, it introduces a filter enabling third parties to enqueue custom scripts in the Donor Dashboard and allowing Blink to render its own `PaymentMethodControl` component.

## Affects
Donor Dashboard

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/DonorDashboards/App.php
+++ b/src/DonorDashboards/App.php
@@ -175,6 +175,8 @@ class App
             [],
             null
         );
+
+        do_action('give_donor_dashboard_enqueued_assets');
     }
 
     /**
--- a/src/DonorDashboards/App.php
+++ b/src/DonorDashboards/App.php
@@ -129,8 +129,9 @@ class App
     /**
      * Enqueue assets for front-end donor dashboards
      *
+     * @unreleased Add action to allow enqueueing additional assets.
+     * @since      2.11.0 Set script translations.
      * @since 2.10.0
-     * @since 2.11.0 Set script translations.
      *
      * @return void
      */
--- a/src/DonorDashboards/App.php
+++ b/src/DonorDashboards/App.php
@@ -177,7 +177,7 @@ class App
             null
         );
 
-        do_action('give_donor_dashboard_enqueued_assets');
+        do_action('give_donor_dashboard_enqueue_assets');
     }
 
     /**
--- a/src/DonorDashboards/resources/js/app/components/subscription-manager/index.tsx
+++ b/src/DonorDashboards/resources/js/app/components/subscription-manager/index.tsx
@@ -41,6 +41,7 @@ const SubscriptionManager = ({id, subscription}) => {
     const subscriptionStatus = subscription.payment.status?.id || subscription.payment.status.label.toLowerCase();
 
     const showAmountControls = subscription.gateway.can_update;
+    const showPaymentMethodControls = subscription.gateway.can_update_payment_method;
     const showPausingControls =
         subscription.gateway.can_pause && !['Quarterly', 'Yearly'].includes(subscription.payment.frequency);
 
@@ -109,11 +110,13 @@ const SubscriptionManager = ({id, subscription}) => {
                     onChange={setAmount}
                 />
             )}
-            <PaymentMethodControl
-                forwardedRef={gatewayRef}
-                label={__('Payment Method', 'give')}
-                gateway={subscription.gateway}
-            />
+            {showPaymentMethodControls && (
+                <PaymentMethodControl
+                    forwardedRef={gatewayRef}
+                    label={__('Payment Method', 'give')}
+                    gateway={subscription.gateway}
+                />
+            )}
 
             {loading && <DashboardLoadingSpinner />}
 
--- a/src/DonorDashboards/resources/js/app/components/subscription-manager/index.tsx
+++ b/src/DonorDashboards/resources/js/app/components/subscription-manager/index.tsx
@@ -41,7 +41,7 @@ const SubscriptionManager = ({id, subscription}) => {
     const subscriptionStatus = subscription.payment.status?.id || subscription.payment.status.label.toLowerCase();
 
     const showAmountControls = subscription.gateway.can_update;
-    const showPaymentMethodControls = subscription.gateway.can_update_payment_method;
+    const showPaymentMethodControls = subscription.gateway.can_update_payment_method ?? showAmountControls;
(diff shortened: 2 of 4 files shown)


src/DonorDashboards/App.php
```diff
@@ -175,6 +176,8 @@ public function loadAssets()
             [],
             null
         );
+
+        do_action('give_donor_dashboard_enqueued_assets');
```
jonwaldstein, line 180: Can we make this _enqueue_ instead of _enqueued_?

```suggestion
        do_action('give_donor_dashboard_enqueue_assets');
```

src/DonorDashboards/resources/js/app/components/subscription-manager/index.tsx
```diff
@@ -37,6 +40,11 @@ const SubscriptionManager = ({id, subscription}) => {
 
     const subscriptionStatus = subscription.payment.status?.id || subscription.payment.status.label.toLowerCase();
 
+    const showAmountControls = wp.hooks.applyFilters(
+        'give_donor_dashboard_subscription_manager_show_amount_controls',
+        true,
+        subscription
+    );
```
jonwaldstein, line 47: I think we might be able to simplify this:

What if instead of a front-end filter we tied into the gateway api and added a new front-end gateway property [here](https://github.com/impress-org/give-recurring/blob/c1b04da4dd74cc99bb52a4bfb94bb265930018cd/src/DonorDashboard/Routes/SubscriptionsRoute.php#L146) called 
```php
'can_update_payment_method' => $subscription->can_update()
``` 

Then within this subscription manager we would check for `gateway.can_update` and `gateway.can_update_payment_method` that would show the necessary controls.

Then by simply using the gateway api methods these would show up accordingly. You then wouldn't have to hack that `give_subscription_can_update_subscription` filter in blink as that may actually have some side effects.

jonwaldstein (changes requested):

src/DonorDashboards/resources/js/app/components/subscription-manager/index.tsx
pauloiankoski: I completely agree with you. In fact, one thought I had was: What if the `can_update` property were an array specifying what it can update? That way, we would only need to verify what could be affected by the change and ensure it doesn’t break any existing code.

jonwaldstein: Cool, maybe in the future we can go the array route or something similar.  For now, i'd prefer to keep the update properties explicit since we only support updating the amount and payment method at the moment.

pauloiankoski: Implemented on [a36cdee](https://github.com/impress-org/givewp/pull/7632/commits/a36cdee20c1d32b04cba2dc554dc427488f3014d) alongside [e542123](https://github.com/impress-org/give-recurring/pull/1212/commits/e542123a288a8879e8502353eb34f7cec691d50f) on Give Recurring

src/DonorDashboards/resources/js/app/components/subscription-manager/index.tsx
```diff
@@ -95,14 +99,16 @@ const SubscriptionManager = ({id, subscription}) => {
...
+                <AmountControl
+                    currency={subscription.payment.currency}
+                    options={options}
+                    max={max}
+                    min={min}
+                    value={amount}
+                    onChange={setAmount}
+                />
+            )}
             <PaymentMethodControl
```
jonwaldstein, line 112: I was thinking we should add a conditional for `gateway.can_update_payment_method` here as well, what do you think?

src/DonorDashboards/resources/js/app/components/subscription-row/index.tsx
```diff
@@ -52,14 +52,14 @@ const SubscriptionRow = ({subscription}) => {
                         {__('View Subscription', 'give')} <FontAwesomeIcon icon="arrow-right" />
                     </Link>
                 </div>
-                {gateway.can_update && (
+                {(gateway.can_update || gateway.can_update_payment_method) && (
```
jonwaldstein, line 55: Looks like we can make a const to be reused in this file like: 

```php
const gatewayCanUpdateSubscription = gateway.can_update || gateway.can_update_payment_method;
```

src/DonorDashboards/resources/js/app/components/subscription-manager/index.tsx
jonwaldstein: I just realized the gateway properties are coming from recurring, we might want to either use the nullish operator `gateway?.can_update_payment_method` or set some default prop values up the chain in the file when accessing the properties so they don't cause an error if someone is not on the latest version of recurring.

src/DonorDashboards/resources/js/app/components/subscription-manager/index.tsx
pauloiankoski: Added a `showPaymentMethodControl` constant, defaulting it to the same value of `showAmountControl` in case the value of the prop is undefined (when not in the latest version of Recurring) keeping the same behavior users have now.

src/DonorDashboards/resources/js/app/components/subscription-manager/index.tsx
pauloiankoski: Since `gateway` is a prop that we have already implemented, it is not supposed to be nullable. I am only accessing the new `can_update_payment_method` prop, which can be `undefined` at this point, and I am adding a fallback for that case.

src/DonorDashboards/resources/js/app/components/subscription-row/index.tsx
pauloiankoski: Yeah, much clearer.

src/DonorDashboards/resources/js/app/components/subscription-manager/index.tsx
jonwaldstein: thanks!

PR #7664, opened 2025-01-02
Feature: add API for gateway webhook events 

glaubersilva:
Resolves #

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Framework/PaymentGateways/ServiceProvider.php
+++ b/src/Framework/PaymentGateways/ServiceProvider.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace Give\Framework\PaymentGateways;
+
+use Give\Donations\ValueObjects\DonationStatus;
+use Give\Framework\PaymentGateways\Webhooks\EventHandlers\DonationAbandoned;
+use Give\Framework\PaymentGateways\Webhooks\EventHandlers\DonationCancelled;
+use Give\Framework\PaymentGateways\Webhooks\EventHandlers\DonationCompleted;
+use Give\Framework\PaymentGateways\Webhooks\EventHandlers\DonationFailed;
+use Give\Framework\PaymentGateways\Webhooks\EventHandlers\DonationPending;
+use Give\Framework\PaymentGateways\Webhooks\EventHandlers\DonationPreapproval;
+use Give\Framework\PaymentGateways\Webhooks\EventHandlers\DonationProcessing;
+use Give\Framework\PaymentGateways\Webhooks\EventHandlers\DonationRefunded;
+use Give\Framework\PaymentGateways\Webhooks\EventHandlers\DonationRevoked;
+use Give\Helpers\Hooks;
+use Give\ServiceProviders\ServiceProvider as ServiceProviderInterface;
+
+/**
+ * @unreleased
+ */
+class ServiceProvider implements ServiceProviderInterface
+{
+    /**
+     * @unreleased
+     */
+    public function register()
+    {
+        // TODO: Implement register() method.
+    }
+
+    /**
+     * @unreleased
+     */
+    public function boot()
+    {
+        $registeredPaymentGatewayIds = give()->gateways->getPaymentGateways();
+        foreach ($registeredPaymentGatewayIds as $gatewayId) {
+            $this->registerDonationEventHandlers($gatewayId);
+        }
+    }
+
+    /**
+     * @unreleased
+     */
+    private function registerDonationEventHandlers(string $gatewayId)
+    {
+        foreach (DonationStatus::values() as $status) {
+            switch ($status) {
+                case $status->isAbandoned():
+                    $this->addDonationStatusEventHandler($gatewayId, $status, DonationAbandoned::class);
+                    break;
+                case $status->isCancelled():
+                    $this->addDonationStatusEventHandler($gatewayId, $status, DonationCancelled::class);
+                    break;
+                case $status->isComplete():
+                    $this->addDonationStatusEventHandler($gatewayId, $status, DonationCompleted::class);
+                    break;
+                case $status->isFailed():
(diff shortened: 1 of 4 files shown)


src/Framework/PaymentGateways/Webhooks/WebhookEvents.php
```diff
@@ -0,0 +1,126 @@
...
+    {
+        $this->gatewayId = $gatewayId;
+    }
+
+    /**
+     * @unreleased
+     *
+     * @return int The webhook event ID. Zero if there was an error setting the event.
+     */
+    public function setDonationStatus(
```
jonwaldstein, line 32: I'm wondering if these methods are still too contextual and instead we should explore using our Gateway Api Commands like [PaymentComplete](https://github.com/impress-org/givewp/blob/e6895a23e2918d45266902845b492b0edd902ca7/src/Framework/PaymentGateways/Commands/PaymentComplete.php#L8) and [PaymentCompleteHandler](PaymentCompleteHandler) to abstract out some of the nuance and let the logic be more declarative.

Some pseudo examples:

```php
$gateway->webhookEvents()->paymentComplete($webhookNotification->gatewayPaymentId);
```

```php
$gateway->webhook()->markPaymentComplete($webhookNotification->gatewayPaymentId);
```

Open for discussion!

glaubersilva: @jonwaldstein  Initially, I went this way to have fewer methods and make the options provided by this API simpler. But thinking better, I think you are right because having more methods (one for each status/action) can simplify it even more and we can keep using the current methods under the hood. 

I liked the idea of your last sample because we can expand it (maybe in a subsequent PR) to support other common features when implementing the webhook logic.

For example: 

```
$gateway->webhook()->getUrl();
```

This way we can remove the necessity to create a route in the gateway integration class. But a change like that also makes me think that we need a specific interface for the _"webhookNotificationsListener"_ route/method.

For example, we could have something like this:

```
class OffSiteGateway extends PaymentGateway implements WebhookNotificationsListener
```

What do you think?

glaubersilva:
@jonwaldstein Your last comment inspired me to refactor it and now we have the following approach for methods related to webhook events:

```
TestGateway::webhook()->events->paymentCompleted($webhookNotification->gatewayPaymentId);

// or...

$this->webhook->events->paymentCompleted($webhookNotification->gatewayPaymentId);
```

For general methods related to webhooks, we have this:

```
TestGateway::webhook()->getNotificationUrl();

// or...

$this->webhook->getNotificationUrl();
```

This is ready for re-review.

glaubersilva:
@jonwaldstein Can you re-review this old PR? I would like to use this new approach in our GoCardless refactor. 🙂

Let me know your thoughts!

src/Framework/PaymentGateways/ServiceProvider.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        add_action('give_init', function () {
+            $registeredPaymentGatewayIds = give()->gateways->getPaymentGateways();
+            foreach ($registeredPaymentGatewayIds as $gatewayId) {
+                $this->addDonationStatusEventHandlers($gatewayId);
+                $this->addSubscriptionStatusEventHandlers($gatewayId);
+                $this->addSubscriptionFirstDonationEventHandler($gatewayId);
+                $this->addSubscriptionRenewalDonationEventHandler($gatewayId);
+            }
+        }, 999);
```
jonwaldstein: So, I know webhooks are _very_ common and most if not all gateways use them, however I'm wondering if gateways should opt-in to webhook support before adding support out of the box.  What do you think?

glaubersilva: @jonwaldstein I think it makes sense. Do you have any suggestions about the best way to do this opt-in? 

My initial idea is to enable it only for gateways that implement the `WebhookNotificationsListener` interface.

jonwaldstein: @glaubersilva yeah I do have some ideas that I think we should chat about.


Here's my initial thoughts:

**Option 1.** `WebhookNotificationsListener` interface. I I like how easy adding the interface is.  It does a good job of abstracting away the webhook logic, but I have a slight concern that it could also be too clever for webhooks specifically and maybe too limiting in terms of the url and logic. I'm also not sure the gateway route methods supports POST requests? will have to double check that.
**Option 2.** Some kind of explicit webhook registrar like `$gateway->registerWebhooks([$url, $listener])` (Maybe similiar to our subscription module?). This would make the webhooks more clear in their intention and flexible with the ability to add more than one, customize the url and listener, etc.  It would also be useful to have a full list of all registered webhooks that admins could be aware of.

glaubersilva: @jonwaldstein  I think option two will remove the simplicity of the approach implemented on this PR - the main idea here is to encapsulate the logic that gets repeated across multiple add-ons/gateways. 

However, if some add-on/gateway requires something too specific, we still can create a customized webhook URL and listener like we do currently, but it will be the exception since our current webhook event handler classes cover the most common cases.

But I like the idea of having some way to customize the default URL used to listen to the webhook notifications, and also would like to explore if it supports POST requests - I think we have a request about it in some place.

src/Framework/PaymentGateways/Actions/GenerateGatewayRouteUrl.php
```diff
@@ -28,7 +28,7 @@ public function __invoke(string $gatewayId, string $gatewayMethod, array $args =
 
         return esc_url_raw(add_query_arg(
             $queryArgs,
-            home_url()
+            'https://kubkavqjkm.sharedwithexpose.com/' //home_url()
```
jonwaldstein: @glaubersilva 💬 😄

PR #7681, opened 2025-01-17
Feature: add renewal methods to subscription model

jonwaldstein:
## Description

This adds some quality of life methods to subscription model including:
- `createRenewal`
- `shouldCreateRenewal`
- `shouldEndSubscription`

The goal is to unify the way we are dealing with renewals across our gateways.

I noticed throughout different gateways, we are duplicating a lot of the same logic for handling the renewals and are not consistent with the properties being used in the creation of the donation model.  For example, sometimes we use the donor's first name vs the initial donation's billing first name, the billing address is not often accounted for, various properties are missing, etc..  With a unified api we can be consistent with the renewal values and which properties are being carried over from the subscription, initial donation, donor, invoice, etc.

This was largely inspired by a [SubscriptionModelDecorator](https://github.com/impress-org/givewp/blob/8adc13337a1289e1a30f52dc9ec632c9b543d0d9/src/PaymentGateways/Gateways/Stripe/StripePaymentElementGateway/Webhooks/Decorators/SubscriptionModelDecorator.php#L14) I created for Stripe webhooks which is being used in production.  

There is also the[ webhook event handlers](https://github.com/impress-org/givewp/blob/304c495bd0a727b0d8e1149c54e7fb1bfecdbcb0/src/Framework/PaymentGateways/Webhooks/EventHandlers/SubscriptionRenewalDonationCreated.php#L75) that @glaubersilva made that could make use of these methods directly.

## Affects

This only introduces new logic to the subscription model and repositories, however will affect gateways in the future as they adopt these methods.

## Visuals

N/A

## Testing Instructions

- Programmatically test the methods

## Pre-review Checklist

-   [x] Includes unit tests

### DIFF
--- a/src/Subscriptions/Models/Subscription.php
+++ b/src/Subscriptions/Models/Subscription.php
@@ -186,7 +186,11 @@ class Subscription extends Model implements ModelCrud, ModelHasFactory
      */
     public function shouldCreateRenewal(): bool
     {
-        return $this->status->isActive() && ($this->isIndefinite() || $this->totalDonations() < $this->installments);
+        if (!$this->status->isActive()) {
+            return false;
+        }
+
+        return $this->isIndefinite() || $this->totalDonations() < $this->installments;
     }
 
     /**
--- a/src/Subscriptions/Models/Subscription.php
+++ b/src/Subscriptions/Models/Subscription.php
@@ -186,10 +186,15 @@ class Subscription extends Model implements ModelCrud, ModelHasFactory
      */
     public function shouldCreateRenewal(): bool
     {
-        $billTimes = $this->installments;
-        $totalPayments = count($this->donations);
+        return $this->status->isActive() && ($this->isIndefinite() || $this->totalDonations() < $this->installments);
+    }
 
-        return $this->status->isActive() && (0 === $billTimes || $totalPayments < $billTimes);
+    /**
+     * @unreleased
+     */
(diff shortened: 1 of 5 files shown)


src/Subscriptions/Models/Subscription.php
```diff
@@ -172,6 +172,36 @@ public static function create(array $attributes): Subscription
...
+        return give()->subscriptions->createRenewal($this, $attributes);
+    }
+
+    /**
+     * @unreleased
+     */
+    public function shouldCreateRenewal(): bool
+    {
+        $billTimes = $this->installments;
+        $totalPayments = count($this->donations);
```
JasonTheAdams, line 190: This PR looks handy, @jonwaldstein! What do you think about making a `$subscription->totalDonations()` method to make this declarative (and in the other method)? Also, does `$this->donations` query and instantiate every single donation? That could be quite costly for some subscriptions (e.g. giving weekly for years). Is it possible to do something like `$this->donations()->count()`?

jonwaldstein: Oh yeah, good call i'll update it

jonwaldstein: updated c719b1bd

JasonTheAdams (approved):
This is such a great addition, @jonwaldstein! Great call! 🎉

PR #7698, opened 2025-01-27
Fix: Prevent NaN error in Campaign Overview Stat Widget

kjohnson:
Resolves GIVE-2031

## Description

This PR updates the `<StatWidget>` component to prevent a `NaN` error while fetching data.

Additionally, a non-breaking space is used to prevent content shifting.

## Visuals

Peek 2025-01-27 15-39-2

## Testing Instructions

- Hard-reload the page (with cache disabled via browser tools)
- Toggle between tabs in the Overview screen

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats/index.tsx
+++ b/src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats/index.tsx
@@ -93,7 +93,7 @@ const StatWidget = ({label, values, description, formatter = null}) => {
             </header>
             <div className={styles.statWidgetAmount}>
                 <DisplayText>
-                    {formatter?.format(values[0]) ?? values[0]}
+                    {!isNaN(values[0]) ? formatter?.format(values[0]) ?? values[0] : (<span>&nbsp;</span>)}
                 </DisplayText>
                 {!! values[1] && (
                     <PercentChangePill value={values[0]} comparison={values[1]} />
--- a/src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats/index.tsx
+++ b/src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats/index.tsx
@@ -3,7 +3,7 @@ import {useEffect, useState} from "react";
 import RevenueChart from "../RevenueChart";
 import GoalProgressChart from "../GoalProgressChart";
 import apiFetch from '@wordpress/api-fetch';
-import { addQueryArgs } from '@wordpress/url';
+import {addQueryArgs} from '@wordpress/url';
 import HeaderText from '../HeaderText';
 import HeaderSubText from '../HeaderSubText';
 import DefaultFormWidget from "../DefaultForm";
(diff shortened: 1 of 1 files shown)


src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats/index.tsx
```diff
@@ -93,7 +93,7 @@ const StatWidget = ({label, values, description, formatter = null}) => {
             </header>
             <div className={styles.statWidgetAmount}>
                 <DisplayText>
-                    {formatter?.format(values[0]) ?? values[0]}
+                    {!isNaN(values[0]) ? formatter?.format(values[0]) ?? values[0] : (<span>&nbsp;</span>)}
```
jonwaldstein, line 96: @kjohnson can you help us understand what could cause one of these values to be considered NaN?  Is it an issue with data still loading or is it possible for a value to resolve as a non number?

kjohnson: The isssue here is that the data is loading. The endpoint would only return a numeric value.

jonwaldstein: Okay, If that's the case then I feel it would be more sensible for the stat widget to have a loading state.  You could then add a css class for the loading state to be styled.  

Without the loading state we are making the assumption the value is always resolved when passed to the component, which is not true here - raising questions of why it could be a non number.

kjohnson: This passes a non-breaking space as a placeholder if the value is not provided, which is during loading. While not a _proper_ loading state, this does cut down on the amount of visual change while loading. I think it would be more distracting for a row of widgets to flash loading spinners or something.

Note that I did update the check from a NaN check to an undefined check, see https://github.com/impress-org/givewp/pull/7698/commits/d3077e7d6192bbefd8e30d8305b883145e5cf0cb.

jonwaldstein: Regardless of the design, what is the reason for _not_ having a loading state in the logic? 

Loading states tell us when our data is resolved and ready to be used. They are useful in the application's logic for conditionally rendering data.  In this scenario we have data that is not resolved yet but we are attempting to render it regardless.  There's lots of ways we can go about this visually but to me the important part is we being intentional about our data loading and rendering.

glaubersilva: About the design for the "loading state" maybe we can use something similar to the `give-skeleton` class, here are a few links for reference:

https://github.com/impress-org/givewp/blob/develop/src/DonationForms/AsyncData/AsyncDataHelpers.php#L34

https://github.com/impress-org/givewp/blob/develop/src/DonationForms/AsyncData/resources/loadAsyncData.scss

And here is the screenshot with the class in action:

image

glaubersilva: Or even better, we can use a package like this one: https://www.npmjs.com/package/react-loading-skeleton

kjohnson: I really don't think this needs a special loading state. It is a simple widget which shows a number returned by the server. We aren't running exspensive queries and we aren't loading tables of data. There isn't a fancy display to be stubbed by a skeleton.

This PR prevents a flashed `NaN` and additionally prevents a content shift when the data is loaded.

kjohnson: Circling back on this, the point was raised that this bug would have been prevented if we had accounted for loading state higher up in the flow of data. While I agree with this point, my goal with this PR was to immediately address the issue of the flashed `NaN` bug as it was presented and refined by the team in accordance with the allotted time of the development cycle. Being that there has been feedback as to how to improve the flow of data, it is my intention that this component architecture be updated at a later time. The intent of which is to balance the cadence of addressing reported issues with the deferred cost of accrued technical debt while meeting negotiated timelines.

To move this work forward it is my suggestion that we continue with the check for undefined data within the nested component while further discussing the best approach for handling undefined data in a subsequent pull request.

kjohnson: @jonwaldstein @glaubersilva any other thoughts?

jonwaldstein: @kjohnson that plan sounds good to me, thanks!

glaubersilva (approved):
👍

jonwaldstein (approved):

kjohnson:
Build for QA, see https://github.com/impress-org/givewp/actions/runs/13139955039.

rickalday (approved):
Passed manual QA tests.

PR #7699, opened 2025-01-28
Feature: add new routes for donations and donors

glaubersilva:
Related to [GIVE-1392] and [GIVE-1393]

## Description

This PR implements 4 new REST API endpoints to retrieve `Donations` and `Donors`. In the endpoints that return multiple entries, is possible to filter the returned data using custom parameters in the request and also is possible use pagination and sort the results using the `page`, `per_page`, `sort` and `direction` parameters.

Another thing to consider is that sensitive data will be returned **only** if the user making the request is the site administrator and the `includeSensitiveData` param is set to `true`.

**Sensitive data for donations:**

```
$sensitiveData = [
    'donorIp',
    'email',
    'phone',
    'billingAddress',
];
```

**Sensitive data for donors:**

```
$sensitiveData = [
    'userId',
    'email',
    'phone',
    'additionalEmails',
];
```

**The new endpoints to retrieve a single entry:**

```
/givewp/v3/donations/(?P<id>[0-9]+)

/givewp/v3/donors/(?P<id>[0-9]+)
```

**The new endpoints to retrieve multiple entries:**

```
/givewp/v3/donations

/givewp/v3/donors
```

**Important:** These endpoints allow filtering the returned data through the `campaignId` parameter. It's also possible to use the `includeSensitiveData` (only available for admins) or `anonymousDonations` and `anonymousDonors` (only admins can include anonymous donations/donors without redacted data) parameters to exclude from the results the donations/donors that made anonymous donations and/or sensitive data that can be used to contact/localize the donors. Beyond that, on the `/givewp/v3/donors` endpoint, it is possible to use the `onlyWithDonations` parameter to retrieve all donors or just the ones that have valid donations completed. For more details, check this comment: https://github.com/impress-org/givewp/pull/7699#discussion_r1988065812

**Sample request including anonymous donations in the results with redact data:**

```
$request = new WP_REST_Request('GET' 'givewp/v3/donations');

$request->set_query_params(
    [
        'anonymousDonations' => 'redact',   // Data that can identify donors will be displayed as "anonymous"
        //'campaignId' => $campaign1->id, // Uncomment this line to filter by campaign
    ]
);
```

**Sample request including anonymous donors in the results with redact data:**

```
$request = new WP_REST_Request('GET' 'givewp/v3/donors');

$request->set_query_params(
    [
        'anonymousDonors' => 'redact',   // Data that can identify donors will be displayed as "anonymous"
        //'campaignId' => $campaign1->id, // Uncomment this line to filter by campaign
    ]
);
```

**Sample request to retrieve the 5 most recent donations:**

```
$request = new WP_REST_Request('GET' 'givewp/v3/donations');

$request->set_query_params(
    [
        'page' => 1,
        'per_page' => 5,
        'sort' => 'createdAt',
        'direction' => 'DESC',
        'anonymousDonations' => 'redact', // Data that can identify donors will be displayed as "anonymous"
        //'campaignId' => $campaign1->id, // Uncomment this line to filter by campaign
    ]
);
```

**Sample request to retrieve the top 5 donors:** 

```
$request = new WP_REST_Request('GET' 'givewp/v3/donors');

$request->set_query_params(
    [
        'page' => 1,
        'per_page' => 5,
        'sort' => 'totalAmountDonated',
        'direction' => 'DESC',
        'anonymousDonors' => 'redact',   // Data that can identify donors will be displayed as "anonymous"
        //'campaignId' => $campaign1->id, // Uncomment this line to filter by campaign
    ]
);
```

## Affects

GiveWP Rest API endpoints available for public use.

## Testing Instructions

In your terminal, run the following commands:

```
composer test -- --filter GetDonationRouteTest

composer test -- --filter GetDonationsRouteTest

composer test -- --filter GetDonorRouteTest

composer test -- --filter GetDonorsRouteTest
```

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-1392]: https://stellarwp.atlassian.net/browse/GIVE-1392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GIVE-1393]: https://stellarwp.atlassian.net/browse/GIVE-1393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Donors/Controllers/DonorRequestController.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        $page = $request->get_param('page');
+        $perPage = $request->get_param('per_page');
+
+        $query = give(DonorRepository::class)->prepareQuery();
+
+        if ($campaignId = $request->get_param('campaignId')) {
+            $query->select(['donationmeta1.donation_id', 'donationId'])
+                ->distinct()
+                ->join(function (JoinQueryBuilder $builder) use ($campaignId) {
```
kjohnson, line 49: Should this use the `attachMeta` functionality of the Query Builder?

src/Donors/Controllers/DonorRequestController.php
```diff
@@ -0,0 +1,103 @@
...
+
+                    $builder->innerJoin('give_donationmeta', 'donationmeta2')
+                        ->joinRaw("ON donationmeta2.meta_key = '" . DonationMetaKeys::CAMPAIGN_ID . "' AND donationmeta2.meta_value = {$campaignId}");
+                })->whereIn('donationmeta1.donation_id', function (QueryBuilder $builder) {
+                    $builder
+                        ->select('ID')
+                        ->from('posts')
+                        ->whereRaw("WHERE ID = donationmeta1.donation_id AND post_type = 'give_payment' AND post_status = 'publish'");
+                });
+        }
```
kjohnson, line 61: If I understand this corrently, this is querying the Donors table and joining in the Donations associated with a specific Campaign, then limiting those results based on the status of the joined Donation?

Would it make more sense to query the Donors table where in a list of Donations associated with the Campaign? This would encapsulate the donation query, instead of conflating the two.

The end result is the same, but the structure of the relationship is different.

From a maintenance perspective I think this would be easier to understand when returning to this code.

Thinking out loud, the `select` for the `donationId` and the use of `distinct` stood out as possible red flags.

kjohnson: We could also leverage the [CampaignDonationQuery](https://github.com/impress-org/give/blob/18049358a9b75ea37a57f04cd4567b18ab0647d5/src/Campaigns/CampaignDonationQuery.php#L14), which would also account for the payment mode and include subscription donations.

```php
(new CampaignDonationQuery($campaign))
    ->joinDonationMeta(DonationMetaKeys::DONOR_ID, 'donorId')
    ->select('donorId')
    ->distinct()
```

src/Donations/Controllers/DonationRequestController.php
```diff
@@ -0,0 +1,91 @@
...
+     */
+    public function getDonations(WP_REST_Request $request): WP_REST_Response
+    {
+        $page = $request->get_param('page');
+        $perPage = $request->get_param('per_page');
+
+        $query = give(DonationRepository::class)->prepareQuery();
+
+        if ($campaignId = $request->get_param('campaignId')) {
+            $query->distinct()->join(function (JoinQueryBuilder $builder) {
```
kjohnson, line 45: This should probably use the `CampaignDonationQuery`, which would cover subscription donations and account for the payment mode (live vs test).

src/Donations/Controllers/DonationRequestController.php
```diff
@@ -0,0 +1,91 @@
...
+            ->limit($perPage)
+            ->offset(($page - 1) * $perPage);
+
+        $donations = $query->getAll() ?? [];
+        $totalDonations = empty($donations) ? 0 : Donation::query()->count();
+        $totalPages = (int)ceil($totalDonations / $perPage);
+
+        $response = rest_ensure_response($donations);
+        $response->header('X-WP-Total', $totalDonations);
+        $response->header('X-WP-TotalPages', $totalPages);
```
kjohnson, line 61: Cool to see headers and links used to add pagination! You should consider extracting this for re-use as a Friday project.

src/Donors/Controllers/DonorRequestController.php
```diff
@@ -0,0 +1,103 @@
...
+
+        $query
+            ->limit($perPage)
+            ->offset(($page - 1) * $perPage);
+
+        $donors = $query->getAll() ?? [];
```
kjohnson, line 67: Does this need to account for Anonymous Donations? Should those be removed or redacted?

kjohnson (commented):
Cool seeing headers and links for pagination! This is often overlooked, but good that we are starting to leverage this feature.

I added some feedback on the queries. In particular, we have the `CampaignDonationQuery` which we should consider using - or at least update what is here to account for subscriptions and test payments.

src/Donors/Controllers/DonorRequestController.php
pauloiankoski: Btw, in the Donors Block, we have a “Show Anonymous” setting where the admin can decide whether to show or hide anonymous donations. That might be an argument for this route. Additionally, we might have the opportunity to set an `orderby` argument, as we’ll need to retrieve the top donors based on the amount they donated.

src/Donations/Routes/RegisterDonationRoutes.php
```diff
@@ -0,0 +1,104 @@
...
+        register_rest_route(
+            DonationRoute::NAMESPACE,
+            DonationRoute::DONATION,
+            [
+                [
+                    'methods' => WP_REST_Server::READABLE,
+                    'callback' => function (WP_REST_Request $request) {
+                        return $this->donationRequestController->getDonation($request);
+                    },
+                    'permission_callback' => '__return_true',
```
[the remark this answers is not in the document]
glaubersilva: @jonwaldstein Good call! I implemented a function to restrict the data returned to users who didn't have the `manage_options` capability. 

Beyond the validation, this function also ensures that we always return the data as an array instead of returning it as model objects so we can keep the consistency of the returns.

See: bfe2a6f230457998b764cb26b72e84ccf0dafc44 and 4dcb2d23e46079c1a45691cac2c4cf429d596eb9

About the GiveWp API credentials, I think it can be implemented in a subsequent PR to not increase the scope of this one.

src/Donations/Controllers/DonationRequestController.php
glaubersilva: Since this endpoint is related to the `Donation` resource, I would prefer to use the `Donation::query()` to generate the base query and then add the proper where conditions and filters attached to it. Even because the $campaingId parameter is optional.

Resolved in 7a086f52fdae9c9df1da630d9967e10931ef4d15 and 4fca5af2d7ecc1665e6f4f951c01efca53652b02

src/Donors/Controllers/DonorRequestController.php
glaubersilva: @kjohnson I added a new `hideAnonymousDonations` param here: 3ada42b77054d82363f10c2459ea87d321135567

@pauloiankoski The orderBy still needs to be implemented.

glaubersilva: After my last comment, I realized that you folks are commenting in the `DonorRequestController` class and I implemented the new param on the `DonationRequestController` class. However, I'm working to implement a similar approach in the Donor controller as well. I'll comment here when it's done.

glaubersilva: Now is possible to exclude anonymous donations/donors and also order by the results using different columns and and directions.

Since I did a bunch of commits with these changes, please re-check the PR description because I included all the details about these changes.

src/Donors/Controllers/DonorRequestController.php
glaubersilva: I fully refactor how this query works...

Since I did a bunch of commits with these changes, I'll not copy and paste all of them here because it will require a new full review in this PR.

src/Donations/Routes/RegisterDonationRoutes.php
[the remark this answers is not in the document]
glaubersilva: I included more details in the PR description about these changes and also added a couple of other unit tests to make sure the sensitive data are returned for admin users doing the request.

JasonTheAdams:
Out of curiosity, is the `give-api` namespace pre-existing? The `-api` suffix part feels redundant since this is part of the REST API path.

glaubersilva:
@JasonTheAdams Yes, it's a pre-existing thing. 

image

JasonTheAdams:
I'm seriously considering recommending a `give/v3` or something that deviates from the old REST API.

src/Donors/Controllers/DonorRequestController.php
```diff
@@ -0,0 +1,173 @@
...
+
+        // Donors only can be donors if they have donations associated with them
+        if ($request->get_param('onlyWithDonations')) {
+            $query->join(function (JoinQueryBuilder $builder) {
+                // The donationmeta1.donation_id should be used in other "donationmeta" joins to make sure we are retrieving data from the proper donation
+                $builder->innerJoin('give_donationmeta', 'donationmeta1')
+                    ->joinRaw("ON donationmeta1.meta_key = '" . DonationMetaKeys::DONOR_ID . "' AND donationmeta1.meta_value = ID");
+
+                // Include only current payment "mode"
+                $mode = give_is_test_mode() ? 'test' : 'live';
```
JasonTheAdams, line 55: @glaubersilva I don't advise this. RESTful APIs should not be stateful. They simply return the data as requested without determining internal state. This should be an enum query parameter that is either "test" or "live". I _could_ be talked into an "auto" enum that checks this, but it should not be the default — though I would prefer it remaining truly stateless.

src/Donations/Controllers/DonationRequestController.php
```diff
@@ -0,0 +1,147 @@
...
+            $query->where('give_donationmeta_attach_meta_campaignId.meta_value', $campaignId);
+        }
+
+        if ($request->get_param('hideAnonymousDonations')) {
+            // Exclude anonymous donations from results
+            $query->where('give_donationmeta_attach_meta_anonymous.meta_value', 0);
+        }
+
+        // Include only current payment "mode"
+        $query->where('give_donationmeta_attach_meta_mode.meta_value', give_is_test_mode() ? 'test' : 'live');
```
JasonTheAdams, line 55: See my other comment on why I don't recommend using `give_is_test_mode()`.

src/Donations/Controllers/DonationRequestController.php
```diff
@@ -0,0 +1,147 @@
...
+
+        if ($campaignId = $request->get_param('campaignId')) {
+            // Filter by CampaignId
+            $query->where('give_donationmeta_attach_meta_campaignId.meta_value', $campaignId);
+        }
+
+        if ($request->get_param('hideAnonymousDonations')) {
+            // Exclude anonymous donations from results
+            $query->where('give_donationmeta_attach_meta_anonymous.meta_value', 0);
+        }
```
JasonTheAdams, line 52: I recommend keeping parameters in the positive as it avoids double negatives (I want to not hide...). Let's rename this to `includeAnonymousDonations` and make it true by default.

src/Donors/Controllers/DonorRequestController.php
```diff
@@ -0,0 +1,173 @@
...
+
+            if ($campaignId = $request->get_param('campaignId')) {
+                // Filter by CampaignId - Donors only can be filtered by campaignId if they donated to a campaign
+                $query->join(function (JoinQueryBuilder $builder) use ($campaignId) {
+                    $builder->innerJoin('give_donationmeta', 'donationmeta3')
+                        ->joinRaw("ON donationmeta3.meta_key = '" . DonationMetaKeys::CAMPAIGN_ID . "' AND donationmeta3.meta_value = {$campaignId} AND donationmeta3.donation_id = donationmeta1.donation_id");
+                });
+            }
+
+            if ($request->get_param('hideAnonymousDonors')) {
```
JasonTheAdams, line 69: See other comment on keeping parameters positive.

JasonTheAdams (changes requested):
Left some thoughts, @glaubersilva! Great work!

rickalday (approved):
Passed manual QA tests.

JasonTheAdams:
@glaubersilva Why did this go to QA with requested changes pending? 😅

glaubersilva:
@JasonTheAdams It didn't go to QA. I think @rickalday got confused because this PR is related to other tasks that were to QA

JasonTheAdams:
Hahah! Got it. 😆

JasonTheAdams:
> > @glaubersilva Why did this go to QA with requested changes pending? 😅
> 
> My bad. I posted on the wrong PR.

Shame

src/Donors/Controllers/DonorRequestController.php
glaubersilva: Resolved in ed34c899c01c02c7980802de61fab60d1ed1e907

src/Donations/Controllers/DonationRequestController.php
glaubersilva: Resolved in ed34c899c01c02c7980802de61fab60d1ed1e907

src/Donations/Controllers/DonationRequestController.php
glaubersilva: Resolved in ded974c38462d50bb6a55d9a2d2e9aa5a7a6aa8e

src/Donors/Controllers/DonorRequestController.php
glaubersilva: Resolved in ded974c38462d50bb6a55d9a2d2e9aa5a7a6aa8e

glaubersilva:
@JasonTheAdams I liked the idea of renaming `/give-api/v2/` to `/give/v3/` for new routes, aligning with the new approach we introduced in the Campaigns domain. This change would ensure consistency across all implementations using the new standards, which are designed to support entities. 

I believe it would be appropriate to apply this replacement everywhere, including the routes implemented in this PR as well as those for Campaigns. 

So, I think we can move forward with this change unless the other devs have concerns or objections about it, let's check with them just to make sure we are not missing something here.

JasonTheAdams:
Sounds great, @glaubersilva! I like the idea of retroactively applying this so long as the endpoints we're applying them to are:
1. _Truly_ RESTful
2. Not in production

src/Donations/Routes/RegisterDonationRoutes.php
```diff
@@ -0,0 +1,132 @@
...
+                        'enum' => ['live', 'test'],
+                    ],
+                    'campaignId' => [
+                        'type' => 'integer',
+                        'default' => 0,
+                    ],
+                    'includeAnonymousDonations' => [
+                        'type' => 'boolean',
+                        'default' => false,
+                    ],
```
JasonTheAdams, line 127: @glaubersilva Do we have any restrictions around this parameter? If our goal is to protect anonymous donations, then we may want to include some permission parameters around this. Presently, this is a fully public API, and these parameters are discoverable, so it's not hard for someone to flip the switch on this. There are a few contexts, here:
1. Donor wall
2. Admin list tables (donations list and campaign donations)
3. Campaign donations
4. 3rd Party usage

I'm guessing this will result in 3 forms of output:
1. Anonymous donations are included and donor info revealed (admin-side)
2. Anonymous donations are included but donor information is redacted (donor wall)
3. Anonymous donations are prohibited

Now, we may actually be fine with anyone in the world being able to query anonymous donations so long as the donor information is redacted (dropping option 3). But we'll still need a way to grab all information for the admin side with proper authorization.

Note: this is true of both the collection endpoint and single donation resource endpoint.

cc: @jonwaldstein

JasonTheAdams: Exactly. It should throw a 403 if the authorization is invalid for a specific parameter.

glaubersilva: We already are blocking access to sensitive data for non-admin users:

```
$sensitiveData = [
    'donorIp',
    'email',
    'phone',
    'billingAddress',
 ];
```

If I understand it well, we also should block access to the "anonymous data" when we are dealing with anonymous donations.

So, I refactored the `escDonation()` method and implemented a new logic there to also remove these properties:

```
if ($donation->anonymous) {
    $anonymousData = [
        'donorId',
        'honorific',
        'firstName',
        'lastName',
        'company',
    ];

    $sensitiveData = array_merge($sensitiveData, $anonymousData);
}
```

I considered here "anonymous data" any property from the donation model that can be used to identify a donor, which makes me wonder if we should remove the "comment" property as well. Currently, I kept it on the result, so would like to know your thoughts on it... Should we keep it or not? I'm asking because depending on the content of the comment is possible to identify the donor (I guess).

Also, I wrote more unit tests to check if the access to these "anonymous data" is being blocked by default and allowed for admin users, see: 7def493999265a45cb5d79574386fcdeadaf1bf9 and 238fd0a69a59f9bb86b6d7c936e69dcb1e4f8f99

src/Donors/Routes/RegisterDonorRoutes.php
```diff
@@ -0,0 +1,134 @@
...
+                    ],
+                    'onlyWithDonations' => [
+                        'type' => 'boolean',
+                        'default' => true,
+                    ],
+                    'campaignId' => [
+                        'type' => 'integer',
+                        'default' => 0,
+                    ],
+                    'includeAnonymousDonations' => [
```
glaubersilva, line 126: @jonwaldstein @JasonTheAdams Here in the Donors query, I think we should remove this parameter that allows include anonymous donations (this is applied only when the `onlyWithDonations` param is set) because if a donor made a donation as anonymous, retrieving the donor should not be possible.

src/Donations/Routes/RegisterDonationRoutes.php
glaubersilva: One more thing to consider here is that right now we are removing the "anonymous data" from the results. I was wondering if we should replace the content instead of removing it.

For example:

```
$newValue [
    'donorId' => 'anonymous',
    'honorific' => 'anonymous',
    'firstName' => 'anonymous',
    'lastName' => 'anonymous',
    'company' => 'anonymous',
];
```

jonwaldstein: Thanks @glaubersilva!

I think your'e on the right track.  I would say donation comment can be public but written by _anonymous_.  

Regarding preserving result keys, I'm leaning toward just removing results if not authenticated as I would prefer only receiving real values for my results rather than auto-generated results based on state if that make sense lol

JasonTheAdams: I'm thinking that it may be useful to turn `anonymousDonations` into an enum:
* `exclude` — (default) simply excludes the donations from the collection
* `redact` — includes the donations and redacts the sensitive information, replacing with "anonymous"
* `include` — includes the donations with full details (requires authorization)

For `include`, if authorization fails we simply 403 and call it a day, as the client is trying to do something beyond their pay grade.

And I agree, Jon, that the donation comment is public. It's presented to the donor as something public when they chose to write a comment.

What do you guys think?

src/Donors/Routes/RegisterDonorRoutes.php
glaubersilva: @jonwaldstein @JasonTheAdams  I was thinking better about it and I think we need to keep it... 

Let's use this sample to retrieve the top 5 donors of a campaign:

```
$request = new WP_REST_Request('GET' 'givewp/v3/donors');

$request->set_query_params(
    [
        'page' => 1,
        'per_page' => 5,
        'sort' => 'totalAmountDonated',
        'direction' => 'DESC',
        'campaignId' => $campaign->id,
    ]
);
```

If one of the top donors retrieved made only anonymous donations, we need to retrieve the information related to the `totalAmountDonated` but without any other data that can be used to identify the donor. So, I think we shouldn't remove this pram, but do something similar to the approach we are doing for donations related to redact data.

In that scenario, I think makes sense to replace the original data with something like this:

```
$newValues = [
    'donorId' => 'anonymous',
    'honorific' => 'anonymous',
    'firstName' => 'anonymous',
    'lastName' => 'anonymous',
    'company' => 'anonymous',
];
```

src/Donations/Routes/RegisterDonationRoutes.php
JasonTheAdams: So this is a RESTful design issue, @glaubersilva . You want to think of REST endpoints as similar to a pure function. That is, you get what you expect based on your parameters. Having the shape of the data change based on conditional things such as authorization means that the result is stateful — which is a RESTful no-no. The result should be predictable by the user.

The reason why the enum is a useful approach is that it gives the client explicit control over the shape of the response. The authorization, required for the `include` value, is explicit but requires authorization to work. So either it returns exactly what the client expects or does a 403 because they're not allowed to do that.

So it's not about having a use case where the admin may want the data redacted or not; it's about creating a system wherein the client can explicitly determine what they want so we don't have to worry about the scenario (i.e. state) they're working from. That said, if I'm logged into a site and viewing the donor wall, it should be redacted even if I'm logged in as an admin — otherwise, the inconsistent behavior becomes confusing. If the _client_ wants this behavior, then it can adjust the query parameter based on its own conditions.

Hope this helps!

src/Donors/Routes/RegisterDonorRoutes.php
```diff
@@ -0,0 +1,134 @@
...
+                    ],
+                    'mode' => [
+                        'type' => 'string',
+                        'default' => 'live',
+                        'enum' => ['live', 'test'],
+                    ],
+                    'onlyWithDonations' => [
+                        'type' => 'boolean',
+                        'default' => true,
+                    ],
```
JasonTheAdams, line 121: @glaubersilva You alluded to this parameter. This weirds me out. This would make sense if this was a `constituents` endpoint, but having a "I want donors that made donations" is weird. It just naturally raises the question: "How is a person a donor if they didn't donate?" 😆

src/Donors/Controllers/DonorRequestController.php
```diff
@@ -0,0 +1,173 @@
...
+        if ($request->get_param('onlyWithDonations')) {
+            $query->join(function (JoinQueryBuilder $builder) use ($mode) {
+                // The donationmeta1.donation_id should be used in other "donationmeta" joins to make sure we are retrieving data from the proper donation
+                $builder->innerJoin('give_donationmeta', 'donationmeta1')
+                    ->joinRaw("ON donationmeta1.meta_key = '" . DonationMetaKeys::DONOR_ID . "' AND donationmeta1.meta_value = ID");
+
+                // Include only current payment "mode"
+                $builder->innerJoin('give_donationmeta', 'donationmeta2')
+                    ->joinRaw("ON donationmeta2.meta_key = '" . DonationMetaKeys::MODE . "' AND donationmeta2.meta_value = '{$mode}' AND donationmeta2.donation_id = donationmeta1.donation_id");
+            });
```
JasonTheAdams, line 58: @glaubersilva @jonwaldstein 
Thinking ahead to database changes, I _strongly_ recommend that we update (or create) a `DonorQueryBuilder` that has declarative methods for things like this. So this craziness becomes:
```php
$query->whereDonorsHaveDonations();
```
It's cool for the REST APIs to use the query builder, but we want to reduce modifications that are directly coupled to the database structure (such as referencing meta like we are here).

src/Donors/Routes/RegisterDonorRoutes.php
JasonTheAdams: Is there a specific use-case for this filter?

jonwaldstein:
@glaubersilva there's one more update i'd like to request that I was discussing with @JasonTheAdams recently:

Let's move the actual registration of the rest endpoints into a shared top-level domain.  The reason is our REST API is really it's own domain that should live outside of the business logic of our other domains. It also makes it much easier to manage 😄 

It looks like we already have an `API`  domain so we could simply add a `REST/V3` folder inside and have the structure look something like this:

Screenshot 2025-02-18 at 5 20 35 PM

We do also have [an existing Service Provider ](https://github.com/impress-org/givewp/blob/develop/src/ServiceProviders/RestAPI.php)that registers our v2 REST endpoints.  I'm not opposed of doing something similar within our new `API\REST\V3` domain as it pretty clearly registers each route programatically.

src/Donations/Controllers/DonationRequestController.php
```diff
@@ -0,0 +1,189 @@
...
+{
+    /**
+     * @var Boolean
+     */
+    private $isAdmin;
+
+    /**
+     * @var Boolean
+     */
+    private $isAnonymousDonationsRedacted;
```
JasonTheAdams, line 25: I recommend using local variables and parameters rather than class properties, as these are really part of the handling of a single request flow, rather than a distinct attribute of the class itself.

src/Donations/Controllers/DonationRequestController.php
```diff
@@ -0,0 +1,189 @@
...
+        return new WP_REST_Response($this->escDonation($donation));
+    }
+
+    /**
+     * @unreleased
+     */
+    public function getDonations(WP_REST_Request $request): WP_REST_Response
+    {
+        $anonymousDonations = $request->get_param('anonymousDonations');
+        $this->isAnonymousDonationsRedacted = 'redact' === $anonymousDonations;
```
JasonTheAdams, line 49: Let's make an actual Enum class for this:
```php
$anonymousMode = new AnonymousDonationMode($request->get_param('anonymousDonations'));

$anonymousMode->isRedacted();
$anonymousMode->isExcluded();
$anonymousMode->isIncluded();
```

This will provide further validation and make things nice and declarative.

src/Donations/Controllers/DonationRequestController.php
JasonTheAdams, line 1: I won't die on this hill, but I'm not personally a fan of separating the request registration from the callbacks. First, it's not the convention. Second, it makes it harder to figure out the expected parameters. I think it feels necessary now because there's so much query building going on, but once we abstract out the query details it will greatly reduce the length of the callback code.

JasonTheAdams (changes requested):
Appreciate the continued work here, @glaubersilva! 🎉

src/Donations/Routes/RegisterDonationRoutes.php
glaubersilva: Resolved in 5bd2cc90cd8b99bf22fc531d11bdab4d3524841b 95acf68edd37cf3cee52d9aa6f58e988a13a7e22 06c849945a5d83d494ce824490fb727cf46e427a 8ff58df661b802dfb3f603dbf124e823b3be0624

src/Donations/Controllers/DonationRequestController.php
glaubersilva: Resolved in 06c849945a5d83d494ce824490fb727cf46e427a

src/Donations/Controllers/DonationRequestController.php
glaubersilva: Resolved in 06c849945a5d83d494ce824490fb727cf46e427a

src/Donors/Routes/RegisterDonorRoutes.php
glaubersilva: @JasonTheAdams In my local dev site, it happens all the time when I delete all donations using the list table bulk actions but don't do the same for donors - check the attached screenshots for reference. 

I believe that this scenario it's an edge case for real users, but it can happen.

So if we need for some reason to return all donors to run a specific action or routine regardless of it they have or not donations, we can set this param to `false`. 

This was my logic to implement it here. Thoughts? 😊

image

image

src/Donations/Routes/RegisterDonationRoutes.php
```diff
@@ -0,0 +1,162 @@
...
+                        ],
+                    ],
+                    'anonymousDonations' => [
+                        'type' => 'string',
+                        'default' => 'exclude',
+                        'enum' => [
+                            'exclude',
+                            'include',
+                            'redact',
+                        ],
```
JasonTheAdams, line 75: @glaubersilva Can you help me understand the two modes, here? What's each one for? Why two?

glaubersilva: @JasonTheAdams The `sensitiveData` is applied to all donations and the `anonymousDonations` is only for anonymous donations. 

Previously, I was checking inside the `escDonation()` method if the current user was logged as admin, and if not I was removing the `sensitiveData` from the results. 

So to keep the consistency and the concept that we applied for the `mode` and `anonymousDonations` parameters, I moved this state control logic to the new `sensitiveData` param.

Therefore, the reason to have it separated into two parameters is that the admins should be able to retrieve it if they need it but these values should never be visible for non-admin users:

```
$sensitiveData = [
    'donorIp',
    'email',
    'phone',
    'billingAddress',
]; 
```

src/Donors/Controllers/DonorRequestController.php
glaubersilva: @JasonTheAdams If you don't mind, I would like to refactor this in a subsequent PR because these endpoints will not be immediately used in the codebase, but I'm afraid to spend too much time on it and delay the merge into the epic even more - as we are changing the namespace of the campaign routes, we need to merge it before the public release. 

Also, I would like to let the folder structure changes mentioned by @jonwaldstein  [**here**](https://github.com/impress-org/givewp/pull/7699#issuecomment-2667062607) be included in this subsequent PR as well.

JasonTheAdams: So long as the API itself is finalized, I'm cool with moving under-the-hood changes to a subsequent PR. BUT! You must agree to actually **plan** to do it. None of this, "oh yeah, we'll do that later for sure" stuff. 😁

src/Donations/Routes/RegisterDonationRoutes.php
JasonTheAdams: I see! Thanks for explaining, @glaubersilva! So do we foresee a mode wherein folks would want to distinguish anonymous data from other sensitive data? Could it all just live under a single "sensitiveData" parameter?

glaubersilva: Since we don't need to redact the sensitive data (we need only include or exclude it) I simplify it to use a boolean value instead of an enum. See: 183ec8c273f462332d1b179030fe56219e3deb2a

glaubersilva: @JasonTheAdams I think about it a LOT because my initial idea was to put it together but I concluded that it's not possible (or maybe I'm missing something here) because the `anonymousDonations` param is designed to `include` (or included with `redact` info) or `exclude` entries from the query. But the `includeSensitiveData` doesn't have the power to change the query, it only changes (include or remove) the properties from donations or donors that will be visible in the results.

So, as admin, I can:

- **Exclude** anonymous donations/donors **with** sensitive data available
- **Exclude** anonymous donations/donors **without** sensitive data available

- **Include** anonymous donations/donors **with** sensitive data available
- **Include** anonymous donations/donors **without** sensitive data available

- **Redact** anonymous donations/donors **with** sensitive data available
- **Redact** anonymous donations/donors **without** sensitive data available

The difference here between the _sensitive data_ and the _anonymous data_ is that the first one is too personal and can be used to **contact/localize** the donors so only admins can have access to it and the second one is more light info that the donor can choose display or not (Redact) - this last one change how the query is build.

src/Donors/Controllers/DonorRequestController.php
glaubersilva: Sure! We can do it later. 

TrueTrollGIF

Ok, I'm kidding! 😅 

My plan is to start a new PR as soon the Campaigns tasks (related to the public release) pile gets down, so probably on the next fun Friday I'll already open a PR for it.

glaubersilva:
@JasonTheAdams @jonwaldstein I think I implemented the most important things we discussed above and other items will be implemented in a subsequent PR as discussed [**here**](https://github.com/impress-org/givewp/pull/7699#discussion_r1987405818), so I updated the PR description and now this is ready for re-review.

src/Donors/Routes/RegisterDonorRoutes.php
jonwaldstein: That is a valid reason I have not thought about as a donor's data could technically be deleted.  It's an edge case for sure but possible lol

jonwaldstein (approved):
This is looking good to me!  This is exciting to have first class REST endpoints and I know there's still a wish list of items we want to include / refactor but I think this is a really good start.  Let's get this into the campaigns epic so we can actually test it out and plan on improving it as we use it internally 😄

src/Donors/Routes/RegisterDonorRoutes.php
```diff
@@ -0,0 +1,156 @@
...
+        register_rest_route(
+            DonorRoute::NAMESPACE,
+            DonorRoute::DONORS,
+            [
+                [
+                    'methods' => WP_REST_Server::READABLE,
+                    'callback' => function (WP_REST_Request $request) {
+                        return $this->donorRequestController->getDonors($request);
+                    },
+                    'permission_callback' => '__return_true',
```
JasonTheAdams, line 94: We need to update the permission callbacks to check for the sensitive data parameters and authentication. If a client tries to get more data than they're authenticated to have, then we return false and the REST API will 403.

src/Donations/Routes/RegisterDonationRoutes.php
```diff
@@ -0,0 +1,154 @@
...
+        register_rest_route(
+            DonationRoute::NAMESPACE,
+            DonationRoute::DONATION,
+            [
+                [
+                    'methods' => WP_REST_Server::READABLE,
+                    'callback' => function (WP_REST_Request $request) {
+                        return $this->donationRequestController->getDonation($request);
+                    },
+                    'permission_callback' => '__return_true',
```
JasonTheAdams, line 53: We need to update this one as well.

JasonTheAdams (changes requested):

src/Donations/Controllers/DonationRequestController.php
```diff
@@ -0,0 +1,210 @@
...
+            );
+        }
+
+        $donationAnonymousMode = new DonationAnonymousMode($request->get_param('anonymousDonations'));
+        if ( ! $isAdmin && $donation->anonymous && ! $donationAnonymousMode->isRedacted()) {
+            return new WP_REST_Response(
+                ['message' => __('You do not have permission to include anonymous donations.', 'give')],
+                403
+            );
+        }
```
JasonTheAdams, line 46: @glaubersilva This shouldn't be happening here, to tie this together. The request handler should be able to assume that all authentication and such has happened. It's concern is purely in handling and returning the request appropriately.

src/Donations/Controllers/DonationRequestController.php
```diff
@@ -0,0 +1,210 @@
...
+
+        // Include only valid statuses
+        $query->whereIn('post_status', ['publish', 'give_subscription']);
+
+        $query
+            ->limit($perPage)
+            ->offset(($page - 1) * $perPage)
+            ->orderBy($sortColumn, $sortDirection);
+
+        $donations = $query->getAll() ?? [];
```
jonwaldstein, line 103: One thing I just noticed, is we are directly returning the query results.  The Model query builder returns an array of models so not all the properties will be readable.  So we actually will need to map over the results and call `$donation->toArray()`.

jonwaldstein: oh wait is that what escDonation is doing?  maybe there's just a few properties we need to adjust for.

glaubersilva: @jonwaldstein Exactly! the `escDonation()` method is doing that.

src/Donors/Routes/RegisterDonorRoutes.php
glaubersilva: Resolved in 0bc52d67abe9047987a915fc436ddab1cfbc91c4

src/Donations/Routes/RegisterDonationRoutes.php
glaubersilva: Resolved in 0bc52d67abe9047987a915fc436ddab1cfbc91c4

src/Donations/Controllers/DonationRequestController.php
glaubersilva: Resolved in 0bc52d67abe9047987a915fc436ddab1cfbc91c4

glaubersilva:
@JasonTheAdams This is ready for re-review.

src/Donations/Routes/RegisterDonationRoutes.php
```diff
@@ -0,0 +1,200 @@
...
+            ]
+        );
+    }
+
+    /**
+     * @unreleased
+     */
+    public function permissionsCheck(WP_REST_Request $request)
+    {
+        $isAdmin = current_user_can('manage_options');
```
JasonTheAdams, line 166: @glaubersilva @jonwaldstein I'm not sure if this can be resolved well now, given our roles and capabilities are a bit garbled, but this is very likely too high of a bar. There are GiveWP roles that are not admins that can view the donor and donation lists.

JasonTheAdams (approved):
I left one comment on how the strictness of the capability check, but I'll let be optionally fixed. Great work, @glaubersilva! Way to stick it out on this one! 🎉

glaubersilva:
@JasonTheAdams Thanks, man! 😊 I'll consider your last comment also on my next PR after discussing with the team how should be our patter to define if a user is a _"GiveWP admin"_ or not and maybe we even can create a helper function to be re-used in other places.

src/Donations/Routes/RegisterDonationRoutes.php
jonwaldstein: Yeah for sure.  GET request to Donations will probably be `current_user_can('view_give_payments')`.  The potential exception would be a a role of `give_donor` who is requesting a donation that belongs to them.

PR #7700, opened 2025-01-30
Feature: Add Campaign Donors block

pauloiankoski:
Resolves [GIVE-1392]

## Description
This pull request adds the Campaign Donors block, with options to show the top donors (by amount) or the recent donors (by date).

## Visuals
CleanShot 2025-01-30 at 17 46 15
__Block Settings__

CleanShot 2025-01-30 at 17 54 10
__Recent donors list__

CleanShot 2025-01-30 at 17 53 35
__Top donors list__

## Pre-review Checklist

-   [ ] Includes unit tests

… scripts

[GIVE-1392]: https://stellarwp.atlassian.net/browse/GIVE-1392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Campaigns/Blocks/CampaignDonorsBlock/render.php
```diff
@@ -0,0 +1,261 @@
...
+        }
+
+        /**
+         * @unreleased
+         */
+        private function formatDonorsData(array $donors): array
+        {
+            return array_map(static function ($entry) {
+                if (isset($entry->date)) {
+                    $entry->date = Carbon::parse($entry->date)->diffForHumans();
```
jonwaldstein, line 116: @pauloiankoski we have to be careful about introducing php libraries because they can conflict with other plugins.  this is the reason we Struass.  However, do we even need this one? we have functions in WP like [human_time_diff](https://developer.wordpress.org/reference/functions/human_time_diff/) and our own [Temporal](https://github.com/impress-org/givewp/blob/3fe9f4dd7f82cefe6c41c37df9e00713dba10c28/src/Framework/Support/Facades/DateTime/TemporalFacade.php#L15) class

src/Campaigns/Blocks/CampaignDonorsBlock/render.php
```diff
@@ -0,0 +1,261 @@
...
+        {
+            $sortBy = $this->attributes['sortBy'] ?? 'top-donors';
+            $blockTitle = $sortBy === 'top-donors' ? __('Top Donors', 'give') : __(
+                'Recent Donors',
+                'give'
+            );
+            $donateButtonText = $this->attributes['donateButtonText'] ?? __('Join the list', 'give');
+
+            $blockInlineStyles = sprintf(
+                '--givewp-campaign-donors-block__primary-color: %s; --givewp-campaign-donors-block__secondary-color: %s;',
```
jonwaldstein, line 137: Can we use `--givewp-primary-color` and `--givewp-secondary-color` here?

wordpress-scripts-webpack.config.js
```diff
@@ -65,6 +66,18 @@ module.exports = {
         adminBlocks: path.resolve(process.cwd(), 'blocks', 'load.js'),
         campaignBlocks: srcPath('Campaigns/Blocks/blocks.ts'),
     },
+    plugins: [
+        ...defaultConfig.plugins,
+        new CopyWebpackPlugin({
```
jonwaldstein, line 71: What's happening here?

src/Campaigns/Blocks/CampaignDonorsBlock/render.php
```diff
@@ -0,0 +1,261 @@
...
+use Give\Campaigns\Repositories\CampaignRepository;
+use Give\DonationForms\Blocks\DonationFormBlock\Controllers\BlockRenderController;
+use Give\Donations\ValueObjects\DonationMetaKeys;
+use Give\Framework\Support\ValueObjects\Money;
+
+/**
+ * @unreleased
+ */
+if ( ! class_exists(CampaignDonorsBlockRenderer::class)) {
+    class CampaignDonorsBlockRenderer
```
jonwaldstein, line 17: Shouldn't this be in it's own namespace file so you don't have to check if class exists?

src/Campaigns/Blocks/CampaignDonorsBlock/render.php
```diff
@@ -0,0 +1,268 @@
...
+                $entry->amount = Money::fromDecimal($entry->amount, give_get_currency());
+
+                return $entry;
+            }, $donors);
+        }
+
+        /**
+         * @unreleased
+         */
+        private function renderBlockHtml($donors): void
```
jonwaldstein, line 127: This is interesting that we are rendering the entire html in php, I was imagining the front-end to be a React instance.  I realize that's a pivotal shift at this point but something we should discuss for these blocks in the future.

src/Campaigns/Blocks/CampaignDonorsBlock/render.php
pauloiankoski: Thanks for sharing that function. I do agree with you, btw.

Refactored on 60aa32e0f8d54b9d0b3e375bd22afe52de934953

src/Campaigns/Blocks/CampaignDonorsBlock/render.php
pauloiankoski: Renamed on 4b7c5f1ad3454f5c8b1a97a1c0f9a27b45e74c05

wordpress-scripts-webpack.config.js
pauloiankoski: Copying every `style.css` and `script.js` files found in the blocks folders to their respective build folder.

src/Campaigns/Blocks/CampaignDonorsBlock/render.php
pauloiankoski: I had to add that check after receiving a fatal error on the editor due to the class being registered multiple times.

src/Campaigns/Blocks/CampaignDonorsBlock/render.php
pauloiankoski: We are still beginning with the blocks. Maybe that discussion is better to happen sooner than later.

src/Campaigns/Blocks/CampaignDonorsBlock/edit.tsx
```diff
@@ -0,0 +1,112 @@
...
+        donateButtonText,
+        sortBy,
+        donorsPerPage,
+        loadMoreButtonText,
+    } = attributes;
+
+    return (
+        <div {...blockProps}>
+            <CampaignSelector attributes={attributes} setAttributes={setAttributes}>
+                <ServerSideRender block="givewp/campaign-donors-block" attributes={attributes} />
```
alaca, line 46: 👍

src/Campaigns/Blocks/CampaignDonorsBlock/edit.tsx
```diff
@@ -0,0 +1,112 @@
...
+            </CampaignSelector>
+
+            {hasResolved && campaign && (
```
alaca, line 49: I'm not sure that the returned result would be null for the campaign. It might be better to do something like campaign?.id.

src/Campaigns/Blocks/CampaignDonorsBlock/edit.tsx
```diff
@@ -0,0 +1,112 @@
...
+                            label={__('Sort by', 'give')}
+                            value={sortBy}
+                            options={[
+                                {label: __('Top Donors', 'give'), value: 'top-donors'},
+                                {label: __('Recent Donors', 'give'), value: 'recent-donors'},
+                            ]}
+                            onChange={(value) => setAttributes({sortBy: value})}
+                            help={__('The order donors are displayed in.', 'give')}
+                        />
+                        {/* TODO: Revert the label and help text back to what are in the designs once the backend for pagination is ready */}
```
alaca, line 91: Have you resolved this comment?

src/Campaigns/Blocks/CampaignDonorsBlock/edit.tsx
```diff
@@ -0,0 +1,112 @@
...
+                        <NumberControl
+                            label={__('Limit', 'give')}
+                            value={donorsPerPage}
+                            min={1}
+                            max={100}
+                            onChange={(value) => setAttributes({donorsPerPage: parseInt(value)})}
+                            help={__('The maximum number of donors to display.', 'give')}
+                        />
+                        {/* TODO: Revert the field back once the backend for pagination is ready
```
alaca, line 100: Same here

alaca (approved):
Looks good to me. I have a few questions, but overall, good job man 👍

src/Campaigns/Blocks/CampaignDonorsBlock/edit.tsx
pauloiankoski: These comments are part of the changes I made while temporarily disabling pagination (load more button) on this block. It's a comment for the future.

src/Campaigns/Blocks/CampaignDonorsBlock/edit.tsx
pauloiankoski: These comments are part of the changes I made while temporarily disabling pagination (load more button) on this block. It's a comment for the future.

src/Campaigns/Blocks/CampaignDonorsBlock/edit.tsx
pauloiankoski: Modded on da79d91796308727f67d0a5cb0e347092378f467

src/Campaigns/Blocks/CampaignDonorsBlock/render.php
pauloiankoski: Refactored on 9392c1085bb153535a92e0fae8d0c3ce9b478e47

src/Campaigns/Blocks/CampaignDonors/index.ts
```diff
@@ -0,0 +1,14 @@
...
+
+const {name} = metadata;
+
+export {metadata, name};
+export const settings = {
+    icon,
+    edit: Edit,
+};
+
+export const init = () => initBlock({name, metadata, settings});
```
alaca, line 14: I'm not sure we need this init-block functionality. We can just include the block root file inside blocks.ts.  Like I did it here https://github.com/impress-org/givewp/blob/674531f04a99386c5af072e6a9b17f1698a2029b/src/Campaigns/Blocks/CampaignCover/index.tsx

src/Campaigns/Blocks/CampaignDonors/resources/views/render.php
alaca, line 1: Campaign donors block will eventually have pagination, right? I'm not sure if plain HTML is the direction we want here. This should probably be a React app.

src/Campaigns/Blocks/blocks.ts
alaca, line 1: Why don't we just include the root block file here and remove the additional layer of complexity?

wordpress-scripts-webpack.config.js
```diff
@@ -1,6 +1,7 @@
 /**
  * External Dependencies
  */
+const fs = require('fs');
```
alaca, line 4: We don't need this anymore

alaca (changes requested):

src/Campaigns/Blocks/CampaignDonors/index.ts
pauloiankoski: That's the approach Gutenberg uses to register their blocks. What I like on it is that we do not need to think too much on all the pieces necessary to register the block.

src/Campaigns/Blocks/CampaignDonors/resources/views/render.php
pauloiankoski: Too late to change that approach. This PR is already ready to be merged and the changes are supposed to be only regarding frontend scripts and styles enqueueing. We'll introduce the React app in a later PR.

src/Campaigns/Blocks/CampaignDonors/index.ts
alaca: So it is easier to add complexity to handle block registration (which we already have - registerBlockType), define an object using this function, and then import it somewhere else, and also add it to another list that handle the registration, than just importing the file? 
What is the benefit of having this?

src/Campaigns/Blocks/CampaignDonors/resources/views/render.php
alaca: Make sense.

src/Campaigns/Blocks/CampaignDonors/index.ts
alaca: We can move forward here and address this another time. I just don't like this approach of abstracting the core functionality without any real benefit. It adds unnecessary complexity so I need to figure out how to register a block, and I already know how to do that.

pauloiankoski: Refactored on 0ac8456179c6abc8786eb6e0c7fc0b47ccc23d10

src/Campaigns/Blocks/blocks.ts
pauloiankoski: Refactored on 0ac8456179c6abc8786eb6e0c7fc0b47ccc23d10

wordpress-scripts-webpack.config.js
pauloiankoski: Fixed on 21570442e5270389a139362a565d60dbee5c1149

src/Campaigns/Blocks/CampaignDonors/index.ts
alaca: Hmm, I think I wasn't clear enough. We don't need getBlock().forEach functionality. What I'm proposing here is that each block registers the block. Example https://github.com/impress-org/givewp/blob/674531f04a99386c5af072e6a9b17f1698a2029b/src/Campaigns/Blocks/CampaignCover/index.tsx

alaca: Okay, lets move this bad boy to QA

alaca (approved):
Ready for QA!!!

PR #7710, opened 2025-02-05
Feature: add locale support for visual form builder

glaubersilva:
Resolves [GIVE-2124]

## Description

This PR adds support for v3 forms to work with different locales through the use of multilingual plugins. As a bonus, it also allows the creation of new forms respecting the language selected in the WP admin bar through the selectors provided by the WPML and Polylang plugins.

**How it's working?**

Before, in the `render()` method of the `BlockRenderController()` class we are returning a `div` with some data attributes like `data-form-url` that are used in the frontend. Now we also are sending a new `data-form-locale` attribute to be used in the frontend for subsequent requests (`donation-form-view`, `donation-form-view-preview`, and `donation-confirmation-receipt-view`).

**Why it works?** 

Because when the `render()` method of the `BlockRenderController()` class is triggered it has the right context of the selected locale, but the subsequent requests made to the server by the frontend (`donation-form-view`, `donation-form-view-preview`, and `donation-confirmation-receipt-view`) didn't have it before but now they have because the new locale attribute is passed alongside to the subsequent requests and retrieved back in the `Give\Framework\Routes\Router` class and then used in the views to make sure we are rendering them with the proper locale.

It was inspired by this feature from the Polylang Pro plugin: https://polylang.pro/doc/rest-api/

## Affects

The v3 forms.

## Visuals

**V3 forms - multilingual plugins support**

https://www.loom.com/share/99d359f094f4448194965d506ea937ef?sid=bddc70ff-5dee-48fd-a4a2-9b7345c1fbe7

**[Bonus Feature] Creating new V3 forms respecting the language selected on the WPML language selector or Polylang Language selector**

https://www.loom.com/share/bbd4328a2b2b409784dcce6d3ae3871d?sid=d9e31524-1437-4737-8e46-cfbf5a000ea4

## Testing Instructions

**V3 forms - multilingual plugins support**

1. Install WPML or Polylang plugin;
2. Setup the plugin installed Previously to use two languages;
3. Create two v3 forms, one for each language setup previously; 
4. Use Loco Translate to translate strings that can't be translated using the Visual Form Builder;
5. Create a new page for the Language 1 and insert the proper form shortcode on it;
6. Create a new page for the Language 2 and insert the proper form shortcode on it;
7. Access the page of the Language 1 and make sure everything is being displayed as expected;
8. Use the language selector provided by the plugin (on the frontend) to switch to the Language 2 page and make sure everything is being displayed as expected;

**[Bonus Feature] Creating new V3 forms respecting the language selected on the WPML language selector or Polylang Language selector**

1. Install WPML or Polylang plugin;
2. Setup the plugin installed Previously to use two languages;
3. Use Loco Translate to translate strings that can't be translated using the Visual Form Builder;
4. Go to the `GiveWP > All Forms` page;
5. Select the wished language in the language selector localized in the WP admin bar;
6. Click on "Add Form" and make sure the new form was created respecting the language selected previously.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2124]: https://stellarwp.atlassian.net/browse/GIVE-2124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DonationForms/ViewModels/DonationFormViewModel.php
+++ b/src/DonationForms/ViewModels/DonationFormViewModel.php
@@ -252,12 +252,19 @@ class DonationFormViewModel
      * 5. Finally, call the specific WP function wp_print_footer_scripts()
      *  - This will only print the footer scripts that are enqueued within our route.
      *
+     * @unreleased Add locale support
      * @unreleased Adds class for form design
      * @since 3.11.0 Sanitize customCSS property
      * @since 3.0.0
      */
-    public function render(): string
+    public function render(string $locale = ''): string
     {
+        $currentLocale = get_locale();
+
+        if ($locale != $currentLocale) {
+            switch_to_locale($locale);
+        }
+
         $this->enqueueGlobalStyles();
 
         $this->enqueueFormScripts(
--- a/src/DonationForms/ViewModels/DonationFormViewModel.php
+++ b/src/DonationForms/ViewModels/DonationFormViewModel.php
@@ -252,8 +252,8 @@ class DonationFormViewModel
      * 5. Finally, call the specific WP function wp_print_footer_scripts()
      *  - This will only print the footer scripts that are enqueued within our route.
      *
-     * @unreleased Add locale support
-     * @unreleased Adds class for form design
+     * @unreleased Add locale support     
+     * @since 3.20.0 Adds class for form design
      * @since 3.11.0 Sanitize customCSS property
      * @since 3.0.0
      */
--- a/src/DonationForms/ViewModels/DonationFormViewModel.php
+++ b/src/DonationForms/ViewModels/DonationFormViewModel.php
@@ -252,18 +252,14 @@ class DonationFormViewModel
      * 5. Finally, call the specific WP function wp_print_footer_scripts()
      *  - This will only print the footer scripts that are enqueued within our route.
      *
-     * @unreleased Add locale support     
+     * @unreleased Add locale support
      * @since 3.20.0 Adds class for form design
      * @since 3.11.0 Sanitize customCSS property
      * @since 3.0.0
      */
     public function render(string $locale = ''): string
     {
-        $currentLocale = get_locale();
-
-        if ($locale != $currentLocale) {
-            switch_to_locale($locale);
-        }
+        Language::switchToLocale($locale);
 
         $this->enqueueGlobalStyles();
 
--- a/src/DonationForms/ViewModels/DonationFormViewModel.php
+++ b/src/DonationForms/ViewModels/DonationFormViewModel.php
@@ -252,15 +252,12 @@ class DonationFormViewModel
      * 5. Finally, call the specific WP function wp_print_footer_scripts()
      *  - This will only print the footer scripts that are enqueued within our route.
      *
-     * @unreleased Add locale support
      * @since 3.20.0 Adds class for form design
      * @since 3.11.0 Sanitize customCSS property
      * @since 3.0.0
      */
-    public function render(string $locale = ''): string
+    public function render(): string
     {
-        Language::switchToLocale($locale);
-
         $this->enqueueGlobalStyles();
 
         $this->enqueueFormScripts(
--- a/src/FormBuilder/Routes/CreateFormRoute.php
+++ b/src/FormBuilder/Routes/CreateFormRoute.php
@@ -9,6 +9,7 @@ use Give\DonationForms\ValueObjects\DonationFormStatus;
 use Give\FormBuilder\Actions\GenerateDefaultDonationFormBlockCollection;
 use Give\FormBuilder\FormBuilderRouteBuilder;
 use Give\Helpers\Hooks;
+use Give\Helpers\Language;
 
 /**
  * Route to create a new form
@@ -16,6 +17,7 @@ use Give\Helpers\Hooks;
 class CreateFormRoute
 {
     /**
+     * @unreleased Add locale support
      * @since 3.1.0 updated default form blocks to be generated from block models instead of json
      * @since 3.0.0
      *
@@ -25,12 +27,18 @@ class CreateFormRoute
     public function __invoke()
     {
         if (isset($_GET['page']) && FormBuilderRouteBuilder::SLUG === $_GET['page']) {
+            $locale = $_GET['locale'] ?? '';
+
             // Little hack for alpha users to make sure the form builder is loaded.
             if (!isset($_GET['donationFormID'])) {
-                wp_redirect(FormBuilderRouteBuilder::makeCreateFormRoute());
+                wp_redirect(FormBuilderRouteBuilder::makeCreateFormRoute($locale)->getUrl());
                 exit();
             }
             if ('new' === $_GET['donationFormID']) {
+                // Make sure the Form will be created using the proper locale
+                $locale = $_GET['locale'] ?? '';
+                Language::switchToLocale($locale);
+
                 $form = new DonationForm([
                     'title' => __('GiveWP Donation Form', 'give'),
                     'status' => DonationFormStatus::DRAFT(),
@@ -45,7 +53,7 @@ class CreateFormRoute
 
                 $form->save();
 
-                wp_redirect(FormBuilderRouteBuilder::makeEditFormRoute($form->id));
+                wp_redirect(FormBuilderRouteBuilder::makeEditFormRoute($form->id, $locale)->getUrl());
                 exit();
             }
         }
--- a/src/Helpers/Language.php
+++ b/src/Helpers/Language.php
@@ -76,6 +76,13 @@ class Language
             $locale = $wpmlLocale != $locale ? $wpmlLocale : $locale;
         }
 
+        // When in the admin area and Polylang is installed, retrieve the language selected in the Polylang language selector of the WordPress admin bar
+        if (function_exists('pll_current_language') && is_admin()) {
+            $pllCurrentLangCode = pll_current_language('slug');
+            $pllCurrentLang = PLL()->model->get_language($pllCurrentLangCode);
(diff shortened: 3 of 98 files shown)


src/Helpers/Language.php
```diff
@@ -52,4 +55,46 @@ public static function getRelativePath(): string
...
+
+        // When in the admin area and Polylang is installed, retrieve the language selected in the Polylang language selector of the WordPress admin bar
+        if (is_admin() && function_exists('pll_current_language') && function_exists('PLL')) {
+            $pllCurrentLangCode = pll_current_language('slug');
+            $pllCurrentLang = PLL()->model->get_language($pllCurrentLangCode);
+            $locale = $pllCurrentLang && $pllCurrentLang->locale != $locale ? $pllCurrentLang->locale : $locale;
+        }
+
+        return $locale;
+    }
```
JasonTheAdams, line 87: @glaubersilva As a recommendation, rather than baking 3rd party dependencies right into our code, I'd recommend making use of a filter, here. We can then have a `ThirdPartySupport` or some sort of domain with `WPML` and `Polylang` sub-domains. These then hook into this and adjust the locale.

This would keep our code distinct from 3rd party support — even if we're the ones doing it. The upside is that an additional 3rd party can also find this and use the same hook themselves. 😄

jonwaldstein: Agreed! 

@glaubersilva and I did talk briefly about this, having a place that any language service can hook in would be ideal.

Love that this already includes support for the additional Polylang 🥇

src/DonationForms/ViewModels/DonationFormViewModel.php
```diff
@@ -252,12 +252,15 @@ public function exports(): array
...
      *
+     * @unreleased Add locale support
      * @since 3.20.0 Adds class for form design
      * @since 3.11.0 Sanitize customCSS property
      * @since 3.0.0
      */
-    public function render(): string
+    public function render(string $locale = ''): string
     {
+        Language::switchToLocale($locale);
```
kjohnson, line 262: I'm wondering if we can move the locale switching higher up the chain of the request to avoid passing `$locale` as an argument. We aren't using `$locale` directly in the View Model, which is a possible red flag.

kjohnson: For example, we register the route `Route::post('donation-form-view')`. The request is parsed into `DonationFormRouteData` (which now has the locale). This DTO is passed the `DonationFormViewController`, specifically the `show()` method, which instantiates the `DonationFormViewModel` and passes the locale from the DTO to the View Model's render method. The new `$locale` argument is only used to set the locale on the session.

| Request Lifecycle | |
| --- | --- |
| `POST /donation-form-view-preview` | `?locale=en_US` |
| DonationFormRouteData | `$self->locale = $request['locale']` |
| DonationFormViewController | `$viewModel->render($data->locale)` |
| DonationFormViewModel | `Language::switchToLocale($locale)` |

---

Instead, could we switch the locale in the `Route::post('donation-form-view-preview')` callback? At this point, we've identified that this is a Give request.

```diff
Route::get('donation-form-view', static function (array $request) {
+   Language::switchToLocale($_REQUEST['locale']);
    return give(DonationFormViewController::class)->show(
        DonationFormViewRouteData::fromRequest($request)
    );
});
```

src/Helpers/Language.php
kjohnson: Re: Hooks

The [get_locale()](https://developer.wordpress.org/reference/functions/get_locale/) function already implements the [locale](https://developer.wordpress.org/reference/hooks/locale/) filter hook.

So we could re-use that filter, if these plugins are not already using it.

kjohnson (changes requested):
@glaubersilva I think we can simplify this by moving the locale switching higher up the request chain in the places where we are passing the locale from Request to Controller to View Model.

src/Helpers/Language.php
glaubersilva: Resolve in 88698fab20d90979a7cba57e036a95c09e3d1ed1

@kjohnson, I think it makes sense to have our hook because we can have more control over it.

Also, I added the `WPML::getLocale()` and `Polylang::getLocale()` methods that use the `get_locale()` function as a fallback. So, if we choose to use the default `locale` filter would be necessary to add a new parameter to these methods to disable the fallback so we could use them inside the default `locale` filter without running into an infinite loop.

src/DonationForms/ViewModels/DonationFormViewModel.php
glaubersilva: @kjohnson Nice catch! This is one of the things that we do when our brain is already overloaded and we keep coding. 
😅 Thanks for noticing that! Resolved in f3381561796c0111cc80fcadde0de428a91b6c2e

src/FormBuilder/Routes/CreateFormRoute.php
```diff
@@ -25,12 +27,18 @@ class CreateFormRoute
...
             // Little hack for alpha users to make sure the form builder is loaded.
             if (!isset($_GET['donationFormID'])) {
-                wp_redirect(FormBuilderRouteBuilder::makeCreateFormRoute());
+                wp_redirect(FormBuilderRouteBuilder::makeCreateFormRoute($locale)->getUrl());
                 exit();
             }
             if ('new' === $_GET['donationFormID']) {
+                // Make sure the Form will be created using the proper locale
+                $locale = $_GET['locale'] ?? '';
+                Language::switchToLocale($locale);
```
kjohnson, line 40: @glaubersilva is the form actually being created for a specific locale? Or is this a convenience for translating values for form settings?

I'm wondering if we are coupling locale too closely to the forms.

src/Helpers/Language.php
kjohnson: Are these plugins not already hooking into the `locale` filter?

src/DonationForms/ViewModels/DonationFormViewModel.php
kjohnson: I definitly understand. Review is a different perspective than development. These things often are revealed by the code review process.

kjohnson: Looking at https://github.com/impress-org/givewp/commit/f3381561796c0111cc80fcadde0de428a91b6c2e, does the locale need to be a part of the DTO? I understand that the locale is appended to the URL, but should that belong to the DTO (business logic) or is it an additional parameter on the request?

In theory, supporting multilingual plugins should change as little of our business logic as possible. We aren't integrating multilingual forms as a core feature. This should be more of a compatibility layer, right?

src/Helpers/Language.php
glaubersilva: @kjohnson Yes, they are hooking it in different scenarios, but not in the one we need that is too specific to the behavior we are implementing on the GiveWP side:

```
/**
 * When in the admin area and WPML or Polylang is installed, retrieve the language 
 * selected in the language selector of the WordPress admin bar added by them
 */
add_filter('givewp_locale', function ($locale) {
    if ( ! is_admin()) {
        return $locale;
    }

    $wpmlLocale = WPML::getLocale();
    if ($wpmlLocale != $locale) {
        return $wpmlLocale;
    }

    $polylangLocale = Polylang::getLocale();
    if ($polylangLocale != $locale) {
        return $polylangLocale;
    }

    return $locale;
});
```

This language selector in the WP admin bar is used to filter the posts created using the selected language. 

As we are using a custom list view instead of the default ones added by CPT, we can't filter the Forms created with specific locales because our list table always displays all forms regardless of the locale.

However, if you switch to the legacy view, you'll see that the filter by language is working because these plugins implement this behavior in the default lit table for CPTs - check the attached screenshots for reference.

So what we are doing here is taking advantage of this language selector from the WP admin bar to provide a convenient way to customers create new forms in the selected language because before they needed to translate everything since the forms always were created in the default language.

This way, using our custom `givewp_locale` filter instead of the default locale is safer because we can ensure that we'll apply this behavior only in the Give context since other plugins can use the default `locale` filter in different contexts that can land to unexpected outcomes.

image

image

src/FormBuilder/Routes/CreateFormRoute.php
glaubersilva: @kjohnson This Is just a convenience  for translating values as described here: https://github.com/impress-org/givewp/pull/7710#discussion_r1952884538

However, if you use the "Add Form" button from the legacy list table, the post will be created in the selected language because it's handled by these multilingual plugins. So, in the legacy list view, also is possible to filter the posts (only the ones created through the legacy view) by language using this language selector from the WP admin bar.

src/DonationForms/ViewModels/DonationFormViewModel.php
glaubersilva: Resolved in 4c5d3a576ac27fa64df473256103ec60f396eaae

kjohnson:
Looks like this broke a test:

`Give\Tests\Unit\DonationForms\Endpoints\TestListDonationForms::testShouldReturnListWithSameData`
```diff
+ 'edit' => '?locale=en_US'
- 'edit' => null
```

The edit link no longer matches. Was ir previously null?

src/FormBuilder/Routes/CreateFormRoute.php
kjohnson: In part my concern is an icrease in scope. The goal here was to add suppoprt for locale in v3 views and routes so that the forms are translatable. Modifying how a form is created was not accounted for in refinment and prioritization.

glaubersilva:
@kjohnson The broken test was fixed here: 920f1460e1295a331d9ed811548383bac68df0ce

This is ready for re-review.

kjohnson (approved):

rickalday (approved):
Passed manual QA tests.

PR #7714, opened 2025-02-07
Fix: Customize editor back button for Campaign Page post type

kjohnson:
Resolves GIVE-2142

## Description

This PR customized the block editor back button to point to the edit URL of the Campaign related to the Campaign Page.

## Visuals

| Before | After |
| --- | --- |
| Campaign Page list table | Campaign Details page |
| Screenshot from 2025-02-20 21-58-15 | Screenshot from 2025-02-20 21-57-47 |

## Testing Instructions

From the Campaign Details page, click the "Edit campaign page" then navigate back to the Campaign Details page by clicking the back button (top left) in the editor.

## Pre-review Checklist

-   [ ] Includes unit tests


jonwaldstein:
@kjohnson nice job! I would just like to get @jdghinson's opinion of what icon we should put in that spot as I think this is a good time to make that decision.  For reference:

Current PR has a close icon:

Screenshot 2025-02-07 at 2 44 37 PM

I do see in our donation form we have a link (to campaign overview)

Screenshot 2025-02-07 at 2 43 46 PM

jdghinson:
@jonwaldstein We can use similar style like we have in the form builder however if we have a small space to work with we can use our Give logo icon in replace of the WordPress or "X" icon

Cc: @kjohnson

jonwaldstein:
@kjohnson can we try using the Give logo instead of the close icon? We should have an svg in the repo already.

kjohnson:
@jonwaldstein @jdghinson I tried adding the "Campaign Overview" link from the Form Builder, but the issue is that the core block editor enforces a fixed width on that slot (using a `grid-template` on the parent element).

So I'm leaning towards an icon button instead of hacking more space into the editor header.

That said, what is the reason for adding the Give logo over a Close Icon? I would think that the Close Icon better communicates the bahavior of closing the editor.

kjohnson:
@jonwaldstein @jdghinson Also, I just realized that the "Edit campaign page" button has a `target="_blank"`, so it opens the editor in a new tab/window. In which case we don't necessarily need a "back" button here.

Thoughts?

jonwaldstein:
@kjohnson 
>That said, what is the reason for adding the Give logo over a Close Icon? I would think that the Close Icon better communicates the bahavior of closing the editor.

From my perspective, we are not _closing_ the editor but instead _navigating back_ to the campaign.  That however, is me coming from a WordPress background where all other post editors have this UI and its what i'm familiar with and expecting (not saying it's right or wrong, just my own intuition).  So thats why I personally would be in favor of using a logo icon to be consistent with that experience in WordPress.

I do like that you brought this new perspective into it though and it's pushing me to think about it differently 😄 

One potential confusing part about the close icon is that it would be right next to the block inserter icon which can also be a close icon lol

Screenshot 2025-02-14 at 1 08 47 PM

jonwaldstein:
@kjohnson also good note about the landing page opening in a new tab, I didn't realize that.  In my opinion that should be opening in the same tab as we're navigating though our own UI and not an external page.

jdghinson:
@jonwaldstein The reason we made it open a new tab was because we didn't have a back option in the editor. However, since we have that option there's no need to having it open a new tab.

Secondly, I agree with Jon's submission on the icon usage, it's a pattern with these editors and even on the web where clicking on the logo takes you back home.

Cc: @kjohnson

kjohnson:
@jonwaldstein @jdghinson updated to use the Give Logo as the icon. I also update the "Edit campaign page" button so that it no longer opens in a new tab/window.

The flow is much better now.

jdghinson (approved):
All good!

src/Campaigns/Actions/EnqueueCampaignPageEditorAssets.php
```diff
@@ -0,0 +1,48 @@
...
+     */
+    public function __invoke()
+    {
+        $currentPost = get_post();
+
+        if($currentPost->post_type !== 'give_campaign_page') {
+            return;
+        }
+
+        $campaignPage = CampaignPage::find($currentPost->ID);
```
jonwaldstein, line 24: Models are nullable so we should return early if not found to avoid fatal errors trying to access property on null.

jonwaldstein (approved):
This is awesome, great job!  Just had one more suggestion, otherwise good to go 🌈

src/Campaigns/Actions/EnqueueCampaignPageEditorAssets.php
kjohnson: In this situation, the `CampaignPage` model represents the post being edited, so it would have to exist.

Probably still worth handling here, so an early return is probably sufficient.

kjohnson:
Build for QA, see https://github.com/impress-org/givewp/actions/runs/13461532026.

PR #7721, opened 2025-02-11
Fix: View Campaign Button

alaca:
## Description
This PR resolves the issue with the View Campaign button leading nowhere because the campaign page URL was not set. The issue is resolved by adding the campaign page URL to the button. Also, missing logic for updating the campaign page ID is added to the campaign model and repository. 

## Testing Instructions

Create Campaign
Edit campaign page (campaign page is created after the user clicks on the campaign page)
View Campaign

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Campaigns/Actions/AssociateCampaignPageWithCampaign.php
+++ b/src/Campaigns/Actions/AssociateCampaignPageWithCampaign.php
@@ -16,7 +16,10 @@ class AssociateCampaignPageWithCampaign
      */
     public function __invoke(CampaignPage $campaignPage){
         $campaign = Campaign::find($campaignPage->campaignId);
-        $campaign->pageId = $campaignPage->id;
-        $campaign->save();
+
+        if ($campaign) {
+            $campaign->pageId = $campaignPage->id;
+            $campaign->save();
+        }
     }
 }


src/Campaigns/resources/admin/components/CampaignDetailsPage/index.tsx
```diff
@@ -172,6 +177,14 @@ export default function CampaignsDetailsPage({campaignId}) {
         })();
     };
 
+    const getCampaignPageUrl = (campaignPage: { id: any; slug: any; link: any; }) => {
```
jonwaldstein, line 180: Let's avoid using the `any` type when we can as its not very helpful.  Also I believe this whole function can be moved outside of the component so its not re-created between re-renders.

```suggestion
    const getCampaignPageUrl = (campaignPage: { id: number; slug: string; link: string; }) => {
```

src/Campaigns/Actions/AssociateCampaignPageWithCampaign.php
```diff
@@ -0,0 +1,19 @@
...
+use Give\Campaigns\Models\Campaign;
+use Give\Campaigns\Models\CampaignPage;
+
+class AssociateCampaignPageWithCampaign
+{
+    /**
+     * @throws Exception
+     */
+    public function __invoke(CampaignPage $campaignPage){
+        $campaign = Campaign::find($campaignPage->campaignId);
```
jonwaldstein, line 15: Friendly reminder that our models are nullable and will cause a fatal error when accessing properties on null.  This one looks safe but might as well ask if `campaignId` from the campaignPage could potentially be null or campaign not found. If so, we should make sure `$campaign` exists before accessing properties.

src/Campaigns/Actions/AssociateCampaignPageWithCampaign.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Give\Campaigns\Actions;
+
+use Exception;
+use Give\Campaigns\Models\Campaign;
+use Give\Campaigns\Models\CampaignPage;
+
+class AssociateCampaignPageWithCampaign
```
jonwaldstein, line 9: Can we add a since tag

src/Campaigns/ServiceProvider.php
```diff
@@ -100,6 +101,7 @@ private function registerActions(): void
     {
         Hooks::addAction('givewp_campaign_deleted', DeleteCampaignPage::class);
         Hooks::addAction('givewp_donation_form_creating', FormInheritsCampaignGoal::class);
+        Hooks::addAction('givewp_campaign_page_created', AssociateCampaignPageWithCampaign::class);
```
jonwaldstein, line 104: Out of curiosity, what order are we creating these objects (`Campaign` & `CampaignPage`) in?

src/Campaigns/Actions/AssociateCampaignPageWithCampaign.php
alaca: Resolved https://github.com/impress-org/givewp/pull/7721/commits/9e375e3f10a5f58accfda9028943426552af54c8

src/Campaigns/Actions/AssociateCampaignPageWithCampaign.php
alaca: Resolved https://github.com/impress-org/givewp/pull/7721/commits/1466a32c7a87efe2edd8911cf3d38cf7f7771e21

src/Campaigns/ServiceProvider.php
jonwaldstein: Ah okay, thanks

jonwaldstein (approved):
nice job!

rickalday (approved):
Passed manual QA tests.

PR #7722, opened 2025-02-11
Refactor: remove "Add Form" options

glaubersilva:
Resolves [GIVE-2109]

## Description

This PR removes the _"Add Form"_ options outside the context of campaigns (eg: the _"All Forms"_ list tables and GiveWP main menu). Now, is not possible anymore to create campaigns without a valid campaignId associated with the form on the URL params.

So if you try to directly access any of the old URLs previously used to add new forms:

```
https://givewp.local/wp-admin/post-new.php?post_type=give_forms

https://givewp.local/wp-admin/edit.php?post_type=give_forms&page=givewp-form-builder
```

You'll be redirected to: 

```
https://givewp.local/wp-admin/edit.php?post_type=give_forms&page=give-campaigns&new=campaign
```

The "Add campaign form" button from the "Forms" tab inside the Campaign details page keeps working as before.

This PR also does a few replacements:

- Updates the places that were pointing directly to the old links to point to the new link.

- Update the link from the WP admin to add a new item, before it was _"Donation Form"_ and now is _"Donation Campaign"_.

## Affects

The "Add Form" options.

## Visuals

image

image

image

image

## Testing Instructions

1. Make sure the `GiveWP > Add Form` menu item doesn't exist anymore;
2. Go to the `GiveWP  > All Forms` page and make sure the "Add Form" button is not displayed, switch to the legacy view and make sure it is also not displayed there;
3. Click on the `+` icon in the WP admin area and make sure the new _"Donation Campaign"_ string is displayed instead of the old _"Donation Form"_ string
4. Try to access the old links to create forms (V2 and V3) directly and make sure you are redirected to the new link to create a campaign instead;
5. Open a campaign details page and in the Forms tab use the _"Add campaign form"_ button to add V2 and V3 forms and make sure it keeps working as before.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2109]: https://stellarwp.atlassian.net/browse/GIVE-2109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kjohnson (changes requested):
@glaubersilva my understanding was that these paths for adding new forms would just be removed. This PR increases the scope by introducing new paths for creating Campaigns. While this is probably something that will need to be done, my concern is that the quiet increase in scope is not being accounted for in review and QA.

If there are obvious places that we should consider introducing paths for creating new Campaigns then we should note those for refinemnt. Angela will need input on priority, Jeffery will need input on Design, and Ben will likely need input on wording (for example this PR introduces the new term "Donation Campaign").

glaubersilva:
@kjohnson About you comment:

> my understanding was that these paths for adding new forms would just be removed. 

If you just remove it visually and don't prevent the old links for creating forms without a campaign from working, so user would be able to still create a form without a campaign which is not allowed anymore. So, my understanding here is to not only remove the options visually but also the functionality in fact.

> This PR increases the scope by introducing new paths for creating Campaigns.

> If there are obvious places that we should consider introducing paths for creating new Campaigns then we should note those for refinemnt. 

I don't see it that way. In fact, this task was firstly set with lower story points and we increased the points later because I mentioned in the refinement session that we needed to check other places using these "Add Form" links. 

At that point in the refinement session, we mentioned the "Manual Donations" add-on as a sample of a possible place (but not the only one) to check - if you take a look at the Jira task will see the "_Confirm whether Manual Donations gives the option to Add Form."_ comment made by Angela when we had this discussion.

So, I checked and could see the "Manual Donations" add-on doesn't allow creating new forms, but in the other places where it's allowed, I replaced the link because we are not supporting "Add Form" outside the scope of a campaign which I think is not an increase in scope but part of the task.

> Angela will need input on priority, Jeffery will need input on Design

From my understanding, the priority already was set when we scoped it and about design, there is no change on it.

> and Ben will likely need input on wording (for example this PR introduces the new term "Donation Campaign").

I agree about it. In fact, I think keeping only "Campaign" would be enough, but we need to check with Angela to make sure. 

However, the scope of the task says to remove this menu item but while implementing the changes I thought would be better just to replace the functionality to create a new campaign instead of a new form.

So, about this one, we can double-check with Angela what she thinks is better.

src/Campaigns/Actions/PreventAddFormWithoutCampaign.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace Give\Campaigns\Actions;
+
+use Give\Campaigns\Models\Campaign;
+
+/**
+ * @unreleased
+ */
+class PreventAddFormWithoutCampaign
```
kjohnson, line 10: Can we rename this to reflect what it does (in addition to why)? Maybe something like `RedirectLegacyCreateFormRequest`?

kjohnson: Or, where is this needed? Shouldn't the use cases just be removed?

includes/post-types.php
```diff
@@ -66,7 +66,7 @@ function give_setup_post_types() {
 			'not_found_in_trash' => __( 'No forms found in Trash.', 'give' ),
 			'parent_item_colon'  => '',
 			'menu_name'          => apply_filters( 'give_menu_name', __( 'Donations', 'give' ) ),
-			'name_admin_bar'     => apply_filters( 'give_name_admin_bar_name', __( 'Donation Form', 'give' ) ),
+			'name_admin_bar'     => apply_filters( 'give_name_admin_bar_name', __( 'Campaign', 'give' ) ),
```
kjohnson, line 69: Let's just remove this from the Admin Bar for now, especially since this is hijacking the Donation Form post type definition in the legacy code. We can look at adding this at a later time as a feature of the Campaign domain.

kjohnson, line 69: Let's just remove this from the Admin Bar for now, especially since this is hijacking the Donation Form post type definition in the legacy code. We can look at adding this at a later time as a feature of the Campaign domain.

```suggestion

```

kjohnson (changes requested):
Let's remove the admin bar from Donation Forms for now and reconsider for Campaigns separatly.

includes/post-types.php
glaubersilva: I liked the filter idea, but to keep consistency it should apply to the other items of this array that we already changed previously:: `Add New Campaign Form` and `Edit Campaign Form` - check the attached screenshot for reference.

Resolved in b9944a0b28902791252b95b23c34812c29f2cb43

Regarding removing the `name_admin_bar` label from the "+ New" menu, my understanding of it, from the product management and design perspective, is that they preferred **not** to remove it. 

So, I would prefer to keep it there now, but we can discuss it again with the team and if we decide to remove this option from the "+ New" menu in the WP admin bar, we can just remove the references to the `name_admin_bar` label in the `post-types.php` and `ReplaceGiveFormsCptLabels.php` files.


image


image


image

src/Campaigns/Actions/PreventAddFormWithoutCampaign.php
glaubersilva: Resolved in 20a371b09bc913af8830b274a02713088a3867ca

kjohnson (approved):
@glaubersilva thanks for the back and forth, and the team discussion!

rickalday (approved):
Passed QA tests

PR #7723, opened 2025-02-11
Feature: add field validation hook and update Akismet integration

jonwaldstein:
Partially resolves [GIVE-1380]

## Description

These updates will resolve the following issues:
- Multi-step form designs not catching non-field external validation errors in between steps
- Gateways like PayPal that trigger their own validation endpoint presuming validity

This adds a new action to both the donate and validation routes, allowing us to intercept both the requests with one action:
```php 
  /**
   * @unreleased
   * 
   * @param array $data validated values in key value pairs
   */
  do_action('givewp_donation_form_fields_validated', $validatedValues);
 ```

This also updates our validation route with the following:
- Catch additional exceptions so all requests have a proper response (this was creating false positives on the front-end)
- Integrate Akismet during this request

Finally, this updates the existing Akismet logic to accept specific params instead of the `DonateControlleData`.

## Affects

- The validation of multi-step forms + Akismet
- The validation with PayPal smart buttons + Akismet

## Visuals

N/A

## Testing Instructions

Zip: https://github.com/impress-org/givewp/actions/runs/13318711776

**First setup Akismet and enable globally.**

**Multi-step forms**
1. Create a v3 form w/ multi-step design
2. Fill in data with [Akismet  error data](https://akismet.com/support/getting-started/confirm/) on step 2 of 3
3. Make sure step 2 throws an error

**PayPal commerce**
1. Create a v3 form with PayPal commerce smart buttons enabled
2. Fill in data with [Akismet  error data](https://akismet.com/support/getting-started/confirm/)
3. Try clicking the smart button, it should be rejected

**Regression tests**
1. Make sure v3 mult-step forms work as expected (successfully)
2. Make sure PayPal smart buttons work as expected (successfully)

## Pre-review Checklist

-   [x] Includes unit tests

[GIVE-1380]: https://stellarwp.atlassian.net/browse/GIVE-1380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### CODE REVIEW

File: src/DonationForms/Routes/ValidationRoute.php [hunk 1]
```diff
@@ -34,12 +36,18 @@ public function __invoke(array $request): bool
...
-            
+
             $this->handleResponse($response);
         } catch (DonationFormFieldErrorsException $exception) {
             $type = DonationFormErrorTypes::VALIDATION;
             $this->logError($type, $exception->getMessage(), $formData);
             $this->sendJsonError($type, $exception->getError());
+        } catch (DonationFormForbidden $exception) {
+            wp_die($exception->getMessage(), 403);
+        } catch (Exception $exception) {
```
jonwaldstein, line 47: This resolves an issue with the front-end receiving false positives when hitting the validation endpoint.  Some fields/hooks were throwing an exception that we were not catching here and in some cases was continuing the logic when it shouldn't have been.

kjohnson:
@jonwaldstein this change means that we are hitting the Akismet API on every page change in a multi-step form. For most validation is probably makes sense, but Akismet does require an HTTP request.

image

jonwaldstein:
@kjohnson you are absolutely right, It's something I made peace with during this PR since I feel the benefits outweighs the cons 😅 

However, what do you think of these two ideas:
1. We conditionally hit the api if certain args/fields are present (`email` and/or `comment`)
2. During AKismet validation, we store the args in memory and a flag if it passed, then subsequently we would check if the args are the same & previously passed before hitting the api.

jonwaldstein:
@kjohnson sounds good, thanks!

rickalday (approved):
Passed QA tests.

PR #7724, opened 2025-02-12
Feature: add campaign comments block

JoshuaHungDinh:
Resolves: [GIVE-1394]

## Pre-review Checklist

-   [ ] Includes unit tests


src/Campaigns/Blocks/CampaignComments/Controller/BlockRenderController.php
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * @unreleased
+     */
+    public function render(array $attributes): string
+    {
+        $blockAttributes = BlockAttributes::fromArray($attributes);
+
+        $encodedAttributes = json_encode($blockAttributes->toArray());
+
+        $this->loadScripts($blockAttributes);
```
alaca, line 25: Instead of loading assets on the backend, this is a new approach we are using now ([block.json](https://github.com/impress-org/givewp/blob/b1b67c0485dc351f5e00cfe84894ceb2a2eab9b1/src/Campaigns/Blocks/CampaignGoal/block.json#L21)). With this approach you don't need `loadScripts` method.

alaca (commented):
Hey @JoshuaHungDinh 
I hope you don't mind me checking your PR. I left a comment regarding assets loading.

src/Campaigns/Blocks/CampaignComments/Controller/BlockRenderController.php
JoshuaHungDinh: Hey yea thanks, I've been having difficulties localizing the window without registering it here. Any suggestions or thoughts?

alaca: Hey Joshua, I'd suggest registering a new route - `/campaign/(?P<id>\d+)/comments` and then fetch data from the block app. For data fetching you can use `useSWR` which is great. Here is an example of how to use it https://github.com/impress-org/givewp/blob/eab7e73280c43290913249938cd45abb7afb3566/src/Campaigns/Blocks/DonateButton/edit.tsx#L32

Also, the current approach might not be the best if we want to introduce pagination for comments. Imagine there are hundreds of comments, which is unlikely but, the current approach won't be that efficient because all comments data will be loaded every single time. And if there are multiple blocks on the same page, but from different campaigns, every single one of them will use the same global object `GiveCampaignCommentsBlockWindowData`.

jonwaldstein: @alaca I think that's a good suggestion for now, however in the future we should consider using the donations endpoint to get the comments (when those endpoints are ready).  We can then filter by `campaignId` and [HAL](https://stateless.group/hal_specification.html) to include donor data, checkout the slack thread on this topic https://lw.slack.com/archives/C04SLRDD9CK/p1738879925746179

src/Campaigns/Routes/GetCampaignComments.php
```diff
@@ -0,0 +1,109 @@
...
+        register_rest_route(
+            CampaignRoute::NAMESPACE,
+            CampaignRoute::CAMPAIGN . '/comments',
+            [
+                [
+                    'methods'             => WP_REST_Server::READABLE,
+                    'callback'            => [$this, 'handleRequest'],
+                    'permission_callback' => function () {
+                        return current_user_can('manage_options');
+                    },
```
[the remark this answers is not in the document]
JoshuaHungDinh: Hey that shouldn't be there 😄 ! 2da7430d0843be6507218ff5b75416477d782d69

src/Campaigns/Blocks/CampaignComments/block.json
```diff
@@ -0,0 +1,95 @@
...
+            "type": "boolean",
+            "default": true
+        },
+        "showName": {
+            "type": "boolean",
+            "default": true
+        },
+        "commentLength": {
+            "type": "number",
+            "default": 11
```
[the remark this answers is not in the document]
JoshuaHungDinh: Good point, I was also thinking similarly if we should change this to word count.
This commit updates the default char limit: adf24d73998323035452c0dc5d70f152b5140992

src/Campaigns/Blocks/CampaignComments/resources/app.tsx
```diff
@@ -0,0 +1,12 @@
+import {createRoot} from '@wordpress/element';
+import CampaignComments from './shared/components/CampaignComments';
+
+const roots = document.querySelectorAll('.givewp-campaign-comment-block');
```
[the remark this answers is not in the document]
JoshuaHungDinh: Ah, now that you mention it - I think I have seen that suggestion. Thanks!
Resolved: 1951d216d1f0e09831239bc2d846f9cdab8667aa

JoshuaHungDinh:
Thanks @jonwaldstein I also cleaned up the empty state UI .  3e625cd6b0444557ef01c2e7bfabc9165b6837e1

src/Campaigns/Blocks/CampaignComments/block.json
```diff
@@ -0,0 +1,95 @@
...
+    "attributes": {
+        "blockId": {
+            "type": "string"
+        },
+        "campaignId": {
+            "type": "number"
+        },
+        "title": {
+            "type": "string",
+            "default": "Share Your Support"
```
jonwaldstein, line 18: The fun little downside to using `block.json` is that attribute default strings are not translatable, so this would be hardcoded in english by default.  To get around this (and continue using `block.json`) I would suggest making the default an empty string and assign these as default params during destructuring of the attributes on edit like this:

```js
const {
      title = __('Share Your Support', 'give'),
      readMoreText = __('Read More', 'give'),
  } = attributes;
```

Screenshot 2025-02-17 at 10 12 14 AM

src/Campaigns/Blocks/CampaignComments/resources/edit.tsx
```diff
@@ -0,0 +1,96 @@
...
+import useCampaign from '../../shared/hooks/useCampaign';
+import {CampaignSelector} from '../../shared/components/CampaignSelector';
+import {useEffect} from 'react';
+
+export default function Edit({
+    attributes,
+    setAttributes,
+    clientId,
+}: BlockEditProps<{
+    blockId: string;
```
jonwaldstein, line 15: Looks like you can share the `AttributeProps` you created elsewhere, maybe make a types file so both can access

jonwaldstein (changes requested):

src/Campaigns/Blocks/CampaignComments/block.json
JoshuaHungDinh: Ok got it. What are your thoughts on assigning the string values within our useEffect in `edit.tsx`?

    useEffect(() => {
        if (!attributes.blockId) {
            setAttributes({blockId: clientId});
        }
        if (!attributes.title) {
            setAttributes({title: __('Share your support', 'give')});
        }
        if (!attributes.readMoreText) {
            setAttributes({readMoreText: __('Read more', 'give')});
        }
    }, []);
    
In my case destructuring in edit does not affect my frontend  & the way attributes are passed I would need to destructure in both edit and each component.

src/Campaigns/Blocks/CampaignComments/resources/edit.tsx
JoshuaHungDinh: Sounds good. I pulled the CommentData out & put the types inline as well. 53bb7453150382b8aca3a71b0cfba1db5ccfeb5e

src/Campaigns/Blocks/CampaignComments/block.json
jonwaldstein: So that approach _could_ work too but you will have to update the conditional to:
```js
if (attributes.title === undefined)
```
This is so its still possible to set the field to be empty.  Or else, If someone tries to delete the text in that field, the next render of the block would replace it with the default text, being an unexpected side affect.

You will also want to make sure that you remove the [block.json default key](https://github.com/impress-org/givewp/pull/7724/files#diff-2912ec6f539328b3e0e3b9dd78c36aafbed72d2b8d776f579e479a415f480a5f) completely for these so that they are undefined on first render (and not an empty string).

You also want to make sure you [BlockAttributes DTO](https://github.com/impress-org/givewp/pull/7724/files#diff-ad2226d3bd99257f4d6ebf7752c087d1b605aa23b1677f3f738ea7107679568f) does not impose a final fallback value or else the field can never be empty.

JoshuaHungDinh: Resolved: c5543ef7901bdd3c214dc54dac58fe95f4321c14
Thanks!

jonwaldstein (approved):
@JoshuaHungDinh nice job man!

rickalday (approved):

PR #7730, opened 2025-02-13
Refactor: replace "Donation Form" with "Campaign Form" in the onboarding wizard

glaubersilva:
Resolves [GIVE-2108]

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Campaigns/Repositories/CampaignRepository.php
+++ b/src/Campaigns/Repositories/CampaignRepository.php
@@ -76,10 +76,13 @@ class CampaignRepository
 
         Hooks::doAction('givewp_campaign_creating', $campaign);
 
-        $dateCreated = Temporal::withoutMicroseconds($campaign->createdAt ?: Temporal::getCurrentDateTime());
+        $currentDate = Temporal::getCurrentDateTime();
+
+        $dateCreated = Temporal::withoutMicroseconds($campaign->createdAt ?: $currentDate);
         $dateCreatedFormatted = Temporal::getFormattedDateTime($dateCreated);
-        $startDateFormatted = Temporal::getFormattedDateTime($campaign->startDate);
-        $endDateFormatted = Temporal::getFormattedDateTime($campaign->endDate);
+
+        $startDateFormatted = Temporal::getFormattedDateTime($campaign->startDate ?: $currentDate);
+        $endDateFormatted = Temporal::getFormattedDateTime($campaign->endDate ?: Temporal::withoutMicroseconds($currentDate->modify('+1 day')));
 
         DB::query('START TRANSACTION');
 
--- a/src/Onboarding/FormRepository.php
+++ b/src/Onboarding/FormRepository.php
@@ -2,11 +2,13 @@
 
 namespace Give\Onboarding;
 
+use Exception;
+use Give\Campaigns\Models\Campaign;
+use Give\Campaigns\ValueObjects\CampaignGoalType;
+use Give\Campaigns\ValueObjects\CampaignStatus;
+use Give\Campaigns\ValueObjects\CampaignType;
 use Give\DonationForms\Models\DonationForm;
-use Give\DonationForms\Properties\FormSettings;
 use Give\DonationForms\ValueObjects\DonationFormStatus;
-use Give\FormBuilder\Actions\GenerateDefaultDonationFormBlockCollection;
-use Give\Log\Log;
 
 /**
  * @since 2.8.0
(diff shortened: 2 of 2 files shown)


src/Campaigns/Repositories/CampaignRepository.php
```diff
@@ -76,10 +76,13 @@ public function insert(Campaign $campaign): void
...
-        $dateCreated = Temporal::withoutMicroseconds($campaign->createdAt ?: Temporal::getCurrentDateTime());
+        $currentDate = Temporal::getCurrentDateTime();
+
+        $dateCreated = Temporal::withoutMicroseconds($campaign->createdAt ?: $currentDate);
         $dateCreatedFormatted = Temporal::getFormattedDateTime($dateCreated);
-        $startDateFormatted = Temporal::getFormattedDateTime($campaign->startDate);
-        $endDateFormatted = Temporal::getFormattedDateTime($campaign->endDate);
+
+        $startDateFormatted = Temporal::getFormattedDateTime($campaign->startDate ?: $currentDate);
+        $endDateFormatted = Temporal::getFormattedDateTime($campaign->endDate ?: Temporal::withoutMicroseconds($currentDate->modify('+1 day')));
```
jonwaldstein, line 85: hmm are we trying to set a campaign end date by default? I don't think we should be setting an end date at all if its null

src/Onboarding/FormRepository.php
```diff
@@ -72,25 +74,38 @@ protected function isFormAvailable($formID)
...
+            'logo' => '',
+            'image' => '',
+            'primaryColor' => '#0b72d9',
+            'secondaryColor' => '#27ae60',
+            'goal' => 1000,
+            'goalType' => CampaignGoalType::AMOUNT(),
+            'status' => CampaignStatus::ACTIVE(),
         ]);
 
+        $form = DonationForm::find($campaign->defaultFormId);
```
jonwaldstein, line 100: Since models are nullable we should always be checking if it exists before accessing properties to avoid fatal errors

src/Campaigns/Repositories/CampaignRepository.php
glaubersilva: Yeah, I agree. I replicated the logic from the frontend here without thinking too much about it. But the logic there was wrong as well because the form submission always was sending an end date value.

Resolved in 496d1e42f87fd688fe147299f1769e9593eebbdc

src/Onboarding/FormRepository.php
glaubersilva: Resolved in 144bde14c13545b9d405d5ab215ed906d13a039e

src/Campaigns/Repositories/CampaignRepository.php
jonwaldstein: Okay good to know, I thought it was a little strange that a campaign would be ending in a day lol.  I do however see we are defaulting to 0 but the database column for end_date is a datetime type, so wouldn't that cause an error?

cc: @alaca

alaca: @jonwaldstein Usually yes, but in our case the `start_date` and `end_date` are nullable columns. `date_created` on the other hand can't be null. Also, passing `0` as a value to the `DATETIME` columns might not work on all servers. It depends on the server configuration.  There is a flag `NO_ZERO_DATE` and if it's set to true, the insert will fail. And if that flag is set to false, then I believe the value will be converted to `0000-00-00`.  Anyway, since `start_date` and `end_date` are nullable columns, I suggest passing `null`.

glaubersilva: @jonwaldstein Using `0` it gets saved without problems.

image

When I replace the `0` with `null` or `$campaign->endDate` or an empty string I get this error when other parts of the code try to retrieve the end date from DB:

```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Give\Framework\Support\Facades\DateTime\TemporalFacade::toDateTime() must be of the type string, null given, called in C:\Users\glaub\Local Sites\givewp\app\public\wp-content\plugins\give\src\Framework\Support\Facades\Facade.php on line 43 and defined in C:\Users\glaub\Local Sites\givewp\app\public\wp-content\plugins\give\src\Framework\Support\Facades\DateTime\TemporalFacade.php:21
```

It happens because the `0` is saved on the database in the proper date string format ( `0000-00-00 00:00:00` ) while the other alternatives that I mentioned above are saved as `NULL`.

jonwaldstein: I'm thinking it would be better for end date to be null/empty and we should conditionally format the endDate column here when Querying our campaigns https://github.com/impress-org/givewp/blob/61756418aa23032548abdd73f75feb195bbd9b58/src/Campaigns/Actions/ConvertQueryDataToCampaign.php#L37

Reason being, the end date of a campaign should be optional and we don't want to add confusion with an end date that is stored as that database fallback (`0000-00-00 00:00:00`)

What do you guys think? @glaubersilva @alaca

jonwaldstein: @glaubersilva great minds think alike!!!

src/Campaigns/Actions/ConvertQueryDataToCampaign.php
```diff
@@ -33,7 +33,7 @@ public function __invoke(object $queryObject): Campaign
             'goal' => (int)$queryObject->goal,
             'goalType' => new CampaignGoalType($queryObject->goalType),
             'startDate' => Temporal::toDateTime($queryObject->startDate),
-            'endDate' => Temporal::toDateTime($queryObject->endDate),
+            'endDate' => $queryObject->endDate ? Temporal::toDateTime($queryObject->endDate) : $queryObject->endDate,
```
jonwaldstein, line 36: To be safe, I would probably just default to null explicitly 

```suggestion
            'endDate' => $queryObject->endDate ? Temporal::toDateTime($queryObject->endDate) : null,
```

src/Campaigns/Repositories/CampaignRepository.php
```diff
@@ -76,10 +76,13 @@ public function insert(Campaign $campaign): void
...
-        $dateCreated = Temporal::withoutMicroseconds($campaign->createdAt ?: Temporal::getCurrentDateTime());
+        $currentDate = Temporal::getCurrentDateTime();
+
+        $dateCreated = Temporal::withoutMicroseconds($campaign->createdAt ?: $currentDate);
         $dateCreatedFormatted = Temporal::getFormattedDateTime($dateCreated);
-        $startDateFormatted = Temporal::getFormattedDateTime($campaign->startDate);
-        $endDateFormatted = Temporal::getFormattedDateTime($campaign->endDate);
+
+        $startDateFormatted = Temporal::getFormattedDateTime($campaign->startDate ?: $currentDate);
+        $endDateFormatted = $campaign->endDate ? Temporal::getFormattedDateTime($campaign->endDate) : $campaign->endDate;
```
jonwaldstein, line 85: Same here 

```suggestion
        $endDateFormatted = $campaign->endDate ? Temporal::getFormattedDateTime($campaign->endDate) : null;
```

jonwaldstein (approved):
@glaubersilva great job! love that we snuck in that end date update - I suggest defaulting to null explicitly otherwise good to go!

src/Campaigns/Actions/ConvertQueryDataToCampaign.php
glaubersilva: Resolved in 43c8a514386deb2e26b564708d89e2bfc1464b82

src/Campaigns/Repositories/CampaignRepository.php
glaubersilva: Resolved in 43c8a514386deb2e26b564708d89e2bfc1464b82

rickalday (approved):
Passed QA tests

PR #7733, opened 2025-02-14
Fix: Add a message for the donor when his card is declined

Genevieve-K:
In the request https://givewp.featureos.app/p/more-context-for-stripe-declined-messages, I suggest two points. This is the first one.


jonwaldstein (approved):
@Genevieve-K thank you!

PR #7734, opened 2025-02-14
Changed: Not display the incomprehensible message for the donor

Genevieve-K:
The message 'Unhandled payment intent status: %s' is incomprehensible for the donor. And there is not status because $paymentIntent is false when there is an exception, and this exception is already handled. 
See my second point the request https://givewp.featureos.app/p/more-context-for-stripe-declined-messages.

Test with card 4000000000000101

### DIFF
--- a/src/PaymentGateways/Gateways/Stripe/Traits/HandlePaymentIntentStatus.php
+++ b/src/PaymentGateways/Gateways/Stripe/Traits/HandlePaymentIntentStatus.php
@@ -22,6 +22,8 @@ trait HandlePaymentIntentStatus
      */
     public function handlePaymentIntentStatus(PaymentIntent $paymentIntent, Donation $donation)
     {
+      // If there is an Exception $paymentIntent is false
+      if ( $paymentIntent ) {
         switch ($paymentIntent->status()) {
             case 'requires_action':
                 $donation->gatewayTransactionId = $paymentIntent->id();
@@ -35,5 +37,6 @@ trait HandlePaymentIntentStatus
                 throw new PaymentIntentException(
                     sprintf(__('Unhandled payment intent status: %s', 'give'), $paymentIntent->status()));
         }
+      }
     }
 }
--- a/src/PaymentGateways/Gateways/Stripe/Traits/HandlePaymentIntentStatus.php
+++ b/src/PaymentGateways/Gateways/Stripe/Traits/HandlePaymentIntentStatus.php
@@ -22,7 +22,7 @@ trait HandlePaymentIntentStatus
      */
     public function handlePaymentIntentStatus(PaymentIntent $paymentIntent, Donation $donation)
     {
-      // If there is an Exception $paymentIntent is false
+      // If there is an Exception, paymentIntentObject is false
       if ( $paymentIntent ) {
         switch ($paymentIntent->status()) {
             case 'requires_action':


src/PaymentGateways/Gateways/Stripe/Traits/HandlePaymentIntentStatus.php
```diff
@@ -22,6 +22,8 @@ trait HandlePaymentIntentStatus
      */
     public function handlePaymentIntentStatus(PaymentIntent $paymentIntent, Donation $donation)
     {
+      // If there is an Exception $paymentIntent is false
+      if ( $paymentIntent ) {
```
jonwaldstein, line 26: I'm surprised this doesn't throw a type error when $paymentIntent is false 🤔

Genevieve-K: Sorry. My test is not correct.
It's not $paymentIntent that is false, but $this->paymentIntentObject in PaymentIntent.php.

Perhaps it's possible a function is PaymentIntent.php:
    public function isset()
    {
        return ( $this->paymentIntentObject );
    }

There is a warning:
PHP Warning:  Attempt to read property "status" on bool in /wp/wp-content/plugins/give/src/PaymentGateways/Gateways/Stripe/ValueObjects/PaymentIntent.php on line 45

Genevieve-K: @jonwaldstein What do you think about that?

jonwaldstein: Yeah I think your on the right track, but it sounds like we need to do a better job here of accounting for errors _before_ attempting to handle the payment status.

Genevieve-K: I agree that my solution is a workaround. As there is an Exception in the function `Create()` in` class-give-stripe-payment-intent.php`, a best solution is to catch this Exception in` CreatePaymentIntent.php` and/or in `CreditCardGateway.php`.
I keep the workaround in my GiveWP code waiting a best solution.

Genevieve-K: The payment Intent is not created due to a Stripe error. In my test: Your card's security code is incorrect.
But I found a simple solution in the file `class-give-stripe-payment-method.php`.
I do the sale think in class-give-stripe-payment-intent.php. I add the line:
`give_send_back_to_checkout( '?payment-mode=' . give_clean( $_GET['payment-mode'] ) );`
The exception is handled and it is not worth doing more and adding a new exception.
I create a new PR and we can close this one.

jonwaldstein: Sounds good, thanks!

Genevieve-K: I can now have a look to another Stripe error. When the error is `card_declined`, the donation is created (with the status `failed`), and the donor receives a confirmation and also an error message: `Payment Failed. Please contact the site owner for assistance.` That's confusing, and one of ours donors asked us a relevant question: `Is my donation successful or not?`

PR #7737, opened 2025-02-14
Enhancement: protect GiveWP upgrading

JasonTheAdams:
## Description

In light of #7735 and the fact that the licensing system broke GiveWP's ability to upgrade itself, I felt it would be a good idea to add a specific bailout to the upgrader function. It checks to see if the plugin being upgrade is GiveWP core itself. If it is, then it just immediately bails out of the add-on upgrader flow. This way if there are any licensing issues in the future we know that GiveWP can always at least be upgraded.

## Affects

The add-on upgrader flow.

## Testing Instructions

1. Install this on a site
2. Manually drop the Version in `give.php` to an older version so it thinks it can upgrade
3. Go to `includes/actions.php` on line 239 and add: `throw new Exception('boom');`
4. Try upgrading GiveWP

## Pre-review Checklist


JasonTheAdams:
This doesn't protect 3rd party plugins from being able to upgrade if there's an issue, but at the very least we can provide a hot fix for GiveWP to resolve the issue that is upgradable.

jonwaldstein (commented):
It looks good in theory, but was not able to confirm if it works.  Might want to check with QA & support as well.

jonwaldstein:
@JasonTheAdams did you ever send this one off to QA to test?

JasonTheAdams:
@jonwaldstein I think so? But if Rick didn't approve I think we should assume no and send it in again.

PR #7740, opened 2025-02-14
Fix: ensure campaign goal progress percentage is an integer and does not exceed 100

jonwaldstein:
Resolves [GIVE-2174]

## Description

This updates the goal progress percentage calculation so we can goal progress is an integer and does not exceed 100.

## Affects

The goal progress widget

## Visuals

Screenshot 2025-02-14 at 4 57 22 PM

## Testing Instructions

Zip: https://github.com/impress-org/givewp/actions/runs/13374370469

- Create a campaign with a small goal of amount
- Make several donations with varying float amounts
- Make sure the goal progress percentage is a whole number and does not exceed 100

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2174]: https://stellarwp.atlassian.net/browse/GIVE-2174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/GoalProgressChart/index.tsx
+++ b/src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/GoalProgressChart/index.tsx
@@ -8,8 +8,15 @@ import {getCampaignOptionsWindowData, amountFormatter} from '@givewp/campaigns/u
 const {currency} = getCampaignOptionsWindowData();
 const currencyFormatter = amountFormatter(currency);
 
-const GoalProgressChart = ({ value, goal }) => {
-    const percentage: number = Math.abs((value / goal) * 100);
+type GoalProgressChartProps = {
+    value: number;
+    goal: number;
+}
+
+const GoalProgressChart = ({ value, goal }: GoalProgressChartProps) => {
+    const progress = Math.ceil((value / goal) * 100);
+    const percentage = Math.min(progress, 100);
+
     return (
             <div className={styles.goalProgressChart}>
                 <div className={styles.chartContainer}>


pauloiankoski:
Honest question: Why can’t the progress percentage exceed 100%? IMHO, it would be nice to see that my campaign was a success and reached 200% of my original goal.

pauloiankoski (approved):
👍🏻

pauloiankoski (approved):
👍🏻

jdghinson:
@jonwaldstein Actually I did a design for this scenario, we just didn't surface it in our scope conversations.

Cc: @pauloiankoski 

image

jonwaldstein:
@jdghinson ah good to know! that's pretty consistent with my reasoning to have some kind of way to show progress past 100. However, I would still raise the question if a circle bar is even appropriate for showing values over 100 as the UI is always going to show a max of a 100 due to the nature of a circle.

jdghinson:
@jonwaldstein There's no UI that will well represent over 100% unless the max is infinite, which doesn't match our use case because there will always be a target goal. That's why in the design, you have the actual goal progress being represented with the percentage and number, and also a supporting text, that makes it clear that you've exceeded your target goal.

jonwaldstein:
@jdghinson Thanks for your insight. I do believe there are other forms of UI that could show progress past 100%, for example: _a bar graph_. However, I understand the nuance and think this is a great UX discussion to have offline from here 😄.

rickalday (approved):
Passed manual QA tests.

PR #7741, opened 2025-02-14
Fix: corrects license handler notice dismissibility and UI

JasonTheAdams:
## Description

I noticed that when an auto activation licensing error occurs it doesn't look quite right. I realized that there was a typo in `isDmissibile()` — should be `dismissible()`, and auto-paragraph should be enabled.

It also provides a fallback reason for an error that might be null and wraps the license in `<code>` tags for readability.

## Affects

Just the notices displayed when a bundled license error occurs.

## Visuals

image

## Testing Instructions

Note, to test you will need an add-on with the `PLUGIN_LICENSE.php` file included in the root. Reach out to @JasonTheAdams  for one.

The easiest way to test this without setting up the GiveWP website locally is to:
* Add the following to `includes/class-give-license-handler.php` on line 348:
    ```php
    $check_license_res['success'] = null;
    ```
    This will simulate a server response failure
* Make sure everything looks right
* Dismiss the notice
* Refresh the page to make sure the notice doesn't display again

## Pre-review Checklist

### DIFF
--- a/includes/class-give-license-handler.php
+++ b/includes/class-give-license-handler.php
@@ -328,7 +328,7 @@ if ( ! class_exists('Give_License') ) :
                 AdminNotices::show(
                     "license-bundle-activation-error-$license_key",
                     "An error occurred while attempting to activate license $license_key: $reason. Please activate the license manually."
-                )->asError()->isDismissible();
+                )->autoParagraph()->asError()->dismissible();
             };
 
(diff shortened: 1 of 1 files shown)


JoshuaHungDinh:
Good ol'e "Dmissibile" 😄

JoshuaHungDinh (approved):

PR #7744, opened 2025-02-18
Fix: Update PHPDoc for Log Method

yiedpozi:
## Description

Update PHPDoc for Log method.

## Visuals

image


jonwaldstein (approved):
Thanks for the update!

PR #7748, opened 2025-02-19
Fix: Skip Recording Payment Status Change If No Modification

yiedpozi:
## Description

- Skip recording a payment status change in the donation note if the status remains unchanged.

## Visuals

image

Updating donation status to Complete, but the donation status is already set as Complete. A donation note for payment status change still will be recorded.

## Testing Instructions

Try updating the donation status programatically, example:

```
$donation_status = DonationStatus::COMPLETE(); // use current donation status, ie: complete

$donation->status = $donation_status;
$donation->save();
```

No more redundant updates on payment status change.


jonwaldstein (approved):
Thanks!

PR #7753, opened 2025-02-20
Return to the checkout form when the payment intent is created

Genevieve-K:
When the payment intent cannot be create due to a Stripe error (Ex:  `Your card's security code is incorrect`.), there is an Exception.  Some functions as `handlePaymentIntentStatus()` cannot access to status/id, and a new exception is created. There is no point in going any further, and we can return directly to the checkout form as in `class-give-stripe-payment-method.php`.

Errors in log:
```
[19-Feb-2025 13:03:38 UTC] PHP Warning:  Attempt to read property "id" on bool in /wp/wp-content/plugins/give/src/PaymentGateways/Gateways/Stripe/ValueObjects/PaymentIntent.php on line 44
[19-Feb-2025 13:03:38 UTC] PHP Warning:  Attempt to read property "status" on bool in /wp/wp-content/plugins/give/src/PaymentGateways/Gateways/Stripe/ValueObjects/PaymentIntent.php on line 53
```
Message for the donor from  `handlePaymentIntentStatus()`:
image

A question: if this error occurs in a recurring donation, the error is simply: `Your card's security code is incorrect.` , `There was an issue with your donation transaction:...`

### DIFF
--- a/includes/gateways/stripe/includes/class-give-stripe-payment-intent.php
+++ b/includes/gateways/stripe/includes/class-give-stripe-payment-intent.php
@@ -73,7 +73,7 @@ if ( ! class_exists( 'Give_Stripe_Payment_Intent' ) ) {
                 		$e->getMessage()
                 	)
                 );
-
+                give_send_back_to_checkout( '?payment-mode=' . give_clean( $_GET['payment-mode'] ) );
                 return false;
             } // End try().
         }


jonwaldstein:
@Genevieve-K thanks for this! I think this is a nice simple solution that _should_ work 😄 

I think your analysis of the nested exception is accurate (knowing this code is considered _legacy_ old and specific to our v2 forms now).  

The proper way this `Give_Stripe_Payment_Intent::create` logic should be reused in our Gateway API is to simply throw an exception with a message for reasoning and let the gateway api respond back to the form.  Instead, we are catching the exception inside the gateway class, which continues the logic while setting errors in session.

For future reference: [here](https://github.com/impress-org/givewp/blob/7b593d9a0a560b4fb618e0288c82992f70cfe6bd/src/LegacyPaymentGateways/Adapters/LegacyPaymentGatewayAdapter.php#L212) is where we are catching all gateway exceptions for v2 forms.  We would need to update the exception message to be dynamic and let gateways throw exceptions instead of catching them.  This is how our v3 forms work with gateways, so all they need to do is throw an exception to be caught later 😄

jonwaldstein (approved):

Genevieve-K:
@jonwaldstein Thank you. We added a cart to GiveWP. So we separate the donation form and the checkout form. A donor can give several gifts before paying. It's difficult to use the v3 forms.

PR #7755, opened 2025-02-21
Feature: Add support for running migration in batches

alaca:
Resolves: [GIVE-2243]

## Description
This PR adds support to run migration in batches using Action Scheduler. 

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2243]: https://stellarwp.atlassian.net/browse/GIVE-2243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Framework/Migrations/Contracts/Migration.php
```diff
@@ -16,9 +16,15 @@ abstract class Migration
...
+     * @unreleased this method is now optional if migration class implements BatchMigration interface
      * @since 2.9.0
      */
-    abstract public function run();
+    public function run()
+    {
+        if ( ! is_subclass_of($this, BatchMigration::class)) {
+            throw new RuntimeException('run method is not defined.');
+        }
+    }
```
JasonTheAdams, line 27: Let's avoid this if we can. This sort of thing can't be caught by static analysis, so the only way to know if there's an issue is at runtime.

src/Framework/Support/Facades/ActionScheduler/AsBackgroundJobsFacade.php
```diff
@@ -42,6 +42,25 @@ public function enqueueAsyncAction(
...
+        bool $unique = false,
+        int $priority = 10
+    ): int {
+        $enqueuedAction = $this->getActionByHookArgsGroup($hook, $args, $group, 'ids');
+        if (empty($enqueuedAction)) {
+            return as_schedule_single_action($timestamp, $hook, $args, $group, $unique, $priority);
+        }
+
+        return $enqueuedAction[0];
+    }
```
JasonTheAdams, line 62: Please provide a description for this. It looks like it conditionally enqueues a single action so long as that action isn't already queued? We may want to call this `safelyEnqueueAction` or something like that.

JasonTheAdams (changes requested):
Love the start, @alaca! A couple thoughts:
1. I really want to use interfaces, but we're going to run into issues because we really should be using inheritance to for sync and async migrations, but interfaces can't enforce static methods until PHP 7.4.
2. So for now, let's:
    - make a `BaseMigration` abstract class with the static `id()` and `timestamp()` methods
    - Change `BatchMigration` to an abstract class that extends this
    - Extend the `Migration` class with this
    - Use `BaseMigration` in the migrations runner and registrar
3. We still want batch migrations to hold up the queue, as all migrations need to run in order, regardless of what kind of migration they are

Love seeing this! So cool!

src/Framework/Migrations/Contracts/Migration.php
JasonTheAdams: If we switch everything to inheritance we shouldn't need to do this.

src/Framework/Migrations/MigrationsServiceProvider.php
```diff
@@ -30,6 +30,8 @@ public function boot()
     {
         Hooks::addAction('admin_init', ManualMigration::class, '__invoke', 0);
         Hooks::addAction('admin_init', MigrationsRunner::class, 'run', 0);
-        Hooks::addAction('give_upgrades', MigrationsRunner::class, 'run', 0);
+        //Hooks::addAction('give_upgrades', MigrationsRunner::class, 'run', 0);
+        // running batch actions via cron doesn't trigger give_upgrades and all registered actions fail
```
JasonTheAdams, line 34: Will this break the existing migrations that are using the old upgrade system?

alaca: I believe not, but to be safe, I left the `give_upgrades` action and changed `admin_init` to `action_scheduler_init`.

src/Framework/Support/Facades/ActionScheduler/AsBackgroundJobsFacade.php
alaca: I removed this method because I don't find this helper class very useful. Mainly because of the limited Action Scheduler API support and performance reasons. If we want to build an AS wrapper, there's a better and much more performant approach we can take (using the `ActionScheduler_Store` instance). Check the new `BatchMigrationRunner` controller.

src/Framework/Migrations/Contracts/Migration.php
alaca: I agree, resolved!

alaca:
There is one big potential issue with running migrations in batches. Batch migration will stop the Migration Runner until it processes all batches, which means migrations that are next in the queue won't be processed. The problem will occur if the next migrations are supposed to change the table structure or add new tables. We will get a bunch of errors if we want to use new properties from the "updated" table or query the table that is not created yet. We have to make sure to run all batch migrations last. Luckily, there is an easy solution for this. So instead of using `strtotime(date)` for batch migration timestamp, we just have to use `time()`. This will ensure that all batch migrations run last.

alaca:
@JasonTheAdams check this https://github.com/impress-org/givewp/pull/7755/commits/c620062b9a0c63160577c5033b6f94b30b25e849
Our migration system was running without any order 😱

JasonTheAdams (changes requested):
Great refactor, @alaca! So much good work here! I left a bunch of comments. Hahah! You're getting my head going with this one, which is fun!

As a note, you mentioned concern over batch migrations causing issues if things like table mutations and the like aren't done in the right order. That's the very reason the migrations are intended to run in sort order, so this can be controlled. And if there's a batch migration that relies on a previous migration, then that should either have an earlier timestamp or else be included in the batch migration on the first step. In any case, I think this system is flexible enough to allow that concern to live in the implementing code, and not something we should try and tackle in the migration framework. And honestly, no, I _would not_ recommend ever using a function like `time()` to create a dynamic return value, as that breaks the whole system.

src/Framework/Migrations/Controllers/BatchMigrationRunner.php
```diff
@@ -0,0 +1,169 @@
...
+    private $actionSchedulerStore;
+
+    /**
+     * @unreleased
+     */
+    public function __construct(BatchMigration $migration)
+    {
+        $this->migration = $migration;
+        $this->actionSchedulerStore = ActionScheduler_Store::instance();
+        $this->registerAction();
```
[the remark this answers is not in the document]
alaca: Make sense!

src/API/Endpoints/Migrations/RunMigration.php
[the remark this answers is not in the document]
JasonTheAdams: I'm talking about the migration system. I guess we do need to think about what happens if batch migration enters into a partial state — wherein some of the steps have succeeded and others have failed. 🤔 

To be clear, at no point should two migrations be running — even if they're batch migrations.

alaca:
> As a note, you mentioned concern over batch migrations causing issues if things like table mutations and the like aren't done in the right order. That's the very reason the migrations are intended to run in sort order, so this can be controlled. And if there's a batch migration that relies on a previous migration, then that should either have an earlier timestamp or else be included in the batch migration on the first step.

Make sense. We should be extra careful with timestamps then. I was setting a dynamic timestamp as a safe bet, so I don't have to worry about it. And it kinda makes sense. I mean, the purpose of batch migration is to move/update data and the logical order for that action is to be the last step.

src/API/Endpoints/Migrations/RunMigration.php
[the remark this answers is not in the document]
alaca: > To be clear, at no point should two migrations be running — even if they're batch migrations.

Yeah, that is not happening. Not even in our current broken system 😅

JasonTheAdams:
> Make sense. We should be extra careful with timestamps then. I was setting a dynamic timestamp as a safe bet, so I don't have to worry about it. And it kinda makes sense. I mean, the purpose of batch migration is to move/update data and the logical order for that action is to be the last step.

Exactly. The best way to think of the timestamp is as a unique and sortable identifier across all systems (core plus add-ons). Something like setting a simple step integer (1, 2, 3) would be gnarly to keep track of. Instead, this relies on the simple fact that when the dev builds the migration they're making the assumption of the shape of the database at that moment in time. So by setting the time to that moment, it locks that migration into that step of the process. Make sense?

src/API/Endpoints/Migrations/RunMigration.php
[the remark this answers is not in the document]
JasonTheAdams: Happy to hop on a call, at a time that isn't interrupting work, and think through this a bit more.

alaca:
> So by setting the time to that moment, it locks that migration into that step of the process. Make sense?

Yes, it makes perfect sense.

Let's imagine this scenario: 
The user is running Give 3.21.1, but the most recent Give version is 4.0.
There are many versions between these two where we introduced many new features, new tables, models, etc.
The user decides to update to 4.0
What will happen?

You mentioned that using new tables, models, etc. should be an implementation detail. I agree, but we will have to constantly do checks if something is implemented. That's why batch migrations are a little bit different than regular migrations. batch migrations shouldn't stop migration runner if some of the migration actions are failed.

src/Framework/Migrations/MigrationsRunner.php
```diff
@@ -103,19 +106,53 @@ public function run()
...
-
             try {
                 /** @var Migration $migration */
                 $migration = give($migrationClass);
 
-                $migration->run();
-
-                // Save migration status
-                $migrationLog->setStatus(MigrationLogStatus::SUCCESS);
+                if (is_subclass_of($migration, BatchMigration::class)) {
```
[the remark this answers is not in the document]
alaca: These are not migration class instances 😀

src/Framework/Migrations/Controllers/BatchMigrationRunner.php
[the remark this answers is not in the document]
alaca: Resolved: https://github.com/impress-org/givewp/pull/7755/commits/6698d318a4876abac25a273c182c6fa76d6f00c6

src/API/Endpoints/Migrations/GetMigrations.php
```diff
@@ -175,6 +177,7 @@ public function handleRequest(WP_REST_Request $request)
                 'run_order' => $this->migrationHelper->getRunOrderForMigration($migration::id()),
                 'source' => $migration::source(),
                 'title' => $migration::title(),
+                'isBatchMigration' => is_subclass_of($migration, BatchMigration::class)
```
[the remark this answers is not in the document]
alaca: These are not migration class instances

alaca:
@JasonTheAdams I believe we are close to the finish line. I really like we are passing a range of IDs to each batch which makes me feel better knowing no data will be left behind 😅
The are a couple of minor things I want to discuss, mainly used wording and the last migration step, but overall, I'm pleased 😁

alaca:
@JasonTheAdams Hey man, do you have time to take a look at this? There are a couple of new changes.

src/Framework/Migrations/MigrationsRunner.php
```diff
@@ -103,19 +97,55 @@ public function run()
...
-                $migrationLog->setStatus(MigrationLogStatus::SUCCESS);
+                if ($migration instanceof BatchMigration) {
+                    $status = (new BatchMigrationRunner($migration))->run();
+
+                    if ($status === MigrationLogStatus::RUNNING) {
+                        give()->notices->register_notice(
+                            [
+                                'id' => $migrationId,
+                                'description' => sprintf(
+                                    esc_html__('Running DB migration: %s', 'give'),
```
JasonTheAdams, line 114:
```suggestion
                                    esc_html__('GiveWP is running the "%s" migration in the background. You will be notified as soon as it completes.', 'give'),
```

src/Framework/Migrations/MigrationsRunner.php
```diff
@@ -103,19 +97,55 @@ public function run()
...
+                        break;
+                    }
+
+                    if ($status === MigrationLogStatus::INCOMPLETE) {
+                        give()->notices->register_notice(
+                            [
+                                'id' => $migrationId,
+                                'type' => 'warning',
+                                'description' => sprintf(
+                                    esc_html__('Incomplete DB migration: %s', 'give'),
```
JasonTheAdams, line 135: What does this mean for the admin? What should they do at this point?

src/Framework/Migrations/readme.md
```diff
@@ -0,0 +1,205 @@
...
+### Standard migration
+We use these migrations each time we want to:
+- Create new tables
+- Edit existing tables (add/remove/update table columns)
+- Delete tables
+
+### Batch migration
+Batch migrations, as name suggests, are processed in batches. We use them __only when we need to process large data sets__.
+
+__Important__: You will never use this type of migration for table creation, or table editing.
```
JasonTheAdams, line 21:
```suggestion
__Important__: Do not use this type of migration for table creation or table editing.
```

src/Framework/Migrations/readme.md
JasonTheAdams, line 1: We don't necessarily need to do it in this PR, but I'd like to add a "Best Practices" section based on our conversations. It's clear that the best built migration system can still be fragile if good practices aren't followed.

Also, let's document the various migration log statuses and what they indicate.

src/Campaigns/Migrations/Donations/AddCampaignId.php
```diff
@@ -86,7 +104,56 @@ public function run()
...
+            ->getAll();
+
+        if ( ! $items) {
+            return null;
         }
+
+        return [
+            min($items)->ID,
+            max($items)->ID,
+        ];
```
JasonTheAdams, line 139: It's faster to use a subquery and return only the first and last values:
```sql
SELECT MIN(id) AS first_id, MAX(id) AS last_id
FROM (
    SELECT id
    FROM posts
    WHERE id > 200
    ORDER BY id ASC
    LIMIT 100
) AS batch;
```

src/Campaigns/Migrations/Donations/AddCampaignId.php
```diff
@@ -86,7 +104,56 @@ public function run()
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function hasIncomingData($lastProcessedId): ?bool
+    {
+        return $this->query()
+            ->where('ID', $lastProcessedId, '>')
+            ->count();
```
JasonTheAdams, line 157: Wouldn't this query everything after that ID? Shouldn't we apply `->limit(1)`, here?

src/Framework/Migrations/Contracts/BatchMigration.php
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * The purpose of this method is to check if we have new items that came during the migration.
+     *
+     * @unreleased
+     *
+     * @param $lastProcessedId
+     *
+     * @return bool - true if there are new items, otherwise false
+     */
+    abstract public function hasIncomingData($lastProcessedId): ?bool;
```
JasonTheAdams, line 59: Let's call this `hasMoreItemsToBatch()`

src/Framework/Migrations/Contracts/BatchMigration.php
```diff
@@ -0,0 +1,60 @@
...
+    abstract public function getItemsCount(): int;
+
+
+    /**
+     *
+     * Get the first and the last item ID for a batch
+     *
+     * @unreleased
+     *
+     * @return array<int, int>
```
JasonTheAdams, line 34: I don't think this syntax is correct. I believe this is for an associative array where the keys are integers and the values are integers. I believe this is correct:
```suggestion
     * @return array{0: int, 1: int} the first value is the first id and the second value is the last id of a batch
```

src/Framework/Migrations/Controllers/BatchMigrationRunner.php
```diff
@@ -0,0 +1,214 @@
...
+                return MigrationLogStatus::SUCCESS;
+            }
+
+            $current = 0;
+            $batches = ceil($itemsCount / $this->migration->getBatchSize());
+
+            // Register migration action for each batch
+            for ($i = 0; $i < $batches; $i++) {
+                if ($items = $this->migration->getBatchItemsAfter($current)) {
+                    [$firstId, $lastId] = $items;
```
JasonTheAdams, line 67: Let's add a validation step here that throws an exception if it's not a 2-element array with integers.

src/Framework/Migrations/Controllers/BatchMigrationRunner.php
```diff
@@ -0,0 +1,214 @@
...
+
+        $pendingActions = (int)$this->actionSchedulerStore->query_actions([
+            'group' => $this->getGroup(),
+            'status' => [
+                ActionScheduler_Store::STATUS_RUNNING,
+                ActionScheduler_Store::STATUS_PENDING,
+            ],
+        ], 'count');
+
+        if ($pendingActions) {
```
JasonTheAdams, line 84: Is this the total number of actions running? Should we compare this to the `$batches` number to make sure they're equal?

Same question below for `$failedActions`.

src/Framework/Migrations/Controllers/BatchMigrationRunner.php
```diff
@@ -0,0 +1,214 @@
...
+            $this->actionSchedulerStore->delete_action($actionId);
+        }
+
+        return MigrationLogStatus::SUCCESS;
+    }
+
+    /**
+     * @unreleased
+     */
+    public function getHook(): string
```
JasonTheAdams, line 116: Seems like this can be private?

src/Framework/Migrations/Controllers/BatchMigrationRunner.php
```diff
@@ -0,0 +1,214 @@
...
+    {
+        return 'givewp-batch-' . $this->migration::id();
+    }
+
+    /**
+     * @unreleased
+     */
+    public function getGroup(): string
```
JasonTheAdams, line 124: Same. Looks like this can be private.

JasonTheAdams (changes requested):
Whew! Finally found some time to do a review! I think this is mostly little nits at this point. I agree that we're really narrowing in here. 👍

src/Framework/Migrations/MigrationsRunner.php
alaca: They should go to the migrations list table where they can re-run the failed migrations. We should probably add a link to the migrations list table here.

src/Campaigns/Migrations/Donations/AddCampaignId.php
alaca: Yep, good catch!

src/Framework/Migrations/Controllers/BatchMigrationRunner.php
alaca: The thing with the last ID parameter is that it will be null on the last step (if there is more data). Check https://github.com/impress-org/givewp/blob/24162f82fd016d24b183407c48c0b7d98d79c0fd/src/Framework/Migrations/Controllers/BatchMigrationRunner.php#L158-L159

We are pulling the last ID from the last action. That last ID will be the first ID of the new action. We don't need the last ID for new action because we will process all the items we found. See https://github.com/impress-org/givewp/blob/24162f82fd016d24b183407c48c0b7d98d79c0fd/src/Campaigns/Migrations/Donations/AddCampaignId.php#L77

src/Framework/Migrations/Controllers/BatchMigrationRunner.php
alaca: Resolved https://github.com/impress-org/givewp/pull/7755/commits/39639508d9282b6b0f385c9947666f9946cb7bfd

src/Framework/Migrations/Controllers/BatchMigrationRunner.php
alaca: Resolved https://github.com/impress-org/givewp/pull/7755/commits/39639508d9282b6b0f385c9947666f9946cb7bfd

src/Framework/Migrations/Contracts/BatchMigration.php
alaca: Resolved https://github.com/impress-org/givewp/pull/7755/commits/144450bf1331e555f2259ff3b80b5179c9867ac7

src/Campaigns/Migrations/Donations/AddCampaignId.php
alaca: Resolved https://github.com/impress-org/givewp/pull/7755/commits/d844f0d1c3bdb8ab925950ea08db699a98ad54fd

src/Framework/Migrations/Contracts/BatchMigration.php
alaca: Resolved https://github.com/impress-org/givewp/pull/7755/commits/d844f0d1c3bdb8ab925950ea08db699a98ad54fd

src/API/Endpoints/Migrations/RunMigration.php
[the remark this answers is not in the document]
JasonTheAdams: @alaca I'd like to switch to a RESTful PUT/PATCH API in a subsequent PR.

src/Framework/Migrations/MigrationsRunner.php
JasonTheAdams: Yup, then let's update this description to provide instructions.

src/Framework/Migrations/Controllers/BatchMigrationRunner.php
JasonTheAdams: The good news is we know when it should be what based on the iteration of the loop, and can therefore validate it accordingly. So:
* If $i is less than $batches, it should be a 2-index array of integers
* If $i is equal to $batches, it should be null

src/Campaigns/Migrations/Donations/AddCampaignId.php
JasonTheAdams: Any thoughts on this, @alaca?

src/Framework/Migrations/Controllers/BatchMigrationRunner.php
alaca: That's true when we set the initial number of batches. What is a bit different is that if the `hasMoreItemsToBatch` function returns true, the BatchMigrationRunner will schedule another batch where all remaining items will be processed. In that situation we don't need to know the range of IDs, we just need to know the last processed item ID. Also, actions/batches don't even know what is the current batch number, they just have the range of IDs for which they are responsible.

src/Campaigns/Migrations/Donations/AddCampaignId.php
alaca: Love it, I just have to implement it 😁

src/Framework/Migrations/Controllers/BatchMigrationRunner.php
JasonTheAdams: Fair enough. If we really just don't have a way of validating here, then YOLO. 🚬 🧐

alaca: What really matters is that we pass the correct IDs, the batch number is not even important. It's used to determine how many batches we should have at that point in time.

src/Campaigns/Migrations/Donations/AddCampaignId.php
alaca: Hmm, I don't see a reason to use the subquery here. We can simplify this by doing this:

```sql
SELECT MIN(id) AS first_id, MAX(id) AS last_id
    FROM posts
    WHERE id > 200
    ORDER BY id ASC
    LIMIT 100
```

I think it should return the same result... I'll check

alaca: Yep, it does work the same

alaca: Resolved https://github.com/impress-org/givewp/pull/7755/commits/62d5a167e6954e8aeb0465bbfad53b0c34438c5c

JasonTheAdams: Nice! I don't know why I thought it would need that. 😆

src/Framework/Migrations/MigrationsRunner.php
alaca: Resolved https://github.com/impress-org/givewp/pull/7755/commits/057ea159288bc94947079c914038841690545a3b

rickalday (approved):
Passed QA tests.

PR #7761, opened 2025-02-26
Feature:  Campaign Page template

alaca:
## Description

Registers a new campaign page template that will be used in themes that support FSE.
Also, the default Gutenberg title block is hidden and replaced with the Campaign title block.

## Affects

Campaign page

## Testing Instructions

Install FSE theme
Create Campaign
Preview the campaign page on frontend
Edit site appearance

## Pre-review Checklist

-   [ ] Includes unit tests


alaca:
@pauloiankoski I added you as a reviewer

src/Campaigns/Actions/CampaignPageTemplate.php
```diff
@@ -0,0 +1,57 @@
...
+        ) {
+            $template = $this->canRegisterBlockTemplate()
+                ? $template
+                : GIVE_PLUGIN_DIR . '/src/Campaigns/resources/views/campaign-page-template.php';
+
+            return locate_block_template($template, 'campaign-page-template', ['campaign-page-template.php']);
+        }
+
+        return $template;
+    }
```
pauloiankoski, line 48: When hooking into `template_include`, we should first check if the user’s theme provides a template before falling back to our own. In this case, `single-give_campaign_page.php` should be prioritized if it exists in the theme.

src/Campaigns/Blocks/CampaignTitle/editor.scss
```diff
@@ -11,3 +11,8 @@
         }
     }
 }
+
+// Hide the default Gutenberg's post title block as we don't use that
+.edit-post-visual-editor__post-title-wrapper {
+    display: none;
+}
```
pauloiankoski, line 18: Ideally, this should be in an editor stylesheet, but for now, it works to keep it within the `CampaignTitle` block since it is being loaded all the time.

pauloiankoski (commented):
Left some points to discuss before approving but nothing critical has been found.

src/Campaigns/Actions/CampaignPageTemplate.php
alaca: The whole purpose of this function is to load our template for themes that support FSE because they load a ton of stuff that we don't want to see on our campaign page. That's the point of this PR. Users can always change the layout from the editor if they don't like it. For sites that don't support FSE, the dev can create `single-give_campaign_page.php` and it will be loaded normally.

src/Campaigns/Blocks/CampaignTitle/editor.scss
alaca: We use our title block only on campaign pages. This makes sense as it is loaded only on campaign pages.

src/Campaigns/Actions/CampaignPageTemplate.php
pauloiankoski: CleanShot 2025-02-28 at 13 00 56

This is an example of a theme without support for FSE. I thought that campaign-page-template.php would surface exactly in cases like that. It sounds like we are implementing it partially, so we’ll need to communicate that clearly to our users.

pauloiankoski (approved):

src/Campaigns/Blocks/CampaignTitle/editor.scss
pauloiankoski: Yeah, but the styles are not being applied to the CampaignTitle block but to the page title element instead. That's what I meant.

rickalday (approved):
Passed QA tests.

PR #7739, opened 2025-02-14
Refactor: remove option to delete a campaign

jonwaldstein:
Resolves [GIVE-2150]

## Description

This removes the option to delete a campaign from the campaign list table.

This ensures we avoid unexpected behaviour by deleting a campaign that we have not accounted for yet (ie. abandoned donations, donors, forms, etc.)

## Affects

The campaign list table

## Visuals

Screenshot 2025-02-14 at 4 35 57 PM

Screenshot 2025-02-14 at 4 40 12 PM

## Testing Instructions

ZIP: https://github.com/impress-org/givewp/actions/runs/13379005587

- Create a campaign
- Make sure you don't see an option to delete that campaign in list table including bulk actions

## Pre-review Checklist

-   [ ] Acceptance criteria satisfied and marked in related issue
-   [ ] Relevant `@unreleased` tags included in DocBlocks
-   [ ] Includes unit tests
-   [ ] Reviewed by the designer (if follows a design)
-   [ ] [Self Review](https://give.gitbook.io/development-manual/devops/github/code-reviews#self-review) of code and UX completed

[GIVE-2150]: https://stellarwp.atlassian.net/browse/GIVE-2150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7765, opened 2025-02-27
Fix: Prevent trashing a campaign default form through bulk actions

pauloiankoski:
Resolves [GIVE-2280]

## Description

Although we have removed the inline option to delete a campaign’s default form, users are still able to do so through bulk actions. This pull request prevents that by:
1.	Adding support for a new `isIdSelectable` method, allowing each action to determine whether an ID should be included in the list of IDs to be processed.
2.	Implementing the `isIdSelectable` method in the `trash` action to filter out the default form ID if it has been selected.

If only the default form is selected, the behavior remains unchanged—it will not open the modal, as there are no valid selections. Ideally, we should provide more context to users in the future when this occurs. However, for now, this solution is sufficient, as the primary goal is to prevent users from deleting their campaign’s default form.

## Affects
List Table's bulk actions but specifically for Campaign Forms

## Visuals
CleanShot 2025-02-27 at 15 28 39 2

## Testing Instructions
1.	On the Campaign Details page, navigate to the Forms tab.
2.	Attempt to remove the default form using bulk actions.
3.	Create additional forms and try including the default form in the list of selected forms for the bulk action.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2280]: https://stellarwp.atlassian.net/browse/GIVE-2280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DonationForms/V2/resources/components/DonationFormsListTable.tsx
+++ b/src/DonationForms/V2/resources/components/DonationFormsListTable.tsx
@@ -240,7 +240,7 @@ const donationFormsBulkActions: Array<BulkActionsConfig> = [
         value: 'trash',
         type: 'danger',
         isVisible: (data, parameters) => parameters.status !== 'trash' && data?.trash,
-        isIdSelectable: (id, data) => data.defaultForm == null || data.defaultForm !== +id,
+        isIdSelectable: (id, data) => !(typeof data?.defaultForm === 'number') || data.defaultForm !== Number(id),
         action: async (selected) => await API.fetchWithArgs('/trash', {ids: selected.join(',')}, 'DELETE'),
         confirm: (selected, names) => (
             <div>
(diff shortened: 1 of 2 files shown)


alaca (changes requested):
I believe this check should be added on the backend. Regarding the frontend, I'd expect that the checkbox for the default form shouldn't be selectable or visible, but that's just cosmetics and it's up to you. Bulk actions are handled through API so if we really want to prevent someone from deleting the default form, we should do that in the API endpoint.

pauloiankoski:
@alaca I agree with you. However, this PR is a follow-up to https://github.com/impress-org/givewp/pull/7739, where we just removed the row action while keeping it in the bulk actions. It’s important to keep in mind that all these List Tables might be revisited in the near future for a refactoring and are very likely to have their API endpoints refactored as well, following up on Glauber’s work to establish the first truly RESTful API.

alaca (approved):
Looks good 👍

rickalday (approved):
Passed manual QA tests.

PR #7786, opened 2025-03-11
Feature: Support Multiple Goal Types in Campaign Goal Progress Chart

pauloiankoski:
Resolves [GIVE-2290]

## Description
This Pull Request enhances the Campaign Goal Progress Chart component by adding support for multiple goal types. Previously, the chart always displayed progress as “Amount Raised”, regardless of the selected goal type.

With this update:
- The widget correctly displays progress based on the selected goal type.
- Currency formatting is now applied only for `amount` and `amountFromSubscriptions` goal types.

Under the hood:
- Added a new `CampaignSubscriptionQuery` class.
- Replaced all instances of the `GoalType` enum from `DonationForms` in the `CampaignGoalData` class with the `CampaignGoalType` enum from `Campaigns`.
- Updated the `getActual` method in the `CampaignGoalData` class to implement distinct query builders depending on the selected campaign goal type (`CampaignDonationQuery` or `CampaignSubscriptionQuery`). Goal types related to subscriptions have been incorporated into the case conditional.

This improves accuracy and clarity in representing different campaign goal types in the Campaign Overview tab.

## Affects
Initially, the Campaign Overview tab, but as a side effect, any other Campaign Goal widget as well.

## Visuals
https://github.com/user-attachments/assets/d7375346-c7bf-4131-b8a9-4241f028639a

## Testing Instructions
To test this PR, you’ll need a campaign with some donations and subscriptions. Then, go to the Settings tab of the campaign, switch between the Goal Type options, and verify that they are correctly reflected in the Campaign Goal widget in the Overview tab.

## Pre-review Checklist

-   [x] Includes unit tests

[GIVE-2290]: https://stellarwp.atlassian.net/browse/GIVE-2290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Campaigns/CampaignSubscriptionQuery.php
+++ b/src/Campaigns/CampaignSubscriptionQuery.php
@@ -30,6 +30,7 @@ class CampaignSubscriptionQuery extends QueryBuilder
         // Include only forms associated with the Campaign.
         $this->joinDonationMeta(DonationMetaKeys::CAMPAIGN_ID, 'campaignId');
         $this->where('campaignId.meta_value', $campaign->id);
+
     }
 
     /**
@@ -37,13 +38,13 @@ class CampaignSubscriptionQuery extends QueryBuilder
      */
     public function between(DateTimeInterface $startDate, DateTimeInterface $endDate): self
     {
-        $this->whereBetween(
+        $query = clone $this;
+        $query->whereBetween(
             'created',
             $startDate->format('Y-m-d H:i:s'),
             $endDate->format('Y-m-d H:i:s')
         );
-
-        return $this;
+        return $query;
     }
 
     /**
@@ -55,7 +56,7 @@ class CampaignSubscriptionQuery extends QueryBuilder
      */
     public function sumInitialAmount()
     {
-        return $this->sum('initial_amount');
+        return (clone $this)->sum('initial_amount');
     }
 
     /**
@@ -63,7 +64,7 @@ class CampaignSubscriptionQuery extends QueryBuilder
      */
     public function countDonations(): int
     {
-        return $this->count('donation.ID');
+        return (clone $this)->count('donation.ID');
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/Campaigns/CampaignSubscriptionQuery.php
```diff
@@ -0,0 +1,109 @@
...
+        $this->where('campaignId.meta_value', $campaign->id);
+
+    }
+
+    /**
+     * @unreleased
+     */
+    public function between(DateTimeInterface $startDate, DateTimeInterface $endDate): self
+    {
+        $query = clone $this;
```
jonwaldstein, line 41: The only thing i'm not sure about is why we are cloning the query for a lot of these methods.  I saw this pattern in the `CampaignDonationsQuery` and it confused me.  It doesn't have to hold up the PR but something to think about if it's necessary.

jonwaldstein (approved):
Nice job man!  Let's have QA test this one with various campaign subscriptions 🌈

src/Campaigns/CampaignSubscriptionQuery.php
pauloiankoski: Yeah, honestly, I don’t think it’s necessary. Though, it’s been added to the `CampaignDonationQuery` methods for some reason, so I kept that pattern. Would you like me to remove those?

jonwaldstein: yeah I think it's safe to remove them.  I don't like cloning objects unless its absolutely necessary and I'm not sure why it would be needed here.

pauloiankoski: All removed on https://github.com/impress-org/givewp/pull/7786/commits/d99d34bf2d3c9e2e398dfe45ad5b26a74393854d

pauloiankoski: @jonwaldstein Oh, I see what’s happening here. Those methods are not just meant to append instructions to the main query but to execute it and return a result. By cloning the query, we can use the same CampaignDonationQuery object to perform multiple requests with methods like `countDonations`, `countDonors`, and so on. Otherwise, we would need to create a new object for each request. I think we'll need to revert that commit back.

jonwaldstein: ah I see

rickalday (approved):
Passed QA tests.

PR #7793, opened 2025-03-12
Fix: campaign migration dates

jonwaldstein:
## Description

The current form to campaign migration is settings the start and end dates as the form goal dates.  This is not correct as these are separate concepts.  The start date of a campaign should be when the form was created and there should be no end date (end date is not a concept yet).

This will ensure the lifespan of a form is migrated to a campaign and the revenue tracking will be accurate because it uses the same dates.

This also protects against potential type errors using arrap_map as results can return null.

## Affects

- The Form to Campaign migration

## Visuals

N/A

## Testing Instructions

- Migrate forms to campaigns and make sure the dates are correct.

## Pre-review Checklist

-   [ ] Includes unit tests


alaca (approved):
Looks good 👍

PR #7801, opened 2025-03-17
Feature: add support for async data on the campaigns list tables

glaubersilva:
Resolves [GIVE-2348]

**Based on:** https://github.com/impress-org/givewp/pull/7483

## Description

This PR fixes two problems:

**Problem 1**

The "Forms" tab from the campaigns details page wasn't loading the `Goal`, `Donations`, and `Revenue` columns using the async approach implemented [**here**](https://github.com/impress-org/givewp/pull/7483) previously for the forms list views, but now they are.

**Problem 2**

Websites with large amounts of data can face problems in loading the `Goal`, `Donations`, and `Revenue` columns of the campaigns list table.

So, to solve the problem a new approach to loading the campaign stats data was implemented. Now, we load the admin campaign list view without loading the campaign stats - these data are loaded through subsequent async requests to the server only after the page gets loaded. 

In a test site, the initial page load was taking **12 minutes** but after the changes introduced on this PR the same site now is loading in **4.8 seconds** - check the visuals section for reference.

If the customer (or the support team) wishes to disable the async approach to the admin campaign list views is possible to do that by adding the following options on the `wp-config.php` file:

```php
define("GIVE_IS_ALL_STATS_COLUMNS_ASYNC_ON_ADMIN_CAMPAIGN_LIST_VIEW", false);
//define("GIVE_IS_GOAL_COLUMN_ASYNC_ON_ADMIN_CAMPAIGN_LIST_VIEW", false);
//define("GIVE_IS_DONATIONS_COLUMN_ASYNC_ON_ADMIN_CAMPAIGN_LIST_VIEW", false);
//define("GIVE_IS_REVENUE_COLUMN_ASYNC_ON_ADMIN_CAMPAIGN_LIST_VIEW", false);
```

**IMPORTANT:**

Please note that some servers have rules to prevent too many requests at the same time for the same client as described here in this Cloudflare forum thread: https://community.cloudflare.com/t/error-429-message/625392

So, what can happen is that some servers can return a 429 error when some rules like that are enabled. It also will return a Retry-After header with the time the client needs to wait until trying to do a new request, usually 60 minutes. More details here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429#response_containing_retry-after_header

So, to try to avoid this error, a throttling technique was implemented to limit the requests made at once when the list pages get loaded. By default, only one async request will be run per time.

But is possible to disable the throttling by adding the following in the `wp-config.php` file:

`define('GIVE_ASYNC_DATA_THROTTLING', false);`

However, it's recommended to let the throttling feature be enabled. Also, It's important to understand that even with this feature enabled the client can get 429 errors due to highly restrictive rate-limiting rules applied to the Server/Firewall - for these cases, the rules should be adjusted or removed.

## Affects

- The "Forms" tab of the campaign details page.
- The admin campaign list view.

## Visuals

**12 minutes** to be able to interact with the campaign list view page **BEFORE** the changes implemented on this PR:

campaigns-loaded

**4.8 seconds** to be able to interact with the campaign list view **AFTER** the changes implemented on this PR:

campaigns-list-table-loading

## Testing Instructions

1. Setup a site with a lot of data that has the problem - you can find a site sample in the attached Jira task;
2. Setup the Give version of this branch on the problematic site;
3. Access the campaign list views on the admin area;
4. Make sure the page from the previous step got loaded in a reasonable time;

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2348]: https://stellarwp.atlassian.net/browse/GIVE-2348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


glaubersilva:
Close in favor of https://github.com/impress-org/givewp/pull/7819 and https://github.com/impress-org/givewp/pull/7832

PR #7809, opened 2025-03-18
Refactor: Update PayPal Commerce to capture order within the Gateway API

jonwaldstein:
Resolves [GIVE-2175]

## Description

This updates the PayPal Commerce gateway to capture the PayPal order server-side through the gateway api.  This will ensure we are never actually capturing the payment (on the front-end) until the form processing navigates to the gateway's server request which is always after full form validation. 

## Affects

PayPal Commerce

## Visuals

**One-time**

Screenshot 2025-03-22 at 10 39 34 PM

Screenshot 2025-03-22 at 10 58 16 PM

**Recurring**

Screenshot 2025-03-18 at 5 34 13 PM

## Testing Instructions

- Test PayPal commerce on both v2 and v3 forms
- Make sure payment is successful with smart buttons and hosted fields for both one-time payments and subscriptions

Checklist:
- [x] PayPal commerce works on v2 forms for one-time donations using smart buttons & hosted fields
- [x] PayPal commerce works on v2 forms for recurring donations using smart buttons
- [x] PayPal commerce works on v3 forms for one-time donations using smart buttons & hosted fields
- [x] PayPal commerce works on v3 forms for recurring donations using smart buttons
- [x] On v3 forms, PayPal commerce does not capture a payment when validation fails (payment is only authorized)
- [x] On v3 forms, PayPal commerce respects form value changes when using hosted fields
- [x] On v3 forms, PayPal commerce respects form value changes when using smart buttons (credit / debit fields)

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2175]: https://stellarwp.atlassian.net/browse/GIVE-2175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jonwaldstein:
@glaubersilva this is ready for review! I tried my best to keep the scope of this as small as possible even though it was tempting to refactor _everything_ 😄 .  I did some thorough testing already to make sure there are no regression issues and that v3 forms is authorizing payments and capturing them later & that subscriptions still work.  

The only thing I wasn't sure about was the `updateOrderAmount` variable which might not be necessary anymore 🤔 .

Anyway, happy to walk you through these changes together if that's easier.

glaubersilva:
@jonwaldstein I was testing it and could see that the status is set to `pending` when we authorize it on the frontend and only gets changed to `complete` when we capture it in the backend. 

So, it might be possible to change the amount in the `createPayment()` method, but it would work only for V3 forms since v2 forms keep capturing it on the front end, so the transactions originated from v2 forms will always get the complete status.

This way, I think we still need the `updateOrderAmount` and I wouldn't touch on it right now to keep the compatibility with v2 forms as well.

image

glaubersilva (approved):
@jonwaldstein Great work! For sure PayPal transactions now will be more stable. I tested all scenarios and it worked as expected. 

The only thing that we still need to confirm is if the PayPal smart button for credit/debit cards will always open in a new modal/tab or if there is any scenario where still is possible to open it inside the same page. If the new modal/tab is the new PayPal standard, so I think we can remove the `updateOrderAmount` logic.

Anyway, I'll approve it and after confirmation from the PayPal team, we can decide whether to remove it or not.

khairo1956:
I have tried it and it works well, the only problem and drawback is that it opens a new tab page when choosing to pay with debit /credit card for Paypal button only, this made me lose a lot of donors if not to say 95% of my donors ,Because it is long and makes the customer fill in all his information and this takes a lot of time, then it suddenly disappears and the site reloads from the beginning. I wish you'll solve solve this problem only..

PR #7843, opened 2025-03-28
Fix: Edit Donation Form should not open campaign creation modal

alaca:
## Description
This PR resolves the issue with the redirect logic for campaign forms. The issue was that the `RedirectLegacyCreateFormToCreateCampaign` class was not accounting for the p2p campaign forms.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Campaigns/Actions/RedirectLegacyCreateFormToCreateCampaign.php
+++ b/src/Campaigns/Actions/RedirectLegacyCreateFormToCreateCampaign.php
@@ -14,49 +14,17 @@ class RedirectLegacyCreateFormToCreateCampaign
      */
     public function __invoke()
     {
-        if ( ! $this->isAddingNewForm()) {
-            return;
+        if (
+            $GLOBALS['pagenow'] === 'post-new.php'
+            && isset($_GET['post_type']) && $_GET['post_type'] === 'give_forms'
+        ) {
+            if (
+                ! isset($_GET['campaignId'])
+                || ! (Campaign::find(absint($_GET['campaignId'])))
+            ) {
+                wp_safe_redirect(admin_url('edit.php?post_type=give_forms&page=give-campaigns&new=campaign'));
+                exit;
+            }
         }
-
-        if ($this->isEditingCampaignForm()) {
-            return;
-        }
-
-        if ( ! isset($_GET['campaignId']) || ! (Campaign::find(absint($_GET['campaignId'])))) {
-            wp_safe_redirect(admin_url('edit.php?post_type=give_forms&page=give-campaigns&new=campaign'));
-            exit;
-        }
-    }
-
-    /**
-     * @unreleased
-     */
-    private function isAddingNewForm(): bool
-    {
-        global $pagenow;
-
-        $isOptionBasedFormEditorPage = $pagenow === 'post-new.php';
-        $isVisualFormBuilderPage = $pagenow === 'edit.php' && isset($_GET['page']) && 'givewp-form-builder' === $_GET['page'];
-        $isGiveFormsCpt = isset($_GET['post_type']) && $_GET['post_type'] === 'give_forms';
-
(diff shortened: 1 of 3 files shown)


glaubersilva:
@alaca Just for reference, the function that was preventing the bug reported in my last comment was this one: https://github.com/impress-org/givewp/pull/7722/files#diff-dc3e755da9c9426f96f2c1a99270a0449d13776199070ee2e0a22ac48da8a2c8R45-R61

alaca:
@glaubersilva, @pauloiankoski - could you guys review this? Thanks!

glaubersilva:
@alaca I pulled all your last changes, and I still can reproduce the exact same bug demonstrated in my last video. I saw you fixed the problem for P2P forms, but for regular forms using VFB, I'm still being redirected to the create campaign modal.

alaca:
@glaubersilva 

https://github.com/user-attachments/assets/cf188809-a69e-4389-b461-dae1f5f687fd

glaubersilva (approved):
@alaca Thanks for the last changes and for refactor it to use declarative methods, much easier to ready and understand the logic now. In my tests, everything works as expected. Ready to Go! 🚀

src/Campaigns/Actions/RedirectLegacyCreateFormToCreateCampaign.php
```diff
@@ -33,30 +38,73 @@ public function __invoke()
...
 
-        if ($formId && $isGiveFormsCpt) {
-            return (bool)Campaign::findByFormId($formId);
+    /**
+     * @unreleased
+     */
+    private function isP2PCampaignFormIdInvalidOrMissing(): bool
+    {
+        if (class_exists(\GiveP2P\P2P\Repositories\CampaignRepository::class)) {
+            return ! isset($_GET['donationFormID']) || ! give(\GiveP2P\P2P\Repositories\CampaignRepository::class)->getByFormId($_GET['donationFormID']);
```
jonwaldstein, line 97: @alaca do we need to pull in the p2p repository or can we just check if the campaign type is p2p?

src/Campaigns/Actions/RedirectLegacyCreateFormToCreateCampaign.php
```diff
@@ -33,30 +38,73 @@ public function __invoke()
...
+               && isset($_GET['page']) && $_GET['page'] === 'givewp-form-builder';
+    }
 
-        return ($isOptionBasedFormEditorPage || $isVisualFormBuilderPage) && $isGiveFormsCpt;
+    /**
+     * @unreleased
+     */
+    private function isP2P(): bool
+    {
+        return isset($_GET['p2p']);
```
jonwaldstein, line 62: Is there anyway we don't have to have this like can we just see if the form is associated to a campaign and if the type is p2p?

src/Campaigns/Actions/RedirectLegacyCreateFormToCreateCampaign.php
alaca: @jonwaldstein P2P doesn't use `campaign_type` column.  We added that to the core.

jonwaldstein: What about this migration that adds the p2p type though?  When someone upgrades to this version of Give that should be available to use right?

https://github.com/impress-org/givewp/blob/b13f5ff2706d435b573afb770fcdf2d67cb021f0/src/Campaigns/Migrations/P2P/SetCampaignType.php#L51

src/Campaigns/Actions/RedirectLegacyCreateFormToCreateCampaign.php
alaca: As I said, P2P doesn't use the `campaign_type` column. 
It would be easy if we added p2p meta to the form, but that is not the case. We need something to identify where to form belongs, and I believe this is our only option.

PR #7846, opened 2025-03-28
Fix: campaign list table when empty

jonwaldstein:
Resolves [GIVE-2372]

## Description

This fixes an error on the the campaign list table when it's empty by adding an empty conditional to the campaign data query when there are no ids to be used in a WHER IN clause.

## Affects

The campaign list able.

## Visuals

N/A

## Testing Instructions

- Create a fresh install, dismiss onboarding and view campaign list table. There should be no errors

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2372]: https://stellarwp.atlassian.net/browse/GIVE-2372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Campaigns/Routes/GetCampaignsListTable.php
+++ b/src/Campaigns/Routes/GetCampaignsListTable.php
@@ -97,8 +97,19 @@ class GetCampaignsListTable implements RestRoute
         $this->request = $request;
         $this->listTable = give(CampaignsListTable::class);
 
-        $campaigns = $this->getCampaigns();
         $campaignsCount = $this->getTotalCampaignsCount();
+
+        if ($campaignsCount === 0) {
+            return new WP_REST_Response(
+                [
+                    'items' => [],
+                    'totalItems' => 0,
+                    'totalPages' => 0,
+                ]
+            );
+        }
+
+        $campaigns = $this->getCampaigns();
         $pageCount = (int)ceil($campaignsCount / $request->get_param('perPage'));
 
         $ids = array_map(function (Campaign $campaign) {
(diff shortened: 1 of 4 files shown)


jonwaldstein:
@alaca this conditional technically solves the problem, it also kind of defeats the purpose of this query builder as we are attempting to use it when there are no campaignIds.  So we can either do that here or up the chain during implementation, I figured this would be a little easier 😄

alaca (commented):
Yeah, we shouldn't call this class if we don't have campaign IDs anyway. It might be better to update the campaigns route and add a check there instead of doing it here. The problem with this is that we are still running queries even though we don't have campaigns.

jonwaldstein:
@alaca agreed, let me update that

jonwaldstein:
@alaca updated to return early during implementation

src/Campaigns/Routes/GetCampaignsListTable.php
```diff
@@ -97,8 +97,19 @@ public function handleRequest(WP_REST_Request $request): WP_REST_Response
         $this->request = $request;
         $this->listTable = give(CampaignsListTable::class);
 
-        $campaigns = $this->getCampaigns();
         $campaignsCount = $this->getTotalCampaignsCount();
+
```
alaca, line 101: Nice optimization here

alaca (approved):
Looks good 👍

PR #7849, opened 2025-03-31
Fix: prevent delete default form

glaubersilva:
Resolves [GIVE-2381]

## Description

This PR prevents the accidental exclusion of Campaign default forms through the _"All Forms"_ legacy page.

## Affects

The legacy "All Forms" page

## Testing Instructions

1. Acces the "All Forms" legacy page using the direct URL - example: `https://givewp.local/wp-admin/edit.php?post_type=give_forms&page=give-forms` or `https://givewp.local/wp-admin/edit.php?post_type=give_forms`
2. Try to delete campaign default forms and make sure it is not possible to do that
3. Try to delete other forms and make sure it keeps working as before

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2381]: https://stellarwp.atlassian.net/browse/GIVE-2381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Campaigns/ServiceProvider.php
```diff
@@ -123,6 +124,9 @@ private function registerActions(): void
         Hooks::addAction('givewp_campaign_page_created', AssociateCampaignPageWithCampaign::class);
         Hooks::addAction('give_form_duplicated', Actions\AssignDuplicatedFormToCampaign::class, '__invoke', 10, 2);
 
+        Hooks::addAction('givewp_donation_form_deleting', PreventDeleteDefaultForm::class);
```
jonwaldstein, line 127: This action passes a donation model instance, looks like your trying to access an ID.

Also, I don't think the legacy tables use the form model do they? What if we just hid the trash link instead of being heavy handed?

jonwaldstein (changes requested):

src/Campaigns/ServiceProvider.php
glaubersilva: @jonwaldstein  You are right! This mistake happened because I was testing different hooks and did the push using the wrong one. The `before_delete_post action` works for both list views. Resolved in 07cc80e8c191723831d568ffe2e738f16e1ecee6

glaubersilva: @jonwaldstein The advantage of this approach, instead of hiding the link, is that it also blocks accidental exclusions through bulk action for both list views.

src/Campaigns/Actions/PreventDeleteDefaultForm.php
```diff
@@ -0,0 +1,53 @@
...
+    public function __invoke($postId)
+    {
+        if (get_post_type($postId) !== 'give_forms') {
+            return;
+        }
+
+        $campaign = Campaign::findByFormId($postId);
+
+        if ($campaign && $campaign->defaultFormId == $postId) {
+            wp_die(sprintf(__('The form %s with ID %d cannot be deleted because it is the default form for a campaign.',
```
jonwaldstein, line 24: Does this return a response to the user in the UI or does it just cause a 500 error if someone tries to trash a default form?

glaubersilva: @jonwaldstein I know this is not the best UX, but as a note, when we try to delete the default form using the bulk actions of the Forms tab on the campaign details page, there is also no return in the UI. The only difference here is that we are returning a 500 error.

jonwaldstein: Okay thanks, I think we need to come up with a way to have the UI respond for customers or else this could be confusing for them 😏

glaubersilva: @jonwaldstein Do you think a simple alert would be enough? Since this page is "hidden" and we are recommending not to use it anymore, I think it would be enough for now.

jonwaldstein: ya that would work 👍

glaubersilva: Resolved in 9e9e96a678d438935ec0fa12f66490b665cdd29f

glaubersilva:

jonwaldstein (approved):
@glaubersilva looks good, thanks! ready for QA 😄

rickalday (approved):
Passed QA tests.

PR #7877, opened 2025-04-14
Fix: Add argument type check

alaca:
## Description
This pull request includes an update to the `ArrayDataSet` class in the `src/Helpers/ArrayDataSet.php` file. The main change introduces a check to ensure that the input is an array before attempting to rename its keys.

## Affects

ArrayDataSet helper class

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Helpers/ArrayDataSet.php
+++ b/src/Helpers/ArrayDataSet.php
@@ -11,6 +11,7 @@ class ArrayDataSet
      * You can pass a multi dimensional array but only zero level array keys will be renamed.
      *
      * @since 2.7.0
+     * @unreleased Add is_array check
      *
      * @param array $renameTo Pass array of existing key name as key and new key name as value.
      *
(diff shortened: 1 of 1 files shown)


src/Helpers/ArrayDataSet.php
```diff
@@ -20,6 +21,10 @@ class ArrayDataSet
      */
     public static function renameKeys($array, $renameTo)
     {
+        if ( ! is_array($array)) {
+            return [];
```
pauloiankoski, line 25: I’m just wondering if we shouldn’t return the original variable like a WP filter does, or maybe even convert it into an array as that's our return type. I know that anyone using this method with a non-array variable is getting an error, but to me, it feels like we should return the original input like “here’s your variable back, I couldn’t modify it.” What do you think?

alaca: I was thinking the same, the only problem with that approach is that the function is expected to return an array. If we return something else, like null, everything after that will break.

pauloiankoski: Yeah, then we could return it like `[$array]`

alaca: That doesn't make any sense. You are returning this `[0=>null]` if someone passes `null`.

alaca: Honestly, I would like to add an array as a type so that whenever someone passes null or something other than an array, it will break. But this happens for some reason, and I'm not entirely sure where this happens. So this is basically a band-aid.

pauloiankoski (approved):

PR #7879, opened 2025-04-15
Fix: The Campaigns strings were not displayed translated

DAnn2012:
Fix: The Campaigns strings were not displayed translated


glaubersilva:
@DAnn2012 Thanks for your help! We merged it into the PR [#7914](https://github.com/impress-org/givewp/pull/7914) that also has other fixes for campaigns missing strings. It still will need to pass through our QA process before getting merged to the develop branch, but it should be there soon. 🙂

PR #7888, opened 2025-04-22
Feature: add currency switcher listeners to core

jonwaldstein:
Related: [GIVE-2516], [GIVE-859], https://github.com/impress-org/give-currency-switcher/pull/321

## Description

This moves the actions that handle updating currency meta into core.  Previously all this logic lived in the currency switcher add-on however, as more of the currency concepts actually live in core it makes sense for this logic to live there as well.  This will ensure that core will handle associating currency data to the donation, eventually leaving the currency switcher add-on to simply supply settings like exchange rates.

## Affects

- creating donations with currency settings (currency switcher)

## Visuals

https://www.loom.com/share/01fb49c1491c483c97c559fb2c367421?sid=90566c00-457c-4422-b223-7f5c1d6f190f

## Testing Instructions

ZIP: https://github.com/impress-org/givewp/actions/runs/14604325546

- create a campaign with both v2 and v3 forms
- install currency switcher and setup settings to enable currency switcher on both forms
- create donation using the currency switcher and make sure the meta for donation is accurate for exchange_rate, as well as the revenue table.

## Pre-review Checklist

-   [ ] Acceptance criteria satisfied and marked in related issue
-   [ ] Relevant `@unreleased` tags included in DocBlocks
-   [ ] Includes unit tests
-   [ ] Reviewed by the designer (if follows a design)
-   [ ] [Self Review](https://give.gitbook.io/development-manual/devops/github/code-reviews#self-review) of code and UX completed

[GIVE-2516]: https://stellarwp.atlassian.net/browse/GIVE-2516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GIVE-859]: https://stellarwp.atlassian.net/browse/GIVE-859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7893, opened 2025-04-24
Feature: Add migration to recalculate exchange rates

pauloiankoski:
Resolves [GIVE-2516]

## Description
This PR implements a migration to update exchange rates for existing donations where the exchange rate was not properly stored. This is a follow-up to the fix implemented in #7888, which addressed the issue of exchange rates not being stored properly in donations made through legacy forms.

## Affects
Donations' Exchange Rate

## Testing Instructions
1. Using a previous version of GiveWP, create donations using legacy forms with Currency Switcher enabled and a different currency selected.
2. In the admin area, navigate to the Donation Details page and verify that the Exchange Rate is displayed as "1".
3. Update GiveWP to this version.
4. In GiveWP > Tools > Data > Database upgrades, confirm that the migration completed successfully without any errors.
5. Return to the Donation Details page and verify that the Exchange Rate has been updated to reflect the approximate exchange rate from the time of the donation.

## Pre-review Checklist

-   [x] Includes unit tests

[GIVE-2516]: https://stellarwp.atlassian.net/browse/GIVE-2516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Donations/Migrations/RecalculateExchangeRate.php
+++ b/src/Donations/Migrations/RecalculateExchangeRate.php
@@ -51,18 +51,22 @@ class RecalculateExchangeRate extends BatchMigration
         // and either the exchange rate does not exist or is equal to 1
 
         return DB::table('posts', 'posts')
-            ->leftJoin('give_donationmeta as baseCurrencyMeta', 'posts.ID', 'baseCurrencyMeta.donation_id')
-            ->leftJoin('give_donationmeta as paymentCurrencyMeta', 'posts.ID', 'paymentCurrencyMeta.donation_id')
-            ->leftJoin('give_donationmeta as exchangeRateMeta', 'posts.ID', 'exchangeRateMeta.donation_id')
+            ->attachMeta(
+                'give_donationmeta',
+                'ID',
+                'donation_id',
+                [DonationMetaKeys::AMOUNT, 'amount'],
+                [DonationMetaKeys::CURRENCY, 'currency'],
+                [DonationMetaKeys::BASE_AMOUNT, 'baseAmount'],
+                [DonationMetaKeys::EXCHANGE_RATE, 'exchangeRate'],
+                ['_give_cs_base_currency', 'baseCurrency']
+            )
             ->where('posts.post_type', 'give_payment')
-            ->where('baseCurrencyMeta.meta_key', '_give_cs_base_currency')
-            ->whereIsNotNull('baseCurrencyMeta.meta_value')
-            ->where('paymentCurrencyMeta.meta_key', DonationMetaKeys::CURRENCY)
-            ->whereRaw('AND baseCurrencyMeta.meta_value != paymentCurrencyMeta.meta_value')
-            ->where('exchangeRateMeta.meta_key', DonationMetaKeys::EXCHANGE_RATE)
+            ->whereIsNotNull('give_donationmeta_attach_meta_baseAmount.meta_value')
+            ->whereRaw('AND give_donationmeta_attach_meta_baseCurrency.meta_value != give_donationmeta_attach_meta_currency.meta_value')
             ->where(function($query) {
-                $query->where('exchangeRateMeta.meta_value', '1')
-                    ->orWhereIsNull('exchangeRateMeta.meta_value');
+                $query->where('give_donationmeta_attach_meta_exchangeRate.meta_value', '1')
+                    ->orWhereIsNull('give_donationmeta_attach_meta_exchangeRate.meta_value');
             });
     }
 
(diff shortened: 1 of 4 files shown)


src/Donations/Migrations/RecalculateExchangeRate.php
```diff
@@ -0,0 +1,200 @@
...
+                    'donation_id',
+                    [DonationMetaKeys::AMOUNT, 'amount']
+                )
+                ->attachMeta(
+                    'give_donationmeta',
+                    'ID',
+                    'donation_id',
+                    [DonationMetaKeys::CURRENCY, 'currency']
+                )
+                ->attachMeta(
```
alaca: Check this https://github.com/impress-org/givewp/tree/develop/src/Framework/QueryBuilder#attachmeta

pauloiankoski: https://github.com/impress-org/givewp/pull/7893/commits/4b173db6432fff34488ac3ec1e20ca04f73aa15a

src/Donations/Migrations/RecalculateExchangeRate.php
```diff
@@ -0,0 +1,180 @@
...
+    {
+        return 100;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function hasMoreItemsToBatch($lastProcessedId): ?bool
+    {
+        return $this->query()
```
alaca, line 175: This is not very performant because we are using attachMeta in the query, and we don't need that here.

src/Donations/Migrations/RecalculateExchangeRate.php
```diff
@@ -0,0 +1,180 @@
...
+    {
+        return $this->query()->count();
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getBatchItemsAfter($lastId): ?array
+    {
+        $item = clone $this->query()
```
alaca, line 145: Also here, we don't need all that stuff from the query method; we need just the first id and the last id

alaca (changes requested):

src/Donations/Migrations/RecalculateExchangeRate.php
pauloiankoski: As the query method has been pretty much simplified, I don't think this will be a problem now.

1769236f0821a9a640d8de59e54bf0ca7db330fd

src/Donations/Migrations/RecalculateExchangeRate.php
```diff
@@ -0,0 +1,180 @@
...
+
+            $donations = $query->getAll();
+
+            foreach ($donations as $donation) {
+                try {
+                    $amount = Money::fromDecimal($donation->amount, $donation->currency);
+                    $baseAmount = Money::fromDecimal($donation->baseAmount, $donation->baseCurrency);
+
+                    if ($baseAmount->formatToDecimal() === '0') {
+                        throw new DatabaseMigrationException(
```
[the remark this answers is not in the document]
pauloiankoski: Replaced the error with a much simpler `continue` command.

1769236f0821a9a640d8de59e54bf0ca7db330fd

src/Donations/Migrations/RecalculateExchangeRate.php
pauloiankoski: `attachMeta` is now only present in the `runBatch` method.

1769236f0821a9a640d8de59e54bf0ca7db330fd

src/Donations/Migrations/RecalculateExchangeRate.php
```diff
@@ -0,0 +1,162 @@
...
+        return $this->query()->count();
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getBatchItemsAfter($lastId): ?array
+    {
+        $item = $this->query()
+            ->select('MIN(ID) AS first_id, MAX(ID) AS last_id')
```
alaca, line 129: This won't work. To get the min ID and max ID, you have to use a sub-select query to get the range of IDs, then you can select the min and max ID from that range.

Example:
```sql
SELECT MIN(ID) AS first_id, MAX(ID) AS last_id FROM (SELECT ID FROM table WHERE ID > someID ORDER BY ID ASC LIMIT batchSize)
```

alaca (changes requested):

src/Donations/Migrations/RecalculateExchangeRate.php
pauloiankoski: I don't think there's a problem there but lemme know what you think. Here is some tests I did before rewriting that query.

CleanShot 2025-04-25 at 10 40 35

CleanShot 2025-04-25 at 10 41 17

pauloiankoski: Here is the test using your SQL query:
CleanShot 2025-04-25 at 10 44 18

Same results.

alaca: What is the last ID in database? Without limit

pauloiankoski: I see what's the issue here. Limit is applied to the number of rows being returned, which in this case is just 1, so it does not applies to trim the range of IDs.

pauloiankoski: Updated on https://github.com/impress-org/givewp/pull/7893/commits/ccfb90da152727c9f4fb3ce25afaee2e2faa6962

src/Donations/Migrations/RecalculateExchangeRate.php
```diff
@@ -0,0 +1,163 @@
...
+     * @inheritDoc
+     */
+    public function getBatchItemsAfter($lastId): ?array
+    {
+        $subquerySQL = $this->query()
+            ->where('ID', $lastId, '>')
+            ->limit($this->getBatchSize())
+            ->getSQL();
+
+        $item = DB::get_row("SELECT MIN(ID) AS first_id, MAX(ID) AS last_id FROM ($subquerySQL) as batch");
```
alaca, line 133: Nice 👍

alaca (approved):

PR #7898, opened 2025-04-25
Feature: make campaign header sticky

jdghinson:
## Description
Currently, the campaign header scrolls with the page, which means when you scroll down to make changes, you’ll have to scroll all the way to the top to access the save changes button. This could lead to confusion, as users might not know where to save their changes. By making the header sticky, users will always have access to key actions like save changes, edit campaign page, and so on, regardless of how far they scroll down.

## Affects
This affects the header including the tabs in the campaign details page.

## Visuals
Before
https://www.loom.com/share/2c3a8991d86a4a2c876f3316b10fc041?sid=ae310217-5f92-46de-a151-323d855fda1d

After
https://www.loom.com/share/0a244e2218f54280a2bcf95e570563a2?sid=e8aa7764-f915-49dc-acd6-cb999d75d5ac

## Testing Instructions

Go to the campaigns section, scroll down, and check if the header remains at the top of the screen. Additionally, ensure that the header behaves the same way responsively.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Campaigns/resources/admin/components/CampaignDetailsPage/CampaignDetailsPage.module.scss
+++ b/src/Campaigns/resources/admin/components/CampaignDetailsPage/CampaignDetailsPage.module.scss
@@ -139,14 +139,15 @@ select[name="campaignId"] {
     display: none;
 }
 
-.pageHeader,
-.tabMenu{
+.pageHeader {
     position: sticky;
     top: var(--givewp-admin-bar-height);
     z-index: 2;
 }
 
 .tabMenu{
+    position: sticky;
+    z-index: 1;
     top: calc(var(--givewp-admin-bar-height) + var(--header-height));
 }
 
@@ -610,7 +611,7 @@ select[name="campaignId"] {
         display: flex;
         flex-direction: column;
         gap: var(--givewp-spacing-1);
-        z-index: 9999;
+        z-index: 3;
         padding: var(--givewp-spacing-1);
         top: 50px;
         width: 203px;


JasonTheAdams:
Nice, @jdghinson! Does the header remain sticky on mobile? Unless it takes up very little vertical space, I don't think it should.

src/Campaigns/resources/admin/components/CampaignDetailsPage/index.tsx
```diff
@@ -63,6 +65,37 @@ export default function CampaignsDetailsPage({campaignId}) {
...
 
+    useEffect(() => {
+        // Function to update header height
+        const updateHeaderHeight = () => {
+          if (headerRef.current) {
+            const height = headerRef.current.offsetHeight;
+            setHeaderHeight(height);
+            
+            // Update CSS variable directly
+            document.documentElement.style.setProperty('--header-height', `${height}px`);
```
jonwaldstein, line 76: very clever!!

jonwaldstein (commented):
@jdghinson this is cool! couple things though:

**1. On small screens the header is too far from the top of the page:**

Screenshot 2025-04-25 at 4 52 51 PM


**2. How would you feel about adding a little box-shadow to the bottom?**

Something like this:

```css
box-shadow: 1px 2px 4px rgba(0, 0, 0, 0.1);
```

Screenshot 2025-04-25 at 4 49 42 PM

3. I just realized the tabs do not live in the header, but I feel like they should.  I have a feeling there is a way to separate the tab list from the tab content - we might want to explore that to make the header styling easier to maintain https://react-spectrum.adobe.com/react-aria/Tabs.html

src/Campaigns/resources/admin/components/CampaignDetailsPage/index.tsx
jdghinson: Thanks @jonwaldstein 

1. I just realised the wp-header scrolls in the mobile view, that's why you have that offset. I'll fix that
2. I thought the border worked well, I'll try the shadow approach too.
3. Yes, I agree because that was tough to style, I had to find a way to make it work for now because I didn't want to do a lot of refactor.

jdghinson:
> Nice, @jdghinson! Does the header remain sticky on mobile? Unless it takes up very little vertical space, I don't think it should.

@JasonTheAdams It takes up little vertical space, we can still have the sticky on mobile.

jonwaldstein:
@jdghinson there's an issue with the hamburger menu, the dropdown items are behind the header

Screenshot 2025-05-09 at 11 23 29 AM

jonwaldstein (approved):
@jdghinson looks good, nice job man!

rickalday (approved):
Passed QA tests.

PR #7900, opened 2025-04-29
Feature: update forms query

alaca:
## Description

This PR adds eager loading to the donation forms list table. Also, the async data thing is removed from the list table.

## Affects

Donation forms list table

## Testing Instructions

Generate 10000 donations and see how fast the forms list table is loading.

## Pre-review Checklist

-   [ ] Includes unit tests


src/DonationForms/Adapter/Form.php
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * @inheritdoc
+     */
+    protected $properties = [
+        'id' => 'int',
+        'title' => 'string',
+        'createdAt' => DateTime::class,
+        'updatedAt' => DateTime::class,
+        'status' => DonationFormStatus::class,
+        'goalSettings' => GoalSettings::class,
```
jonwaldstein, line 33: @alaca honestly man, I think we can just modify the existing v2 donation form model to add this `goalSettings` param and whatever else you need here.  It's safe to do so because it's a ready only model and we're just adding more properties.

You already did the work here, just need to merge what you have from this adapter into the v2 form repository [prepareQuery](https://github.com/impress-org/givewp/blob/5690f034f0a8c3718d9d61ee4bb099beed44a67a/src/DonationForms/V2/Repositories/DonationFormsRepository.php#L36) & [DonationFormQueryData](https://github.com/impress-org/givewp/blob/e301dea8e5c3d11b03b2768e49cce711b459245f/src/DonationForms/V2/DataTransferObjects/DonationFormQueryData.php#L18) and we can just continue using the v2 form model with the added properties.

Then we don't have to have a 3rd form model 😅

src/DonationForms/Repositories/DonationFormDataRepository.php
```diff
@@ -0,0 +1,169 @@
...
+            ? $this->subscriptionAmounts
+            : $this->amounts;
+
+        foreach ($data as $row) {
+            if (isset($row['form_id']) && $row['form_id'] == $form->id) {
+                return (int)$row['sum'];
+            }
+        }
+
+        return 0;
```
jonwaldstein, line 67: Remember, we don't want to cast this into a int to avoid php rounding float values.

```suggestion
     * @return int|float
     */
    public function getRevenue(Form $form)
    {
        $data = $form->goalSettings->goalType->isSubscriptions()
            ? $this->subscriptionAmounts
            : $this->amounts;

        foreach ($data as $row) {
            if (isset($row['form_id']) && $row['form_id'] == $form->id) {
                return $row['sum'];
            }
        }

        return 0;
```

src/DonationForms/Repositories/DonationFormDataRepository.php
```diff
@@ -0,0 +1,169 @@
+<?php
+
+namespace Give\DonationForms\Repositories;
+
+
+use Give\DonationForms\DonationFormDataQuery;
+use Give\DonationForms\Adapter\Form;
+use Give\DonationForms\ValueObjects\GoalType;
+
+class DonationFormDataRepository
```
jonwaldstein, line 10: Let's add a Unit test for this, you can duplicate the one I started for campaigns and just swap it out for forms

https://github.com/impress-org/givewp/blob/ddf306c22d5def2f24dc11b7d2fd624cf7bd1498/tests/Unit/Campaigns/Repositories/CampaignsDataRepositoryTest.php#L19

src/DonationForms/DonationFormDataQuery.php
```diff
@@ -0,0 +1,126 @@
...
+use Give\Donations\ValueObjects\DonationMetaKeys;
+use Give\Framework\QueryBuilder\JoinQueryBuilder;
+use Give\Framework\QueryBuilder\QueryBuilder;
+
+/**
+ * Class used for loading the number of donors, donations and revenue amounts for multiple forms
+ *
+ * @unreleased
+ */
+class DonationFormDataQuery extends QueryBuilder
```
jonwaldstein, line 14: Let's add a Unit test for this, you can duplicate the one I started for Campaigns and replace with forms

https://github.com/impress-org/givewp/blob/ddf306c22d5def2f24dc11b7d2fd624cf7bd1498/tests/Unit/Campaigns/CampaignsDataQueryTest.php#L18

jonwaldstein (changes requested):

src/DonationForms/Adapter/Form.php
alaca: Yeah, you're right. Much easier to understand than having a new model. 
Resolved: https://github.com/impress-org/givewp/pull/7900/commits/0cc11416991da7c973f23ebd02a3ccef6b91f478

src/DonationForms/DonationFormDataQuery.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/7900/commits/a78d5c95d17e94202fb703d0f826eb03e9c8123d

src/DonationForms/Repositories/DonationFormDataRepository.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/7900/commits/a78d5c95d17e94202fb703d0f826eb03e9c8123d

src/DonationForms/DonationFormDataQuery.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/7900/commits/05058e84f07e38d9afd02929e8f42cdcc93d2faa

jonwaldstein (changes requested):

src/DonationForms/Repositories/DonationFormDataRepository.php
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * @unreleased
+     *
+     * Get goal data for form
+     *
+     * @param DonationForm $form
+     *
+     * @return array{actual: int, goal: int, actualFormatted: string, goalFormatted:string, percentage:float}
+     */
+    public function getGoalData(DonationForm $form): array
```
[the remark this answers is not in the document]
alaca: > Using this will also solve the formatting issue with Subscription amount raised

@jonwaldstein It will also significantly decrease performance because `DonationFormGoalData` will run queries for each goal column in the list table. That defies the purpose of this PR. We need a different approach if our goal is to improve performance here.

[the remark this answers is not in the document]
alaca: When would we cache those values, and where? We should probably eager-load campaign goals for forms that use campaign goals. I don't see any other way if we want to keep this performant. It will be slightly less performant than what we have here, but that's okay. It will still be better than doing expensive queries in each goal column.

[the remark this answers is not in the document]
jonwaldstein: Yeah, I was thinking we could check if the form goal type is campaign and use your `CampaignsDataRepository` instead

alaca:
@glaubersilva, how to remove async data loading from the donation forms list table?

glaubersilva:
@alaca You can do that by changing the default value of the `isAllStatsColumnsAsync()` method from the `Give\DonationForms\AsyncData\AdminFormListView\AdminFormListViewOptions` class to `false`.

**Reference:** https://github.com/impress-org/givewp/blob/6a4270a1b681c60e388480dbc5fc1b3a3e04e70c/src/DonationForms/AsyncData/AdminFormListView/AdminFormListViewOptions.php#L16

alaca:
@glaubersilva, just to be clear, this will disable async loading for the entire list table?

glaubersilva:
@alaca Yes, exactly! It disables the async load for all columns from both new and legacy list tables, but keeps the feature enabled for the form grid.

alaca:
@glaubersilva this disables the JS part, but what about the filters, they are still filtering the data. My question is, if I remove the filters from the ServiceProvider, will that also affect the FormGrid?

src/DonationForms/Repositories/DonationFormDataRepository.php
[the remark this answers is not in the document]
alaca: @jonwaldstein I got it working, but I'm not sure what we gained with this because I have to load a campaign for forms that are using campaign goal. Also, to access that data later, I need the campaign id prop on the form model. I don't like the current implementation of this part, but it's more performant than using another join just to get the campaign id, and sometimes we don't even need it. I'm reconsidering using the `DonationFormGoalData`.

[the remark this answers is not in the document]
jonwaldstein: Yeah there's a lot going on haha, but I think it's working pretty good as far as I can tell.  The good news is for list tables, it will only end up pulling in 1 campaign ID that all the forms share so that should be pretty performant too. 
 Campaign ID makes sense on a form model anyway so all good there.  

I just tested it out and did a quick fix for a discrepancy in amount formatting [here]([09b5804](https://github.com/impress-org/givewp/pull/7900/commits/09b58041e45325d3091ba6911d7abcd8f23cd691)) - otherwise, should be good to go with a little more testing...


Also noticed something on donation form view that I created a PR for:
https://github.com/impress-org/givewp/pull/7920

alaca:
@jonwaldstein I'm not sure why the test is failing. Could you take a look?

src/DonationForms/V2/DataTransferObjects/DonationFormQueryData.php
```diff
@@ -127,4 +152,83 @@ public function getDonationFormLevels($object): array
...
+                );
+
+                return GoalSettings::fromArray([
+                    'goalSource' => $settings->goalSource->getValue(),
+                    'enableDonationGoal' => $settings->enableDonationGoal,
+                    'goalType' => $goalType,
+                    'goalAmount' => $goalType->isOneOf(GoalType::AMOUNT(), GoalType::AMOUNT_FROM_SUBSCRIPTIONS())
+                        ? Money::fromDecimal($campaign->goal, $currency)->formatToDecimal()
+                        : $campaign->goal,
+                ]);
```
jonwaldstein, line 185: Something's not right about this for v3 forms as I change goal types on campaign.  I think this part might be the culprit as we should be able to use form settings directly instead of trying to convert them. 

```suggestion
                return GoalSettings::fromArray([
                    'goalSource' => $settings->goalSource->getValue(),
                    'enableDonationGoal' => $settings->enableDonationGoal,
                    'goalType' => $settings->goalType,
                    'goalAmount' => $settings->goalAmount,
                ]);
```

jonwaldstein: Ah yeah, you got it working in the DonationFormGoalData somehow.  If you set a form goal to inherit campaign and change the campaign settings, the form list should match here:

Screenshot 2025-05-13 at 12 33 22 PM

alaca: @jonwaldstein resolved https://github.com/impress-org/givewp/pull/7900/commits/acb7926ace988390b237b43b59f8511bd3d4bfd8

jonwaldstein:
@alaca this is looking good now except the form grid goal is stuck in loading state

Screenshot 2025-05-14 at 8 50 57 AM

alaca:
@jonwaldstein resolved https://github.com/impress-org/givewp/pull/7900/commits/943d3e3e743c5cb506bbebed6a6f5a952d406e36

jonwaldstein (approved):
@alaca great work man! as we discussed we'll need to open a separate ticket to explore form grid more and potentially implement this eager loading there as well.

rickalday (approved):
Passed QA tests

PR #7901, opened 2025-04-30
Fix: forms to campaigns migration duplicate entry error and change query that retrieves V2 upgraded forms

glaubersilva:
Resolves [GIVE-2529]

## Description

This PR fixes the duplicated entry exception that is happening in some cases where the `migrate_forms_to_campaign_forms` try to add the same data in the `give_campaign_forms` table.

It also updates the query from the `getUpgradedV2FormsData()` method to prevent the association of the same `formId` with multiple campaigns. For cases where the user tried to upgrade the same form multiple times, it also makes sure to associate only the last form upgraded to the campaign.

Beyond that, this PR also added these 3 new test cases for this migration:

- `testUpgradedFormsMoreThanOnceAreNotAssociatedWithMultipleCampaigns()`
- `testUpgradedFormsMoreThanOnceBelongsToSameCampaignFromLastUpgradeAttempt()`
- `testMigrationReRunsWithoutErrors()`

## Affects

The `migrate_forms_to_campaign_forms` migration.

## Visuals

image

## Testing Instructions

**For developers, run the following command in your terminal:** 

`composer test -- --filter MigrateFormsToCampaignFormsTest`

**For QA:**

Test 1:

1. In a fresh install using a GiveWP version before the 4.0 release (without campaigns), create a few V2 and V3 forms;
2. Migrate some V2 forms to V3 forms;
3. Install the Give version of this zip/branch;
4. The migration should convert V2 forms to Campaigns;
5. The migration should associate V2 upgraded forms with Campaigns as non-default forms.

Check this video for better context: https://www.loom.com/share/7bf0d4727f944c0da29fd0df9219d7bc?sid=2b9f341f-c61a-4f13-b088-b8600b09afab

Test 2:

1. In a fresh install using a GiveWP version before the 4.0 release (without campaigns), create a few V2 and V3 forms;
2. Create a V2 form and start migrating it to a V3 form and abort it by **trashing** the new V3 version;
3. Restart the migration for the V2 form from the last step to a new V3 form and finish it now.
4. Install the Give version of this zip/branch;
5. The migration should convert V2 forms to Campaigns;
6. Make sure the V3 form created/migrated in step 3 was converted to a campaign, and the V2 form created in step 2 was associated with this same campaign as a non-default form.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2529]: https://stellarwp.atlassian.net/browse/GIVE-2529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
```diff
@@ -179,15 +197,20 @@ protected function addUpgradedV2FormToCampaign($data): void
...
     {
-        DB::table('give_campaign_forms')
-            ->insert([
-                'form_id' => $formId,
-                'campaign_id' => $campaignId,
-            ]);
+        $table = DB::prefix('give_campaign_forms');
+
+        DB::query(
+            DB::prepare("INSERT INTO {$table} (form_id, campaign_id) VALUES (%d, %d) ON DUPLICATE KEY UPDATE form_id = form_id",
```
alaca, line 208: Hey Glauber, I believe we shouldn't do this. This is a clear indication that our query or logic for migrating forms is wrong.

jonwaldstein:
@glaubersilva I snuck this in here [d4dffab](https://github.com/impress-org/givewp/pull/7901/commits/d4dffab115af7ae3344f04153838a339bc7a1ad6) 😄 

Screenshot 2025-05-02 at 12 16 26 PM

src/Campaigns/Migrations/MigrateFormsToCampaignForms.php
glaubersilva: This was my initial idea of a "just in case" solution for situations where duplicated data is added to this table by accident. On the other hand, it can also prevent us from discovering new problems with those queries if they happen. So, I removed it and rolled back the `whereNotIn` statement in the query of the `getUpgradedV2FormsData()` method to ensure we are not retrieving data already added to the `give_campaign_forms` table previously.

Resolved in f387ca46537cbd2bfe71ca5a136bfea4456ccc0f

glaubersilva:
@alaca This is ready for re-review.

alaca (changes requested):
Okay, there are many confusing things in this migration, and it's unclear what is going on. Also, code comments do not reflect what is happening. The major problem here is that we are excluding v2 forms, but we are including all v3 forms that are created from that v2 form, which is wrong. 

The correct flow would be:
1. Get all v2 and v3 forms, excluding upgraded forms (forms that have `migratedFormId` set), and create campaigns from them
2. Get all upgraded forms (forms with migratedFormId set) and assign them to a campaign (use `migratedFormId` to find campaign ID)

Regarding errors that happen when a user is running this migration twice, that is how it should be. This migration should not be run twice under any circumstances. We should probably add a description saying that if you run this twice, you will break your site. 


The term "upgraded forms" I'm using here is a bit misleading. What I mean by upgraded form is the v3 form that is created when the user upgrades the v2 form

glaubersilva:
@alaca, I didn't understand your suggestions for changes in the last comment.

> Okay, there are many confusing things in this migration, and it's unclear what is going on.

What exactly is confusing you?

> Also, code comments do not reflect what is happening. 

Can you point out what's wrong and **why** it's wrong?

> The correct flow would be...

The workflow you describe is exactly how the migration is working right now. 

The `array_map([$this, 'createCampaignForForm'], $this->getAllFormsData());` line is doing what you suggest on the step 1 and the `array_map([$this, 'addUpgradedV2FormToCampaign'], $this->getUpgradedV2FormsData());` line is doing what you suggested on the step 2. 

Also, there is a demo video attached to the PR description (in the test instructions section) demonstrating it's working.

Do you find anything not working as expected in the current workflow? Can you share the steps to replicate?

> Regarding errors that happen when a user is running this migration twice, that is how it should be. This migration should not be run twice under any circumstances. We should probably add a description saying that if you run this twice, you will break your site.

In the ideal scenario, yes, it should not be running twice. But an unexpected exception can happen, like in the cases we have on the support right now. So, people will try to rerun it to see the success status for the migration. When rerun it the site will not break because the `getAllFormsData()` will not generate any error since it is already excluding forms associated with a campaign previously. This PR is doing the same for the `getUpgradedV2FormsData()` to prevent the duplicate entry error.

If we just add the message as you are suggesting, what should we say to the customers and support team for these cases where the migration failed?

alaca:
> What exactly is confusing you?

https://github.com/impress-org/givewp/blob/8a6d2e0fbe77304cbd3ca947ff9c57a7cf66a4ff/src/Campaigns/Migrations/MigrateFormsToCampaignForms.php#L144

What is this, and why do we need this? I don't understand the logic behind this. Can you explain it to me?

> Can you point out what's wrong and why it's wrong?

This comment https://github.com/impress-org/givewp/blob/8a6d2e0fbe77304cbd3ca947ff9c57a7cf66a4ff/src/Campaigns/Migrations/MigrateFormsToCampaignForms.php#L136

When the user deletes the upgraded form, all the form meta is deleted with it. Yes, that includes the `migratedFormId`.
Duplicate rows are not the result of some meta key leftover, they are the result of the query itself. 


So for the second part, we should keep things simple and understandable. I'll comment on the actual code.

glaubersilva:
@alaca Here is a video demonstrating how we can have multiple `migratedFormId` originated from the same source form (v2), which explains why we need the inner join statement: https://www.loom.com/share/d2b7687a5231411aa33f01579f3dc39f

I can see that my comment can generate confusion because I said "deleted" instead of "trashed"

alaca (approved):
Great job, @glaubersilva

rickalday (approved):
Passed QA tests.

PR #7929, opened 2025-05-15
If they are a donor and logged in give them access to donor dash

kadencewp:
I ran into this issue during testing with Give memberships. If a user signs up for a membership, their role can be different from a Give donor role. I'm not entirely sure I understand why there is a check for role, but I'm hoping a solution like this PR or something like it can be found for users with other custom roles, or even a subscriber role can view their donor dashboard if they've donated.


jonwaldstein (approved):
makes sense, thanks!

PR #7936, opened 2025-05-20
Feature: Add header and tabs to Donor Details page

pauloiankoski:
Resolves [GIVE-2583]

## Description
This pull request implements the initial scaffold for the Donor Details page, including Header and the tab navigation system.

## Affects
Donor Details page

## Visuals
https://github.com/user-attachments/assets/0bdfd10b-af78-4beb-bcc6-e733da82080e

## Pre-review Checklist

-   [ ] Includes unit tests


src/Donors/resources/admin/components/Notifications/Notification.tsx
```diff
@@ -0,0 +1,74 @@
+import {useEffect} from '@wordpress/element';
```
alaca, line 1: Hey Paulo, why do we need to duplicate this? Do we have to do this for every admin screen? I don't think so. We should extract the shared logic so that it can be used on every screen. We don't have to reinvent the wheel each time we add a new screen.

alaca (changes requested):
Instead of duplicating the code from campaigns and making it work explicitly for Donors, we should extract shared components, hooks, helper functions, etc., and make them reusable on every screen. What you have here is not doing that. With this approach, every new screen will require a new set of everything you have here, which is not something we would like to maintain.

src/Donors/resources/admin/components/Notifications/Notification.tsx
pauloiankoski: @alaca Refactored this and created a generic `AdminDetailsPage` component that hold the main shared logic while still allowing customizations when needed.

CleanShot 2025-05-22 at 16 27 00

src/Donors/resources/admin/components/DonorDetailsPage/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+            <span className={styles.statusBadgeText}>{statusMap[status]}</span>
+        </div>
+    );
+};
+
+export default function DonorDetailsPage({ donorId }) {
+    const { adminUrl } = getDonorOptionsWindowData();
+    const [showConfirmationDialog, setShowConfirmationDialog] = useState<boolean>(false);
+
+    const useObjectEntityRecord = (donorId: number) => {
```
alaca, line 49: Why do we need another wrapper?

src/Donors/resources/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+    return apiFetch({
+        url: window.GiveDonorOptions.adminUrl + '/admin-ajax.php?action=' + id,
+        method: 'POST',
+    })
+}
+
+/**
+ * @unreleased
+ */
+export function amountFormatter(currency: Intl.NumberFormatOptions['currency'], options?: Intl.NumberFormatOptions): Intl.NumberFormat {
```
alaca, line 48: Do we need this helper function in almost every domain? Can we extract this and put it in one place so that we don't have to write it multiple times.

src/Views/Components/AdminDetailsPage/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+            if (showContextMenu) {
+                return;
+            }
+
+            if (
+                e.target instanceof HTMLElement &&
+                !e.target.closest(`.${styles.donorButtonDots}`) &&
+                !e.target.closest(`.${styles.contextMenu}`)
+            ) {
+                setShowContextMenu(false);
```
alaca, line 86: We shouldn't be using `querySelector`, it's inefficient and completely opposite of the declarative way of doing things in React. I'd suggest using `useRef`.

alaca (commented):
Hey, man. Good job. I have a couple of questions and one suggestion. Otherwise, looks good.

src/Donors/resources/utils.ts
pauloiankoski: Yeah, absolutely. Here it is: 6c7d0dbc22eecc4978c13a0a15d90738e56d2f62

src/Donors/resources/admin/components/DonorDetailsPage/index.tsx
alaca: Yeah, I was a bit confused why we need a wrapper. I would go with record, but if you think this makes sense for semantic reasons, then ok.

alaca (approved):
This was a huge PR 😅
Looks good.

PR #7944, opened 2025-05-29
Feature: add donor notes backend

glaubersilva:
Resolves [GIVE-2592]

## Description

This PR implements all necessary endpoints to handle donor notes on the front end.

Also, it adds Repository and Model classes for donor notes.

To keep consistency, we used in the `DonorNoteType` the same values used in `DonationNoteType` value object.

So we have the following options:

`DonorNoteType::ADMIN` - should be used for **private** notes (available only in the WP admin area)

`DonorNoteType::DONOR` - should be used for **public** notes (the donor will be notified through email-  not implemented on this PR) 

## Affects

The `givewp/v3` endpoints

## Visuals

image

image

## Testing Instructions

Run the tests on your terminal or install the [WP OpenAPI plugin](https://wordpress.org/plugins/wp-openapi/) to play around with the endpoints as demonstrated in the attached screenshots in the _Visuals_ section.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2592]: https://stellarwp.atlassian.net/browse/GIVE-2592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/API/REST/V3/Routes/Donors/DonorNotesController.php
```diff
@@ -0,0 +1,473 @@
...
+            [
+                'methods' => WP_REST_Server::CREATABLE,
+                'callback' => [$this, 'create_item'],
+                'permission_callback' => [$this, 'create_item_permissions_check'],
+                'args' => $this->get_endpoint_args_for_item_schema(WP_REST_Server::CREATABLE),
+                'schema' => [$this, 'get_public_item_schema'],
+            ],
+        ]);
+
+        register_rest_route($this->namespace, '/' . $this->rest_base . '/(?P<donorId>[\d]+)/notes/(?P<id>[\d]+)', [
```
JasonTheAdams, line 58: One thing to be aware of is that, as of yet, I've not been able to get entities to work nicely with nested resources like this. I did some digging into Gutenberg and they never built a system for it — the closest being revisions, but they decided (boo!) to make a specific solution for revisions rather than a generic one. The only way I got it to work for me is by having the base resource ids static in the `baseURL` when the entity is registered:
```php
dispatch( coreStore ).addEntities( [
	{
		name,
		kind,
		baseURL: `/partner-gateway/v1/sites/${$siteId}/zones/${$zoneId}/records`,
		plural: 'records',
		label: 'Record',
		supportsPagination: false,
	}
] );
```

If that's a problem for y'all, you'll probably need to make a `donor-notes` root resource that has a required `donor_id` parameter.

glaubersilva: @alaca Let me know if 👆 it will work on your end. If not, I can refactor it.

PR #7959, opened 2025-06-06
Feature: Private notes

alaca:
## Description

This PR adds private notes functionality to donor details screen.
## Visuals

https://github.com/user-attachments/assets/f36f0414-3a41-4443-8260-e0d985e69dd4

## Testing Instructions

Go to edit donor screen
Try to add, edit and delete note.
Try show more button (requires at least 6 notes)

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Views/Components/AdminDetailsPage/store/index.ts
+++ b/src/Views/Components/AdminDetailsPage/store/index.ts
@@ -17,7 +17,7 @@ export const store = createReduxStore('givewp/admin-details-page-notifications',
             case 'ADD_NOTIFICATION':
                 const notificationExist = state.filter((notification: { id: string }) => notification.id === action.notification.id);
                 if (!notificationExist.length) {
-                    state.push(action.notification);
+                    return [...state, action.notification];
                 }
                 return state;
 


src/Admin/components/PrivateNotes/index.tsx
```diff
@@ -0,0 +1,350 @@
...
+                await mutate(response);
+                dispatch.addSnackbarNotice({
+                    id: 'delete-note',
+                    content: __('Private note deleted successfully', 'give'),
+                });
+            });
+    };
+
+    const editNote = (id: number, content: string) => {
+        apiFetch({path: `/givewp/v3/donors/${donorId}/notes/${id}`, method: 'POST', data: {id, content}})
```
pauloiankoski: We don’t need to send the `id` again in the request body since it’s already part of the URL. Also, I think it might be better to use `PATCH` instead of `POST`, since we’re only updating part of the note. That would follow REST conventions more closely.

src/Admin/components/PrivateNotes/index.tsx
```diff
@@ -0,0 +1,350 @@
...
+                                };
+                            });
+
+                            await mutate(endpoint);
+                        }}>
+                        {__('Show more', 'give')}
+                    </button>
+                )}
+            </div>
+            <NotificationPlaceholder type="snackbar" />
```
pauloiankoski: I think we might not need this here — the `AdminDetailsPage` wrapper already includes the `NotificationPlaceholder` component, so we’re probably rendering it twice.

src/Admin/components/PrivateNotes/index.tsx
```diff
@@ -0,0 +1,350 @@
...
+            >
+                {currentlyEditing ? (
+                    <>
+                        <div className={style.addNoteContainer}>
+                        <textarea
+                            className={style.textarea}
+                            onChange={(e) => setContent(e.target.value)}
+                        >
+                            {note.content}
+                        </textarea>
```
pauloiankoski: The `textarea` value is set as children instead of using the value prop, making it an uncontrolled component.

src/Admin/components/PrivateNotes/index.tsx
```diff
@@ -0,0 +1,350 @@
...
+            )}
+            {data?.data?.length ? (
+                <>
+                    {data.data.map((note) => {
+                        return (
+                            <Note
+                                note={note}
+                                onDelete={(id: number) => deleteNote(id)}
+                                onEdit={(id: number, content: string) => editNote(id, content)}
+                            />
```
pauloiankoski: Missing `key` prop in the loop.

src/Admin/components/PrivateNotes/index.tsx
```diff
@@ -0,0 +1,350 @@
...
+        const data = await response.json();
+        return {
+            data,
+            totalPages: Number(response.headers.get('X-WP-TotalPages')),
+            totalItems: Number(response.headers.get('X-WP-Total')),
+        };
+    }, {revalidateOnFocus: false});
+
+    const saveNote = () => {
+        setState({isSavingNote: true});
```
pauloiankoski, line 64: I can’t find any implementation using `isSavingNote`. Also, it’s never being reset — once we save a note, it stays `true` forever. By the way, it would be great to show some loading state in the UI so users know the action is being processed.

src/Donors/resources/admin/components/DonorDetailsPage/Tabs/Overview.tsx
```diff
@@ -187,10 +188,13 @@ export default function DonorDetailsPageOverviewTab() {
...
                         subtitle={__('This note will be seen by only admins', 'give')}
-                        actionOnClick={() => {}}
+                        actionOnClick={() => setIsAddingNote(true)}
                         actionText={__('Add note', 'give')}
                     />
-                    <PrivateNote />
+                    <PrivateNotes
+                        donorId={donorId}
+                        context={[isAddingNote, setIsAddingNote]}
+                    />
```
pauloiankoski: Just a thought here: we’re creating a piece of state there, but it’s only being used inside the `PrivateNotes` component. That kind of breaks encapsulation and makes `PrivateNotes` dependent on its parent to function properly.

Maybe it would make sense to move the `Header` component (and the state) into `PrivateNotes` so everything stays self-contained? That way, the state logic would live where it’s actually needed, and PrivateNotes could be reused more easily elsewhere.

pauloiankoski (changes requested):
Overall, this looks good! However, there are a few areas that need attention before this can be merged.

src/Donors/resources/admin/components/DonorDetailsPage/Tabs/Overview.tsx
alaca: @pauloiankoski I agree with you, but the thing is, we are triggering adding note functionality outside the private notes component. That's why I need both state and state handler inside private notes component. I mean, I can change this to use two attributes instead of just context.

src/Admin/components/PrivateNotes/index.tsx
alaca: Yeah, I forgot to move this to the Overview component

src/Admin/components/PrivateNotes/index.tsx
alaca: Resolved: https://github.com/impress-org/givewp/pull/7959/commits/0a0185823b5f3d1f9dc5b362b948796b56e909bd

src/Admin/components/PrivateNotes/index.tsx
alaca: Resolved: https://github.com/impress-org/givewp/pull/7959/commits/0a0185823b5f3d1f9dc5b362b948796b56e909bd

src/Admin/components/PrivateNotes/index.tsx
alaca: Resolved: https://github.com/impress-org/givewp/pull/7959/commits/0a0185823b5f3d1f9dc5b362b948796b56e909bd

src/Admin/components/PrivateNotes/index.tsx
alaca: Resolved: https://github.com/impress-org/givewp/pull/7959/commits/0a0185823b5f3d1f9dc5b362b948796b56e909bd

src/Admin/components/PrivateNotes/index.tsx
pauloiankoski: We actually don't need to add this component anywhere here. It's already being entirely provided and managed by the `AdminDetailsPage`.

src/Donors/resources/admin/components/DonorDetailsPage/Tabs/Overview.tsx
pauloiankoski: My suggestion was to do a full-encapsulation into the PrivateNotes component so that it contains everything it needs to work (aka move Header and state into PrivateNotes)

src/Admin/components/PrivateNotes/index.tsx
pauloiankoski: 👀

src/Donors/resources/admin/components/DonorDetailsPage/Tabs/Overview.tsx
alaca: Resolved: https://github.com/impress-org/givewp/pull/7959/commits/1b8f39aeb820c30fdb7af9a5af2628f370dae389

src/Admin/components/PrivateNotes/index.tsx
alaca: Resolved https://github.com/impress-org/givewp/pull/7959/commits/1b8f39aeb820c30fdb7af9a5af2628f370dae389

jonwaldstein:
@pauloiankoski please help us fix that 😄

pauloiankoski (approved):
@alaca I just applied the fix for that store issue so this PR can be merged.

PR #7961, opened 2025-06-06
fetaure/add skeleton loader to campaign loading

jdghinson:
## Description
This was a fun friday project for campaigns where I added a skeleton loader to the campaign overview page and the overview stats when you are filtering by the day range.

## Affects
The campaign overview page and the overview stats card

## Visuals
Before
Uploading Old Loading.mov…

After
https://github.com/user-attachments/assets/7454f94b-31bd-4977-ab51-0449bc1971e0

## Testing Instructions
Create or open an existing campaign, you will see the skeleton loader before the data shows. Same applies to when you are filtering by the day range, you should see the skeleton loader before the number shows up.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats/SkeletonLoader.tsx
+++ b/src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats/SkeletonLoader.tsx
@@ -1,0 +1,49 @@
+import styles from './styles.module.scss';
+
+const SkeletonLoader = () => {
+    return (
+        <>
+            <div className={styles.dateRangeFilter}>
+                {[1, 2, 3, 4, 5].map((i) => (
+                    <div key={i} className={styles.skeletonButton} />
+                ))}
+            </div>
+            <div className={styles.mainGrid}>
+                {/* Stat Widgets */}
+                {[1, 2, 3].map((i) => (
+                    <div key={i} className={styles.statWidget}>
+                        <div className={styles.skeletonHeader} />
+                        <div className={styles.skeletonValue} />
+                        <div className={styles.skeletonFooter} />
+                    </div>
+                ))}
+                
+                {/* Revenue Widget */}
+                <div className={styles.revenueWidget}>
+                    <div className={styles.headerSpacing}>
+                        <div className={styles.skeletonHeader} />
+                        <div className={styles.skeletonSubHeader} />
+                    </div>
+                    <div className={styles.skeletonChart} />
+                </div>
+
+                {/* Nested Grid */}
+                <div className={styles.nestedGrid}>
+                    <div className={styles.progressWidget}>
+                        <div className={styles.headerSpacing}>
+                            <div className={styles.skeletonHeader} />
+                            <div className={styles.skeletonSubHeader} />
+                        </div>
+                        <div className={styles.skeletonChart} />
+                    </div>
+                    <div className={styles.statWidget}>
+                        <div className={styles.skeletonHeader} />
+                        <div className={styles.skeletonValue} />
+                    </div>
+                </div>
+            </div>
+        </>
+    );
+};
+
+export default SkeletonLoader; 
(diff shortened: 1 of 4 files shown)


src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats/index.tsx
```diff
@@ -29,20 +30,23 @@ export const filterOptions = [
...
-        setLoading(true);
+    setLoading(true);
 
-        try {
-            const response = await apiFetch({
-                path: addQueryArgs(`/givewp/v3/campaigns/${campaignId}/statistics`, {rangeInDays: days}),
-            });
+    try {
+        // Add artificial delay of 1.5 seconds
+        await new Promise(resolve => setTimeout(resolve, 300));
```
jonwaldstein: I'm assuming this was for testing purposes, we can't have this in production 😏

src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats/index.tsx
```diff
@@ -54,12 +58,21 @@ const CampaignStats = () => {
      */
     const [stats, setStats] = useState<CampaignOverViewStat[]>([]);
     const {campaign} = useCampaignEntityRecord();
-    const [loading, setLoading] = useState<boolean>(false);
+    const [loading, setLoading] = useState<boolean>(true);
+    const [isInitialLoad, setIsInitialLoad] = useState<boolean>(true);
```
jonwaldstein, line 59: What's this for?

src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats/SkeletonLoader.tsx
```diff
@@ -0,0 +1,49 @@
+import styles from './styles.module.scss';
+
+const SkeletonLoader = () => {
```
jonwaldstein, line 3: Is this for the whole campaign overview page?

jonwaldstein (changes requested):

src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats/index.tsx
jdghinson: Yeah, thanks!

src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats/index.tsx
jdghinson: The "loading" is for the showing the skeletons while the data is being fetched for the overview, and the "isInitialLoad" is to track whether it's the first load or not, to avoid the full skeleton loader when you are changing the date range.

src/Campaigns/resources/admin/components/CampaignDetailsPage/Components/CampaignStats/SkeletonLoader.tsx
jdghinson: Yes

jonwaldstein:
@jdghinson I'm liking this.

Couple things though
1. We have this initial spinner, not sure if we need both? Screenshot 2025-06-11 at 1 35 52 PM
2. The graph is a little jarring after the skeleton loader, it takes a moment to find it's place

jonwaldstein:
@jdghinson couple more thoughts 😄 


1. It looks like on page load, the skeleton perhaps does not account for the height of the header.  So the screen kind of shifts up and then down once the header comes in.

https://github.com/user-attachments/assets/67022ea0-48c5-44c2-b5c8-459e505c278a

2. The stat widgets are shifting their height when the skeleton is added because of the flexbox alignment being baseline with the percentage change.  One option is to make that alignment center

https://github.com/user-attachments/assets/3fe50f44-bd7f-4cbc-9840-6d15f2b49e4a

jdghinson:
@jonwaldstein Fixed the issues
1. I added a skeleton loader for the header considering the height
2. Changed the alignment to center

jonwaldstein:
@jdghinson thanks man, it's looking better now!  I would still like to get your thoughts @pauloiankoski as you are really leading the charge on creating frameworks for our admin screens 😄 - do you like this direction, does it make sense for other screens and if so is there a way we can unify the experience?

JasonTheAdams:
Fancy skeleton loaders! I think  it's nice on the initial page load. I'm not as much of a fan for taking existing content, putting a skeleton in place, and then putting the content back for a couple reasons:
1. It's hard not to jog the adjacent content and cause shifting
2. It's a bit jarring to have content go from text to a block and back to text

For this, I'd do something different, one of:
* No loader. The text just replaces after load. 🤷‍♂️ 
* Use a library like https://number-flow.barvian.me for a slick post-loading animation
* Put a simple spinner or something next to the text

PR #7963, opened 2025-06-07
Fix: ensure useFormData is watching only the address fields that live in the billingAddress block

glaubersilva:
Resolves [GIVE-2626]

## Description

This PR fixes a bug that was preventing the PayPal modal from opening when a custom field with the meta key "country" was added to a v3 form.

Why does it work?

Because our v3 forms will throw the following error when trying to save a field with the same meta key name:

> ERROR: the form was not saved due to a meta key name conflict. A field already exists on this form with the meta key 'country'. Meta key names must be unique. Change the conflicting meta key and try to save again.

So it ensures that if we have the billingAddress block added to the form, only the country field of this block can be present in the same form.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2626]: https://stellarwp.atlassian.net/browse/GIVE-2626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DonationForms/resources/app/hooks/useFormData.ts
+++ b/src/DonationForms/resources/app/hooks/useFormData.ts
@@ -1,6 +1,7 @@
 import type {DonationTotals} from '@givewp/forms/app/store/donation-summary';
 import {useDonationSummaryContext} from '@givewp/forms/app/store/donation-summary';
 import type {subscriptionPeriod} from '@givewp/forms/registrars/templates/groups/DonationAmount/subscriptionPeriod';
+import { useDonationFormState } from '@givewp/forms/app/store';
 
 /**
  * Zero decimal currencies are currencies that do not have a minor unit.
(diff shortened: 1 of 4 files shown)


jonwaldstein:
@glaubersilva I have a simpler suggestion for this:

1. Add the billing address fields (country, zip, etc.) to the [disallowed list](https://github.com/impress-org/givewp/blob/7dc34307ecf9c80f5cdcae46977745a6dcef23fc/src/FormBuilder/ViewModels/FormBuilderViewModel.php#L367) of custom field names.  This will ensure we don't get a conflicting unique field name.  Someone can still add a country field but it will be like `country_1` instead of our `country` field.
2. (optionally) Sanitize the country field value to ensure it matches the 2 digit code syntax we are expecting.  This would just be a safe-guard if we feel it's necessary.

glaubersilva:
@jonwaldstein  Thanks for this suggestion, much simpler! 

Even because I was getting crazy trying to understand why the `dispatch` method is not working inside the billingAddress export.

But I think we can revisit it later because it would be nice to have a way to add references to the global context.


However, for now, it works great, so I'll close this PR.

Closing in favor of https://github.com/impress-org/givewp/pull/7965

PR #7982, opened 2025-06-17
refactor: Optimize donor dashboard queries

jfn4th:
## Description

This aims to greatly improve donor dashboard load times, particularly for sites with very large donation histories. A site we maintain with 130k donations and 6.3 million rows of donation meta was seeing timeouts when hitting `/wp-json/give-api/v2/donor-dashboard/donations`, and locally (M1 Pro Macbook Pro) was taking 30-40 seconds to resolve. This was due to `getDonationIds` and calls to `getDonationAggregate` each taking 5-8 seconds to run. Here, we remove an (as far as I am aware) unnecessary join to `wp_give_revenue` in `getDonationIds` and cache the result so we can reuse the IDs. Then consolidate the aggregate queries into a single query, taking advantage of the ID cache to remove the need for expensive joins.  Altogether, this reduces the response time from 30-40 seconds to 1-2 seconds on my machine.

Also replaces the usage of the deprecated `Give\ValueObjects\Money` class with `Give\Framework\Support\ValueObjects\Money` updates method typedocs to reflect possibility of null `count`, `revenue`, and `average`, and removes some stray whitespace characters.

## Affects

This affects the internal queries used in the `/wp-json/give-api/v2/donor-dashboard/donations` donor dashboard endpoint.

## Testing Instructions

1. Load the donor dashboard before applying these changes and measure response time to `/wp-json/give-api/v2/donor-dashboard/donations`
2. Do the same after applying changes.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/give.php
+++ b/give.php
@@ -6,8 +6,8 @@
  * Description: The most robust, flexible, and intuitive way to accept donations on WordPress.
  * Author: GiveWP
  * Author URI: https://givewp.com/
- * Version: 4.3.2
- * Requires at least: 6.5
+ * Version: 4.4.0
+ * Requires at least: 6.6
  * Requires PHP: 7.4
  * Text Domain: give
  * Domain Path: /languages
@@ -251,6 +251,7 @@ final class Give
         Give\FeatureFlags\OptionBasedFormEditor\ServiceProvider::class,
         Give\ThirdPartySupport\ServiceProvider::class,
         Give\API\REST\V3\Routes\ServiceProvider::class,
+        Give\Framework\PaymentGateways\ServiceProvider::class,
     ];
 
     /**
@@ -421,7 +422,7 @@ final class Give
     {
         // Plugin version.
         if (!defined('GIVE_VERSION')) {
-            define('GIVE_VERSION', '4.3.2');
+            define('GIVE_VERSION', '4.4.0');
         }
 
         // Plugin Root File.
--- a/readme.txt
+++ b/readme.txt
@@ -2,10 +2,10 @@
 Contributors: givewp, dlocc, webdevmattcrom, ravinderk, mehul0810, kevinwhoffman, jason_the_adams, henryholtgeerts, kbjohnson90, alaca, benmeredithgmailcom, jonwaldstein, joshuadinh, glaubersilvawp, pauloiankoski
 Donate link: https://go.givewp.com/home
 Tags: donation, donate, recurring donations, fundraising, crowdfunding
-Requires at least: 6.5
+Requires at least: 6.6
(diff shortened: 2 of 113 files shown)


jonwaldstein:
@jfn4th thanks for the PR! 🙌   

I'll admit, our donor dashboard queries are a bit outdated with some of our current performance updates we're making so this is a welcome update 😄.

We're a small team and currently working on some big projects but i'll try to prioritize reviewing this within a week or so.

jfn4th:
Thanks for the info! I realized I missed that `getDonationAggregate` filtered by a narrower post status than `getDonationIDs` did, so I've updated the PR to include that and DRY up the `getDonationAggregate` and my new `getDonationsSummary` methods since they ended up so similar.

src/DonorDashboards/Repositories/Donations.php
```diff
@@ -67,12 +75,76 @@ public function getAverageRevenue($donorId)
 
         return $aggregate ?
             $this->getAmountWithSeparators(
-                Money::ofMinor($aggregate->result, $currencyCode)->getAmount(),
+                (new Money((int) round((float) $aggregate->result), $currencyCode))->formatToDecimal(),
```
alaca: The amount will be incorrect if we cast it to an `int`.

src/DonorDashboards/Repositories/Donations.php
```diff
@@ -83,20 +155,42 @@ public function getAverageRevenue($donorId)
...
+     * Query donation revenue data from the revenue table for a specific donor
+     *
+     * @unreleased
+     *
+     * @param int    $donorId
+     * @param string $selectClause
+     *
+     * @return object|null
+     */
+    private function queryDonationRevenueData($donorId, $selectClause)
```
alaca: The idea of passing SQL as a parameter makes me anxious 😅
I'd suggest refactoring this method to something like `getRevenueDonationDataQuery`, which will return a `QueryBuilder` object with a base query. Then, when you call it, it will look something like this:
```
$query = $this->getRevenueDonationDataQuery($ids);
$query->select('whatever');
$data = $query->getAll();
```
If you are not familiar with our QueryBuilder library, you can check it here  https://github.com/impress-org/givewp/blob/develop/src/Framework/QueryBuilder/README.md

alaca (commented):
Hey @jfn4th, great work. It will improve the performance of the Donor Dashboard. Thank you. I have a couple of suggestions, other than that, looks good 👍

alaca (approved):
Good job, @jfn4th. This looks good to me 👍 
Thank you for your help, we appreciate it.

PR #7985, opened 2025-06-23
Feature: add donation overview page

JoshuaHungDinh:
Resolves #

## Pre-review Checklist

-   [ ] Includes unit tests


pauloiankoski (approved):
Great job! I’m approving this 👏
There are just a couple of small things we might want to revisit later:
- I’m wondering if it wouldn’t be better for gateways to provide their own view URL, maybe as a method in the Gateway API. That way, given a `transactionId`, each gateway could return its formatted URL. The current approach might make it harder for other gateways (including 3rd-party ones) to have their view URLs shown in the Donation Details page.
- Also, we’re not showing the gateway title properly at the moment. Maybe we could get the title directly from the gateway and include it already formatted in the API response?

Nothing blocking for now, just some ideas to consider moving forward 🙂

JoshuaHungDinh:
@pauloiankoski Thanks! For the Payment method title - I actually did update where we were pulling that information from but did not update the visuals for you in time. Its updated now but I am however working with Jeffrey to get icons for all the gateways.

I'll make sure to circle back on how we are providing the urls! Could be a good Fun-Friday thing I might bring you in on 😄

PR #7995, opened 2025-06-27
Fix: Divi compatibility issues

alaca:
## Description

This PR resolves the issue with the Divi theme builder. The problem with the Divi theme builder is that it deregisters WP core scripts that are used by GiveWP entities. The issue is resolved by deregistering GiveWP entities on Divi theme builder pages. 

## Testing Instructions

Make sure GiveWP 4.0 + and the Divi theme are installed.

In WP Admin, Go to Divi > Theme Builder.

In any template, click Add Global Body.

Click Build Global Body.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/ThirdPartyCompatibility/Divi/DeregisterEntityScripts.php
+++ b/src/ThirdPartyCompatibility/Divi/DeregisterEntityScripts.php
@@ -9,6 +9,10 @@ class DeregisterEntityScripts
 {
     public function __invoke()
     {
+        if ( ! isset($_GET['page']) || $_GET['page'] !== 'et_theme_builder') {
+            return;
+        }
+
         global $wp_scripts;
 
         $registered_scripts = $wp_scripts->registered;
--- a/src/ThirdPartyCompatibility/Divi/DeregisterEntityScripts.php
+++ b/src/ThirdPartyCompatibility/Divi/DeregisterEntityScripts.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Give\ThirdPartyCompatibility\Divi;
+
+/**
+ * @unreleased
+ */
+class DeregisterEntityScripts
+{
+    public function __invoke()
+    {
+        $entities = [
+            'givewp-campaign-entity',
+            'givewp-form-entity',
+            'givewp-donor-entity',
+        ];
+
+        if (
+            isset($_GET['page'])
+            && $_GET['page'] == 'et_theme_builder'
+        ) {
+            foreach ($entities as $entity) {
+                wp_dequeue_script($entity);
+                wp_deregister_script($entity);
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/ThirdPartyCompatibility/Divi/DeregisterEntityScripts.php
```diff
@@ -0,0 +1,28 @@
...
+            'givewp-donor-entity',
+        ];
+
+        if (
+            isset($_GET['page'])
+            && $_GET['page'] == 'et_theme_builder'
+        ) {
+            foreach ($entities as $entity) {
+                wp_dequeue_script($entity);
+                wp_deregister_script($entity);
```
glaubersilva: While this is working for the current entities, it will not work for new ones, so it would be necessary to edit the `$entities` array every time a new one is added to the project.

So, I would prefer replace this loop here with a  `deregisterGiveWpEntityScripts` function that would have as content something like this:

```
    private function deregisterGiveWpEntityScripts()
    {
        global $wp_scripts;
    
        $registered_scripts = $wp_scripts->registered;

        // Filter for GiveWP entity scripts using the pattern 'givewp-*-entity'
        foreach ($registered_scripts as $handle => $script) {
            if (preg_match('/^givewp-.*-entity$/', $handle)) {
                wp_dequeue_script($handle);
                wp_deregister_script($handle);
            }
        }
    }
```

I didn't test this code to check if it's working, but the main idea is to make the current logic less fragile and check the pattern name we are using to register entities.

alaca: Great idea, Glauber. I'll implement that

alaca: Resolved: https://github.com/impress-org/givewp/pull/7995/commits/4ae63db7b371eee33f43b001c4449056c92c4728

src/ThirdPartyCompatibility/Divi/DeregisterEntityScripts.php
```diff
@@ -0,0 +1,23 @@
...
+ */
+class DeregisterEntityScripts
+{
+    public function __invoke()
+    {
+        global $wp_scripts;
+
+        $registered_scripts = $wp_scripts->registered;
+
+        foreach ($registered_scripts as $handle => $script) {
```
glaubersilva, line 20: @alaca I think we still need the condition to check if the current page is `et_theme_builder`, otherwise it will deregister the entities for all pages.

alaca: Yeah, I removed that by accident

alaca: @glaubersilva resolved https://github.com/impress-org/givewp/pull/7995/commits/99b2465aab217236655d5d7c7fbb15a7c31275a6

PR #7997, opened 2025-07-01
Feature: donation details refund

jonwaldstein:
Resolves [GIVE-2685]

## Description

This adds the ability to refund a donation from the new donation details screen.

Some highlights:
- A new `PaymentGatewaysRefundable` interface was added to the gateway API
- Any core gateways that were implementing a `refundDonation` method in their gateway class that was not functional has been removed to make sure we don't show the refund button until their logic is updated.
- The existing gateway controller refund method was too specific to how the old ajax system works and was left alone in favor of a method directly inside the REST endpoint.

The old way of refunding donations through gateways works completely different than this new way.  Meaning there are several gateways that would need to be updated to support this new API.

The following gateways will not be compatible with this new refund button until we choose to update them.  

- PayPalStandard
- All legacy Stripe gateways
- Stripe Payment Element
- PayPal commerce

## Affects

- Refund functionality in the gateway api

## Visuals

https://www.loom.com/share/46e7f0884ef844a2ba981f2c5cd3cbe8?sid=f8cff19b-546b-4dbd-9de7-54b528902ca5

## Testing Instructions

ZIP: https://github.com/impress-org/givewp/actions/runs/16119430837

- Create a donation with a gateway that supports donations
- Use the refund button and make sure it works as expected

## Pre-review Checklist

-   [x] Includes unit tests

[GIVE-2685]: https://stellarwp.atlassian.net/browse/GIVE-2685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Framework/PaymentGateways/Actions/HandleGatewayPaymentCommand.php
+++ b/src/Framework/PaymentGateways/Actions/HandleGatewayPaymentCommand.php
@@ -67,16 +67,9 @@ class HandleGatewayPaymentCommand
         if ($command instanceof PaymentRefunded) {
             $handler = new PaymentRefundedHandler($command);
             $handler->handle($donation);
-            $url = isset($_REQUEST['_wp_http_referer']) ? home_url($_REQUEST['_wp_http_referer']) : null;
+            $url = isset($_REQUEST['_wp_http_referer']) ? home_url($_REQUEST['_wp_http_referer']) : home_url('/');
 
-            if ($url) {
-                return new RedirectResponse($url);
-            }
-
-            return new JsonResponse([
-                'status' => 'success',
-                'message' => __('Refund completed successfully', 'give'),
-            ]);
+            return new RedirectResponse($url);
         }
 
         if ($command instanceof RedirectOffsite) {


glaubersilva:
@jonwaldstein  I got your point, but was thinking of the following scenario: 

1) The user changes the donation status to "refunded" on the admin UI and expects it to also make the refund on the gateway side, like in the legacy screen.
2) He noted that the refund on the gateway side didn't work, so he clicked on the 3 dots menu to make the refund, and couldn't see the option to refund because the status was changed previously.
3) So it's necessary to change the status again to "completed" to enable the real refund feature again in the 3 dots menu.
4) It can generate confusion.

I don't know how frequently it can happen, but since the customer already has the habit of making the refund just by changing the status, I suspect that it can happen frequently.

So, if we decide to keep it in a separate endpoint like it is now, I think we need to add at least some way to inform the user that just changing the status will not perform the refund on the gateway side.


image

jonwaldstein:
@glaubersilva okay so you're thinking more from a UI/UX perspective, good thoughts!  It's a valid point about the way people might be used to refunding and how this new interface works.  However, I don't think that actually correlates to the API endpoints we use as we can guide the user in the UI by making it more intuitive and/or have docs/feedback messaging.

As an aside, I really disliked the old interface for refunding and found that to _not_ be intuitive because you have to select that past-tense status of _refunded_ with that little checkbox for interacting with the gateway.  Honestly, i'm hopeful this interface will actually be more intuitive but I think we can bring up your concerns about potential old habit from the old interface up for discussion with @jdghinson 😄

glaubersilva (approved):
@jonwaldstein Great work! Ready to go! 🚀

PR #8008, opened 2025-07-08
Renewal date not set if period is quarter

Genevieve-K:
`strtotime()` doesn't work with "quarter". Ex:. `strtotime("+1 quarter")` is null.
I suggest to replace it by "3 months".

### DIFF
--- a/src/LegacySubscriptions/includes/give-subscription.php
+++ b/src/LegacySubscriptions/includes/give-subscription.php
@@ -656,7 +656,11 @@ class Give_Subscription {
 		}
 
 		$last_day   = cal_days_in_month( CAL_GREGORIAN, date( 'n', $base_date ), date( 'Y', $base_date ) );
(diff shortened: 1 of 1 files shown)


jonwaldstein (approved):
Thanks @Genevieve-K 🙌

Genevieve-K:
Hi @jonwaldstein . I also see a problem in give-recurring, but I cannot access to GitHub.
This concern the email tags for Renewal Receipt Email. 
The next renewal date is sometimes wrong. I understood the issue.
This is in file `plugins/give-recurring/includes/give-recurring-emails.php`.

As the donation is a renewal, the subscription is not found with this line:
```
$subscription_id = give_recurring_get_subscription_by( 'payment', $tag_args['payment_id'] );`
```
I suggest:
```
$donation = Donation::find( $tag_args['payment_id'] );
$subscription_id = $donation->subscriptionId;
```

PR #8010, opened 2025-07-09
Feature: add reverse method and prevent DonationMetaKeys::CAMPAIGN_ID records duplicates

glaubersilva:
Resolves [GIVE-2696]

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Campaigns/Migrations/Donations/AddCampaignId.php
+++ b/src/Campaigns/Migrations/Donations/AddCampaignId.php
@@ -6,13 +6,14 @@ use Give\Donations\ValueObjects\DonationMetaKeys;
 use Give\Framework\Database\DB;
 use Give\Framework\Database\Exceptions\DatabaseQueryException;
 use Give\Framework\Migrations\Contracts\BatchMigration;
+use Give\Framework\Migrations\Contracts\ReversibleMigration;
 use Give\Framework\Migrations\Exceptions\DatabaseMigrationException;
 use Give\Framework\QueryBuilder\QueryBuilder;
 
 /**
  * @since 4.0.0
  */
-class AddCampaignId extends BatchMigration
+class AddCampaignId extends BatchMigration implements ReversibleMigration
 {
     /**
      * @inheritDoc
@@ -50,6 +51,10 @@ class AddCampaignId extends BatchMigration
 
     /**
      * @inheritDoc
+     * 
+     * @unreleased Prevent DonationMetaKeys::CAMPAIGN_ID records duplicates
+     * @since 4.0.0
+     * 
      * @throws DatabaseMigrationException
      */
     public function runBatch($firstId, $lastId)
@@ -82,6 +87,18 @@ class AddCampaignId extends BatchMigration
             }
 
             $donations = $query->getAll();
+
+            if (empty($donations)) {
+                return;
+            }
+
+            $donationIds = array_column($donations, 'ID');
+            
+            // Delete existing DonationMetaKeys::CAMPAIGN_ID records for these donations to prevent duplicates
+            DB::table('give_donationmeta')
+                ->whereIn('donation_id', $donationIds)
+                ->where('meta_key', DonationMetaKeys::CAMPAIGN_ID)
+                ->delete();
 
             $donationMeta = [];
 
@@ -156,5 +173,17 @@ class AddCampaignId extends BatchMigration
             ->where('ID', $lastProcessedId, '>')
             ->limit(1)
             ->count();
+    }
+
+    /**
+     * @inheritDoc
+     *
+     * @unreleased
+     */
+    public function reverse(): void
+    {
+        DB::table('give_donationmeta')
+            ->where('meta_key', DonationMetaKeys::CAMPAIGN_ID)
+            ->delete();
     }
 }


src/Campaigns/Migrations/Donations/AddCampaignId.php
```diff
@@ -83,6 +88,14 @@ public function runBatch($firstId, $lastId)
 
             $donations = $query->getAll();
 
+            if (empty($donations)) {
+                return;
+            }
+
+            // Delete existing DonationMetaKeys::CAMPAIGN_ID records for these donations to prevent duplicates
+            $donationIds = implode(', ', array_column($donations, 'ID'));            
+            DB::query("DELETE FROM " . DB::prefix('give_donationmeta') . " WHERE donation_id IN ($donationIds) AND meta_key = '" . DonationMetaKeys::CAMPAIGN_ID . "'");
```
jonwaldstein: @glaubersilva I don't think this migration should be concerned with deleting existing duplicates.  That would be a side effect of previously running this more than once.  The reverse method will solve that.  

The purpose of the migration is to add campaign IDs however, we're using the DB insert method on the meta table which if re-run would add the duplicate.  

So I would suggest simply updating the insert meta sql to either use the built-in WP update_meta function or do a `ON DUPLICATE KEY UPDATE`

Screenshot 2025-07-09 at 11 44 25 AM

jonwaldstein (changes requested):

src/Campaigns/Migrations/Donations/AddCampaignId.php
glaubersilva: @jonwaldstein My first idea was to use `ON DUPLICATE KEY UPDATE`, but it's not possible for this case because the primary key is `meta_id`, which never gets duplicated in this insert statement.

So my next ideas were:

Use the `give_update_payment_meta()` or `Give()->payment_meta->update_meta()` to make the update, but I notice that it will decrease the migration performance since would be necessary loop trough all donations and call one of these function that performs one operation per time on the DB while the current insert statement is receiving the `$donationMeta` which is a array that allow us perform all inserts with one single operation on DB.

This way, my next idea was to add some logic to insert the meta key only for the donations without it and update the ones already added previously, but it was getting too complicated and performed 3 + n queries.

After that, in my last idea, I realized that the best approach to not hurt performance would be to add this simple delete query before the insert to simplify things and make sure we'll always perform only 3 queries.

jonwaldstein: @glaubersilva thanks for the thought process! you make a great point about performance here and something I addressed in a [similar migration](https://github.com/impress-org/givewp/blob/ff6d563137d11ad6dae0bb6e47b74c18506cb288/src/Subscriptions/Migrations/BackfillMissingCampaignIdForDonations.php#L20).

I think you're approach is actually a pretty simple way to effectively do a bulk upsert in meta. There is also `REPLACE INTO` but it's limited to MySQL and MariaDB so there's an edge case if someone is using a different DB. 

A couple things to think about with this one though:

1. Is the process of bulk deleting being batched as well? I can't tell by looking at this but if not - that could be new problem people run into.
2. With the addition of the reverse functionality, is this even still necessary?

glaubersilva: @jonwaldstein  Oh cool! I didn't know about the `REPLACE INTO`. 

About your questions:

1. Yes, the bulk delete is made for the donations from the current batch only.
2. Yes, it is necessary because the reverse button will only be displayed in cases where the migration failed, while this delete logic will prevent the duplicate entries when someone re-runs it.

So, I think it is better to keep both functionalities because for the support team, it would be easier to solve the problem in sites that already have the duplicated entries, as all they need to do is re-run this migration, and everything should be fixed. In other words, they won't need to manually change the status of the migration on the DB to "failed" to see the reverse button.

jonwaldstein: @glaubersilva sounds good thanks!

I think i'm on board with your approach 😄 - this migration is very important to get right and I really want the least amount of friction for our customers and support.

I'll go ahead and approve this so we can get into QA.  I would like to get the eyes of @alaca just as a final look since I believe he created this migration.

alaca: In my opinion, the `runBatch` method should do only one thing: insert data. I'd rather update our migration list table and allow users to roll back successful migrations than do stuff like this. Regarding performance, yeah, there is not much we can do. We are already doing it in the fastest way possible. If we allow users to rollback the successful migration, then we don't have to worry about the duplicate meta because the user has a way of resolving that.

jonwaldstein: @alaca I agree with you about the purpose of the runBatch function.  It's purpose is to insert the campaign ID.  I'm still tempted to treat this one as a special use case because of how important it is.  On the rollback topic - I thought we chose not to allow rollback of successful migration because it could cause weird side effects if being reversed out of order with other migrations.

glaubersilva: @alaca @jonwaldstein I understand when you guys says that this delete logic doesn't seem to be something for the `runBatch` method handle because the goal of this migration is just inset the data. 

However, if we provide a re-run feature that allow user to execute it again, I thinks, yes, is responsibility of this method take care of it and prevent duplicates. Otherwise, we should not have the Re-run migration option which I think is not a good idea. 

In other words, if we allow the user to run a migration again, I think the migration needs to check before if the data was added previously by the migration itself to prevent unexpected results.

I know that this is a feature for advanced users and they need do the re-run by their own risk and do a website backup before, but if we can provide a way to mitigate the possible problems that it can generate, I can't see a good reason to not do that since the performance is not too much affect with the current approach, especially because this is a batch migration.

alaca: No, we don't need to do that if we enable the rollback option for successful migration.

glaubersilva: @jonwaldstein @alaca Resolved in dbad2fc9d6b4147483005c1aa60196b4f46bf2f0

As discussed in the call today, I removed the logic to delete meta keys from the `runBatch` method and enabled the `Reverse Update` button even when the migration has a success status, and the final result is this:

image

Both buttons, `Re-Run Update` and `Reverse Update`, are displayed for successful migrations that implemented the `ReversibleMigration` interface.

This is the expected behavior, right?

jonwaldstein: @glaubersilva yup! we're giving folks the keys 😄

jonwaldstein (approved):
@glaubersilva thanks for working through this one man!

PR #8011, opened 2025-07-09
Feature: add frontend requirements for trashing donations

JoshuaHungDinh:
Resolves [GIVE-2686]

## Description
This PR adds support for trashihng donations from the Overview page. It introduces a custom hook to handle the API call for deletion using the v3 api. The confirmation dialog opened by the row action "delete" on the DonationListTable has also been updated for parity & to enable trashing rather then permanent deletion. 

Note: Can be tested with 
- branch `feature/v3-donations-support-trash`
- pull request https://github.com/impress-org/givewp/pull/8018
## Affects
Donation Overview page
Deleting Donations

## Testing Instructions
- Create several donations.
- From the DonationListTable hover over a donation row and select delete.
- Verify the new wording and confirm delete.
- Select edit on another donation.
- From the top right settings select delete donation.
- View the database and confirm that the donations have been moved to trash status.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2686]: https://stellarwp.atlassian.net/browse/GIVE-2686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Donations/resources/admin/components/DonationDetailsPage/Tabs/Overview/DonationReceipt/index.tsx
+++ b/src/Donations/resources/admin/components/DonationDetailsPage/Tabs/Overview/DonationReceipt/index.tsx
@@ -11,18 +11,6 @@ import type { Donation } from '@givewp/donations/admin/components/types';
  * @unreleased
  */
 export default function DonationReceipt({ donation }: { donation: Donation }) {
-  const billingInfo = {
-    name: donation?.firstName + ' ' + donation?.lastName,
-    email: donation?.email,
-    address: {
-      country: donation?.billingAddress?.country,
-      address1: donation?.billingAddress?.address1,
-      address2: donation?.billingAddress?.address2,
-      city: donation?.billingAddress?.city,
-      state: donation?.billingAddress?.state,
-      zip: donation?.billingAddress?.zip,
-    },
-  };
   return (
     <OverviewPanel>
       <aside
@@ -32,7 +20,7 @@ export default function DonationReceipt({ donation }: { donation: Donation }) {
       >
         <div className={styles.content}>
           <header className={styles.header} role="banner">
-            <h2 className={styles.title} id="donation-details-title">{__('Fundraising Form', 'give')}</h2>
+            <h2 className={styles.title} id="donation-details-title">{donation?.formTitle}</h2>
             <p>{__('Below is a detailed breakdown of this donation.', 'give')}</p>
           </header>
 
@@ -43,7 +31,7 @@ export default function DonationReceipt({ donation }: { donation: Donation }) {
 
             <section className={styles.address} aria-labelledby="billing-information">
               <h3 id="billing-information">{__('Billing information', 'give')}</h3>
-              <BillingInformation {...billingInfo} />
+              <BillingInformation name={`${donation?.firstName} ${donation?.lastName}`.trim()} email={donation?.email} address={donation?.billingAddress} />
             </section>
           </div>
         </div>


src/Donations/resources/hooks/useDonationDelete.tsx
```diff
@@ -0,0 +1,42 @@
...
+    async function deleteDonation(donationId: number) {
+      if (!donationId) return;
+  
+      setIsResolving(true);
+      setHasResolved(false);
+      setError(null);
+  
+      try {
+        await apiFetch({
+          path: `give-api/v2/admin/donations/delete?ids=${donationId}`,
```
jonwaldstein: This is a good way to get this working since it already exists - however, we should really be updating our v3 endpoint for delete functionality.  Perhaps we can follow up on this UI focused PR with the updated endpoint functionality and swap this out.

JoshuaHungDinh: @jonwaldstein Yea there's also an issue with using `wp_delete_post(id, false)` it still deletes the post not trash it. I was eventually able to get it to work with `wp_trash_post` but could not pin down why the former would not trash it properly.

src/Donations/resources/admin/components/DonationDetailsPage/index.tsx
```diff
@@ -85,6 +87,20 @@ export default function DonationDetailsPage() {
...
     };
 
+    /**
+     * @unreleased
+     */
+    const handleDelete = async () => {
+        try {
+            await deleteDonation(donation?.id);
+            setShowConfirmationDialog(null);
+            window.location.href = `${adminUrl}edit.php?post_type=give_forms&page=give-payment-history`;
```
jonwaldstein: this full url would be convenient to send through the window something like: `donationsAdminUrl`

src/Donations/resources/hooks/useDonationDelete.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+      setIsResolving(true);
+      setHasResolved(false);
+      setError(null);
+
+      try {
+        await apiFetch({
+          path: `/givewp/v3/donations/${donationId}?force=${force}`,
+          method: 'DELETE',
+        });
```
jonwaldstein: I discovered we can use [deleteEntityRecord](https://developer.wordpress.org/block-editor/reference-guides/data/data-core/#deleteentityrecord) here instead of fetch.

It will make a the DELETE request and we would include the force param.

There's a nifty little tutorial [here](https://github.com/WordPress/gutenberg/blob/trunk/docs/how-to-guides/data-basics/5-adding-a-delete-button.md) on how to use it but it's pretty similiar to our other entity record usages

src/Donations/resources/admin/components/DonationDetailsPage/index.tsx
JoshuaHungDinh: Hey - It actually looks like this url is already being passed down from the window!
Resolved: cd6661217cbaba2e2e790e0a68c8f8c54bf765ff

src/Donations/resources/hooks/useDonationDelete.tsx
JoshuaHungDinh: Since we were not using anything else from the useDonationDelete custom hook I replaced it with inline usages.
Resolved ca2d215780ca5ac2d7a0c291df22fa06ae0ea104

There was also an issue while deleting a donation where the pageTitle would show NAN so I pushed a quick fix - so that it does not display. 4fe90fb9066ddadfaf3bf069b02cbcb12f4dbb5e

jonwaldstein: @JoshuaHungDinh awesome, glad that entity function worked!

PR #8013, opened 2025-07-10
Fix: bulk actions selector on campaign forms list

jonwaldstein:
Resolves #

## Description

**Problems with the original code:**

- No cleanup function - The event listener was never removed, causing memory leaks
- Logic error - The show.contextMenu check was using stale closure values
- Always active - The listener was always active on the document, interfering with other dropdowns
- Missing dependencies - The effect wasn't reactive to state changes

**The fix:**

- Conditional effect - Now only runs when show.contextMenu is true
- Proper cleanup - Returns a cleanup function that removes the event listener
- Correct dependencies - Added show.contextMenu to the dependency array
- Better event handling - Extracted the handler into a separate function
- Now the document click listener is only active when the context menu is actually open, preventing interference with other dropdown elements in your UI. The listener is automatically removed when the context menu closes or when the component unmounts.

## Affects

The bulk actions selector on campaign forms list

## Visuals

N/A

## Testing Instructions

Make sure the bulk actions selector works & you can also still close the campaign hamburger menu by clicking outside

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Campaigns/resources/admin/components/CampaignDetailsPage/index.tsx
+++ b/src/Campaigns/resources/admin/components/CampaignDetailsPage/index.tsx
@@ -71,24 +71,24 @@ export default function CampaignsDetailsPage({campaignId}) {
           if (headerRef.current) {
             const height = headerRef.current.offsetHeight;
             setHeaderHeight(height);
-            
+
             // Update CSS variable directly
             document.documentElement.style.setProperty('--header-height', `${height}px`);
           }
         };
-        
+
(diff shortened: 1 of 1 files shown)


JoshuaHungDinh:
Everything looks good to me. 👍 👍  


https://github.com/user-attachments/assets/6475b40a-39a3-4f6a-bdcc-e7e62d2892e0

JoshuaHungDinh (approved):

PR #8016, opened 2025-07-10
Fix: This string is not formatted correctly for translation

DAnn2012:
Fix: This string is not formatted correctly for translation

Fix https://wordpress.org/support/topic/this-string-is-not-formatted-correctly-for-translation-8/


jonwaldstein (approved):
@DAnn2012 thank you!

PR #8018, opened 2025-07-11
Feature:  Update V3 DonationController to support trashed donations

JoshuaHungDinh:
Resolves [GIVE-2686]

## Description
Enhance DonationController to Support Soft and Hard Deletes with force Parameter

This PR improves the Donation REST API by introducing support for soft deletes (trashing) and hard deletes (permanent removal) of donations via a new optional force parameter on delete endpoints.

### Key Changes:
- Added force boolean parameter to both single and batch donation delete routes.
- When `force` is `false` or omitted (default), donations are soft deleted by moving them to trash (`wp_trash_post`).
- When `force` is `true`, donations are permanently deleted from the database.
- Updated the REST route argument definitions to include the force parameter for single and multiple deletion endpoints.
- Updated the `delete_item` and `delete_items` methods to handle the `force` parameter accordingly.

## Affects
- Deleteing donation/'s
- v3 API Donations Controller
- Donation Status

## Visuals
https://www.loom.com/share/7be2fb998b484fa4a18ef9d545dc9ea9?sid=590b1f4d-5e15-4f28-9e02-75fae55d3a0b

## Testing Instructions
- Create several donations.
- While on the Donation Overview page - useing the 3 dot menu on the top right delete a donation.
- View the donation in the` Post` table of the database to verify the its status is now `Trashed`
## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2686]: https://stellarwp.atlassian.net/browse/GIVE-2686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Donations/Endpoints/ListDonations.php
```diff
@@ -245,6 +245,9 @@ protected function getWhereConditions(QueryBuilder $query): array
 
         $hasWhereConditions = $search || $start || $end || $campaignId || $donor;
 
+        $query->where('post_type', 'give_payment');
+        $query->where('post_status', 'trash', '<>');
```
JasonTheAdams: We should have a way for folks to choose whether trashed donations are included in the results.

JoshuaHungDinh: Agreed. I’ve was thinking we might need a way to surface trashed donations in the list table and also provide users with a way to 'empty the trash'.

src/Donations/Endpoints/ListDonations.php
```diff
@@ -126,6 +127,16 @@ public function registerRoute()
                             'columns',
                         ],
                     ],
+                    'postStatus' => [
```
jonwaldstein: The post status is something we use under the hood for storage - however, the donation should just refer to this as _status_

```suggestion
                    'status' => [
```

src/Donations/Endpoints/ListDonations.php
```diff
@@ -126,6 +127,16 @@ public function registerRoute()
...
+                        'type' => 'string',
+                        'required' => false,
+                        'default' => 'active',
+                        'enum' => [
+                            'active',
+                            'trash',
+                        ],
+                        'description' => 'Filter donations by WordPress post status: active (all except trash), or trash.'
```
jonwaldstein: Again here let's keep the context of the donation - not WordPress posts (which is the storage we use under the hood and will change in the future)

```suggestion
                        'description' => 'Filter donations by Donation status: active (all except trash), or trash.'
```

src/API/REST/V3/Routes/Donations/DonationController.php
```diff
@@ -447,11 +465,21 @@ public function delete_items($request): WP_REST_Response
...
-                $errors[] = ['id' => $id, 'message' => __('Failed to delete donation', 'give')];
-            }
+                if ($force) {
+                    if ($donation->delete()) {
+                        $deleted[] = ['id' => $id, 'previous' => $item];
+                    } else {
+                        $errors[] = ['id' => $id, 'message' => __('Failed to delete donation', 'give')];
+                    }
+                } else {
+                    $trashed = wp_trash_post($donation->id);
```
jonwaldstein: I'm hesitant to use `wp_trash_post` directly here as we want this endpoint to interact with the donation model & abstract awa the data layer as much as possible.  In the future when the storage of a donation changes - this would be something we would have to update.

`wp_trash_post` does a couple things:
1. Changes the status to `trash`
2. Stores the previous status & time in meta `_wp_trash_meta_status`, `_wp_trash_meta_time` (as a way to restore to previous status)

All that is something we can support directly in the donation repository with a new `trash()` method and/or build it into the existing delete method.

At the very least if we simply changed the status of the donation to trash and saved - then that would be better for maintainability here.

jonwaldstein (changes requested):

src/Donations/Endpoints/ListDonations.php
JoshuaHungDinh: Resolved: 4b24aa62c811897cdaa5127f1859cdd294d0c5ce

src/Donations/Endpoints/ListDonations.php
JoshuaHungDinh: Resolved: 4b24aa62c811897cdaa5127f1859cdd294d0c5ce

src/API/REST/V3/Routes/Donations/DonationController.php
JoshuaHungDinh: Thanks for the context! That makes sense in terms of long-term maintainability and keeping the data layer abstracted.

I pushed up commits here to reflect that approach: 59a1f8b43180d54eb857a31e5bd28aa1f95729e1 eb9efd4ad565db7e18b631587b2f7f4284acfe79. Let me know if this looks more in line with what you were thinking!

src/Donations/Repositories/DonationRepository.php
```diff
@@ -352,6 +352,43 @@ public function delete(Donation $donation): bool
...
+
+        Hooks::doAction('givewp_donation_trashing', $donation);
+
+        try {
+            $previousStatus = DB::table('posts')
+                ->where('ID', $donation->id)
+                ->value('post_status');
+
+            give()->payment_meta->update_meta($donation->id, '_wp_trash_meta_status', $previousStatus);
+            give()->payment_meta->update_meta($donation->id, '_wp_trash_meta_time', time());
```
jonwaldstein, line 372: Just a note about this to help us think about the future.

I think these are okay to keep wp specific as our donations are still technically custom post types and this will inform WP that they are in fact in trash status 👍  

The important thing here is all of this is abstracted away in our donation repository so we're safe to call $donation->trash() and not have to worry about what goes on under the hood.

In the future, when the database changes - we'll be going through our repositories and making changes that will reflect the new storage.

src/API/REST/V3/Routes/Donations/DonationController.php
```diff
@@ -431,6 +453,7 @@ public function delete_item($request): WP_REST_Response
     public function delete_items($request): WP_REST_Response
     {
         $ids = $request->get_param('ids');
+        $force = filter_var($request->get_param('force'), FILTER_VALIDATE_BOOLEAN);
```
jonwaldstein: @JoshuaHungDinh fun fact - the WP REST API sanitizes values for us using the property type defined in schema.  So since `force` is a type of boolean it will sanitize as boolean up stream allowing you to safely access the value here without having to manually sanitize again.  This means we don't need this 😄 

Ref: https://github.com/WordPress/WordPress/blob/86b7fc84bdff5a9db05a120bba7858d5f4d32364/wp-includes/rest-api.php#L2872

JoshuaHungDinh: Resolved: 07e57b040702bce504a36066aee63d776b09e772

JoshuaHungDinh:
> @JoshuaHungDinh would you mind providing a loom for this? I'm seeing a mix of delete and trash in the UI
> 
> Screenshot 2025-07-17 at 2 23 51 PM Screenshot 2025-07-17 at 2 24 07 PM

You're right! I must have skipped the pane displaying the change -  It should be updated to Trash.  Ihttps://app.zeplin.io/project/63f3ba3677982b0ead24ecb7/screen/68388d82807b907aa9e04f61

Also I included the video to the visuals!

jonwaldstein:
@JoshuaHungDinh this works! one thought about the UX:

There's a little gap between when the donation is done being trashed and redirecting to the list table.  I think if you just don't close the modal after deleting - it would look more seamless.

Also, there's a prop called `isConfirming` on the [ConfirmationDialog modal](https://github.com/impress-org/givewp/blob/feature/v3-donations-support-trash/src/Donations/resources/admin/components/DonationDetailsPage/index.tsx#L125) that you can use for a loading status.

So i'm suggesting here you can probably remove `setShowConfirmationDialog(null);` during success so there is no flash between redirect - in conjuction with the isConfirming prop 😄 

Screenshot 2025-07-17 at 3 49 11 PM

JoshuaHungDinh:
> @JoshuaHungDinh nice job! I added that UX suggestion but feel free to enhance later 😄

Resolved: 6c295636221215e2dabcf9b09c98aea502c98b33

jonwaldstein:
@JoshuaHungDinh thanks! I have to merge this to the epic but feel free to enhance later 😄

PR #8027, opened 2025-07-17
Fix: donation API endpoint permission checks for GiveWP roles

pauloiankoski:
## Description

This PR fixes the permission checks for the donation API endpoints to properly align with GiveWP's custom user roles and capabilities. The existing implementation was using `manage_options` which only allowed WordPress administrators, but the donation management functionality should be accessible to GiveWP-specific roles.

The changes ensure that:
- Permission checks match the legacy admin interface behavior
- GiveWP Managers and Accountants can access donation data appropriately  
- Workers are properly restricted (matching existing admin page access)
- Sensitive data access is controlled based on proper capabilities

## Affects

- Donation API endpoints (`/donations` collection and individual donation routes)
- Permission checks for viewing, editing, deleting, and refunding donations
- Sensitive data access controls for different user roles
- Integration between new API and existing GiveWP role system

## Testing Instructions

1. Create test users with different GiveWP roles (Manager, Accountant, Worker)
2. Test API access for each role:
   - **Manager**: Should have full access to all endpoints and sensitive data
   - **Accountant**: Should have access to view/edit donations but limited sensitive data
   - **Worker**: Should be blocked from API access (matches admin interface)
   - **Administrator**: Should retain full access
3. Verify that `includeSensitiveData` parameter works correctly for each role
4. Test anonymous donation handling with different permission levels

## Pre-review Checklist

- [ ] Includes unit tests

### DIFF
--- a/src/API/REST/V3/Routes/Donations/DonationController.php
+++ b/src/API/REST/V3/Routes/Donations/DonationController.php
@@ -609,10 +609,11 @@ class DonationController extends WP_REST_Controller
      */
     public function permissionsCheck(WP_REST_Request $request)
     {
-        $isAdmin = current_user_can('manage_options');
+        $canEditDonations = current_user_can('manage_options') ||
+            (current_user_can('edit_give_payments') && current_user_can('view_give_payments'));
 
         $includeSensitiveData = $request->get_param('includeSensitiveData');
-        if ( ! $isAdmin && $includeSensitiveData) {
+        if ( !$canEditDonations && $includeSensitiveData) {
             return new WP_Error(
                 'rest_forbidden',
                 esc_html__('You do not have permission to include sensitive data.', 'give'),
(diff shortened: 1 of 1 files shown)


src/API/REST/V3/Routes/Donations/DonationController.php
```diff
@@ -639,7 +640,7 @@ public function permissionsCheck(WP_REST_Request $request)
      */
     public function update_item_permissions_check($request)
     {
-        if (!current_user_can('manage_options')) {
+        if (!current_user_can('edit_give_payments') && !current_user_can('manage_options')) {
```
jonwaldstein: Shouldn't this be an _or_ check?  

1. If admin then pass
2. if can't edit payment then fail

```suggestion
        if (!current_user_can('edit_give_payments') || !current_user_can('manage_options')) {
```

pauloiankoski: Not really, changing it to an OR would require both capabilities to be a valid statement. That means, if you are an give_manager but not an WP administrator, it would fails. I created a simple test in JS showing the differences where the first implementation uses AND and the last one uses OR:

CleanShot 2025-07-18 at 12 46 30

jonwaldstein: @pauloiankoski oh yeah I see your point - I have personally got confused by these permission conditionals in the past.  On a recent update I started just making it easier for me to understand [like this](https://github.com/impress-org/givewp/blob/epic/donation-details-admin-page/src/Donations/Endpoints/Endpoint.php#L62):

```php
 if (current_user_can('manage_options') || current_user_can('edit_give_payments')) {
      return true;
  }
  
  return WP_Error;
```

pauloiankoski: Got ya. I'll rewrite those that way.

PR #8032, opened 2025-07-23
Refactor: subscriptions REST API V3 endpoints

glaubersilva:
Resolves [GIVE-2675]

## Description

This PR implements a comprehensive REST API system for managing subscriptions and their associated donations, providing frontend developers with complete control over the subscription lifecycle and donation management.

🔄 Complete CRUD Operations for Subscriptions

**GET** `/wp-json/givewp/v3/subscriptions` - List subscriptions with filtering and pagination
**POST** `/wp-json/givewp/v3/subscriptions` - Create new subscriptions
**GET** `/wp-json/givewp/v3/subscriptions/{id}` - Retrieve individual subscription details
**PUT** `/wp-json/givewp/v3/subscriptions/{id}` - Update subscription properties
**DELETE** `/wp-json/givewp/v3/subscriptions/{id}` - Soft delete or permanently remove subscriptions
**DELETE** `/wp-json/givewp/v3/subscriptions` - Bulk delete multiple subscriptions

💰 New Donation Creation Endpoint with Subscription Support

**GET** `/wp-json/givewp/v3/donations` - Retrieve donations filtered by `subscriptionId`
**POST** `/wp-json/givewp/v3/donations` - Create new donations and renewals just by passing `subscriptionId` and` type` (set to _"renewal"_) parameters

Features:

- Added support for `subscriptionId` parameter to create donations associated with subscriptions
- Implemented support for creating renewal donations linked to existing subscriptions
- Added comprehensive validation rules for subscription-donation relationships
- Implemented automatic gateway validation to ensure compatibility with subscription requirements

## Affects

The `subscriptions` and `donations` routes.

## Testing Instructions

To test the new subscription endpoints, you can run these commands in your terminal:

```
composer test -- --filter SubscriptionRouteCreateItemTest
composer test -- --filter SubscriptionRouteDeleteItemsTest
composer test -- --filter SubscriptionRouteDeleteItemTest
composer test -- --filter SubscriptionRouteGetItemsTest
composer test -- --filter SubscriptionRouteGetItemTest
composer test -- --filter SubscriptionRouteUpdateItemTest
```

To test the new donation endpoint and the refactor for the previous one, you can run these commands in your terminal:

```
composer test -- --filter DonationRouteCreateTest
composer test -- --filter GetDonationsRouteTest
```

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2675]: https://stellarwp.atlassian.net/browse/GIVE-2675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/API/REST/V3/Routes/Donations/DataTransferObjects/DonationCreateData.php
+++ b/src/API/REST/V3/Routes/Donations/DataTransferObjects/DonationCreateData.php
@@ -252,9 +252,11 @@ class DonationCreateData
         
         $subscription = Subscription::find($this->subscriptionId);
         
-        // For renewal donations, we only need minimal attributes
-        // The subscription's createRenewal method will handle the rest
-        $renewalAttributes = [];
+        // Pass the processed attributes to allow overriding values from the request
+        // Filter out auto-generated fields and subscription-specific fields
+        $renewalAttributes = array_filter($this->attributes, function ($key) {
+            return !in_array($key, ['id', 'createdAt', 'updatedAt', 'subscriptionId', 'type'], true);
+        }, ARRAY_FILTER_USE_KEY);
 
         return $subscription->createRenewal($renewalAttributes);
     }
--- a/src/API/REST/V3/Routes/Donations/DataTransferObjects/DonationCreateData.php
+++ b/src/API/REST/V3/Routes/Donations/DataTransferObjects/DonationCreateData.php
@@ -5,15 +5,10 @@ namespace Give\API\REST\V3\Routes\Donations\DataTransferObjects;
 use DateTime;
 use Exception;
 use Give\API\REST\V3\Routes\Donations\Exceptions\DonationValidationException;
+use Give\API\REST\V3\Routes\Donations\Helpers\DonationFields;
 use Give\Donations\Models\Donation;
-use Give\Donations\Properties\BillingAddress;
-use Give\Donations\ValueObjects\DonationMode;
-use Give\Donations\ValueObjects\DonationStatus;
 use Give\Donations\ValueObjects\DonationType;
-use Give\Framework\Exceptions\Primitives\InvalidArgumentException;
-use Give\Framework\Support\ValueObjects\Money;
 use Give\Subscriptions\Models\Subscription;
-use WP_Error;
 use WP_REST_Request;
 
 /**
@@ -351,73 +346,10 @@ class DonationCreateData
                 continue;
             }
 
-            $processedAttributes[$key] = $this->processFieldValue($key, $value);
+            $processedAttributes[$key] = DonationFields::processValue($key, $value);
         }
 
         return $processedAttributes;
-    }
-
-    /**
-     * Process field values for special data types
-     *
-     * @unreleased
-     *
-     * @param string $key
-     * @param mixed $value
-     * @return mixed
-     */
-    private function processFieldValue(string $key, $value)
-    {
-        switch ($key) {
-            case 'amount':
-            case 'feeAmountRecovered':
-                if (is_array($value)) {
-                    // Handle Money object array format: ['amount' => 100.00, 'currency' => 'USD']
-                    if (isset($value['amount']) && isset($value['currency'])) {
-                        return Money::fromDecimal($value['amount'], $value['currency']);
-                    }
-                }
-                return $value;
-
-            case 'status':
-                if (is_string($value) && DonationStatus::isValid($value)) {
-                    return new DonationStatus($value);
-                }
-                return $value;
-
-            case 'type':
-                if (is_string($value) && DonationType::isValid($value)) {
-                    return new DonationType($value);
-                }
-                return $value;
-
-            case 'mode':
-                if (is_string($value) && DonationMode::isValid($value)) {
-                    return new DonationMode($value);
-                }
-                return $value;
-
-            case 'billingAddress':
-                if (is_array($value)) {
-                    return BillingAddress::fromArray($value);
-                }
-                return $value;
-
-            case 'createdAt':
-                try {
-                    if (is_string($value)) {
-                        return new DateTime($value, wp_timezone());
-                    } elseif (is_array($value)) {
-                        return new DateTime($value['date'], new \DateTimeZone($value['timezone']));
-                    }
-                } catch (\Exception $e) {
-                    throw new InvalidArgumentException("Invalid date format for {$key}: {$value}.");
-                }
-                return $value;
-
-            default:
-                return $value;
-        }
     }
 
     /**
--- a/src/API/REST/V3/Routes/Donations/DataTransferObjects/DonationCreateData.php
+++ b/src/API/REST/V3/Routes/Donations/DataTransferObjects/DonationCreateData.php
@@ -1,0 +1,435 @@
+<?php
+
+namespace Give\API\REST\V3\Routes\Donations\DataTransferObjects;
+
+use DateTime;
+use Exception;
+use Give\API\REST\V3\Routes\Donations\Exceptions\DonationValidationException;
+use Give\Donations\Models\Donation;
+use Give\Donations\Properties\BillingAddress;
+use Give\Donations\ValueObjects\DonationMode;
+use Give\Donations\ValueObjects\DonationStatus;
+use Give\Donations\ValueObjects\DonationType;
+use Give\Framework\Exceptions\Primitives\InvalidArgumentException;
+use Give\Framework\Support\ValueObjects\Money;
+use Give\Subscriptions\Models\Subscription;
+use WP_Error;
+use WP_REST_Request;
+
+/**
+ * @unreleased
+ */
+class DonationCreateData
+{
+    /**
+     * @var array
+     */
+    private $attributes;
+
+    /**
+     * @var bool
+     */
+    private $isRenewal;
+
+    /**
+     * @var DonationType|null
+     */
+    private $type;
+
+    /**
+     * @var int
+     */
+    private $subscriptionId;
+
+    /**
+     * @since 3.0.0
+     */
+    public function __construct(array $attributes)
+    {
+        $this->attributes = $this->processAttributes($attributes);
+        $this->subscriptionId = $this->attributes['subscriptionId'] ?? 0;
+        $this->type = $this->attributes['type'] ?? null;
+        $this->isRenewal = $this->determineIfRenewal();
+    }
+
+    /**
+     * Create DonationCreateData from REST request
+     *
+     * @unreleased
+     *
+     * @param WP_REST_Request $request
+     * @return DonationCreateData
+     */
+    public static function fromRequest(WP_REST_Request $request): DonationCreateData
+    {
+        return new self($request->get_params());
+    }
+
+    /**
+     * Validate data for creating a single donation
+     *
+     * @unreleased
+     *
+     * @throws DonationValidationException
+     */
+    public function validateCreateDonation(): void
+    {
+        if ($this->isRenewal) {
+            throw new DonationValidationException(
+                __('Cannot create single donation for renewal type', 'give'),
+                'invalid_donation_type',
+                400
+            );
+        }
+
+        $requiredFields = ['donorId', 'amount', 'gatewayId', 'mode', 'formId', 'firstName', 'email'];
+
+        foreach ($requiredFields as $field) {
+            if (!isset($this->attributes[$field])) {
+                throw new DonationValidationException(
+                    sprintf(__('Missing required field: %s', 'give'), $field),
+                    'missing_required_field',
+                    400
+                );
+            }
+        }
+    }
+
+    /**
+     * Validate data for creating a renewal donation
+     *
+     * @unreleased
+     *
+     * @throws DonationValidationException
+     */
+    public function validateCreateRenewal(): void
+    {
+        if (!$this->isRenewal) {
+            throw new DonationValidationException(
+                __('Cannot create renewal donation for non-renewal type', 'give'),
+                'invalid_donation_type',
+                400
+            );
+        }
+
+        $requiredFields = ['subscriptionId', 'type'];
+
+        foreach ($requiredFields as $field) {
+            if (!isset($this->attributes[$field])) {
+                throw new DonationValidationException(
+                    sprintf(__('Missing required field: %s', 'give'), $field),
+                    'missing_required_field',
+                    400
+                );
+            }
+        }
+
+        // Validate subscription exists
+        $subscription = Subscription::find($this->subscriptionId);
+        if (!$subscription) {
+            throw new DonationValidationException(
+                __('Subscription not found', 'give'),
+                'subscription_not_found',
+                404
+            );
+        }
+
+        // Ensure total donations don't exceed subscription installments
+        if ($subscription->installments > 0 && $subscription->totalDonations() >= $subscription->installments) {
+            throw new DonationValidationException(
+                __('Cannot create donation: subscription installments limit reached', 'give'),
+                'subscription_installments_exceeded',
+                400
+            );
+        }
+    }
+
+    /**
+     * Validate subscription-related rules
+     *
+     * @unreleased
+     *
+     * @throws DonationValidationException
+     */
+    public function validateSubscriptionRules(): void
+    {
+        // When subscriptionId is greater than zero, type must be "subscription" or "renewal"
+        if ($this->subscriptionId > 0) {
+            if (!$this->type || !in_array($this->type->getValue(), ['subscription', 'renewal'], true)) {
+                throw new DonationValidationException(
+                    __('When subscriptionId is provided, type must be "subscription" or "renewal"', 'give'),
+                    'invalid_donation_type_for_subscription',
+                    400
+                );
+            }
+
+            // Validate subscription exists
+            $subscription = Subscription::find($this->subscriptionId);
+            if (!$subscription) {
+                throw new DonationValidationException(
+                    __('Subscription not found', 'give'),
+                    'subscription_not_found',
+                    404
+                );
+            }
+
+            // When creating a donation associated with subscriptionId, ensure type is not "subscription"
+            // if a donation of that type already exists for this subscription
+            if ($this->type->getValue() === 'subscription' && $subscription->totalDonations() > 0) {
+                throw new DonationValidationException(
+                    __('A subscription donation already exists for this subscription', 'give'),
+                    'subscription_donation_already_exists',
+                    400
+                );
+            }
+
+            // When creating a subscription or renewal donation, ensure gatewayId matches the subscription's gateway
+            if (in_array($this->type->getValue(), ['subscription', 'renewal'], true)) {
+                $donationGatewayId = $this->attributes['gatewayId'] ?? null;
+                if ($donationGatewayId && $subscription->gatewayId && $donationGatewayId !== $subscription->gatewayId) {
+                    throw new DonationValidationException(
+                        __('Gateway ID must match the subscription gateway for subscription and renewal donations', 'give'),
+                        'gateway_mismatch_for_subscription_donation',
+                        400
+                    );
+                }
+            }
+        } else {
+            // When subscriptionId is zero, type can only be "single" (if provided)
+            if ($this->type && $this->type->getValue() !== 'single') {
+                throw new DonationValidationException(
+                    __('When subscriptionId is zero, type can only be "single"', 'give'),
+                    'invalid_donation_type_for_single',
+                    400
+                );
+            }
+
+            // Set type to single if not provided
+            if (!$this->type) {
+                $this->attributes['type'] = DonationType::SINGLE();
+            }
+        }
+    }
+
+    /**
+     * Convert to Donation model
+     *
+     * @unreleased
+     *
+     * @return Donation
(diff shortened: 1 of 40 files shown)


src/API/REST/V3/Routes/Donors/DonorController.php
```diff
@@ -2,7 +2,7 @@
 
 namespace Give\API\REST\V3\Routes\Donors;
 
-use Give\API\REST\V3\Routes\CURIE;
+use Give\API\REST\V3\Helpers\CURIE;
```
jonwaldstein: I suggest we avoid the "Helpers" trap 😄 , once you start creating a helpers folder and file it becomes a bit of a dumping ground.

Instead, the way I like to approach this is to create a "Support" folder inside my domain and then you can add top level classes and also sub folders.

This is what we have in the [Give framework folder](https://github.com/impress-org/givewp/blob/f44bd73aa7a71fbbfaba7b3dbf096e97297bf7bc/src/Framework/Support)

This approach was inspired by the [Laravel framework structure ](https://github.com/laravel/framework/tree/12.x/src/Illuminate/Support)

src/API/REST/V3/Routes/Subscriptions/Helpers/SubscriptionPermissions.php
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * @unreleased
+     *
+     * @param WP_REST_Request $request
+     *
+     * @return true|WP_Error
+     */
+    public static function validationForDeleteMethods(WP_REST_Request $request)
+    {
+        if (! self::canEdit()) {
```
jonwaldstein:
```suggestion
        if (! self::canDelete()) {
```

src/API/REST/V3/Routes/Donations/DonationController.php
```diff
@@ -385,6 +458,18 @@ private function processFieldValue(string $key, $value)
                 }
                 return $value;
 
+            case 'type':
+                if (is_string($value)) {
```
jonwaldstein: Might want to add additional isValid check for ValueObjects

```suggestion
                if (is_string($value) && DonationType::isValid()) {
```

src/API/REST/V3/Routes/Donations/DonationController.php
```diff
@@ -408,6 +493,138 @@ private function processFieldValue(string $key, $value)
         }
     }
 
+    /**
+     * Process attributes for creating a new donation.
+     *
+     * @unreleased
+     */
+    private function processCreateAttributes(array $params)
```
jonwaldstein: This feels like it should be a DTO that takes in the params and has the ability to declaratively validate and return data in the proper format.  This approach can help distinguish the preparing of data in the proper format, validation, and creating objects.

I see a lot of opportunity for something like this that could be utilized in these controllers as there are various places we are preparing data in a different format.

Here's essentially what i'm thinking:

```php
$data = DonationControllerData::fromRequest($request);

if ($data->type->isSingle()) {
  $data->validateCreateDonation();
  $donation = $data->toDonation();
  $donation->save();
}

if ($data->type->isRenewal()) {
  $data->validateCreateRenewal();
  $renewal = $data->toRenewal();
  $renewal->save();
}
````

I did something similar in the form processing route a while back

- [Route data ](https://github.com/impress-org/givewp/blob/3e94d1b91880b7b11f1d4402237b81adc944d3f0/src/DonationForms/Routes/DonateRoute.php#L48)
- [Controller Data](https://github.com/impress-org/givewp/blob/f13f109dcf20d73a6f231e8708a78b90267b4bd7/src/DonationForms/DataTransferObjects/DonateControllerData.php#L22)
- [Controller data Implementation](https://gitkraken.dev/link/dnNjb2RlOi8vZWFtb2Rpby5naXRsZW5zL2xpbmsvci8xMDIzZTU4OWE4YTNhNDFiZWEwNGJkYTY4ZjA1ZDkwOWQ5ODJmNmI5L2Yvc3JjL0RvbmF0aW9uRm9ybXMvQ29udHJvbGxlcnMvRG9uYXRlQ29udHJvbGxlci5waHA%2FdXJsPWh0dHBzJTNBJTJGJTJGZ2l0aHViLmNvbSUyRmltcHJlc3Mtb3JnJTJGZ2l2ZXdwLmdpdA%3D%3D?origin=gitlens)

File: src/API/REST/V3/Routes/Donations/DonationController.php [hunk 5]
```diff
@@ -408,6 +493,138 @@ private function processFieldValue(string $key, $value)
...
+                // Skip these fields as they are auto-generated
+                continue;
+            }                        
+            
+            $processedValue = $this->processFieldValue($key, $value);
+            $attributes[$key] = $processedValue;
+        }            
+        
+        // Validate subscription-related rules
+        $subscriptionValidation = $this->validateSubscriptionRules($attributes);
```
jonwaldstein: @glaubersilva I think the majority of these validation can be handled directly from the json schema using conditionals.  Check the docs here, it's pretty powerful and would be handled up the chain:

https://json-schema.org/understanding-json-schema/reference/conditionals

We did this in the [Campaigns schema](https://github.com/impress-org/givewp/blob/8b06b4db839d14456ec311a31b448ef7b983e70d/src/API/REST/V3/Routes/Campaigns/RegisterCampaignRoutes.php#L355) a while back 

cc: @JasonTheAdams in case you have anymore insight into conditional schema validation

src/API/REST/V3/Routes/Subscriptions/Helpers/SubscriptionFields.php
```diff
@@ -0,0 +1,79 @@
...
+ * @unreleased
+ */
+class SubscriptionFields
+{
+    /**
+     * Process field values for special data types before setting them on the subscription model.
+     *
+     * @unreleased
+     */
+    public static function processValue(string $key, $value)
```
jonwaldstein, line 22: This is another one of those things I think could be handled by a higher level DTO (probably just need the one) so you can simply return the values in the format expected.

```php
$data = SubscriptionControllerData::fromRequest($request->get_params());

$subscription = $data->toSubscription();
$subscription->save();
```

src/API/REST/V3/Routes/Subscriptions/SubscriptionController.php
```diff
@@ -222,19 +229,97 @@ public function get_item($request)
...
+
+        // Validate required fields
+        foreach ($requiredFields as $field) {
+            $value = $request->get_param($field);
+            if (empty($value)) {
+                return new WP_REST_Response([
+                    'message' => sprintf(__('Field "%s" is required', 'give'), $field)
+                ], 400);
+            }
+        }
```
jonwaldstein: Shouldn't this validation be handled by the REST args/schema up the chain?

src/API/REST/V3/Routes/Subscriptions/SubscriptionController.php
```diff
@@ -446,99 +495,37 @@ public function get_collection_params(): array
...
-    public function prepare_item_for_response($item, $request): WP_REST_Response
+    public function prepare_item_for_response($item, $request)
     {
         $subscriptionId = $request->get_param('id') ?? $item['id'] ?? null;
 
         if ($subscriptionId) {
             $self_url = rest_url(sprintf('%s/%s/%d', $this->namespace, $this->rest_base, $subscriptionId));
             $donor_url = rest_url(sprintf('%s/%s/%d', $this->namespace, 'donors', $item['donorId']));
-            $donationForm_url = rest_url(sprintf('%s/%s/%d', $this->namespace, 'forms', $item['donationFormId']));
+            $donor_url = add_query_arg([
```
jonwaldstein: We should add a donations link as well

src/Subscriptions/ViewModels/SubscriptionViewModel.php
```diff
@@ -19,18 +24,76 @@ public function __construct(Subscription $subscription)
...
+
+        return $this;
+    }
+
     /**
      * @unreleased
      */
     public function exports(): array
     {
+        $donor = $this->subscription->donor()->get();
```
jonwaldstein: If a model has a defined relationship, then you can just call the property and it will return the model for you.  Alternatively, invoking the method will return the query builder.

```suggestion
        $donor = $this->subscription->donor;
```

src/Subscriptions/ViewModels/SubscriptionViewModel.php
```diff
@@ -19,18 +24,76 @@ public function __construct(Subscription $subscription)
...
+                        break;
+                }
+            }
+        }        
+        
+        if (isset($this->anonymousMode ) && $this->anonymousMode->isRedacted() && $this->subscription->donor->isAnonymous()) {
+            $anonymousDataRedacted = [
+                'donorId',
+                'firstName',
+                'lastName',
```
jonwaldstein, line 82: Is there anything else we need to add here like email?

jonwaldstein (changes requested):

src/API/REST/V3/Routes/Donations/DonationController.php
JasonTheAdams: I'm sad to say this won't work. The JSON schema system in WordPress uses a _much_ older version of JSON Schema that dates before conditionals like this. Honestly, I don't recommend putting conditional validation rules inside the schema — as it just gets too janky and requires that you completely change the shape of the schema. For straightforward validations, the schema is awesome; for conditional validation, it's pretty terrible.

JasonTheAdams: ...And you'll need to update the Campaigns schema. 😅

jonwaldstein: @JasonTheAdams oh lol well last I checked our conditional validation in the Campaign schema _does_ work using the `allOf`approach - it even works on the front-end using the react-hook-form adapter.  @alaca might have some insight into this

JasonTheAdams: `allOf` should work fine. It's the `if` statements that won't in the REST API. The RHF adapter will work great because it supports the latest schema versions. This is just an unfortunate limitation of the REST API, because it's JSON schema validation is hand crafted.

jonwaldstein: Okay @JasonTheAdams so to clarify, we should or should _not_ use `allOf` for conditional validation for this scenario?

Can you also clarify what you mean by this?
> ...And you'll need to update the Campaigns schema. 😅

JasonTheAdams: @glaubersilva The use of `allOf` and `oneOf` are perfectly fine. Those are supported. If you look at the Campaigns code you'll see `if...then` statements in the schema — and this is what's not supported.

So when I say the Campaign schema will need to be updated I meant that in another PR the `if...then` statements will need to be removed in the schema and validated programatically.

src/API/REST/V3/Routes/Donors/DonorController.php
glaubersilva: Resolved in 232af9194f1a151bc2348caff4cecfd0838d4b4f

src/API/REST/V3/Routes/Subscriptions/Helpers/SubscriptionPermissions.php
glaubersilva: Resolved in 1b10f3e2d73b930a11cd193882e15f30a8c87d7e

src/API/REST/V3/Routes/Donations/DonationController.php
glaubersilva: Resolved in ffbce0968b5bc3d3bf02c179b81c85dfeb49e081 and 3e2936488dc3668876223baec9559f0b8b45adf0

src/Subscriptions/ViewModels/SubscriptionViewModel.php
glaubersilva: Resolved in 2fd4d66ec9d50b26537674276ec0c56cbc18169b

src/API/REST/V3/Routes/Donations/DonationController.php
jonwaldstein: Ah okay, so it looks like WP REST API supports _a subset of_ [json-schema draft 4](https://json-schema.org/specification-links#draft-4) so we just need to align with that spec when writing conditional validation

src/Subscriptions/ViewModels/SubscriptionViewModel.php
glaubersilva: It's not necessary because the subscription model doesn't have the email property, and we also aren't including it as additional properties (like we are doing for `firstName`, `lastName`, and `gateway`).

References: 

https://github.com/impress-org/givewp/blob/56f23fc4de9fef4bbea9b382e1c1f4bccff6e026/src/Subscriptions/Models/Subscription.php#L52-L68

https://github.com/impress-org/givewp/blob/2fd4d66ec9d50b26537674276ec0c56cbc18169b/src/Subscriptions/ViewModels/SubscriptionViewModel.php#L54-L61

src/API/REST/V3/Routes/Donations/DonationController.php
glaubersilva: The `validateSubscriptionRules` method it's a bunch of if statements. If I understand it well, we don't need to migrate that logic to the schema because it would not work there, so we can keep all validations made by this method. Is that correct?

src/API/REST/V3/Routes/Subscriptions/SubscriptionController.php
glaubersilva: Resolved in a878bef4728c22d4448f725bc87c7d66a1c24600

src/API/REST/V3/Routes/Donations/DonationController.php
jonwaldstein: @glaubersilva I think it's still possible to use the conditional logic in schema for this but would just have to follow the draft 4 spec.  I suggest seeing if cursor can write them for you just tell it to use the json-schema draft 4.  If that works than great, otherwise we can stick with the existing logic.

JasonTheAdams: It's more complicated than following draft 4. 😭 

The key thing to understand about the WP flavor of this is that each property is evaluated as an **independent** JSON schema. This why `'required' => true` is defined _inside_ of the property instead of outside of it — as you'd often see a `'required' => ['last_name', 'email']` kind of thing. So you **cannot** have anything outside of the property array that references multiple arrays — each property must be entirely self-contained as that is how it is validated.

Happy to hop on a call if it's still not clear.

src/API/REST/V3/Routes/Donations/DonationController.php
glaubersilva: Resolved in 168475725f1ff22247acef4c63439c7ccbbf7550 and 7a9a3cd7bc770b88b334e33cba01c6d50d936406 and 662ef0f8831342939a782af05b4cc4e3c2965888

src/API/REST/V3/Routes/Subscriptions/Helpers/SubscriptionFields.php
glaubersilva: This method is also used in the `update_item` method and cannot be a DTO due to the different use case there. However, I created a new `SubscriptionCreateData` DTO to be used in the `create_item` method.

Resolved in 37bc094defada7ed4efb504da54bcb5f584e6dea

src/API/REST/V3/Routes/Subscriptions/SubscriptionController.php
glaubersilva: Resolved in fdad899395ccb3b59921a0749e944ec2fbf2f60d

src/API/REST/V3/Routes/Donations/DonationController.php
glaubersilva: Nice to know about it:

> This why `'required' => true` is defined inside of the property instead of outside of it

I fixed it here: fdad899395ccb3b59921a0749e944ec2fbf2f60d


About other validation changes, since it is not something so straightforward and the sprint time is running out, I think we can revisit any changes related to other schema validations in subsequent PRs.

jonwaldstein: @glaubersilva sounds good

src/API/REST/V3/Routes/Donations/DataTransferObjects/DonationCreateData.php
```diff
@@ -0,0 +1,369 @@
...
+        // Validate donation creation
+        $this->validateCreateDonation();
+        
+        // Filter out auto-generated fields
+        $donationAttributes = array_filter($this->attributes, function ($key) {
+            return !in_array($key, ['id', 'createdAt', 'updatedAt'], true);
+        }, ARRAY_FILTER_USE_KEY);
+
+        return Donation::create($donationAttributes);
+    }
```
jonwaldstein: Hey @glaubersilva so the whole idea about `$data->toDonation()`naming pattern is simply to return a donation instance (not include validation or inserting into database) so that all the logic in the controller is declarative and readable in order like so:

**Controller example**
```php
if ($data->donationType->isRenewal()){
  $data->validateSubscriptionRules();
  
  $donation = $data->toRenewal();
  $donation->save();
  
} else if ($data->donationType->isSingle()){
  $data->validateCreateDonation();
  
  $donation = $data->toDonation();
  $donation->save();
}
```

[Here's an example of that approach](https://github.com/impress-org/givewp/blob/refactor/subscriptions-rest-api-GIVE-2675/src/DonationForms/Controllers/DonateController.php#L53)

If you want to keep your current approach than I would recommend changing the name to `$data->createDonation()` so it's more clear what it's doing.

src/API/REST/V3/Routes/Subscriptions/DataTransferObjects/SubscriptionCreateData.php
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Convert to Subscription model
+     *
+     * @unreleased
+     *
+     * @return Subscription
+     * @throws Exception
+     */
+    public function toSubscription(): Subscription
```
jonwaldstein: same note here about the use of this `toSubscription()` pattern

src/API/REST/V3/Routes/Donations/DataTransferObjects/DonationCreateData.php
```diff
@@ -0,0 +1,369 @@
...
+    {
+        if ($this->isRenewal) {
+            throw new DonationValidationException(
+                __('Cannot create single donation for renewal type', 'give'),
+                'invalid_donation_type',
+                400
+            );
+        }
+
+        $requiredFields = ['donorId', 'amount', 'gatewayId', 'mode', 'formId', 'firstName', 'email'];
```
jonwaldstein, line 80: For these required fields, I think these would apply to anytime we're creating a donation right? if that's the case they can just be added to the schema as required.

glaubersilva: @jonwaldstein We need to keep this validation here because there are differences between the types of donations. Renewals only require the `type` and `subscriptionId` parameters to create a donation, and other donations require these other ones. Since the scheme does not support if statements, we need to check it here.

src/API/REST/V3/Routes/Donations/DataTransferObjects/DonationCreateData.php
glaubersilva: Resolved in a740920b5e99d1c2a2edce2e69e114ced919f9a3

src/API/REST/V3/Routes/Subscriptions/DataTransferObjects/SubscriptionCreateData.php
glaubersilva: Resolved in a740920b5e99d1c2a2edce2e69e114ced919f9a3

src/API/REST/V3/Routes/Donations/DataTransferObjects/DonationCreateData.php
jonwaldstein: Ah okay, I think that's okay for now then

glaubersilva:
@jonwaldstein I'm using support for the top-level directory to add the classes useful for all resources. But for the classes that are specific only for some resources, I thought that the Helpers folder would be more suitable because it's used specially to clean up the controller classes and also to prevent code duplication at the resource level. 

Do you think it is better to rename it to Support as well?

image

jonwaldstein:
@glaubersilva thanks! honestly, a Helpers folder is kind of my last resort for these types of domain specific things. 

Instead, I would rather have a couple specific folders than a general purpose one until it's necessary.

Here's an example of what I mean:

```
Donations/Fields/DonationFields
Subscriptions/Fields/SubscriptionFields
Subscriptions/Permissions/SubscriptionPermissions
```

glaubersilva:
@jonwaldstein I changed it to follow your suggestions. See: https://github.com/impress-org/givewp/pull/8051/commits/d265aa5465b9f404fffe90d555b116e32f547fc5

PR #8035, opened 2025-07-24
Feature: Add Subscription stat cards to Overview

JoshuaHungDinh:
Resolves [GIVE-2699]

## Description
This PR introduces a SubscriptionStats component to the Subscription Details page Overview tab, displaying key metrics about a subscription's donations and payment progress. It also updates StatWidget to support additional value type.

Displays:
- Total contribution so far (sum of all donation amounts).
- Payments completed (completed installments out of total installments, or ∞ if ongoing).

Key Changes
- New Hook: useDonationsBySubscription. Retrieves donation records associated with a given subscription via useEntityRecords.

StatWidget Enhancements
- Supports value as number | string | React.ReactNode.
- Ensures numeric formatting by parsing values when applicable.

### Why
These changes improve the subscription management experience by giving administrators quick visibility into a subscriber’s payment history and total contributions, while keeping the UI consistent with existing dashboard components.
## Affects
Subscription page 

## Visuals

https://github.com/user-attachments/assets/1c44f088-98e7-47b4-8c7d-85041493fdb6

## Testing Instructions
- Create a form with subscriptions -> set a value/number of installments through the settings.
- Process a payment.
- View the subscription page to verify the amount for total contributions is accurate.
- Veriify the number of payments and installments is accurate.
- Set the number of installments to ongoing -> verify the lemniscate symbol is displayed.
- This updates the values passed to the StatWidget component - we will need to test the other screens to ensure no regressions occur.

## Pre-review Checklist 

-   [ ] Acceptance criteria satisfied and marked in related issue
-   [ ] Relevant `@unreleased` tags included in DocBlocks
-   [ ] Includes unit tests
-   [ ] Reviewed by the designer (if follows a design)
-   [ ] [Self Review](https://give.gitbook.io/development-manual/devops/github/code-reviews#self-review) of code and UX completed

[GIVE-2699]: https://stellarwp.atlassian.net/browse/GIVE-2699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8037, opened 2025-07-25
Feature: Add SubscriptionPaymentDetails to Overview

JoshuaHungDinh:
Resolves [GIVE-2668]

## Description
This feature adds a new SubscriptionPaymentDetails component to the subscription overview page, providing administrators with a comprehensive view of subscription payment information. The implementation includes:
Unified Payment Details Component: Created a reusable PaymentDetails component that works for both donations and subscriptions

## Affects
Subscription page

## Visuals

https://github.com/user-attachments/assets/77499a53-e790-4d3a-9150-9fcf3e8b6539

## Testing Instructions
- View the Donation details page & verify there are no regressions.
- View the Subscription details page & verify the data displayed is correct.
- Verify that the links for the Campaign, donor and gateway subvscription work properly.
-  Verify the "next payment" box has a limited & unlimited pill depending on the amount of installments
- Make a subscription with an isntalled gateway / uninstall the gateway if the gateway data does not show a notice should be visible to hand the missing information.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2668]: https://stellarwp.atlassian.net/browse/GIVE-2668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/PaymentGateways/Gateways/Stripe/StripePaymentElementGateway/StripePaymentElementGateway.php
+++ b/src/PaymentGateways/Gateways/Stripe/StripePaymentElementGateway/StripePaymentElementGateway.php
@@ -2,9 +2,14 @@
 
 namespace Give\PaymentGateways\Gateways\Stripe\StripePaymentElementGateway;
 
+use Exception;
 use Give\Donations\Models\Donation;
+use Give\Donations\Models\DonationNote;
 use Give\Framework\PaymentGateways\Commands\GatewayCommand;
+use Give\Framework\PaymentGateways\Commands\PaymentRefunded;
 use Give\Framework\PaymentGateways\Commands\RespondToBrowser;
+use Give\Framework\PaymentGateways\Contracts\PaymentGatewayRefundable;
+use Give\Framework\PaymentGateways\Exceptions\PaymentGatewayException;
 use Give\Framework\PaymentGateways\PaymentGateway;
 use Give\Framework\Support\Scripts\Concerns\HasScriptAssetFile;
 use Give\Helpers\Language;
@@ -14,7 +19,7 @@ use Stripe\Exception\ApiErrorException;
 /**
  * @since 3.0.0
  */
-class StripePaymentElementGateway extends PaymentGateway
+class StripePaymentElementGateway extends PaymentGateway implements PaymentGatewayRefundable
 {
     use StripePaymentElementRepository;
     use HasScriptAssetFile;
@@ -158,4 +163,44 @@ class StripePaymentElementGateway extends PaymentGateway
             ],
         ]);
     }
+
+    /**
+     * @unreleased
+     *
+     * @throws Exception
+     */
+    public function refundDonation(Donation $donation): PaymentRefunded
+    {
+        try {
+            $refund = $this->refundStripePayment($donation);
+
+            if ($refund->status !== 'succeeded') {
+                throw new Exception(__('Refund failed. Please check the Stripe dashboard for more details.', 'give'));
+            }
+
+            DonationNote::create([
+                'donationId' => $donation->id,
+                'content' => sprintf(
+                    __('Donation refunded in Stripe for transaction ID: %s', 'give'),
+                    $donation->gatewayTransactionId
+                ),
+            ]);
+
+            return new PaymentRefunded();
+        } catch (Exception $e) {
+            DonationNote::create([
+                'donationId' => $donation->id,
+                'content' => sprintf(
+                    __(
+                        'Error! Donation %s was NOT refunded. Find more details on the error in the logs at Donations > Tools > Logs. To refund the donation, use the Stripe dashboard tools.',
+                        'give'
+                    ),
+                    $donation->id
+                ),
+            ]);
+
+            throw new PaymentGatewayException(sprintf(__('Stripe API error: %s', 'give'), $e->getMessage()));
+        }
+    }
 }
+
--- a/src/PaymentGateways/Gateways/Stripe/StripePaymentElementGateway/StripePaymentElementRepository.php
+++ b/src/PaymentGateways/Gateways/Stripe/StripePaymentElementGateway/StripePaymentElementRepository.php
@@ -15,6 +15,7 @@ use Stripe\ErrorObject;
 use Stripe\Exception\ApiErrorException;
 use Stripe\Exception\InvalidRequestException;
 use Stripe\PaymentIntent;
+use Stripe\Refund;
 
 trait StripePaymentElementRepository
 {
@@ -30,6 +31,17 @@ trait StripePaymentElementRepository
             $data->toParams(),
             $data->toOptions($stripeConnectAccountId)
         );
+    }
+
+    /**
+     * @unreleased
+     * @throws ApiErrorException
+     */
+    protected function refundStripePayment(Donation $donation): Refund
+    {
+        return Refund::create([
+            'payment_intent' => $donation->gatewayTransactionId,
+        ]);
     }
 
     /**
--- a/src/PaymentGateways/PayPalCommerce/PayPalCommerce.php
+++ b/src/PaymentGateways/PayPalCommerce/PayPalCommerce.php
@@ -2,10 +2,13 @@
 
 namespace Give\PaymentGateways\PayPalCommerce;
 
+use Exception;
 use Give\Donations\Models\Donation;
-use Give\Framework\Exceptions\Primitives\Exception;
+use Give\Donations\Models\DonationNote;
 use Give\Framework\PaymentGateways\Commands\GatewayCommand;
 use Give\Framework\PaymentGateways\Commands\PaymentComplete;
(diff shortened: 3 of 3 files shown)


src/Admin/components/PaymentDetails/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+  infoCardTitle?: string;
+  infoCardBadgeLabel?: string | React.ReactNode;
+  infoCardClassName?: string;
+}
+
+/**
+ * @unreleased
+ * Generic donation | subscription details.
+ */
+export default function PaymentDetails({
```
jonwaldstein: I don't think this should be a shared component, reason being - it's very contextual about donations, subscriptions, which page it's on, and has a limited grid of specific cards that will struggle to scale if you want to add more to it.

I realize these grids of cards are very similar across screens however, they are still too contextual to pull out as a whole.  Instead, just keep the building block generic components like `Grid` and `Card` in the shared components and build out the context specific grid of these within their domains.

Essentially, if we're starting to get very contextual and data-limited in our components, it's usually a sign that they are either too specific to that domain - or pieces of them can be broken down into smaller reusable components that the domains can reuse.  


Here's silly question I like to ask myself when deciding what goes into a shared component library 😄 

>Is this component dumb enough to be shared?

JoshuaHungDinh: @jonwaldstein I think this is a step in the right direction - but it also feels very card specific. Can I get some feedback on this please.

e4547850716ede94946382b8257837c1bf28f805

src/Donations/resources/admin/components/DonationDetailsPage/Tabs/Overview/index.tsx
```diff
@@ -24,9 +24,8 @@ export default function DonationDetailsPageOverviewTab() {
             <DonationStats donation={donation} isResolving={isResolvingDonation} />
 
             <div className={styles.left}>
-                <DonationSummaryGrid
-                    donation={donation}
-                />
+                <DonationPaymentDetails donation={donation}/>
```
jonwaldstein: Honestly, I don't think we should be touching this here as it increasing the scope of this PR to the donation screen.  Was the reason just for the component naming?

src/Admin/components/PaymentDetails/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+                    href={`edit.php?post_type=give_forms&page=give-campaigns&id=${campaign?.id}&tab=overview`}
+                    className={styles.campaignLink}
+                >
+                    {campaign?.title}
+                </a>
+            )}
+        </div>
+    );
+}
+
```
jonwaldstein: Okay so i'll use this one as an example but the same feedback applies to all these components.  

This is something very contextual that is responsible for fetching data and has hardcoded values in it.  For that reason, it's not something I would elevate to a shared component across all admin screens. Instead I would break only the reusable component down like this below and use them as building blocks inside your domain.  It's also _okay_ that not every single thing is a separate component too.

**Example of a shared component**
Notice this is purely presentational, it should display our values - offering us the value of not having to think about the _design_ of the component.

```js
export function GridCard({children, heading, headingId}: {children: React.ReactNode, heading: string, headingId: string}) {

    return (
        <div className={styles.card} role="region" aria-labelledby={`${headingId}-label`}>
            <h3 id={headingId}>{heading}</h3>
            {children}
        </div>
    );
}
```

**Example of domain specific component**
Notice we are now fetching data and providing context specific text and values.  This is something that is fine to repeat in several domains as the implementation will likely be slightly different, and if it's not thats' okay too because the majority of the logic is reusable anyway.

```js
function CampaignCard({donation}: {donation: Donation}) {
    const {campaign, hasResolved: hasResolvedCampaign} = useCampaignEntityRecord(donation?.campaignId);

    return (
        <GridCard
            heading={__('Campaign name', 'give')}
            headingId="campaign-name-label"
        >
            {!hasResolvedCampaign && <Spinner />}
            {hasResolvedCampaign && campaign && (
                <a
                    href={`edit.php?post_type=give_forms&page=give-campaigns&id=${campaign?.id}&tab=overview`}
                    className={styles.campaignLink}
                >
                    {campaign?.title}
                </a>
            )}
        </GridCard>
    );
}
```

For the reasons above, I don't see any reason for us to have a shared `PaymentDetails` components as really all we need are building blocks for a Grid of Cards - everything else is contextual that can remain in it's respective domain:

```js
<Grid>
  <GridCard>
  Campaign Details
  </GridCard>
  <GridCard>
   Donor Details
  </GridCard>
</Grid>
```

cc: @pauloiankoski for visibility

src/Admin/components/PaymentDetails/index.tsx
jonwaldstein: Thanks for the effort! feedback provided here https://github.com/impress-org/givewp/pull/8037#discussion_r2257016397

src/Donations/resources/utils.ts
```diff
@@ -1,4 +1,5 @@
 import {useEntityRecord} from '@wordpress/core-data';
+import { EntityRecordResolution } from '@wordpress/core-data/build-types/hooks/use-entity-record';
```
jonwaldstein: See comment here about this https://github.com/impress-org/givewp/pull/8035#discussion_r2256940078

JoshuaHungDinh: Resolved: https://github.com/impress-org/givewp/pull/8053

src/Donations/resources/admin/components/DonationDetailsPage/Tabs/Overview/index.tsx
JoshuaHungDinh: It was but it's been changed back since we are going with the Grid components!

Screenshot 2025-08-06 at 10 35 18 PM

src/Admin/components/Grid/index.tsx
```diff
@@ -0,0 +1,46 @@
+import { __ } from '@wordpress/i18n';
```
glaubersilva: Let's remove this unused import.

src/Donations/resources/admin/components/DonationDetailsPage/Tabs/Overview/DonationSummaryGrid/index.tsx
```diff
@@ -3,12 +3,14 @@ import classnames from 'classnames';
 import OverviewPanel from '@givewp/src/Admin/components/OverviewPanel';
```
glaubersilva, line 2: Let's remove the `import classnames from 'classnames';` which is not used.

src/Donations/resources/utils.ts
glaubersilva: I think we can remove this unused import as well.

glaubersilva (changes requested):
@JoshuaHungDinh I created a Donation with Braintree, then I deactivated the Braintree add-on, and the gateway information is not displayed, which is expected because the API returns an empty array when the gateway add-on is not available to retrieve the info. 

However, we have access to the `gatewayId` in the subscription or donation models, which I think can be used to inform the admin that the information for that `gatewayId` wasn't found, so maybe we can add a message in this blank space to say that they need to activate the add-on for that gatewayId in specific to be able to see this info.

We can also just add a "Not found payment gateway" message. Maybe worth checking with @jdghinson 

image

src/Admin/components/Grid/index.tsx
JoshuaHungDinh: Resolved: 30870e00cd69fd3fe09a31d6668bf42df9acf7ef

src/Donations/resources/admin/components/DonationDetailsPage/Tabs/Overview/DonationSummaryGrid/index.tsx
JoshuaHungDinh: Resolved: 83060fec1dd7336c5e01c90b291fd36d5c6eca50

src/Donations/resources/utils.ts
JoshuaHungDinh: Yup! Must have came back in a merge accident haha Resolved: 77b2e24957ac3a14fe89215161cd77ac8f2edcde

JoshuaHungDinh:
@glaubersilva So none of the donation?.gateway data is available if the gateway is not active/installed?

glaubersilva:
@JoshuaHungDinh Exactly! As @pauloiankoski  points out in the meeting today, it was generating a fatal error before, so I fixed it today on [**this commit**](https://github.com/impress-org/givewp/pull/8051/commits/be15ee90887fec8a1c6800c8d1f1d838764c989b), and now it's returning an empty array when the gateways' add-ons are not installed.

PR #8038, opened 2025-07-25
Feature: add SubscriptionSummary to Overview

JoshuaHungDinh:
Resolves [GIVE-2668]

## Pre-review Checklist

-   [ ] Includes unit tests


src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Overview/SubscriptionSummary/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+          value: dateI18n('jS M, Y', subscription?.createdAt?.date, undefined),
+        },
+        {
+          label: __('End date', 'give'),
+          value: endDate ? dateI18n('jS M, Y', endDate, undefined) : __('Ongoing', 'give'),
+        },
+        {
+          label: __('Donation form', 'give'),
+          value: (
+            <a className={styles.link} href={`${adminUrl}/edit.php?post_type=give_forms&page=givewp-form-builder&donationFormID=${donation?.formId}&locale=en_US`} target="_blank" rel="noopener noreferrer">
```
jonwaldstein: careful about the hardcoded `locale=en_US` here, might want to ask @glaubersilva about that

src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Overview/SubscriptionSummary/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+import Header from '@givewp/src/Admin/components/Header';
+import OverviewPanel from '@givewp/src/Admin/components/OverviewPanel';
+import SummaryTable, {SummaryItem} from '@givewp/src/Admin/components/SummaryTable';
+import { Donation } from '@givewp/donations/admin/components/types';
+import styles from './styles.module.scss';
+
+/**
+ * @unreleased
+ */
+const calculateEndDate = (subscription: any): string | null => {
```
jonwaldstein, line 25: Let's add some descriptions here to explain how this is being calculated

src/Admin/components/SummaryTable/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const isSingleValue = React.isValidElement(value) || typeof value === 'string';
+
+    if (isObjectWithValues) {
+      return (
+        <div className={styles.summaryTableValues}>
+          <p>{value.value1}</p>
+          <p>{value.value2}</p>
+        </div>
+      );
+    }
```
jonwaldstein: What's the reason a summary item would have 2 values? Forgive my ignorance, but I feel like this could be simplified by just requiring the value to be either a string or function that can dump whatever it wants instead of limiting to 1 or 2 values.

src/Donors/resources/admin/components/DonorDetailsPage/Tabs/Overview/DonorSummary/index.tsx
```diff
@@ -1,26 +1,12 @@
...
- */
-export type SummaryItem = {
-    label: string;
-    value: string | {
-        value1: string;
-        value2: string;
-    };
-    isPill?: boolean;
-};
+import SummaryTable, {SummaryItem} from '@givewp/src/Admin/components/SummaryTable';
```
jonwaldstein: Do we have a better alias for pulling in shared components?  It would be nice to have something like:

```js
import {SummaryTable} from `@givewp/admin/components`;
```

src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Overview/SubscriptionSummary/index.tsx
glaubersilva: We also need consider the `campaignId` and `p2p` params, so I would say that the best approach here would be create the edit link in the backend with all necessary parameters and expose it to the frontend. 

You can do it in the backend this way:

`FormBuilderRouteBuilder::makeEditFormRoute($donationFormId, Language::getLocale())->getUrl()`

glaubersilva: Oh! You don't need explicitly set the locale because the the constructor of the class already do it, so it should be enough:

`FormBuilderRouteBuilder::makeEditFormRoute($donationFormId)->getUrl()`

src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Overview/SubscriptionSummary/index.tsx
JoshuaHungDinh: Resolved: 1a0403ece44ac467f400229e0780165265c2f524

src/Donors/resources/admin/components/DonorDetailsPage/Tabs/Overview/DonorSummary/index.tsx
JoshuaHungDinh: How does this look for export/imports? 9c40d8819fa9dbc80062e2cbdf482d5df47b32d2

jonwaldstein: Solid!

src/Admin/components/SummaryTable/index.tsx
JoshuaHungDinh: Yea I see what you mean. We can pass the jsx from the parent as a single value.
Resolved: ba0534623fd5f16d01614d86b4a415ecb7b3999f

src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Overview/SubscriptionSummary/index.tsx
JoshuaHungDinh: @jonwaldstein @glaubersilva Any recommendations on where we should be passing it from?

jonwaldstein: You can add this to `GiveSubscriptionOptions` [here](https://github.com/impress-org/givewp/blob/0aee182a7beabd7518c39decba728c085bba9092/src/Subscriptions/Actions/LoadSubscriptionAdminOptions.php#L31)

jonwaldstein: @JoshuaHungDinh maybe we should drop the link for now and circle back to this as we don't have access to the formId during page load

JoshuaHungDinh: Yea I was trying to figure out if it made sense to include it in an entityRecord. Maybe through our `useForm` hook?

jonwaldstein: hmm maybe,  on another note It would also be kind of cool to have a short-link generator in the future so you could do something like
`/?givewp-link=donation-form-edit&id=1` and the server would convert it to the correct path. 

I think for now we can either drop the link or drop the locale specifically and circle back to this.  The weird thing about a donation form link is it just links to the form builder which i'm not even sure how useful that is in this context 🤔

JoshuaHungDinh: Resolved: 479f691f5c50aec468888379246f9a26051a5cc4

src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Overview/SubscriptionSummary/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+          value: dateI18n('jS M, Y', subscription?.createdAt?.date, undefined),
+        },
+        {
+          label: __('End date', 'give'),
+          value: endDate ? dateI18n('jS M, Y', endDate, undefined) : __('Ongoing', 'give'),
+        },
+        {
+          label: __('Donation form', 'give'),
+          value: (
+            <a className={styles.link} href={`${adminUrl}/edit.php?post_type=give_forms&page=givewp-form-builder&donationFormID=${donation?.formId}`} target="_blank" rel="noopener noreferrer">
```
glaubersilva: This is generating a link with a double slash, as you can see in the screenshot: 

image

src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Overview/index.tsx
```diff
@@ -2,21 +2,22 @@
 import { __ } from "@wordpress/i18n";
```
glaubersilva: Let's remove this unused import.

src/API/REST/V3/Routes/Donations/DonationController.php
```diff
@@ -266,7 +266,21 @@ public function get_items($request)
...
+            $donation = Donation::find($id);
+        } 
+
+        if($subscriptionId) {
+            $donation = Donation::query()
+                ->where('give_donationmeta_attach_meta_subscriptionId.meta_value', $subscriptionId)
+                ->limit(1)
+                ->get();
+        }
+
```
glaubersilva: I think we can remove it, right? I saw that you are using the proper `get_items` endpoint to retrieve the donations associated with the subscription.

src/Admin/components/SummaryTable/index.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import classnames from 'classnames';
+import Header from '@givewp/src/Admin/components/Header';
+import OverviewPanel from '@givewp/src/Admin/components/OverviewPanel';
```
glaubersilva: Let's remove these 2 unused imports.

src/API/REST/V3/Routes/Donations/DonationController.php
JoshuaHungDinh: get_items does return an array of the donations which I think in theory should all be the same but I was going based off this previous review comment. https://github.com/impress-org/givewp/pull/8035

Screenshot 2025-08-07 at 9 45 58 AM

src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Overview/SubscriptionSummary/index.tsx
JoshuaHungDinh: Nice catch!
Resolved: 8162a1a8d9cd734d0c02fd84bcf3b4a72899afe4

src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Overview/index.tsx
JoshuaHungDinh: 👍 Resolved: 7c009b95507206324f625d106e1987ebcd76319e

src/Admin/components/SummaryTable/index.tsx
JoshuaHungDinh: 👍 Resolved: d4fdcc3c57b8307c2b6d33f75f6d8769fba2d4c4

src/API/REST/V3/Routes/Donations/DonationController.php
glaubersilva: @JoshuaHungDinh About it:

> It does return an array of the donations

I don't think this is right since this logic lives inside the `get_item` endpoint, which is designed to return only a single donation: **/givewp/v3/donations/{id}**

image

However, I can see that in the codebase you are using the `get_items` which is the proper endpoint design to return multiple donations: `/givewp/v3/donations?subscriptionId=${subscriptionId}&mode=${mode}`

If you put breakpoints in the `get_item` and `get_items` methods, and go to the subscription details page will see that this code is never triggered.

So, I think it is safe to remove that since it seems not to be used in any place and conceptually is wrong.

JoshuaHungDinh: I think Im a little confused - I was under the impression we should be able to get a single donation based on the subscriptionId from what Jon said. Is this possible?

JoshuaHungDinh: @glaubersilva 
Resolved: c01b5c02c92515148535bda315464b58fa8c3dfd 5d19b37beec904d2ce276ffc0dd7a1c1ce1864f9 eb97df260515ddad37ec0bd8240313aa0581a84d

Issue #8042, opened 2025-07-29
MASSIVE SECURITY ISSUE - GiveWP plugin is exposing donors name and email addresses directly in the source code


kxkv:
First of all, I am not a user of this plugin myself. However, I have discovered that for several sites that are running this plugin, the account information of the donors are exposed directly in the source code of the site.

Names and email addresses are exposed inside the donors element, in the following piece of code:

```
<script id="give-donation-options-js-extra">
var GiveDonationOptions = {"isAdmin":"", ...  "donors": {}
</script>
```

I do not want to expose any sites directly here, but, if you do a simple search for sites using this plugin it will not take long before you will have several examples of this having happened.

jonwaldstein:
This has been resolved in Give 4.6.1 released today

dschaper:
By the way, no one from GiveWP has replied to my emails nor bothered to advise that this patch was released. I had to find it just by coincidence.

jonwaldstein:
Hey fellow devs! 👋 

First let me say, we really appreciate you all reaching out about this.  We are grateful to have this powerful and helpful open-source community we have built over the past 10 years building GiveWP with people like yourselves 🙂.  I personally also emphasize with your concerns as we really do try our best to ensure we are releasing the most secure and reliable code, so this was not a fun thing to discover on our end.  Luckily the exposed information was basic contact information (name, email, and donor ID) and not something worse like addresses and payment details.  Some of that information is already available publicly if using one of our blocks & short-codes like our Donor Wall - however I understand emails can be more sensitive.

In terms of how this happened, our small dev team recently updated our _10 year old legacy admin screens_ to be more modern, using React and the WP REST API.  This was a **huge** amount of work for us that touched so many parts of our codebase and will effectively remove a great portion of technical debt we have acquired over the years - unlocking more opportunities for newer features 😄.  In order to meet the growing demands of our valued customers we scoped this work in small sprints that we were able to release in pieces.  Some of this work is currently still in progress as we are iterating on some features. 

Although we try our best to peer review our code as thoroughly as possible, along with a QA process that is meant to catch any regression and compatibility bugs - we’re all humans and sometimes things can get overlooked like this.  The donor information specifically was meant to be used in these new admin screens but was not using the proper WP hook for enqueuing in the admin only 🤦. 

[...]

PR #8043, opened 2025-07-29
Fix: Move admin options loading to admin_enqueue_scripts hook

pauloiankoski:
Resolve [GIVE-2774]

## Description

This PR fixes an issue where admin-facing options were being loaded in all pages. The actions were previously hooked to the `init` action, but they should be loaded on `admin_enqueue_scripts` to ensure they're only available in the admin context when admin screens are being set up.

This change ensures admin screen options are properly scoped to admin-only functionality. Additionally, the hooks were wrapped by conditionals, ensuring they are called only where they are needed.

## Affects
- Campaign admin screens and options
- Donation admin screens and options  
- Donor admin screens and options

## Visuals

No visual changes - this is a backend hook optimization.

## Testing Instructions

1. Navigate to any admin screen related to campaigns, donations, or donors
2. Verify that all admin screen options are properly loaded and functional
3. Check that options are not being loaded unnecessarily on the frontend
4. Ensure no JavaScript errors or missing dependencies in admin screens

## Pre-review Checklist

- [ ] Includes unit tests

[GIVE-2774]: https://stellarwp.atlassian.net/browse/GIVE-2774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/Unit/ViewModels/FormBuilderViewModelTest.php
+++ b/tests/Unit/ViewModels/FormBuilderViewModelTest.php
@@ -98,6 +98,7 @@ class FormBuilderViewModelTest extends TestCase
                     'primaryColor' => '',
                     'secondaryColor' => '',
                 ],
+                'showFormGoalNotice' => true,
             ],
             $viewModel->storageData($formId)
         );


src/Campaigns/ServiceProvider.php
```diff
@@ -233,10 +233,11 @@ private function registerCampaignBlocks()
      */
     private function loadCampaignOptions()
     {
-        Hooks::addAction('init', LoadCampaignOptions::class);
+        Hooks::addAction('admin_init', LoadCampaignOptions::class);
```
jonwaldstein: @pauloiankoski If these are specific to loading scripts shouldn't we be using `admin_enqueue_scripts`?

src/Campaigns/ServiceProvider.php
```diff
@@ -229,11 +229,14 @@ private function registerCampaignBlocks()
...
     /**
+     * @unreleased Move to admin_init hook
      * @since 4.0.0
      */
     private function loadCampaignOptions()
     {
-        Hooks::addAction('init', LoadCampaignOptions::class);
+        if (CampaignsAdminPage::isShowingDetailsPage()) {
+            Hooks::addAction('admin_init', LoadCampaignOptions::class);
+        }
```
jonwaldstein: @pauloiankoski couple things here:

1. We should always be accessing php globals inside an action or else this will get checked on every page load since this is the boot of the service provider - so I recommend putting those conditionals _inside_ the action call to scope their usage.

```php
add_action('hook', function(){
    if ($_GET['foo'] === 'bar') {
        // do something
    }
});
```
2. If these are specific to loading scripts shouldn't we be using `admin_enqueue_scripts`?

3. Are these scripts being used in any front-end blocks or just the admin? If just admin - we should consider changing these class names to be more specific like `LoadAdminCampaignScripts` and if necessary additional `LoadPublicCampaignScripts` for anything front facing

jonwaldstein (changes requested):

pauloiankoski:
@jonwaldstein on https://github.com/impress-org/givewp/pull/8043/commits/12e01c5c5427e55481b9fb5b27a9ee9fdd655f5a I’m creating a public-facing localization (which may be refactored or even removed in the future). For now, it’s only being loaded when a Campaign block is present in the current post or page.

Additionally, we’re now also loading the Campaign admin options as part of the ListTable assets load process.

jonwaldstein (approved):
@pauloiankoski nice job man and thanks for your help!

I tested this locally and it appears all the campaign blocks and admin pages are still working properly 👍

DanielRuf:
@centromere yes, see https://github.com/impress-org/givewp/issues/8042

PR #8047, opened 2025-08-01
Feature: campaigns stats cache

alaca:
## Description

This PR adds a caching layer for the CampaignsDataRepository. A batch migration is added to handle the initial cache, and on each donation, an async action will be added to update the cache.

## Affects

- The campaign list table stats (goal, revenue, etc.)
- The campaign grid goal progress

## Visuals

N/A

## Testing Instructions

Create 2,3 campaigns
Generate at least 20K+ donations
Donate to each campaign (check the campaigns list table - the numbers will be updated once the background action runs)
Check the loading speed of the campaigns list table

**We need to battle test this before releasing.**

**Places to check:**
- The campaign list table (loading times and goal/revenue columns)
- The campaign grid block (loading times and goal data)

**Scenarios to check:**

- When an existing site updates to this version, the migration is run successfully and their campaign goal data is accurate.
- When donating via, v2 forms, v3 forms, manual donations, and manual renewals - the campaign goal/revenue data on the list table is accurate (note: this can take 1-2 minutes per transaction to be reflected).  You can look at action scheduler to see what’s being added to the cache.
- When a subscription renewal is recorded from any core gateway include Stripe (legacy and Payment Element) and PayPal the campaign goal/revenue data on the list table is accurate 
- Finally, this is tested on a high volume customer staging site to make sure it will work for edge cases. We also want to confirm the list table load time is significantly reduced.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Campaigns/Actions/CacheCampaignData.php
+++ b/src/Campaigns/Actions/CacheCampaignData.php
@@ -42,7 +42,7 @@ class CacheCampaignData
     public function registerAction()
     {
         add_action('givewp_cache_campaign_data', function (int $campaignId) {
-            $this->handleCache($campaignId);
+            give(CacheCampaignData::class)->handleCache($campaignId);
         });
     }
 
@@ -75,7 +75,6 @@ class CacheCampaignData
         ], $campaignsSubscriptionData);
 
         $donations = CampaignsDataQuery::donations([$campaign->id]);
-        $subscriptions = CampaignsDataQuery::subscriptions([$campaign->id]);
 
         $isCached = $isSubscriptionCached = false;
 
@@ -106,6 +105,9 @@ class CacheCampaignData
 
         // Update campaign subscriptions data
         if (defined('GIVE_RECURRING_VERSION')) {
+
+            $subscriptions = CampaignsDataQuery::subscriptions([$campaign->id]);
+
             foreach ($campaignsSubscriptionData['amounts'] as $i => $data) {
                 if ($data['campaign_id'] == $campaign->id) {
                     $isSubscriptionCached = true;
--- a/src/Campaigns/Actions/CacheCampaignData.php
+++ b/src/Campaigns/Actions/CacheCampaignData.php
@@ -1,0 +1,123 @@
+<?php
+
+namespace Give\Campaigns\Actions;
+
+use Give\Campaigns\CampaignsDataQuery;
+use Give\Campaigns\Models\Campaign;
+use Give\Donations\Models\Donation;
+
+/**
+ * @unreleased
+ *
+ * @uses givewp_donation_created hook
+ * @uses givewp_donation_updated hook
+ *
+ * Action used to update campaign's stats data
+ *
+ */
+class CacheCampaignData
+{
+    /**
+     * @unreleased
+     */
+    public function __invoke(Donation $donation): void
+    {
+        if ( ! $donation->status->isComplete()) {
+            return;
+        }
+
+        $campaign = Campaign::findByFormId($donation->formId);
+
+        if ( ! $campaign) {
+            return;
+        }
+
+        $campaignsData = give_get_option('give_campaigns_data', []);
+        $campaignsSubscriptionData = give_get_option('give_campaigns_subscription_data', []);
+
+
+        $donations = CampaignsDataQuery::donations([$campaign->id]);
+        $subscriptions = CampaignsDataQuery::subscriptions([$campaign->id]);
+
+        $isCached = $isSubscriptionCached = false;
+
+        // Update cache only if it exists for this campaign
+        foreach ($campaignsData['amounts'] as $i => $data) {
+            if ($data['campaign_id'] == $donation->campaignId) {
+                $isCached = true;
+
+                $campaignsData['amounts'][$i] = $donations->collectIntendedAmounts()[0];
+                break;
+            }
+        }
+
+        foreach ($campaignsData['donationsCount'] as $i => $data) {
+            if ($data['campaign_id'] == $donation->campaignId) {
+                $campaignsData['donationsCount'][$i] = $donations->collectDonations()[0];
+                break;
+            }
+        }
+
+        foreach ($campaignsData['donorsCount'] as $i => $data) {
+            if ($data['campaign_id'] == $donation->campaignId) {
+                $campaignsData['donorsCount'][$i] = $donations->collectDonors()[0];
+                break;
+            }
+        }
+
+
+        // Update campaign subscriptions data
+        if (defined('GIVE_RECURRING_VERSION')) {
+            foreach ($campaignsSubscriptionData['amounts'] as $i => $data) {
+                if ($data['campaign_id'] == $donation->campaignId) {
+                    $isSubscriptionCached = true;
+
+                    $campaignsSubscriptionData['amounts'][$i] = $subscriptions->collectInitialAmounts()[0];
+                    break;
+                }
+            }
+
+            foreach ($campaignsSubscriptionData['donationsCount'] as $i => $data) {
+                if ($data['campaign_id'] == $donation->campaignId) {
+                    $campaignsSubscriptionData['donationsCount'][$i] = $subscriptions->collectDonations()[0];
+                    break;
+                }
+            }
+
+            foreach ($campaignsSubscriptionData['donorsCount'] as $i => $data) {
+                if ($data['campaign_id'] == $donation->campaignId) {
+                    $campaignsSubscriptionData['donorsCount'][$i] = $subscriptions->collectDonors()[0];
+                    break;
+                }
+            }
+        }
+
+
+        // Save updated cache
+        if ($isCached) {
+            give_update_option('give_campaigns_data', $campaignsData);
+        }
+
+        if ($isSubscriptionCached) {
+            give_update_option('give_campaigns_subscriptions_data', $campaignsSubscriptionData);
+        }
+
+        if ($isCached || $isSubscriptionCached) {
+            return;
+        }
+
+        give_update_option('give_campaigns_data', [
+            'amounts' => array_merge($campaignsData['amounts'] ?? [], $donations->collectInitialAmounts()),
+            'donationsCount' => array_merge($campaignsData['donationsCount'] ?? [], $donations->collectDonations()),
+            'donorsCount' => array_merge($campaignsData['donorsCount'] ?? [], $donations->collectDonors()),
+        ]);
+
+        if (defined('GIVE_RECURRING_VERSION')) {
+            give_update_option('give_campaigns_subscriptions_data', [
+                'amounts' => array_merge($campaignsSubscriptionData['amounts'] ?? [], ...$subscriptions->collectIntendedAmounts()),
+                'donationsCount' => array_merge($campaignsSubscriptionData['donationsCount'] ?? [], ...$subscriptions->collectDonations()),
+                'donorsCount' => array_merge($campaignsSubscriptionData['donorsCount'] ?? [], ...$subscriptions->collectDonors()),
+            ]);
+        }
+    }
+}
--- a/src/Campaigns/Actions/CacheCampaignData.php
+++ b/src/Campaigns/Actions/CacheCampaignData.php
@@ -35,7 +35,6 @@ class CacheCampaignData
         $campaignsData = give_get_option('give_campaigns_data', []);
         $campaignsSubscriptionData = give_get_option('give_campaigns_subscription_data', []);
 
-
         $donations = CampaignsDataQuery::donations([$campaign->id]);
         $subscriptions = CampaignsDataQuery::subscriptions([$campaign->id]);
 
@@ -43,7 +42,7 @@ class CacheCampaignData
 
         // Update cache only if it exists for this campaign
         foreach ($campaignsData['amounts'] as $i => $data) {
-            if ($data['campaign_id'] == $donation->campaignId) {
+            if ($data['campaign_id'] == $campaign->id) {
                 $isCached = true;
 
                 $campaignsData['amounts'][$i] = $donations->collectIntendedAmounts()[0];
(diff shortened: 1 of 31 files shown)


alaca:
@jonwaldstein, the scope of this task was to resolve the loading issues with the Campaigns list table, which is done by adding a caching layer to the `CampaignsDataRepository`. We use that class for eager loading the campaigns data in the campaigns list table and campaign endpoints. Regarding the Donation model, that is not affected at all. The same goes for the `Give_Payment` and `Give_Subscription` classes. Regarding hooking into `give_insert_payment` and `give_recurring_add_subscription_payment` hooks, there is no need to do that. We are hooking into `givewp_donation_created` and `givewp_donation_updated`. Also, in that action, we are not using the campaign ID, which is not set at that time on the Donation object. We are using `formId` to get the campaign, and we fetch the campaign stats data only for that campaign and store them in cache.

src/Campaigns/Actions/CacheCampaignData.php
```diff
@@ -0,0 +1,161 @@
...
+    /**
+     * @unreleased
+     */
+    public function __invoke(Donation $donation): void
+    {
+        if ( ! $donation->status->isComplete()) {
+            return;
+        }
+
+        as_enqueue_async_action('givewp_cache_campaign_data', [$donation->formId], 'givewp_campaigns_cache');
```
jonwaldstein: Is there any reason we can't just pass the `$donation->campaignId` directly?  It should be [available](https://github.com/impress-org/givewp/blob/ac3ca0772449492f40cee289d4eb8c11e21142f1/src/Donations/Repositories/DonationRepository.php#L433) by the time the model action is fired.

Alternatively we could just pass the `$donation->id` and use the donation model in the handle cache logic as that would be more flexible for the future

src/Campaigns/ServiceProvider.php
```diff
@@ -128,6 +133,15 @@ private function registerActions(): void
...
         Hooks::addAction('transition_post_status', PreventDeleteDefaultForm::class, 'preventTrashStatusChange', 10, 3);
 
+        /**
+         * @unreleased cache campaign stats data
+         */
+        give(CacheCampaignData::class)->registerAction();
+
+        Hooks::addAction('give_insert_payment', CacheCampaignData::class);
+        Hooks::addAction('give_update_payment_status', CacheCampaignData::class);
+        Hooks::addAction('give_recurring_update_subscription_status', CacheCampaignData::class);
```
jonwaldstein: I think this is the wrong hook:

```suggestion
        Hooks::addAction('give_recurring_add_subscription_payment', CacheCampaignData::class);
```

src/Campaigns/Actions/CacheCampaignData.php
```diff
@@ -0,0 +1,162 @@
...
+ */
+class CacheCampaignData
+{
+    /**
+     * @unreleased
+     */
+    public function __invoke(int $donationId): void
+    {
+        $donation = Donation::find($donationId);
+
```
jonwaldstein, line 27: Fallback in case the model is not found:

```suggestion
        $donation = Donation::find($donationId);
        if (!$donation){
          return;
        }

```

alaca: Resolved: https://github.com/impress-org/givewp/pull/8047/commits/6d6fc95202de6718779114932e623fe860596eb1

src/Campaigns/ServiceProvider.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/8047/commits/aca8eab9d5209fb3d1c54122d0d6fdd9c62814f5

jonwaldstein:
@alaca do you think we should remove [this check](https://github.com/impress-org/givewp/blob/8b06b4db839d14456ec311a31b448ef7b983e70d/src/Campaigns/Controllers/CampaignRequestController.php#L82) for 3 campaigns and always use the cache now?

glaubersilva:
@jonwaldstein @alaca  About the Webhooks add-on, I would say that the test classes there can be useful to check how to create donations using legacy code and test how it impacts the cache values. 

References:

https://github.com/impress-org/give-webhooks/blob/develop/tests/Unit/Events/Donations/TestDonationCreated.php#L54

https://github.com/impress-org/give-webhooks/blob/develop/tests/Unit/Traits/Helpers.php#L147-L226

alaca:
> @alaca do you think we should remove [this check](https://github.com/impress-org/givewp/blob/8b06b4db839d14456ec311a31b448ef7b983e70d/src/Campaigns/Controllers/CampaignRequestController.php#L82) for 3 campaigns and always use the cache now?

@jonwaldstein good catch. Now that we are using cache, we can remove that.

jonwaldstein:
@alaca okay sounds good. Please go ahead and update the PR to remove those checks.  The rest of the PR is looking good and I updated the testing instructions to help with QA hand-off.  Let's have @glaubersilva do a final review once you finish the update

src/Campaigns/Migrations/CacheCampaignsData.php
```diff
@@ -0,0 +1,181 @@
...
+                ),
+                'donorsCount' => array_merge(
+                    $campaignsData['donationsCount'] ?? [],
+                    $donations->collectDonors()
+                ),
+            ]);
+
+            // Set subscriptions data
+            if (defined('GIVE_RECURRING_VERSION')) {
+                $subscriptionsData = get_option('give_campaigns_data', []);
```
glaubersilva, line 98: Let's remove this unused variable.

src/Campaigns/Migrations/CacheCampaignsData.php
```diff
@@ -0,0 +1,181 @@
...
+                    $donations->collectDonations()
+                ),
+                'donorsCount' => array_merge(
+                    $campaignsData['donationsCount'] ?? [],
+                    $donations->collectDonors()
+                ),
+            ]);
+
+            // Set subscriptions data
+            if (defined('GIVE_RECURRING_VERSION')) {
```
glaubersilva, line 97: Not sure if we should restrict the update of the `give_campaigns_subscriptions_data` option only when the recurring add-on is installed. If someone that have a recurring add-on deactivated for some reason runs this migration, the info will not be properly filled out. Since the `CampaignsDataQuery::subscriptions` live in Give core, I think we can always run it just to make sure.

src/Campaigns/Actions/CacheCampaignData.php
```diff
@@ -0,0 +1,179 @@
...
+        if (!$donation) {
+            return;
+        }
+
+        if ($donation->status->isComplete() || $donation->status->isRenewal()) {
+            as_enqueue_async_action('givewp_cache_campaign_data', [$donation->campaignId], 'givewp_campaigns_cache');;
```
glaubersilva: This line has a double semicolon at the end. Also, we can use our facade class for it. 

So we would have something like this:

```
AsBackgroundJobs::enqueueAsyncAction(
    'givewp_cache_campaign_data',
    [$donation->campaignId],
    'givewp_campaigns_cache'
);
```

src/Campaigns/Actions/CacheCampaignData.php
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /**
+     * Register AS action
+     *
+     * @unreleased
+     */
+    public function registerAction()
+    {
+        add_action('givewp_cache_campaign_data', function (int $campaignId) {
```
glaubersilva: Let's keep the consistency with other classes, where we always register the actions in the service provider. 

So, we can remove this registerAction() method and add something like this in the service provider: 

```
add_action('givewp_cache_campaign_data', function (int $campaignId) {
    give(CacheCampaignData::class)->handleCache($campaignId);
});
```

src/Campaigns/Actions/CacheCampaignData.php
```diff
@@ -0,0 +1,179 @@
...
+        foreach ($campaignsData['donorsCount'] as $i => $data) {
+            if ($data['campaign_id'] == $campaign->id) {
+                $campaignsData['donorsCount'][$i] = $donations->collectDonors()[0];
+                break;
+            }
+        }
+
+
+        // Update campaign subscriptions data
+        if (defined('GIVE_RECURRING_VERSION')) {
```
glaubersilva, line 96: We are querying the subscriptions a few lines above in this way: 

`$subscriptions = CampaignsDataQuery::subscriptions([$campaign->id]);`

But in this condition, we are preventing the update if the recurring add-on is not installed.

Thus, I think we need this same condition before querying the `$subscriptions` data so we can prevent call this query without a reason for that.

src/Campaigns/Controllers/CampaignRequestController.php
```diff
@@ -78,17 +78,11 @@ public function getCampaigns(WP_REST_Request $request): WP_REST_Response
...
-            ? CampaignsDataRepository::campaigns($ids)
-            : null;
+        $campaignsData = CampaignsDataRepository::campaigns($ids);
 
         $campaigns = array_map(function ($campaign) use ($campaignsData) {
             $view = new CampaignViewModel($campaign);
-
-            if ($campaignsData) {
-                $view->setData($campaignsData);
-            }
```
glaubersilva: Let's add an unreleased tag for these changes.

src/Campaigns/Repositories/CampaignsDataRepository.php
```diff
@@ -47,20 +47,67 @@ class CampaignsDataRepository
     public static function campaigns(array $ids): CampaignsDataRepository
     {
         $self = new self();
+        $campaignsData = get_option('give_campaigns_data', []);
```
glaubersilva, line 53: Let's add an unreleased tag for it.

src/Campaigns/ServiceProvider.php
```diff
@@ -250,4 +258,19 @@ private function addNewBadgeToMenu(): void
     {
         (new AddNewBadgeToAdminMenuItem())();
     }
+
+    /**
+     * @unreleased
+     */
+    private function registerCampaignCache(): void
+    {
+        give(CacheCampaignData::class)->registerAction();
```
glaubersilva: Let's remove it and register the hook/action in the service provider.

tests/Unit/Campaigns/CampaignsDataQueryTest.php
```diff
@@ -2,12 +2,15 @@
 
 namespace Give\Tests\Unit\Campaigns;
 
+use ActionScheduler;
+use ActionScheduler_Action;
```
glaubersilva: We can remove these unused imports.

glaubersilva (changes requested):
@alaca I tested it in a real-world scenario through the data from the GIVE-1079 task and could see that the request to load the list table was reduced from 22 seconds to 4.68 seconds after the cache was set up for the first time. The cache data also seems to be being updated properly.

That said, the functionality seems to be ok, but I found a few things to tweak and improvements on the codebase. Also, the unit tests are failing, so we need to fix that.

src/Campaigns/ServiceProvider.php
glaubersilva: For reference, check this comment: https://github.com/impress-org/givewp/pull/8047#discussion_r2263556053

src/Campaigns/Repositories/CampaignsDataRepository.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/8047/commits/59deef1b7764527cefca51bdeb73e8ac3fa5a01c

src/Campaigns/ServiceProvider.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/8047/commits/368f5aab8e262c224acfa42569fd93e32124cade

src/Campaigns/Actions/CacheCampaignData.php
alaca: The query is not executed until you call other methods on the object. So it really doesn't matter, but I'll update the code.

alaca: https://github.com/impress-org/givewp/pull/8047/commits/9e002d21b8875bb2508862f06c43e471286436af

tests/Unit/Campaigns/CampaignsDataQueryTest.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/8047/commits/07ab38532046764670cef503abcef6168efa32bf

src/Campaigns/Actions/CacheCampaignData.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/8047/commits/368f5aab8e262c224acfa42569fd93e32124cade

src/Campaigns/Actions/CacheCampaignData.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/8047/commits/9c1c8f26f3e113e4e9d2a2350da20aafac57dd24

src/Campaigns/Migrations/CacheCampaignsData.php
alaca: It doesn't matter, the data will be cached on next donation or renewal

src/Campaigns/Actions/CacheCampaignData.php
glaubersilva: I notice we still have the method `registerAction`, but we don't need it because the `givewp_cache_campaign_data` should be registered in the service provider, like I said in my previous comment.

src/Campaigns/ServiceProvider.php
```diff
@@ -250,4 +258,18 @@ private function addNewBadgeToMenu(): void
     {
         (new AddNewBadgeToAdminMenuItem())();
     }
+
+    /**
+     * @unreleased
+     */
+    private function registerCampaignCache(): void
+    {
+        Hooks::addAction('init', CacheCampaignData::class, 'registerAction');
```
glaubersilva: Let's remove it and replace it with this:

```
add_action('givewp_cache_campaign_data', function (int $campaignId) {
    give(CacheCampaignData::class)->handleCache($campaignId);
});
```

We also need to remove the `registerAction` method from the `CacheCampaignData` class.

src/Campaigns/Migrations/CacheCampaignsData.php
glaubersilva: I notice that the `$subscriptionsData` is still not being used in any place.

src/Campaigns/ServiceProvider.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/8047/commits/72a9b0bce6b09eff4f19b35ef4d5aea0bfd88e5e

PR #8031, opened 2025-07-22
Fix: add additional format to REST args 

jonwaldstein:
## Description

This updates the remainder of our string args with a proper format for sanitization.

## Affects

- Donor Endpoint
- Donor Notes Endpoint
- Donation Notes Endpoint

## Visuals

N/A

## Testing Instructions

- Make sure text fields are sanitized and everything still works properly

## Pre-review Checklist

-   [ ] Acceptance criteria satisfied and marked in related issue
-   [x] Relevant `@unreleased` tags included in DocBlocks
-   [ ] Includes unit tests
-   [ ] Reviewed by the designer (if follows a design)
-   [ ] [Self Review](https://give.gitbook.io/development-manual/devops/github/code-reviews#self-review) of code and UX completed

PR #8054, opened 2025-08-07
Fix: Update Rest V3 routes with correct schema implementation

defunctl:
> :rotating_light:  **NOTE:** I couldn't get tests to run locally and they seem to also fail via the GitHub actions due to outdated PHPUnit libraries, otherwise I would have added some tests for this.

## Description

As I started to think about the MCP Server implementation, I will need to provide it accurate JSON schemas for the responses and I noticed a few problems that this resolves:

1. In the V3 Rest controllers, the `schema` callback was incorrectly placed in the wrong spot in the array for `register_rest_route()` to actually display the schema, so it never did.
2. Most schemas were missing the `'$schema' => 'http://json-schema.org/draft-07/schema#',` property.
3. Some endpoints didn't have a schema at all, so I built those based on the responses and using AI.
4. Some endpoints schema did not match the response, and now it should.

> I ran out of time for today, but I wasn't sure where the Form REST controllers were.

Send an `OPTIONS` method request, e.g.:

```shell
curl -X https://local.lndo.site/wp-json/givewp/v3/campaigns
```

### Before

This is missing the `schema` property:

<details>

<summary>Show  JSON Response</summary>

```json
{
    "namespace": "givewp/v3",
    "methods": [
        "GET",
        "POST"
    ],
    "endpoints": [
        {
            "methods": [
                "GET"
            ],
            "args": {
                "status": {
                    "type": "array",
                    "items": {
                        "type": "string",
                        "enum": [
                            "active",
                            "draft",
                            "archived"
                        ]
                    },
                    "default": [
                        "active"
                    ],
                    "required": false
                },
                "ids": {
                    "type": "array",
                    "default": [],
                    "required": false
                },
                "page": {
                    "type": "integer",
                    "default": 1,
                    "minimum": 1,
                    "required": false
                },
                "per_page": {
                    "type": "integer",
                    "default": 30,
                    "minimum": 1,
                    "maximum": 100,
                    "required": false
                },
                "sortBy": {
                    "type": "enum",
                    "enum": [
                        "date",
                        "amount",
                        "donors",
                        "donations"
                    ],
                    "default": "date",
                    "required": false
                },
                "orderBy": {
                    "type": "enum",
                    "enum": [
                        "asc",
                        "desc"
                    ],
                    "default": "desc",
                    "required": false
                }
            }
        },
        {
            "methods": [
                "POST"
            ],
            "args": {
                "title": {
                    "type": "string",
                    "required": true
                },
                "shortDescription": {
                    "type": "string",
                    "required": false
                },
                "startDateTime": {
                    "type": "string",
                    "format": "date-time",
                    "required": false
                },
                "endDateTime": {
                    "type": "string",
                    "format": "date-time",
                    "required": false
                }
            }
        }
    ],
    "_links": {
        "self": [
            {
                "href": "https://local.lndo.site/wp-json/givewp/v3/campaigns"
            }
        ]
    }
}
```

</details>

---

### After

The schema property is now available:

<details>

<summary>Show  JSON Response</summary>

```json
{
    "namespace": "givewp/v3",
    "methods": [
        "GET",
        "POST"
    ],
    "endpoints": [
        {
            "methods": [
                "GET"
            ],
            "args": {
                "status": {
                    "type": "array",
                    "items": {
                        "type": "string",
                        "enum": [
                            "active",
                            "draft",
                            "archived"
                        ]
                    },
                    "default": [
                        "active"
                    ],
                    "required": false
                },
                "ids": {
                    "type": "array",
                    "default": [],
                    "required": false
                },
                "page": {
                    "type": "integer",
                    "default": 1,
                    "minimum": 1,
                    "required": false
                },
                "per_page": {
                    "type": "integer",
                    "default": 30,
                    "minimum": 1,
                    "maximum": 100,
                    "required": false
                },
                "sortBy": {
                    "type": "enum",
                    "enum": [
                        "date",
                        "amount",
                        "donors",
                        "donations"
                    ],
                    "default": "date",
                    "required": false
                },
                "orderBy": {
                    "type": "enum",
                    "enum": [
                        "asc",
                        "desc"
                    ],
                    "default": "desc",
                    "required": false
                }
            }
        },
        {
            "methods": [
                "POST"
            ],
            "args": {
                "title": {
                    "type": "string",
                    "required": true
                },
                "shortDescription": {
                    "type": "string",
                    "required": false
                },
                "startDateTime": {
                    "type": "string",
                    "format": "date-time",
                    "required": false
                },
                "endDateTime": {
                    "type": "string",
                    "format": "date-time",
                    "required": false
                }
            }
        }
    ],
    "schema": {
        "$schema": "http://json-schema.org/draft-07/schema#",
        "title": "campaign",
        "type": "object",
        "properties": {
            "id": {
                "type": "integer",
                "description": "Campaign ID."
            },
            "pageId": {
                "type": "integer",
                "description": "ID of the WordPress page assigned to the campaign."
            },
            "pagePermalink": {
                "type": "string",
                "format": "uri",
                "description": "Permalink for the campaign page."
            },
            "defaultFormId": {
                "type": "integer",
                "description": "ID of the default donation form for this campaign."
            },
            "defaultFormTitle": {
                "type": "string",
                "description": "Title of the default donation form."
            },
            "type": {
                "type": "string",
                "description": "Campaign type, e.g. &quot;core&quot;.",
                "enum": [
                    "core",
                    "p2p"
                ]
            },
            "title": {
                "type": "string",
                "description": "Campaign title.",
                "minLength": 3,
                "maxLength": 128,
                "errorMessage": "Campaign title is required."
            },
            "shortDescription": {
                "type": "string",
                "description": "Campaign short description",
                "maxLength": 120
            },
            "longDescription": {
                "type": "string",
                "description": "Long description of the campaign."
            },
            "logo": {
                "type": "string",
                "description": "URL to the campaign logo image."
            },
            "image": {
                "type": "string",
                "description": "URL to the campaign featured image."
            },
            "primaryColor": {
                "type": "string",
                "description": "Primary color used in campaign design."
            },
            "secondaryColor": {
                "type": "string",
                "description": "Secondary color used in campaign design."
            },
            "goal": {
                "type": "number",
                "description": "Goal amount/metric for the campaign.",
                "errorMessage": "Must be a number."
            },
            "goalType": {
                "enum": [
                    "amount",
                    "donations",
                    "donors",
                    "amountFromSubscriptions",
                    "subscriptions",
                    "donorsFromSubscriptions"
                ],
                "description": "Campaign goal type"
            },
            "goalStats": {
                "type": "object",
                "description": "Statistics about goal progress.",
                "properties": {
                    "actual": {
                        "type": "number",
                        "description": "Actual value raised so far."
                    },
                    "actualFormatted": {
                        "type": "string",
                        "description": "Formatted actual amount."
                    },
                    "percentage": {
                        "type": "number",
                        "description": "Percentage of the goal completed."
                    },
                    "goal": {
                        "type": "number",
                        "description": "Goal amount."
                    },
                    "goalFormatted": {
                        "type": "string",
                        "description": "Formatted goal amount."
                    }
                }
            },
            "status": {
                "enum": [
                    "active",
                    "inactive",
                    "draft",
                    "pending",
                    "processing",
                    "failed",
                    "archived"
                ],
                "description": "Current status of the campaign."
            },
            "startDate": {
                "type": [
                    "string",
                    "null"
                ],
                "format": "date-time",
                "description": "Start date of the campaign."
            },
            "endDate": {
                "type": [
                    "string",
                    "null"
                ],
                "format": "date-time",
                "description": "End date of the campaign (if set)."
            },
            "createdAt": {
                "type": "string",
                "format": "date-time",
                "description": "Timestamp when the campaign was created."
            }
        },
        "required": [
            "id",
            "title",
            "goal",
            "goalType"
        ],
        "allOf": [
            {
                "if": {
                    "properties": {
                        "goalType": {
                            "const": "amount"
                        }
                    }
                },
                "then": {
                    "properties": {
                        "goal": {
                            "type": "number"
                        }
                    },
                    "errorMessage": {
                        "properties": {
                            "goal": "Goal amount must be greater than 0"
                        }
                    }
                }
            },
            {
                "if": {
                    "properties": {
                        "goalType": {
                            "const": "donations"
                        }
                    }
                },
                "then": {
                    "properties": {
                        "goal": {
                            "minimum": 1,
                            "type": "number"
                        }
                    },
                    "errorMessage": {
                        "properties": {
                            "goal": "Number of donations must be greater than 0"
                        }
                    }
                }
            },
            {
                "if": {
                    "properties": {
                        "goalType": {
                            "const": "donors"
                        }
                    }
                },
                "then": {
                    "properties": {
                        "goal": {
                            "minimum": 1,
                            "type": "number"
                        }
                    },
                    "errorMessage": {
                        "properties": {
                            "goal": "Number of donors must be greater than 0"
                        }
                    }
                }
            },
            {
                "if": {
                    "properties": {
                        "goalType": {
                            "const": "amountFromSubscriptions"
                        }
                    }
                },
                "then": {
                    "properties": {
                        "goal": {
                            "minimum": 1,
                            "type": "number"
                        }
                    },
                    "errorMessage": {
                        "properties": {
                            "goal": "Goal recurring amount must be greater than 0"
                        }
                    }
                }
            },
            {
                "if": {
                    "properties": {
                        "goalType": {
                            "const": "subscriptions"
                        }
                    }
                },
                "then": {
                    "properties": {
                        "goal": {
                            "minimum": 1,
                            "type": "number"
                        }
                    },
                    "errorMessage": {
                        "properties": {
                            "goal": "Number of recurring donations must be greater than 0"
                        }
                    }
                }
            },
            {
                "if": {
                    "properties": {
                        "goalType": {
                            "const": "donorsFromSubscriptions"
                        }
                    }
                },
                "then": {
                    "properties": {
                        "goal": {
                            "minimum": 1,
                            "type": "number"
                        }
                    },
                    "errorMessage": {
                        "properties": {
                            "goal": "Number of recurring donors must be greater than 0"
                        }
                    }
                }
            }
        ]
    },
    "_links": {
        "self": [
            {
                "href": "https://local.lndo.site/wp-json/givewp/v3/campaigns"
            }
        ]
    }
}
```

</details>

---

### Compare this to a core WordPress endpoint response

```shell
curl -X https://local.lndo.site/wp-json/wp/v2/font-collections
```

<details>

<summary>Show  JSON Response</summary>

```json
{
    "namespace": "wp/v2",
    "methods": [
        "GET"
    ],
    "endpoints": [
        {
            "methods": [
                "GET"
            ],
            "args": {
                "context": {
                    "description": "Scope under which the request is made; determines fields present in response.",
                    "type": "string",
                    "enum": [
                        "view",
                        "embed",
                        "edit"
                    ],
                    "default": "view",
                    "required": false
                },
                "page": {
                    "description": "Current page of the collection.",
                    "type": "integer",
                    "default": 1,
                    "minimum": 1,
                    "required": false
                },
                "per_page": {
                    "description": "Maximum number of items to be returned in result set.",
                    "type": "integer",
                    "default": 10,
                    "minimum": 1,
                    "maximum": 100,
                    "required": false
                }
            }
        }
    ],
    "schema": {
        "$schema": "http://json-schema.org/draft-04/schema#",
        "title": "font-collection",
        "type": "object",
        "properties": {
            "slug": {
                "description": "Unique identifier for the font collection.",
                "type": "string",
                "context": [
                    "view",
                    "edit",
                    "embed"
                ],
                "readonly": true
            },
            "name": {
                "description": "The name for the font collection.",
                "type": "string",
                "context": [
                    "view",
                    "edit",
                    "embed"
                ]
            },
            "description": {
                "description": "The description for the font collection.",
                "type": "string",
                "context": [
                    "view",
                    "edit",
                    "embed"
                ]
            },
            "font_families": {
                "description": "The font families for the font collection.",
                "type": "array",
                "context": [
                    "view",
                    "edit",
                    "embed"
                ]
            },
            "categories": {
                "description": "The categories for the font collection.",
                "type": "array",
                "context": [
                    "view",
                    "edit",
                    "embed"
                ]
            }
        }
    },
    "_links": {
        "self": [
            {
                "href": "https://local.lndo.site/wp-json/wp/v2/font-collections"
            }
        ]
    }
}
```

</details>

## Affects

This affects most of the `/wp-json/givewp/v3` endpoints, **besides forms**.

## Visuals

All **but the forms** V3 REST endpoints should now include a `schema` property when requested with an `OPTIONS` method:

image

## Testing Instructions

Send a `GET` request to `/wp-json/givewp/v3/` to get all available endpoints.

The endpoints listed here, **besides any with form in the name**, when queried with an `OPTIONS` method should now return a valid `schema` property with a valid JSON schema.

In addition to that, pasting the object from the `schema` property should validate in an online validator, e.g. https://jsonschema.dev/

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/API/REST/V3/Routes/Campaigns/GetCampaignComments.php
+++ b/src/API/REST/V3/Routes/Campaigns/GetCampaignComments.php
@@ -111,7 +111,6 @@ class GetCampaignComments implements RestRoute
     public function getSchema(): array
     {
         return [
-            '$schema' => 'http://json-schema.org/draft-04/schema#',
             'title'   => 'campaign-comment',
             'type'    => 'object',
             'properties' => [
--- a/src/API/REST/V3/Routes/Campaigns/GetCampaignComments.php
+++ b/src/API/REST/V3/Routes/Campaigns/GetCampaignComments.php
@@ -44,6 +44,7 @@ class GetCampaignComments implements RestRoute
                         'default'  => true,
                     ],
                 ],
+                'schema' => [$this, 'getSchema'],
             ]
         );
     }
@@ -103,4 +104,41 @@ class GetCampaignComments implements RestRoute
 
         return new WP_REST_Response($formattedComments);
     }
+
+    /**
+     * @since unreleased
+     */
+    public function getSchema(): array
+    {
+        return [
+            '$schema' => 'http://json-schema.org/draft-07/schema#',
+            'title'   => 'campaign-comment',
+            'type'    => 'object',
+            'properties' => [
+                'donorName' => [
+                    'type'        => 'string',
+                    'description' => esc_html__('Display name of the donor who left the comment.', 'give'),
+                ],
+                'comment' => [
+                    'type'        => 'string',
+                    'description' => esc_html__('The comment text left by the donor.', 'give'),
+                ],
+                'anonymous' => [
+                    'type'        => 'boolean',
+                    'description' => esc_html__('Whether the donor chose to remain anonymous.', 'give'),
+                ],
+                'date' => [
+                    'type'        => 'string',
+                    'description' => esc_html__('Human-readable time since the comment was posted (e.g., "2 hours ago").', 'give'),
+                ],
+                'avatar' => [
+                    'type'        => 'string',
+                    'format'      => 'uri',
+                    'description' => esc_html__('URL to the donor’s avatar image.', 'give'),
+                ],
+            ],
+            'required' => ['donorName', 'comment', 'anonymous', 'date', 'avatar'],
+        ];
+    }
+
 }
--- a/src/API/REST/V3/Routes/Campaigns/GetCampaignComments.php
+++ b/src/API/REST/V3/Routes/Campaigns/GetCampaignComments.php
@@ -111,7 +111,7 @@ class GetCampaignComments implements RestRoute
     public function getSchema(): array
     {
         return [
-            '$schema' => 'http://json-schema.org/draft-07/schema#',
+            '$schema' => 'http://json-schema.org/draft-04/schema#',
             'title'   => 'campaign-comment',
             'type'    => 'object',
             'properties' => [
--- a/src/API/REST/V3/Routes/Donations/DonationController.php
+++ b/src/API/REST/V3/Routes/Donations/DonationController.php
@@ -824,7 +824,7 @@ class DonationController extends WP_REST_Controller
                 'honorific' => [
                     'type' => ['string', 'null'],
                     'description' => esc_html__('Donor honorific/prefix', 'give'),
-                    'enum' => give_get_option('title_prefixes', array_values(give_get_default_title_prefixes())),
+                    'enum' => give_get_option('title_prefixes', array_values(give_get_default_title_prefixes() + [ null ])),
                 ],
                 'email' => [
                     'type' => 'string',
@@ -977,12 +977,11 @@ class DonationController extends WP_REST_Controller
                 'createdAt' => [
                     'type' => ['object', 'null'],
                     'description' => esc_html__('Donation creation date', 'give'),
-                    'format' => 'date-time',
                     'properties' => [
                         'date' => [
                             'type' => 'string',
                             'description' => esc_html__('Date', 'give'),
-                            'format' => 'date-time',
+                            'pattern' => '^\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}\\.\\d{6}$',
                         ],
                         'timezone' => [
                             'type' => 'string',
@@ -998,12 +997,11 @@ class DonationController extends WP_REST_Controller
                 'updatedAt' => [
                     'type' => ['object', 'null'],
                     'description' => esc_html__('Donation last update date', 'give'),
-                    'format' => 'date-time',
                     'properties' => [
                         'date' => [
                             'type' => 'string',
                             'description' => esc_html__('Date', 'give'),
-                            'format' => 'date-time',
+                            'pattern' => '^\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}\\.\\d{6}$',
                         ],
                         'timezone' => [
                             'type' => 'string',
--- a/src/API/REST/V3/Routes/Donations/DonationController.php
+++ b/src/API/REST/V3/Routes/Donations/DonationController.php
@@ -1036,61 +1036,6 @@ class DonationController extends WP_REST_Controller
                         ],
                     ],
                 ],
-                '_links' => [
-                    'type' => 'object',
-                    'description' => esc_html__('Hypermedia links for the resource', 'give'),
-                    'properties' => [
-                        'self' => [
-                            'type' => 'array',
-                            'items' => [
-                                'type' => 'object',
-                                'properties' => [
-                                    'href' => ['type' => 'string', 'format' => 'uri'],
-                                    'targetHints' => [
-                                        'type' => 'object',
-                                        'properties' => [
-                                            'allow' => [
-                                                'type' => 'array',
-                                                'items' => ['type' => 'string'],
-                                            ],
-                                        ],
-                                    ],
-                                ],
-                            ],
-                        ],
-                        'givewp:donor' => [
-                            'type' => 'array',
-                            'items' => [
-                                'type' => 'object',
-                                'properties' => [
-                                    'embeddable' => ['type' => 'boolean'],
-                                    'href' => ['type' => 'string', 'format' => 'uri'],
-                                ],
-                            ],
-                        ],
-                        'givewp:campaign' => [
-                            'type' => 'array',
-                            'items' => [
-                                'type' => 'object',
-                                'properties' => [
-                                    'embeddable' => ['type' => 'boolean'],
-                                    'href' => ['type' => 'string', 'format' => 'uri'],
-                                ],
-                            ],
-                        ],
-                        'curies' => [
-                            'type' => 'array',
-                            'items' => [
-                                'type' => 'object',
-                                'properties' => [
-                                    'name' => ['type' => 'string'],
-                                    'href' => ['type' => 'string', 'format' => 'uri-template'],
-                                    'templated' => ['type' => 'boolean'],
-                                ],
-                            ],
-                        ],
-                    ],
-                ],
             ],
             'required' => [
                 'id',
--- a/src/API/REST/V3/Routes/Donations/DonationController.php
+++ b/src/API/REST/V3/Routes/Donations/DonationController.php
@@ -787,7 +787,7 @@ class DonationController extends WP_REST_Controller
     public function get_item_schema(): array
     {
         return [
-            '$schema' => 'http://json-schema.org/draft-07/schema#',
+            '$schema' => 'http://json-schema.org/draft-04/schema#',
             'title' => 'donation',
             'type' => 'object',
             'properties' => [
--- a/src/API/REST/V3/Routes/Donations/DonationController.php
+++ b/src/API/REST/V3/Routes/Donations/DonationController.php
@@ -787,7 +787,6 @@ class DonationController extends WP_REST_Controller
     public function get_item_schema(): array
     {
         return [
-            '$schema' => 'http://json-schema.org/draft-04/schema#',
             'title' => 'donation',
             'type' => 'object',
             'properties' => [
@@ -814,12 +813,10 @@ class DonationController extends WP_REST_Controller
                 'firstName' => [
                     'type' => 'string',
                     'description' => esc_html__('Donor first name', 'give'),
-                    'format' => 'text-field',
                 ],
                 'lastName' => [
                     'type' => 'string',
                     'description' => esc_html__('Donor last name', 'give'),
-                    'format' => 'text-field',
                 ],
                 'honorific' => [
                     'type' => ['string', 'null'],
@@ -834,12 +831,10 @@ class DonationController extends WP_REST_Controller
                 'phone' => [
                     'type' => ['string', 'null'],
                     'description' => esc_html__('Donor phone', 'give'),
-                    'format' => 'text-field',
                 ],
                 'company' => [
                     'type' => ['string', 'null'],
                     'description' => esc_html__('Donor company', 'give'),
-                    'format' => 'text-field',
                 ],
                 'amount' => [
                     'type' => ['object', 'null'],
@@ -853,7 +848,6 @@ class DonationController extends WP_REST_Controller
                         ],
                         'currency' => [
                             'type' => 'string',
-                            'format' => 'text-field',
                         ],
                     ],
                 ],
@@ -869,13 +863,11 @@ class DonationController extends WP_REST_Controller
                         ],
                         'currency' => [
                             'type' => 'string',
-                            'format' => 'text-field',
                         ],
                     ],
                 ],
                 'eventTicketsAmount' => [
                     'type' => ['object', 'null'],
-                    'readonly' => true,
                     'description' => esc_html__('Event tickets amount', 'give'),
                     'properties' => [
                         'value'             => [
@@ -886,7 +878,7 @@ class DonationController extends WP_REST_Controller
                         ],
                         'currency'          => [
                             'type' => 'string',
-                            'format' => 'text-field',
+
                         ],
                     ],
                 ],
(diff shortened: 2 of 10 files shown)


src/API/REST/V3/Routes/Campaigns/GetCampaignComments.php
```diff
@@ -103,4 +104,41 @@ public function handleRequest($request): WP_REST_Response
...
         return new WP_REST_Response($formattedComments);
     }
+
+    /**
+     * @since unreleased
+     */
+    public function getSchema(): array
+    {
+        return [
+            '$schema' => 'http://json-schema.org/draft-07/schema#',
```
jonwaldstein: Hey @defunctl is draft-07 correct? I thought WordPress uses a subset of draft 4 mentioned [in the docs](https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/)

src/API/REST/V3/Routes/Donations/DonationController.php
```diff
@@ -982,8 +1036,71 @@ public function get_item_schema(): array
                         ],
                     ],
                 ],
+                '_links' => [
```
jonwaldstein: Are we supposed to be manually inserting this into the schema?

src/API/REST/V3/Routes/Donations/DonationController.php
```diff
@@ -982,8 +1036,71 @@ public function get_item_schema(): array
...
+                                    'name' => ['type' => 'string'],
+                                    'href' => ['type' => 'string', 'format' => 'uri-template'],
+                                    'templated' => ['type' => 'boolean'],
+                                ],
+                            ],
+                        ],
+                    ],
+                ],
+            ],
+            'required' => [
```
jonwaldstein, line 1030: Does specifying an array of required properties actually work? I think @JasonTheAdams pointed out these might need to be moved into the individual properties

src/API/REST/V3/Routes/Campaigns/GetCampaignComments.php
defunctl: TIL, I'll get it updated.

src/API/REST/V3/Routes/Donations/DonationController.php
defunctl: Was just doing some reading and according to WordPress these are not typically included as part of the rest schema, so I think I'll just remove them.

src/API/REST/V3/Routes/Donations/DonationController.php
JasonTheAdams: Correct, this does not work. Properties _cannot_ be referenced outside of the property definition — each property is its own schema. As such, `'required' => true` needs to be applied to each required property.

defunctl: @JasonTheAdams can you please explain exactly where this doesn't work? 

Is WordPress using this somewhere to validate the return schema somehow or are you saying in terms of a general JSON schema this doesn't work?

JasonTheAdams: The item schema is used internally by WP in creation and update requests for validation. When it validates a property it uses _only_ the schema for that property (not the whole item schema). That's what I mean when I say "each property is its own schema". So having a required element that references properties by name won't even be seen by the validation system, so it won't know those properties are required.

JasonTheAdams: These are fun quirks completely unique to how WordPress uses and processes the schema in validation. 🤪

src/API/REST/V3/Routes/Donations/DonationController.php
```diff
@@ -796,24 +794,34 @@ public function get_item_schema(): array
...
                 ],
                 'firstName' => [
                     'type' => 'string',
                     'description' => esc_html__('Donor first name', 'give'),
-                    'format' => 'text-field',
                 ],
                 'lastName' => [
                     'type' => 'string',
                     'description' => esc_html__('Donor last name', 'give'),
-                    'format' => 'text-field',
```
jonwaldstein, line 811: Hey @defunctl what's the reason for removing the format on all these properties? We specifically added those for sanitization, most notably for _string_ types as the REST API will sanitize other primitives but not really strings unless you specify the format. 

Docs:
- https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#format
- https://github.com/WordPress/WordPress/blob/86b7fc84bdff5a9db05a120bba7858d5f4d32364/wp-includes/rest-api.php#L2763

Ref:
- https://github.com/impress-org/givewp/pull/8031
- https://github.com/impress-org/givewp/pull/8027

defunctl:
@jonwaldstein @JasonTheAdams okay this creates a few more problems:

It looks like the admin frontend uses Ajv to validate schemas, however because the `schema` key was never in the right place, [this code would never any schema to validate against](https://github.com/impress-org/givewp/blob/4.6.1/src/Views/Components/AdminDetailsPage/index.tsx#L70-L77), but _now_ it does.

**This introduces its own issues:**
a. The Ajv library > [v7.0 does not support `draft-04`](https://ajv.js.org/guide/schema-language.html#draft-04), only the `draft-07` or above. I have removed the `$schema` keys referencing that to make this work for now, noting that I now know that WordPress doesn't support `draft-07`.
b. [The readonly prop](https://core.trac.wordpress.org/ticket/56152) is also not valid JSON schema which WordPress uses, the valid property is `readOnly`. So this also causes Ajv validation to fail. I since removed those, but we can put them back depending on what we decide below.
c. Using the `'required' => true` within a single property is also not valid JSON, which specifies it should be an array, and as I learned, WordPress requires this when using `rest_get_endpoint_args_for_schema()` and not using custom `args` for inputs.
d. Because Ajv was never getting actual schemas to validate against, it was also not apparent that Ajv was missing extended formats (like `email` etc...) and didn't allow union types due to strict mode, which I fixed for now.


_So, we have a major conflict here..._

@glaubersilva it looks like you're the one that originally added [this Ajv code](https://github.com/impress-org/givewp/blame/4.6.1/src/Views/Components/AdminDetailsPage/index.tsx#L70-L77).

I think there's only one path forward here: **Stop using Ajv on the frontend to validate schemas** because:
1. [Ajv doesn't support `draft-04`](https://ajv.js.org/guide/schema-language.html#draft-04) and [WordPress does not support `draft-07`](https://developer.wordpress.org/news/2024/07/json-schema-in-wordpress/#wordpress-rest-api).
2. WordPress has many non-spec JSON properties, which would cause Ajv to fail validation (`readonly` instead of `readOnly`, `required => true` etc...) and cause WordPress REST endpoints to not function as expected.

Considering it wasn't technically working until now, I think removing it would not cause any problems, but I really want everyone's feedback of which direction we should take on this before proceeding any further.

I hope this all makes sense!

src/API/REST/V3/Routes/Donations/DonationController.php
defunctl: @jonwaldstein see my latest comment, these are not valid JSON formats and cause the Ajv schema validation to fail.

If ya'll decide to stop using that validation, we can add them all back.

defunctl: Also, the help doc you provided doesn't list `text-field` as a valid format WordPress supports either, so I'm just a bit confused there as well.

image

Edit: I see in https://github.com/WordPress/WordPress/blob/86b7fc84bdff5a9db05a120bba7858d5f4d32364/wp-includes/rest-api.php#L2763 they added it, I guess the docs are just outdated. Either way, the Ajv issue remains. We could add that as a custom format so Ajv knows about it, but the remaining Ajv problems would remain.

jonwaldstein: @defunctl out of curiosity how were you able to notice the Ajv schema validation was failing?

FYI `text-field` was added in [WP 5.9.0 ](https://github.com/WordPress/WordPress/blob/86b7fc84bdff5a9db05a120bba7858d5f4d32364/wp-includes/rest-api.php#L2756) (I don't the docs are up to date) but you can see in the codebase

https://github.com/WordPress/WordPress/blob/86b7fc84bdff5a9db05a120bba7858d5f4d32364/wp-includes/rest-api.php#L2900

defunctl: Yep! I just edited my comment with the same info :smile: , but as mentioned, we could tell Ajv about this non-spec  JSON format, but the other issues with Ajv would remain.

defunctl: @jonwaldstein I noticed it failing in the dashboard with the console open: 

GiveWP > Donors > Click `Add note` which fires all that off and you'd get console errors.

image

defunctl: If I checkout a commit much earlier before I removed the `readonly` stuff for example as well:

image

PR #8064, opened 2025-08-17
Feature: add slot fill to subscription records tab

glaubersilva:
Resolves [GIVE-2670]

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Records/index.tsx
+++ b/src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Records/index.tsx
@@ -1,10 +1,15 @@
+import { AdminSectionsWrapper } from '@givewp/components/AdminDetailsPage/AdminSection';
+import { RecordsSlot } from '../../slots';
+
 /**
  * @unreleased
  */
 export default function SubscriptionDetailsPageRecordsTab() {
     return (
-        <div>
-            <h2>Records</h2>
-        </div>
+        <>
+            <AdminSectionsWrapper>
+                <RecordsSlot />
+            </AdminSectionsWrapper>
+        </>
     );
 }


src/Admin/components/Notices/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+    
+    const noticeClasses = `${styles.notice} ${
+        type === 'warning' ? styles.warning : type === 'error' ? styles.error : styles.info
+    }`;
+    
+    return (
+        <div className={noticeClasses}>
+            {type === 'warning' && <WarningIcon />}
+            {type === 'error' && <ErrorIcon />}
+            {type === 'info' && <InfoIcon />}
```
pauloiankoski: Nothing wrong here but another possible solution could be something like:

```js
const noticeClasses = `${styles.notice} ${styles[type] ?? styles.info}`;

const NoticeIcon = ({ type }: { type: Props['type'] }) => {
    const icons = {
        warning: WarningIcon,
        error: ErrorIcon,
        info: InfoIcon
    };
    const IconComponent = icons[type] ?? icons.info;
    return <IconComponent />;
};
```

then
```js
<NoticeIcon type={type} />
```

src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Records/index.tsx
```diff
@@ -1,10 +1,29 @@
...
     return (
-        <div>
-            <h2>Records</h2>
-        </div>
+        <>
+            {isDirty && (
+                <div style={{ marginBottom: 'var(--givewp-spacing-4)' }}>
+                    <Notice
+                        type="info"                        
+                    >
```
pauloiankoski: We could keep this in a single line as we only have one attribute here.

pauloiankoski (approved):
Well done. Just left some comments but nothing that blocks this PR.

src/Admin/components/Notices/index.tsx
glaubersilva: Resolv in a5fb0c980152d5b606214ab1a7e53a627173e380

src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Records/index.tsx
glaubersilva: Resolv in a5fb0c980152d5b606214ab1a7e53a627173e380

PR #8065, opened 2025-08-18
'new-offline-donation': preview the correct default message

Genevieve-K:
When we preview the 'new-offline-donation' default message for admin, the 'new-donation' message is used. 
 ==> Use the correct function for the default message.


jonwaldstein (approved):
Thank you @Genevieve-K

PR #8068, opened 2025-08-22
Not apply give_clean() on receipt args

Genevieve-K:
If we display the confirmation page without ID and without email access, an alert is displayed:
`'You are missing the donation id to view this donation receipt.'`

For a site in Russian (non latin characters), we translated this alert: 
`'Отсутствует идентификатор пожертвования для просмотра квитанции о пожертвовании.'`

The string is "\u041e\u0442\u0441\u0443\u0442\u0441\u0442\u0432\u0443\u0435\u0442 \u0438\u0434\u0435\u043d\u0442\u0438\u0444\u0438\u043a\u0430\u0442\u043e\u0440 \u043f\u043e\u0436\u0435\u0440\u0442\u0432\u043e\u0432\u0430\u043d\u0438\u044f \u0434\u043b\u044f \u043f\u0440\u043e\u0441\u043c\u043e\u0442\u0440\u0430 \u043a\u0432\u0438\u0442\u0430\u043d\u0446\u0438\u0438 \u043e \u043f\u043e\u0436\u0435\u0440\u0442\u0432\u043e\u0432\u0430\u043d\u0438\u0438." 

With give_clean(), the \ are removed.
receipt-alert

If I remove give_clean():
receipt-alert2

I let you check if other actions of give_clean() are needed. But it works without this function.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved receipt processing by safely sanitizing shortcode parameters before decoding.
  * Preserved valid receipt request data while preventing unintended filtering of other parameters.


jonwaldstein (changes requested):
Hey @Genevieve-K the main reason we use `give_clean` is to sanitize request data.  Removing this has the potential to add a security issue - although it does look like the data is using json_encode which _might_ make it safer.  Perhaps there is another way to fix the escaping here? I would check what the request is actually returning as it could also be the front-end formatting.

Genevieve-K:
@jonwaldstein Did you see that I suggested another method to solve this issue?

PR #8075, opened 2025-08-29
Fix: move the schema key to the route level and make sure the Subscription schema matches the actual response

glaubersilva:
Resolves [GIVE-2863]

## Description

This PR fixes some inconsistencies in our REST API V3 implementation:

- Move schema definition to the route level (applies to all resources)
- Add schema validation tests (**Subscription only**, as a starting point to add tests for other resources in subsequent PRs)
- Standardize date formatting using ISO 8601 (**Subscription only**)

## Affects

Schemas from our REST API V3

## Visuals

schema-definitions

## Testing Instructions

To check if the schemas are being properly registered, you can do an `OPTIONS` request and check if the schema for each resource is returning properly, or you can use the WP OpenAPI plugin to easily check that all schemas are now being displayed there (before only the schema for campaigns was)

To test the other changes made on this PR related to the `Subscriptions` resource, just run it on your terminal:

`composer test -- --filter SubscriptionSchemaTest`

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2863]: https://stellarwp.atlassian.net/browse/GIVE-2863?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/API/REST/V3/Support/Item.php
```diff
@@ -0,0 +1,169 @@
...
+     */
+    public static function formatDateForResponse($date): ?string
+    {
+        if ($date === null) {
+            return null;
+        }
+
+        // Format DateTime to ISO 8601 string (e.g., "2023-12-25T14:30:00+00:00")
+        if ($date instanceof DateTime) {
+            return $date->format('c');
```
jonwaldstein: Should this be wrapped in `mysql_to_rfc3339()`?

src/API/REST/V3/Support/Item.php
```diff
@@ -0,0 +1,169 @@
...
+            }
+            return $item;
+        }
+
+        // Use specific field lists if provided
+        $item = self::formatDatesForResponse($item, $dateFields);
+        $item = self::formatValueObjectsForResponse($item, $valueObjectFields);
+
+        return $item;
+    }
```
jonwaldstein: @glaubersilva a couple thoughts about this:

1. Can you help me understand what the additional param arrays would be used for? it doesn't look like they are being used
2. I don't think we need to format our ValueObjects because they already implement JsonSerializable and returning the value as seen [here](https://github.com/impress-org/givewp/blob/9d76baab373bd19a5211b1ac789391f271c26119/src/Framework/Support/ValueObjects/BaseEnum.php#L323)
3. Although this is super handy and clever, part of me feels like this could be simplified to just formatting the date properties directly in the `prepare_item_for_response` methods (instead of looping through all the items for every request). Example:
```php
$item['createdAt'] = mysql_to_rfc3339($item['createdAt]->format('c'))

//(or with a helper function)
```

glaubersilva: Resolve in 0d4c7ed485df1ff4a06042c84323c9970979bc6c and f24fc678cd4777fe4f4408342819ca10b777b183

About your questions: 

1. I started to create this class as a generic formatting class, then renamed it to be specific to Items, but kept the old logic there that can be removed since we don't need to cover all scenarios I had imagined before. So, I refactored it to keep things simpler.

2. Our `SubscriptionViewModel` has a function that exports all the properties of the subscription model with additional properties and returns them as an array. This process is happening here: https://github.com/impress-org/givewp/blob/f24fc678cd4777fe4f4408342819ca10b777b183/src/Subscriptions/ViewModels/SubscriptionViewModel.php#L56-L64

However, each key/property of the array returned by the export method still keeps its original type. So, the `$item` var that arrives inside the  `prepare_item_for_response` method still is using `DateTime` and `ValueObjects` types instead of the string type, as is expected by the schema.

image

3. I agree that we don't need to loop through all items all the time, so now, after the refactoring, it is necessary to inform which properties should be formatted to optimize how it works.

src/API/REST/V3/Support/Item.php
glaubersilva: Not really, in fact, the function `mysql_to_rfc3339()` is used to format a string date to the ISO 8601 format, but in our case, the models always return a `DateTime` object, so this approach is enough to get the same output that `mysql_to_rfc3339()` would produce for strings.

I added this comment in the codebase to better explain it:
https://github.com/impress-org/givewp/blob/f24fc678cd4777fe4f4408342819ca10b777b183/src/API/REST/V3/Support/Item.php#L31-L40

glaubersilva:

src/API/REST/V3/Support/Item.php
jonwaldstein: @glaubersilva I did notice the `mysql_to_rfc3339` function affects the time differently than the DateTime format option
Screenshot 2025-09-08 at 11 07 27 AM

src/API/REST/V3/Support/Item.php
jonwaldstein: Hey @glaubersilva re: (2) ValueObjects

When a class implements `JsonSerializable`, it's `jsonSerialize()` method will get invoked when calling `json_encode`. The REST API will call `json_encode` on the raw data `$request->get_data()` which is the request type we will be consuming. So although your right if you were to compare the raw exports object - however the real world scenario will actually be using `json_encode` which is the shape of the data we should be most testing for as other classes like the Money object use `JsonSerializable` as well.

src/API/REST/V3/Support/Item.php
glaubersilva: @jonwaldstein Great catch! Although the PHP docs say this approach converts to `ISO 8601 date`. Seems that the WP way of doing it with the `mysql_to_rfc3339()` function is also removing the time zone.

Resolved in 4d1a2bdb0f8a9e3baf72bc32a66cdb41542187f5

image

src/API/REST/V3/Support/Item.php
glaubersilva: @jonwaldstein  Oh, ok! Got it now, thanks for the clarification. 

I remember we had to tweak the tests previously using this approach:

```
$dataJson = json_encode($response->get_data());
$data = json_decode($dataJson, true);
```

My mistake is that I was relying only on the tests instead of performing a real request to check the output.

So, I updated the tests to use that and also did a bunch of tweaks to make sure everything keeps working as expected. 

The Item class is now much smaller and just handles date properties since the other ones are being handled when the serialization happens.

jonwaldstein: @glaubersilva that's great thanks!  The one thing i'm still debating here is the Item class.  Looking at the implementation i'm not sure it's that much more convenient than just formatting the date properties individually either in the exports array or before the response.  I feel like the current iteration of the Item class helper would be good for formatting unknown date properties but since we know the date properties why don't we just format them before responding?  Here's some examples of what i'm thinking:

**Currently:**

```php
$response = new WP_REST_Response(Item::formatDatesForResponse($item, ['createdAt', 'renewsAt']));
```

**Alternative 1: format known properties**
```php
$item['createdAt'] = Item::formatDate($item['createdAt']);
$item['renewsAt'] = Item::formatDate($item['renewsAt']);

$response = new WP_REST_Response(item));
```


**Alternative 2: handle formatting up the chain, exports will always be run through json_encode so this is a good option that's easier to test**
```php
   //SubscriptionViewModel
    public function exports(): array
    {
        $donor = $this->subscription->donor;

        $data = array_merge(
            $this->subscription->toArray(),
            [
                'firstName' => $donor ? $donor->firstName : '',
                'lastName' => $donor ? $donor->lastName : '',
                'gateway' => $this->getGatewayDetails(),
                'projectedAnnualRevenue' => $this->subscription->projectedAnnualRevenue(),
                'createdAt' => mysql_to_rfc3339($this->subscription->createdAt->format('c')),
                'renewsAt' => mysql_to_rfc3339($this->subscription->renewsAt->format('c')),
            ]
        );
     }
```

glaubersilva: @jonwaldstein Seems the **Alternative 1** is basically the same thing that we have in the current implementation, but instead of informing all properties in the array of the second parameter, we would perform the format individually for each property, which seems a bit verbose.

The **Alternative 2** was my first choice when I started work on this, but I realized that the `SubscriptionViewModel` lives inside the `Subscription` domain and can be used in other contexts in the future, not only in the REST API context. Since this date format is something exclusive for the REST API endpoints, I think it's better to keep it in the `prepare_item_for_response()` method.

With the current approach, we can also reuse the same logic in other resources (Donations, Donors, etc.), all we need to do is inform the properties we would like to format. For Donation, it would be something like this:

`$response = new WP_REST_Response(Item::formatDatesForResponse($item, ['createdAt', 'updatedAt']));`

glaubersilva: @jonwaldstein Just to clarify, I'm not opposed to switching to **Alternative 1**, but I'm wondering if it's necessary. If you think it would be better, I can change it. Not a big deal!

jonwaldstein: @glaubersilva you're right there's not a _big_ difference between what you have and **Alternative 1** - it mainly cuts down on extra loops and type checking which would be more useful for unknown properties. Your current way is slightly more convenient so either way is fine with me 😄 

The `SubscriptionViewModel` is interesting - do we know what the intention of these separate view models is outside the REST API?  I agree, if they are being used elsewhere we want to be careful about additional formatting - however i'm not exactly sure where else they are supposed to be used 😏

glaubersilva: @jonwaldstein Great question! But I don't have a good reply for that. 😅 To be honest, after my last comment, I started to wonder why this view is not inside the REST API domain. I think we just replicated for Subscription endpoints the same pattern used before for the Donations endpoint, where the view also lives inside the Donations domain. 

So, I would say that we can consider moving these views to the REST API domain and make them exclusive to it. However, it seems something out of the scope of this PR.

jonwaldstein: @glaubersilva all good, we can bring up with the devs and yes it's out of scope for this - so I think what you have is fine for now 😄

src/Views/Components/AdminDetailsPage/index.tsx
```diff
@@ -1,34 +1,29 @@
 /**
  * External Dependencies
  */
-import {useEffect, useState, useRef} from 'react';
-import {JSONSchemaType} from 'ajv';
-import {ajvResolver} from '@hookform/resolvers/ajv';
+import {useEffect, useRef, useState} from 'react';
 import {FormProvider, SubmitHandler, useForm, useFormContext, useFormState} from 'react-hook-form';
+import {fetchSchemaAndConfigureResolver} from '../../../Admin/ajv';
```
jonwaldstein: Let's clean up this import
```suggestion
import {fetchSchemaAndConfigureResolver} from '@givewp/admin/ajv';
```

src/Admin/ajv.ts
```diff
@@ -0,0 +1,525 @@
...
+    path: string,
+    setResolver: (resolver: any) => void
+): Promise<void> {
+    try {
+        const {schema}: {schema: JSONSchemaType<any>} = await apiFetch({
+            path,
+            method: 'OPTIONS',
+        });
+
+        const customResolver = (data: any) => {
```
jonwaldstein: I suggest we separate out the custom resolver from the fetch logic to follow the react-hook-form pattern.  That would make this much more flexible or else we're kind of tied to coupling the fetch with the resolver itself which might not always be necessary. 

It should be a fairly simple update to pull out the resolver that receives schema.  

The fetching of OPTIONS is pretty trivial (although, you could keep a small function for it) but setting the resolver in local state is very contextual and not something I feel should be abstracted.

That would bring us back to the standard/existing implementation with only really changing the import to our own resolver.

```js
import {ajvResolver} from '@givewp/admin/ajv';

 apiFetch({
      path: `/givewp/v3/campaigns/${campaignId}`,
      method: 'OPTIONS',
  }).then(({schema}: {schema: JSONSchemaType<any>}) => {
      setResolver({
          resolver: ajvResolver(schema),
      });
  });
```

src/API/REST/V3/Routes/Donations/DonationNotesController.php
```diff
@@ -53,6 +53,7 @@ public function register_routes()
                 'args' => $this->get_endpoint_args_for_item_schema(WP_REST_Server::CREATABLE),
                 'schema' => [$this, 'get_public_item_schema'],
```
alaca: I see the schema has been removed from previous definitions, but we still have it here?

src/API/REST/V3/Routes/Donations/DonationNotesController.php
```diff
@@ -371,7 +371,7 @@ public function get_collection_params(): array
     public function get_item_schema(): array
     {
         $schema = [
-            'schema' => 'http://json-schema.org/draft-07/schema#',
+            '$schema' => 'http://json-schema.org/draft-04/schema#',
```
alaca, line 373: How is AJV handling this field? If I remember correctly, this was causing some errors (`$schema`).

src/Admin/ajv.ts
```diff
@@ -0,0 +1,525 @@
...
+        try {
+            new URL(uri);
+            return true;
+        } catch {
+            return false;
+        }
+    });
+
+    // Validate ISO 8601 date-time format (RFC 3339)
+    ajv.addFormat('date-time', (dateTime: string) => {
```
jonwaldstein: I saw there are a couple nifty packages that [ajv](https://github.com/react-hook-form/resolvers/blob/master/ajv/src/ajv.ts) is using that might also help us out:

- https://ajv.js.org/packages/ajv-errors.html
- https://ajv.js.org/packages/ajv-formats.html

src/Admin/ajv.ts
glaubersilva: Resolve in ee28ac7f0f050d6ae2f30b2e88ea1f308e8b73e7

src/Views/Components/AdminDetailsPage/index.tsx
glaubersilva: Resolved in 294fda8a1629e15d1cba239d8637ff4648f74274

src/Admin/ajv.ts
glaubersilva: @jonwaldstein Thanks for noticing these packages. 

Resolved in e93e341f39d33360d3c818d314eaee3e6bbd92b4 and febc87d148f8c0fa06b1d631e130f67337394d77

src/API/REST/V3/Routes/Donations/DonationNotesController.php
glaubersilva: Nice catch! Resolved in cd22b5f65535f10dbc83b2dfc904ea3ad9c40195

src/Admin/ajv.ts
jonwaldstein: Glad to see those worked and cleaned this up a bit 🎉

src/Admin/ajv.ts
jonwaldstein: Thank you!

src/API/REST/V3/Routes/Donations/DonationNotesController.php
glaubersilva: @alaca Yes, before using your custom AJV resolve, some errors were occurring. Now, we are converting the draft 04 to draft 07 in the frontend, and the errors are not happening anymore.

jonwaldstein: nice!

rickalday (approved):
QA PASS

**Verified**

- All information is properly displayed on the overview tab
- Admin can add renewals manually
- Admin can edit the Fund associated with the Subscription using the Records tab
- Admin can sync subscriptions
- Admin can cancel subscriptions

PR #8081, opened 2025-09-03
Fix: remove conditional on canSync variable

JoshuaHungDinh:
Resolves https://lw.slack.com/archives/C04SLRDD9CK/p1756908964317189?thread_ts=1756843777.960549&cid=C04SLRDD9CK

## Description
Remove the conditional check that would previously show the can sync button if there was no gateway available. This is incorrect though, the button should show only when the gateway allows syncing. 

Also updates the check for the current gateway to see if it has the capability to sync transactions - this will help avoid issues with legacy gateways performing there own sync function through the new UI by hiding the sync subscription button when working with a legacy gateway subscription.

## Affects
Subscription Sync

## Pre-review Checklist

-   [ ] Includes unit tests


glaubersilva (approved):
@JoshuaHungDinh Nice work! Thanks for the improvements.

PR #8084, opened 2025-09-05
Feature: add new data to donations list table response

glaubersilva:
Resolves [GIVE-2751]

## Pre-review Checklist

-   [ ] Includes unit tests


src/Donations/Endpoints/ListDonations.php
```diff
@@ -166,13 +165,23 @@ public function handleRequest(WP_REST_Request $request): WP_REST_Response
...
+        ];
+
+        if ($this->shouldIncludeStats()) {
+            $oneTimeDonationsCount = $this->getOneTimeDonationsCount();
+            $recurringDonationsCount = $this->getRecurringDonationsCount();
+            $data['stats'] = [
+                'donationsCount' => $donationsCount,
+                'oneTimeDonationsCount' => $oneTimeDonationsCount,
+                'recurringDonationsCount' => $recurringDonationsCount,
+            ];
```
jonwaldstein: Is there a way we can make this more performant perhaps using one main query? i'm concerned running 3 extra queries will really slow things down.  Perhaps this would even be better as a separate endpoint for now like `/list-donations/stats`?

glaubersilva: @jonwaldstein  Thanks for the feedback! I've implemented the optimization using a single query approach instead of a separate endpoint.

Key reasons for this decision:

1. Performance: Reduces from 3 queries to 1 query - eliminates 2 extra database calls
2. Simplicity: Keeps all logic in one place, making it easier to maintain
3. No extra HTTP overhead: Avoids additional network requests every time the list updates
4. Data consistency: All statistics are calculated simultaneously, preventing timing issues

Resolved in 51df2490e51d279329bb716db050f2d3d6f637d4

jonwaldstein: Hey @glaubersilva thanks! the query optimization is a big improvement.  However, the team discussed this on daily yesterday (I think when you left) that we should really move this into a separate endpoint so that the stats can load independently from the list table.  

Reason being: we already know load times are an existing problem for people on the list tables and introducing another query is going to make it even slower .  So at least if this was a separate request (that has the potential to be slow on larger sites - think 100k+ donations), the list table itself would likely load sooner than the stats which is more important for this screen as a list table.

Then as an enhancement, we could even explore cache'ing the stats endpoint - which again would be independent from the list table request.

glaubersilva: @jonwaldstein No problem! I refactored it to add the new endpoint for stats. 

During the process, I noticed that the previous logic was wrong because I added where clauses that should be applied only for the list table, but not for the stats info, since it should return the stats for all donations, regardless of what is displayed in the list table.

Now this is ready for a re-review.

src/Donations/Endpoints/DonationStats.php
```diff
@@ -0,0 +1,119 @@
...
+
+use Give\Donations\ValueObjects\DonationMetaKeys;
+use Give\Framework\Database\DB;
+use WP_REST_Request;
+use WP_REST_Response;
+
+/**
+ * @unreleased
+ */
+class DonationStats extends Endpoint
```
jonwaldstein: Can we call this something like `DonationsListStats` just to make it clear it's associated to the donation list endpoint?

src/Donations/Endpoints/DonationStats.php
```diff
@@ -0,0 +1,119 @@
...
+            ->where('post_status', 'trash', '<>'); // Exclude trash items
+
+        // Filter by test mode
+        $dependencies = [DonationMetaKeys::MODE(), DonationMetaKeys::SUBSCRIPTION_ID()];
+        
+        $query->attachMeta(
+            'give_donationmeta',
+            'ID',
+            'donation_id',
+            ...DonationMetaKeys::getColumnsForAttachMetaQueryFromArray($dependencies)
```
jonwaldstein: Since we're optimizing for aggregates, we don't need to pull in _every_ meta key here which will cause unnecessary joins.  So I think we should just specify the meta we need for this query

src/Donations/Endpoints/DonationStats.php
```diff
@@ -0,0 +1,119 @@
...
+     *
+     * @unreleased
+     */
+    public function getDonationStatistics(): array
+    {
+        $testMode = $this->request->get_param('testMode');
+
+        $query = DB::table('posts')
+            ->where('post_type', 'give_payment')
+            ->where('post_status', 'trash', '<>'); // Exclude trash items
```
jonwaldstein: I'm not sure if we talked about this, but we might want to consider only counting _completed_ donations (including renewals) in these stats.  I'm going to ping Angela about this.

As a side note, we could always add a status arg as an array so we have the option

src/Donations/Endpoints/DonationStats.php
glaubersilva: Resolved in 385821cd723ead569a91eda87a1d55ceba2981c2

src/Donations/Endpoints/DonationStats.php
glaubersilva: Resolved in 6fbc9ed85048c87816f6c5e958e0abf4e5acc0d6

src/Donations/Endpoints/DonationStats.php
glaubersilva: @jonwaldstein it makes sense. I refactored it to only include completed donations (one-time and recurring).

I would prefer not to add the status arg now since we don't have plans to use it for now, we can revisit it later when (if) necessary.

Resolved in 5aa1f06934551e8e6723ed009f73cc402de8c4e3

glaubersilva: As agreed in the call today, I reverted it to include all statuses except the _"trash"_ one.

Resolved in f25031042c0317c14bd0fe4ca6635430bfc0ed95

PR #8087, opened 2025-09-12
Hotfix: Ensure campaignId is saved in donation details

pauloiankoski:
Resolves [GIVE-2870]

## Description

This hotfix addresses an issue where a donation could not be updated through the details screen after the campaign field was changed. Changes in this PR include:
	•	Removed `campaignId` from the `nonEditableFields` property in the donations API endpoint, allowing it to be updated.
	•	Moved the fallback logic for `campaignId` assignment to `getCoreDonationMetaForDatabase` to centralize it, removing redundant logic from the `insert` method in the donations repository and the `createRenewal` method in the subscriptions repository.

## Affects
Donation Details and Subscription Details

## Visuals

https://github.com/user-attachments/assets/bf6c936d-b3bb-4611-8efb-920f658a7b6b

_We can now update a donation after removing `campaignId` from the `nonEditableFields` list._

https://github.com/user-attachments/assets/fb59552a-9eb1-40eb-b4b6-1f7ee1a8d3e1

_Showing that we can still create a renewal after refactoring the `campaignId` logic._

https://github.com/user-attachments/assets/2656d1a1-b852-4157-8a85-bfb78d521189

_Throttling my connection to show that campaign data loads only after the donation data._

## Testing Instructions
1.	Update the campaign and form in a donation using the Donation Details screen.
2.	Create a renewal from the Subscription Details screen.
3.	Verify that there are no 404 responses from any API requests on the Subscription Details screen.

## Pre-review Checklist

-   [x] Includes unit tests

[GIVE-2870]: https://stellarwp.atlassian.net/browse/GIVE-2870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Donations/resources/admin/components/DonationDetailsPage/Tabs/Overview/DonationSummaryGrid/index.tsx
+++ b/src/Donations/resources/admin/components/DonationDetailsPage/Tabs/Overview/DonationSummaryGrid/index.tsx
@@ -26,9 +26,11 @@ export type DonationSummaryGridProps = {
 export function CampaignCard({donation}: {donation: Donation}) {
     return (
         <GridCard heading={__('Campaign name', 'give')} headingId="campaign-name">
-            {!donation?.campaignId
+            {!donation
                 ? <Spinner />
-                : <CampaignCardContent campaignId={donation.campaignId} />
+                : donation?.campaignId
+                    ? <CampaignCardContent campaignId={donation.campaignId} />
+                    : <p>{__('No campaign associated with this donation', 'give')}</p>
             }
         </GridCard>
     );
@@ -67,9 +69,11 @@ function CampaignCardContent({campaignId}: {campaignId: number}) {
 export function DonorCard({donation}: {donation: Donation}) {
     return (
         <GridCard heading={__('Associated donor', 'give')} headingId="donor">
-            {!donation?.donorId
+            {!donation
                 ? <Spinner />
-                : <DonorCardContent donorId={donation.donorId} />
+                : donation?.donorId
+                    ? <DonorCardContent donorId={donation.donorId} />
+                    : <p>{__('No donor associated with this donation', 'give')}</p>
             }
         </GridCard>
     );
--- a/src/Donations/resources/admin/components/DonationDetailsPage/Tabs/Overview/DonationSummaryGrid/index.tsx
+++ b/src/Donations/resources/admin/components/DonationDetailsPage/Tabs/Overview/DonationSummaryGrid/index.tsx
@@ -24,38 +24,68 @@ export type DonationSummaryGridProps = {
  * @since 4.6.0
  */
 export function CampaignCard({donation}: {donation: Donation}) {
-    const {campaign, hasResolved: hasResolvedCampaign} = donation?.campaignId
-        ? useCampaignEntityRecord(donation?.campaignId)
-        : {campaign: null, hasResolved: false};
-
     return (
         <GridCard heading={__('Campaign name', 'give')} headingId="campaign-name">
-            {!hasResolvedCampaign && <Spinner />}
-            {hasResolvedCampaign && campaign && (
-                <a
-                    href={`edit.php?post_type=give_forms&page=give-campaigns&id=${campaign?.id}&tab=overview`}
-                    className={styles.campaignLink}
-                >
-                    {campaign?.title}
-                </a>
-            )}
+            {!donation?.campaignId
+                ? <Spinner />
+                : <CampaignCardContent campaignId={donation.campaignId} />
+            }
         </GridCard>
     );
 }
 
 /**
+ * @unreleased
+ */
+function CampaignCardContent({campaignId}: {campaignId: number}) {
+    const {campaign, hasResolved: hasResolvedCampaign} = useCampaignEntityRecord(campaignId);
+
+    return (
+        <>
+            {!hasResolvedCampaign
+                ? <Spinner />
+                : campaign ? (
+                    <a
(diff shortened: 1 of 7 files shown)


src/Donations/resources/admin/components/DonationDetailsPage/Tabs/Overview/DonationSummaryGrid/index.tsx
```diff
@@ -19,11 +19,14 @@ export type DonationSummaryGridProps = {
...
 /**
+ * @unreleased Refactored to only try to load campaign when we have a valid campaignId from the donation
  * @since 4.8.0 export function for SubscriptionSummaryGrid & add GridCard component
  * @since 4.6.0
  */
 export function CampaignCard({donation}: {donation: Donation}) {
-    const {campaign, hasResolved: hasResolvedCampaign} = useCampaignEntityRecord(donation?.campaignId);
+    const {campaign, hasResolved: hasResolvedCampaign} = donation?.campaignId
+        ? useCampaignEntityRecord(donation?.campaignId)
+        : {campaign: null, hasResolved: false};
```
jonwaldstein: Using conditionals with hooks will break the [Rules of Hooks](https://react.dev/reference/rules/rules-of-hooks#only-call-hooks-at-the-top-level).  Basically, we're not supposed to call hooks conditionally, but instead handle any conditional logic inside the hook itself.

pauloiankoski: Fixed on [ae89224](https://github.com/impress-org/givewp/pull/8087/commits/ae89224e96a35bdd74980b390c899081364b943e)

src/Donations/resources/admin/components/DonationDetailsPage/Tabs/Overview/DonationSummaryGrid/index.tsx
```diff
@@ -24,38 +24,68 @@ export type DonationSummaryGridProps = {
...
-            {hasResolvedCampaign && campaign && (
-                <a
-                    href={`edit.php?post_type=give_forms&page=give-campaigns&id=${campaign?.id}&tab=overview`}
-                    className={styles.campaignLink}
-                >
-                    {campaign?.title}
-                </a>
-            )}
+            {!donation?.campaignId
+                ? <Spinner />
```
jonwaldstein: @pauloiankoski If there's no `campaignId`, I feel like we should display the empty message and reserve the spinner for when a request is happening.  What do you think?

```js
<p>{__('Campaign not found', 'give')}</p>
```

pauloiankoski: Check this: https://github.com/impress-org/givewp/pull/8087/commits/95a1bebe4c75053000344907978822e649f0d32e

My idea is to show the spinner first while loading the donation. Once the donation is loaded, we check if it has a campaign ID. If it does, we render the `CardContent` component; otherwise, we display an error message.

Inside the `CardContent` component, we also start by showing a spinner while loading the campaign. If the campaign is found, we render it; otherwise, we display an error message.

In short, in the ideal path, we will have two distinct spinners being rendered — one while loading the donation and the other while loading the campaign — and then we finally render the expected content. If either the donation has no `campaignId`, or no campaign is found for that ID, error messages are displayed.

jonwaldstein: @pauloiankoski your reasoning for the _order_ of loading indicators makes sense - however from a component standpoint it's not clear because these components just receive a Donation prop which does not tell us there is any loading state.  A simple way to make this more clear would be to add a loading prop to the component so the logic inside is clearly accounting for when the donation is loading instead of being null which can be ambiguous. 

However, I do see we are accounting for this further up the chain here which makes me curious if it will prevent this component from being rendered: https://github.com/impress-org/givewp/blob/a6165c0acca3e0e51d8ae96582bd7d6aa5dd82f6/src/Donations/resources/admin/components/DonationDetailsPage/Tabs/Overview/index.tsx#L17

pauloiankoski: Yeah, I see that the whole chain of components (from the `DonationSummaryGrid` to the `CampaignCard` and `DonorCard`) relies on the fact that none of them are rendered while the donation is not resolved. If you think it would be better to handle the loading state at the component level, then I’d add a loading state to all three components. Otherwise, we can keep relying on the current setup — just making sure we don’t accidentally use these components somewhere without managing the loading state individually.

jonwaldstein: Since this PR is meant to be a p1 fix, let's keep these changes smaller and roll back to relying on the parent components that prevent the rendering if there is no donation.

pauloiankoski: I reverted the frontend changes and implemented the same check we have on the Donation Details overview tab onto the Subscription Details overview tab, ensuring the entities are resolved before rendering the components. The previous changes have been moved to a new local branch that I may push separately from this P1 issue. See https://github.com/impress-org/givewp/pull/8087/commits/41f26378fbd54d2f885f99c5de08a2fd19f6ed0b

jonwaldstein (approved):
@pauloiankoski thanks!

PR #8090, opened 2025-09-15
Added Text Domain for one string in donation-options-form-editor.php

DAnn2012:
Added Text Domain for one string in donation-options-form-editor.php


jonwaldstein (approved):
Thank you @DAnn2012

PR #8095, opened 2025-09-15
Fixed String Translation in ListTable.php

DAnn2012:
Fixed String Translation in ListTable.php


jonwaldstein (approved):
Thank you @DAnn2012

PR #8099, opened 2025-09-16
Refactor: Enhanced Donations List Table with Updated Sorting and Filtering

JoshuaHungDinh:
Resolves [GIVE-2814]

## Overview
This PR significantly improves the donations list table by adding comprehensive sorting capabilities and a new custom select component for enhanced filtering functionality.

### 🔢 **Amount Sorting Enhancement**
- **Enhanced Amount Column Sorting**: Implemented amount sorting property in column that handles currency conversion and exchange rates
- **SQL-based Sorting**: Uses `COALESCE(CAST(amount AS DECIMAL) / COALESCE(CAST(exchangeRate AS DECIMAL), 1), 0)` for accurate numerical sorting
- **Multi-currency Support**: Properly sorts donations across different currencies by converting to base amounts
- 
### **Frontend Components**
- **CustomSelect**: New reusable filter component with React Select integration
- **Filter System**: Enhanced filter rendering system supporting multiple input types

## 🎨 **UI/UX Enhancements**
- **Consistent Styling**: All filter components use the new `CustomFilter` component

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2814]: https://stellarwp.atlassian.net/browse/GIVE-2814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Donations/ListTable/Columns/CampaignColumn.php
```diff
@@ -14,6 +14,8 @@
  */
 class CampaignColumn extends ModelColumn
 {
+    protected $sortColumn = 'campaignTitle';
```
jonwaldstein: @JoshuaHungDinh FYI in refinement we decided _not_ to add sorting functionality for the campaign column.  I can see this is also adding more overhead with additional meta storage and migrations that we should revert.

Also, since it's already possible to filter by campaign, it makes the sorting of that column less useful.

Screenshot 2025-09-16 at 4 59 10 PM

JoshuaHungDinh: Oh your right! I do remember speaking about that in a refinement actually.  I've removed it.

jonwaldstein: Thank you sir!

src/Views/Components/ListTable/CustomFilter/useAsyncCampaigns.ts
```diff
@@ -0,0 +1,86 @@
...
+        try {
+            const queryParams = createCampaignQueryParams({
+                perPage: CAMPAIGNS_PER_PAGE,
+                page: currentPage,
+                search: search || undefined,
+            });
+
+            const campaigns = await apiFetch<Campaign[]>({
+                path: `/givewp/v3/campaigns?${queryParams.toString()}`,
+            });
```
jonwaldstein, line 63: @JoshuaHungDinh Is there any reason we're not using the campaigns entity here?

Example: https://github.com/impress-org/givewp/blob/a6165c0acca3e0e51d8ae96582bd7d6aa5dd82f6/src/Campaigns/Blocks/shared/hooks/useCampaigns.ts#L23

pauloiankoski: The component we use here for the async select dropdown handles many of the async process internally, so it doesn’t work very well with the way entities work. While we could make it work, that would require an important refactor, which I suggest doing at a later time since it would also need to be changed in [useDonorAsyncSelect.tsx](https://github.com/impress-org/givewp/blob/be784639587de19d574980640327c1908c9f30ab/src/Donations/resources/admin/components/DonationDetailsPage/Tabs/Records/General/AssociatedDonor/useDonorAsyncSelect.ts).

pauloiankoski (approved):
Other than that question about using entities instead of the `apiFetch` approach, I think this PR is good to go.

src/Views/Components/ListTable/CustomFilter/useAsyncCampaigns.ts
jonwaldstein: Okay, I think it's worth exploring if entities can be used here instead of the manual fetch.  I'm okay refactoring this part later so it's not a blocker.

PR #8101, opened 2025-09-17
Fix: currency rule should respect GiveWP options

glaubersilva:
Resolves [GIVE-2701]

## Description

This PR implements the following items:

- Swap out the generic currency field rule within the donation form with a custom rule that uses the defined currency list & filter for Give
- Audit the existing list and make sure it’s up to date using  ISO-4217 standards
- Update the Money class to support using the same currency list & filter from Give so the amounts are able to be processed correctly

## Affects

- The V3 forms
- The currencies list
- The Money class

## Visuals

image

image

## Testing Instructions

1. In the `Give > General > Currency` settings, check if the [changes made in the audit](https://github.com/impress-org/givewp/blob/ad8923ab60677c828393df0c9b585eb746745684/includes/currencies-list.php#L10-L28) were reflected
2. Select a new currency for tests, like the `Tuvaluan dollar` (TVD)
3. Make some test donations and make sure everything is working as expected.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2701]: https://stellarwp.atlassian.net/browse/GIVE-2701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DonationForms/Rules/CurrencyRule.php
+++ b/src/DonationForms/Rules/CurrencyRule.php
@@ -3,7 +3,9 @@
 namespace Give\DonationForms\Rules;
 
 use Closure;
+use Give\Framework\Support\Currencies\GiveCurrencies;
 use Give\Vendors\StellarWP\Validation\Contracts\ValidationRule;
+use Money\Currency;
 
 /**
  * Custom currency validation rule that uses GiveWP's currency list and filter.
@@ -39,26 +41,20 @@ class CurrencyRule implements ValidationRule
      */
     public function __invoke($value, Closure $fail, string $key, array $values)
     {
-        // Get GiveWP's supported currencies
-        $supportedCurrencies = array_keys(give_get_currencies_list());
-
         // Check format first for better error messaging
         if (!$this->isValidFormat($value)) {
-            $providedValue = is_array($value) ? 'array' : (is_object($value) ? 'object' : (string) $value);
             $fail(
                 sprintf(
-                    __('%s must be a valid 3-letter currency code in uppercase format (example: USD). Provided: %s', 'give'),
-                    '{field}',
-                    $providedValue
+                    __('%s must be a valid 3-letter currency code in uppercase format (example: USD)', 'give'),
+                    '{field}'
                 )
             );
-        } elseif (!in_array($value, $supportedCurrencies, true)) {
+        } elseif (!give(GiveCurrencies::class)->contains(new Currency($value))) {
             $fail(
                 sprintf(
-                    __('%s must be a valid currency. Provided: %s. Valid currencies are: %s', 'give'),
+                    __('%s must be a valid currency. Provided: %s', 'give'),
                     '{field}',
-                    $value,
-                    implode(', ', $supportedCurrencies)
+                    $value
                 )
             );
         }
--- a/tests/Unit/DonationForms/Rules/TestCurrencyRule.php
+++ b/tests/Unit/DonationForms/Rules/TestCurrencyRule.php
@@ -111,8 +111,7 @@ class TestCurrencyRule extends TestCase
         /** @var string $error */
         $this->assertStringContainsString('must be a valid currency', $error);
         $this->assertStringContainsString('Provided: XYZ', $error);
-        $this->assertStringContainsString('Valid currencies are:', $error);
-        $this->assertStringContainsString(implode(', ', $supportedCurrencies), $error);
+        $this->assertStringNotContainsString('Valid currencies are:', $error);
     }
 
     /**
@@ -145,7 +144,7 @@ class TestCurrencyRule extends TestCase
         /** @var string $error */
         $this->assertStringContainsString('must be a valid 3-letter currency code in uppercase format', $error);
         $this->assertStringContainsString('(example: USD)', $error);
-        $this->assertStringContainsString('Provided: ' . $lowercaseCurrency, $error);
+        $this->assertStringNotContainsString('Provided:', $error);
         // Should not contain the full list of currencies in this specific error message
         $this->assertStringNotContainsString('Valid currencies are:', $error);
     }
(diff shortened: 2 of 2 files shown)


src/DonationForms/Rules/CurrencyRule.php
```diff
@@ -0,0 +1,59 @@
...
+use Give\Vendors\StellarWP\Validation\Contracts\ValidationRule;
+
+/**
+ * Custom currency validation rule that uses GiveWP's currency list and filter.
+ * Replaces the generic stellar validation library currency rule to ensure consistency
+ * with GiveWP's configured currencies and filters.
+ * 
+ * @unreleased
+ */
+class CurrencyRule implements ValidationRule
```
jonwaldstein, line 17: Could you create a unit test for this rule?  I have a helper for asserting rules you can checkout this one:

https://github.com/impress-org/givewp/blob/c4fb6a0cd5863c13f6ea3fd577a98c363f36c5e6/tests/Unit/DonationForms/Rules/TestHoneyPotRule.php#L15

src/DonationForms/Rules/CurrencyRule.php
```diff
@@ -0,0 +1,59 @@
...
+    public function __invoke($value, Closure $fail, string $key, array $values)
+    {
+        if (empty($value)) {
+            return;
+        }
+
+        // Get GiveWP's supported currencies
+        $supportedCurrencies = array_keys(give_get_currencies_list());
+
+        if (!in_array($value, $supportedCurrencies, true)) {
```
jonwaldstein: I wonder if we should account for string case meaning our options are:
- a. keep the check for in_array which would imply exact uppercase syntax 
- b. uppercase the value before checking for in_array so the case does not matter

I'm inclined to keep what you have (a) because the Money class we use checks against the exact values without modifying the case & I believe we also have been storing these values as uppercase in the database.  Still would like your opinion 😄

glaubersilva: I agree with keeping the current approach. However, I've added an improvement to provide a better developer experience: 30fe6c3a6520395deeb44daed627b15d7de69cf2

The goal here is to provide a specific, educational error message when someone uses the `give_register_currency` filter with incorrect formatting (wrong case, length, or non-alphabetic characters). This helps developers understand exactly what they're doing wrong when registering custom currencies, rather than just getting a generic _"currency not supported"_ message.

src/DonationForms/Rules/CurrencyRule.php
glaubersilva: Resolved in d2eaf778ea2b626f0b087e21a4367c07c9236fb2

jonwaldstein: thanks!

src/DonationForms/Rules/CurrencyRule.php
```diff
@@ -0,0 +1,79 @@
...
+     * Validates that the currency code is in GiveWP's supported currency list.
+     * Uses give_get_currencies_list() to get the current supported currencies
+     * and provides clear error messages with valid currency options.
+     *
+     * @unreleased
+     */
+    public function __invoke($value, Closure $fail, string $key, array $values)
+    {
+        // Get GiveWP's supported currencies
+        $supportedCurrencies = array_keys(give_get_currencies_list());
```
jonwaldstein: Would it be possible to reuse the new `GiveCurrencies::contains()` method here?

src/DonationForms/Rules/CurrencyRule.php
```diff
@@ -0,0 +1,79 @@
...
+                )
+            );
+        } elseif (!in_array($value, $supportedCurrencies, true)) {
+            $fail(
+                sprintf(
+                    __('%s must be a valid currency. Provided: %s. Valid currencies are: %s', 'give'),
+                    '{field}',
+                    $value,
+                    implode(', ', $supportedCurrencies)
+                )
```
jonwaldstein, line 58: We don't need to list out the supported currencies here or else it would be a giant list.  Just saying the currency is not valid / supported should be enough

src/DonationForms/Rules/CurrencyRule.php
```diff
@@ -0,0 +1,80 @@
...
+        $supportedCurrencies = array_keys(give_get_currencies_list());
+
+        // Check format first for better error messaging
+        if (!$this->isValidFormat($value)) {
+            $providedValue = is_array($value) ? 'array' : (is_object($value) ? 'object' : (string) $value);
+            $fail(
+                sprintf(
+                    __('%s must be a valid 3-letter currency code in uppercase format (example: USD). Provided: %s', 'give'),
+                    '{field}',
+                    $providedValue
```
jonwaldstein: We don't need to respond with the provided value here, just the messaging about the format is enough.

src/DonationForms/Rules/CurrencyRule.php
glaubersilva: Resolved in 388d2e89bcde0175ee63dc4cc9e659fe449842e8

src/DonationForms/Rules/CurrencyRule.php
glaubersilva: Resolved in 388d2e89bcde0175ee63dc4cc9e659fe449842e8

src/DonationForms/Rules/CurrencyRule.php
glaubersilva: Resolved in 388d2e89bcde0175ee63dc4cc9e659fe449842e8

glaubersilva:
@jonwaldstein This is ready for a new review.

PR #8103, opened 2025-09-18
Fix: Subscriptions import

alaca:
## Description
This PR adds subscription import functionality. When you export donations and import them again, previously, if you had subscriptions, they didn't get imported as subscriptions but as regular one-time donations. This PR resolves that, and users can now import donations that have subscriptions. 

## Visuals
https://www.loom.com/share/352378401b5b494aa381a317b4f6913f?sid=fa2d6a38-35c5-4a1e-95c7-3168cac7a038

## Testing Instructions

Create a recurring donation
Export donations
Delete all donations and subscriptions
Import the exported file
Check the donations and subscriptions list table

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/includes/admin/import-functions.php
+++ b/includes/admin/import-functions.php
@@ -954,8 +954,6 @@ function give_save_import_donation_to_db( $raw_key, $row_data, $main_key = [], $
                                 'createdAt' => new \DateTime( $payment_data['post_date'] ),
                             ] );
 
-                            $subscription->save();
-
                             $payment->update_meta( 'subscription_id',  $subscription->id);
                             $payment->update_meta( '_give_subscription_payment',  1);
                             $payment->update_meta( '_give_is_donation_recurring',  1);


jonwaldstein:
FYI I have a [new PR](https://github.com/impress-org/givewp/pull/8104) that adds a new "Import Subscriptions" tool.

I think we can probably scope this PR down to just supporting renewals in the importer and being able to assign the subscription ID but removing the subscription importing logic as that should be handled as a separate importer tool.

includes/admin/import-functions.php
```diff
@@ -911,6 +918,64 @@ function give_save_import_donation_to_db( $raw_key, $row_data, $main_key = [], $
...
+                // Handle subscription import
+                if (
+                    defined('GIVE_RECURRING_VERSION')
+                    && ! empty( $data['payment_type'] ) && strtolower( $data['payment_type'] ) === 'subscription'
+                ) {
+                    global $wpdb;
+
+                    // Check if subscription already exists
+                    $subscriptionExists = DB::get_var(
+                        DB::prepare( "SELECT COUNT(id) FROM {$wpdb->prefix}give_subscriptions WHERE parent_payment_id = %d", $payment->ID )
```
glaubersilva, line 931: This logic will never get renewal donations associated with a subscription.  It can only identify if the first donation is from a subscription. As result, renewals will be imported without any association with the subscription.

image

image

glaubersilva (changes requested):
@alaca There is no way to make it work to associate renewals with the subscription, as I demonstrated in my comment/screenshots. To make it work, it would be necessary to have a "gateway subscription ID" column to associate all donations (the first and renewals) with the proper subscription.

PR #8107, opened 2025-09-19
Fix: create or edit v2 forms resets goal, min/max, donation level amounts

glaubersilva:
Resolves [GIVE-2890]

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/includes/formatting.php
+++ b/includes/formatting.php
@@ -578,6 +578,7 @@ function give_human_format_large_amount($amount, $args = [])
 /**
  * Returns a nicely formatted amount with custom decimal separator.
  *
+ * @unreleased Rollback backward compatibility for amount param - allows both array and individual parameter calls
  * @since 4.9.0 PHP 8 compatibility
  * @since 1.0
  *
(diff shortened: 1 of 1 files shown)


pauloiankoski:
@glaubersilva I see that similar code has been removed from other methods. Aren’t those also necessary to prevent other errors? Are we keeping this PHP 8.3 compatible? If so, please confirm whether the previous _since_ tag still makes sense, since we are rolling back what was done before.

glaubersilva:
@pauloiankoski Yeah, I was checking these other methods right now, and I think we need to roll back the backward compatibility logic for those as well, since it's not throwing any errors or warnings related to PHP 8 in the logs. I'm closing my day now, but next week I'll take a look again.

glaubersilva:
@pauloiankoski This is ready for review. Also, I updated the PR description with more details about the changes.

pauloiankoski (approved):
I checked this by comparing it with the version prior to the compatibility release and can confirm that the approach used here is the correct one.

PR #8111, opened 2025-09-23
Fix: pagination not shown in donation history

Genevieve-K:
In my example, a donor has made 96 donations, but only 16 are completed. Only the 20 first are displayed in the donation history. Explanation: The `give_count_donations_of_donor()` function calls `give_get_donation_stats_by_user( $user )`, and `$donor->purchase_count` is used to count donations. However this attribute counts only completed donations, whereas in the donation history, all donations are displayed (status `any`). As 16 < 20, pagination is not necessary.
I suggested using `$donor->payment_ids` which contains all donations, and counting them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved donation history counts for logged-in and email-token users.
  * Displays zero donations when no matching donor record exists.
  * Updated pagination to accurately reflect the number of donations per page.


includes/user-functions.php
```diff
@@ -261,7 +262,7 @@ function give_count_donations_of_donor( $user = null ) {
 
 	$stats = ! empty( $user ) ? give_get_donation_stats_by_user( $user ) : false;
 
-	return isset( $stats['purchases'] ) ? $stats['purchases'] : 0;
+	return isset( $stats['donations_count'] ) ? $stats['donations_count'] : 0;
```
jonwaldstein: If the donation history shortcode has a status param, you might want to account for that here so the pagination works for `any` or `completed/publish`

```suggestion
    return $status === 'any' ? ($stats['donations_count'] ?? 0) : ($stats['purchases'] ?? 0)
```

jonwaldstein (changes requested):
Thanks for this, and for the writeup — the diagnosis is exactly right and it's a genuinely annoying bug. The donation history table lists every status (`give_get_users_donations( ..., 'any' )` at `templates/history-donations.php:11`) while the pagination divides by `purchase_count`, which only counts completed donations. A donor with 96 donations and 16 completed gets no pagination at all.

My hesitation is with the lever rather than the diagnosis. The change redefines `give_count_donations_of_donor()` for every caller, and core has exactly one:

```
templates/history-donations.php:268    'total' => ceil( give_count_donations_of_donor() / 20 ),
```

The function is public and filtered, so add-ons and customisations may well be calling it and relying on the completed-donation meaning it has had since 1.0. Changing it under them to fix one caller's pagination is a wider blast radius than the bug needs.

It also currently fails CI — these three pass on `develop` and fail here:

```
Tests_Give_Donors::test_increment_stats            '2' vs expected 1
Tests_Give_Donors::test_decrement_stats            '0' vs expected 1
Tests_Give_Donors::test_get_donation_stats_by_user  3 matches expected 2
```

The first two are the contract changing: `increase_purchase_count()` and `decrease_donation_count()` no longer move what the function returns. The third is just the array growing a key.

Could you move the count to the call site instead? `history-donations.php` already does exactly this a few lines up, at line 22, for the email-access throttle check:

```php
$donations_count = count( explode( ',', $donor->payment_ids ) );
```

So computing the pagination total the same way is consistent with the file, fixes the reported bug, and leaves `give_count_donations_of_donor()` and its tests alone.

Two smaller things whenever you next push:

- The new line is indented with spaces; the rest of the file uses tabs.
- Please add a changelog entry — `composer run changelog:add` writes the YAML file, and it gets committed with the change. Happy to add it for you if you'd rather.

Really appreciate you tracking this down — the explanation in the description made it easy to verify.

Genevieve-K:
Hello @jonwaldstein . I use another method, better. 
First I revert my changes in `user_functions.php`, the functions `give_get_donation_stats_by_user()` and  `give_count_donations_of_donor()` are not changed. 
In `history-donations.php`, I don't use the function `give_count_donations_of_donor()`, but I calculate it with the variable `$donations_count`.
there are 3 cases:
1. The user is logged: we can count the donations from the user ID with `count( explode( ',', $donor->payment_ids ) )`
2. Email access: we can count the donations from the user email with `count( explode( ',', $donor->payment_ids ) )`
3. Get the donor from the session: `$donations_count` is already set.

About the history by donation status, the donation history is always for all donations (status 'any').  The parameter status for the shortcode` [donation_history] `is only to display or not the status.

PR #8114, opened 2025-09-25
Fix: custom AJV resolver is throwing errors and warnings in the browser console, and is not ignoring readonly fields from validation

glaubersilva:
Resolves [GIVE-2922]

## Description

This PR edits our custom AJV resolver to support enum fields that can also have `null` values, like the honorific field of the donation REST V3 endpoint. Reference: https://github.com/impress-org/givewp/blob/645381b3ead39f35cd063aadb5c7f91aa5a8a374/src/API/REST/V3/Routes/Donations/DonationController.php#L853-L857

This way, we can prevent this kind of error in the browser console when the field is null: `enum value must be ["array"]`

Also, it adds the `integer` and `boolean` formats to our custom AJV resolver so we can prevent warnings like this in the browser console: `unknown format "integer" ignored in schema`

Beyond that, it also makes the AJV resolver ignore read-only fields from being validated.

## Affects

All screens using the AJV resolver to validate form submissions: 

- Campaign details page
- Donation details page
- Donor details page
- Subscription details page

## Visuals

image

## Testing Instructions

1. Go to the `Donation` details screen
2. Update some information and click on the save button 
3. Make sure no errors or warnings mentioned in the description section were thrown on the browser console
4. Repeat the same steps above for the other details page: `Campaigns`, `Donors`, `Subscriptions`

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2922]: https://stellarwp.atlassian.net/browse/GIVE-2922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Admin/ajv.ts
+++ b/src/Admin/ajv.ts
@@ -15,6 +15,7 @@ import addFormats from 'ajv-formats';
  * Key advantage: WordPress REST API supports most JSON Schema Draft 4 features but lacks
  * some advanced features (if/then/else, allOf, not) that AJV can provide for enhanced frontend validation.
  *
+ * @unreleased Refactor transformWordPressSchemaToDraft7 to handle readonly/readOnly fields and conditionally remove enum from nullable fields when value is null to prevent AJV conflicts
  * @since 4.9.0
  *
  * @param schema - The JSON Schema from WordPress REST API


src/Admin/ajv.ts
```diff
@@ -196,6 +197,29 @@ function transformWordPressSchemaToDraft7(schema: JSONSchemaType<any>): JSONSche
                 delete prop.required;
             }
 
+            // Handle WordPress schema with Array type and enum
+            if (prop && typeof prop === 'object' && !Array.isArray(prop)) {
```
jonwaldstein: I noticed we're doing this check for `prop && typeof prop === 'object'` for every conditional.  Perhaps we can clean this up and just return early if this is false so the rest are clear to checking for specific things.

glaubersilva: @jonwaldstein Good idea! I resolved it in my last commits.

jonwaldstein: Much better, thanks!

jonwaldstein (approved):
@glaubersilva great job!

PR #8115, opened 2025-09-25
Changed: Filter read-only fields from admin detail forms

pauloiankoski:
## Description

This PR refactors the admin detail page forms to properly filter out read-only fields from React Hook Form (RHF) initialization and updates CustomFields components to use entity values instead of form field values.

**HOW this solves the problem:**
- Implements a recursive  utility function that removes read-only fields from entity records before initializing forms
- Updates CustomFields components to access entity data directly instead of through RHF, preventing issues with read-only fields being included in form state
- Adds proper TypeScript interfaces for schema properties to ensure type safety

**WHY these choices were made:**
- Read-only fields should not be part of form state as they cannot be edited by users
- CustomFields need to display actual entity data, not form data, to show accurate information including read-only fields
- Recursive filtering handles nested objects and arrays properly for complex entity structures

## Affects

- **Admin detail pages**: Donations, Donors, Subscriptions, and Campaigns detail forms
- **CustomFields components**: Now use entity data instead of form data
- **Form initialization**: Read-only fields are filtered out before setting default form values
- **Type safety**: New SchemaProperty interface ensures proper typing

## Visuals

No visual changes - this is an internal refactor that improves data handling without affecting the UI appearance.

## Testing Instructions

1. Navigate to any admin detail page (Donation, Donor, Subscription, or Campaign)
2. Verify that:
   - The form loads correctly without read-only fields in the form state
   - CustomFields sections display accurate data
   - Read-only fields like gateway info, projected revenue, etc. are not editable
   - Nested objects (like amount, fund info) are handled properly
   - Form submission works as expected

## Pre-review Checklist

- [ ] Includes unit tests

### DIFF
--- a/src/Views/Components/AdminDetailsPage/index.tsx
+++ b/src/Views/Components/AdminDetailsPage/index.tsx
@@ -28,6 +28,7 @@ import TabPanels from './Tabs/TabPanels';
 import {AdminDetailsPageProps} from './types';
 
 import './store';
+import { filterOutReadOnlyFields } from './utils';
 
 /**
  * @since 4.4.0
@@ -37,7 +38,6 @@ export default function AdminDetailsPage<T extends Record<string, any>>({
     objectType,
     objectTypePlural,
     useObjectEntityRecord,
-    resetForm,
     shouldSaveForm,
     breadcrumbUrl,
     breadcrumbTitle,
@@ -51,6 +51,8 @@ export default function AdminDetailsPage<T extends Record<string, any>>({
 }: AdminDetailsPageProps<T>) {
     const [resolver, setResolver] = useState({});
     const [isSaving, setIsSaving] = useState(false);
+    const [isLoading, setIsLoading] = useState(true);
+    const [schema, setSchema] = useState<JSONSchemaType<any> | null>(null);
     const [showContextMenu, setShowContextMenu] = useState<boolean>(false);
     const contextMenuButtonRef = useRef<HTMLButtonElement>(null);
     const contextMenuRef = useRef<HTMLDivElement>(null);
@@ -68,6 +70,7 @@ export default function AdminDetailsPage<T extends Record<string, any>>({
             path: `/givewp/v3/${objectTypePlural}/${objectId}`,
             method: 'OPTIONS',
         }).then(({schema}: {schema: JSONSchemaType<any>}) => {
+            setSchema(schema);
             setResolver({
                 resolver: ajvResolver(schema),
             });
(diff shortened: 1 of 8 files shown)


src/Views/Components/AdminDetailsPage/utils.ts
```diff
@@ -42,3 +44,38 @@ export function formatDateLocal(dateString: string) {
 
     return `${year}-${month}-${day}`;
 }
+
+/**
+ * @unreleased
+ */
+export function filterOutReadOnlyFields(
```
jonwaldstein: @pauloiankoski nice job! couple thoughts here:

1. I think this is generic enough to RHF and REST schema that will be useful outside of just this component directory -  so we should elevate it perhaps up into `src/Admin/`
2. Since the actual filtering of `readonly` is pretty contextual and is kind of a small part here, I would probably rename this function to be like `prepareEntityRecordFromSchema()` and try to simplify the logic a bit to be easier to understand perhaps with some small utility functions inside for checking `isReadOnly()`, `isObject()`, etc.  Then, if we need to update our filtering in the future we have a dedicated place within this function to do so since it's already parsing the record.

src/Views/Components/AdminDetailsPage/index.tsx
```diff
@@ -110,14 +113,12 @@ export default function AdminDetailsPage<T extends Record<string, any>>({
...
-            } else {
-                reset(record);
-            }
+        if (hasResolved && schema && record) {
+            const filteredRecord = filterOutReadOnlyFields(record, (schema as any)?.properties) as T;
+            reset(filteredRecord);
+            setIsLoading(false);
         }
-    }, [hasResolved]);
+    }, [hasResolved, schema, record]);
```
jonwaldstein: Be careful about adding objects as a dependency in useEffect.  They cause re-renders because to comparing objects will always return false.  Instead, the workaround is to use JSON.stringify() around the object so that you are comparing a string value.

jonwaldstein (changes requested):

src/Views/Components/AdminDetailsPage/utils.ts
pauloiankoski: Resolved on 698957567b4edd4c6b6111ba198ce502b79ef9fb

src/Views/Components/AdminDetailsPage/index.tsx
pauloiankoski: Resolved on dc00d93fc1d21fcac543e319f65f9d5054f84b53

Also, I noticed that we were not preparing the data received after saving the entity, which could cause RHF to be updated with read-only fields again.

src/Admin/utils.ts
```diff
@@ -66,3 +67,50 @@ export function getRelativeTimeString(date: Date): string {
     }
     return formatDistanceToNow(date, {addSuffix: true});
 }
+
+/**
+ * @unreleased
+ */
+export function prepareDefaultValuesFromSchema(
```
jonwaldstein: Personal preference, I like to create separate files for each utility function and then have a `common` or `utilities` folder that exports them all.  It's always easier to find that way or else these utils files get massive and hard to manage 😏

pauloiankoski: Thank you! I have refactored that on 66e513883f08797026d2053950094ecd2518c107

pauloiankoski:
> @pauloiankoski is this still necessary now or can we remove?
> 
> https://github.com/impress-org/givewp/blob/c08e77123d91fce3f812bc2db775f4a59c145de8/src/API/REST/V3/Routes/Subscriptions/Actions/GetSubscriptionItemSchema.php#L151

No, it's not necessary and it's not included in this PR. I'll update my branch with develop and ensure that is removed.

PR #8119, opened 2025-09-29
Feature: Add associated donor to the Subscription Records tab

alaca:
## Pre-review Checklist

-   [ ] Includes unit tests


src/Donations/ServiceProvider.php
```diff
@@ -164,7 +165,7 @@ private function registerDonationEntity()
     private function loadDonationAdminOptions()
     {
         add_action('admin_enqueue_scripts', function () {
-            if (DonationsAdminPage::isShowingDetailsPage()) {
+            if (DonationsAdminPage::isShowingDetailsPage() || SubscriptionsAdminPage::isShowing()) {
```
jonwaldstein: @alaca I know your still working on this but let's try to avoid this as the admin page window data has been intentionally scoped to their own pages.  This is to avoid unwanted data leaks and limit the amount of global objects.

I suspect this has to do with trying to move a field into a shared component.  If that's the case I have a couple suggestions:
1. Locate the stuff it needs from the window and turn it into a prop that can be passed in based on the context (Subscription screen vs Donation screen). 
2. Don't move the whole field into a shared component, instead create a new field for the Subscription page that uses the same underlying shared components but for it's own context
3. Perhaps there is a need for some kind of general settings endpoint instead of using window data

src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Records/SubscriptionDetails.tsx
```diff
@@ -13,13 +14,21 @@ export default function SubscriptionDetails() {
...
+                title={__('Subscription details', 'give')}
+                description={__('This includes the subscription information', 'give')}
+            >
+                <AdminSectionField error={errors.status?.message as string}>
+                    <StatusField statusOptions={subscriptionStatuses} />
+                </AdminSectionField>
+            </AdminSection>
+            <AdminSection
+                title={__('Associated donor', 'give')}
+                description={__('Manage the donor connected to this donation', 'give')}
```
jonwaldstein:
```suggestion
                description={__('Manage the donor connected to this subscription', 'give')}
```

src/Admin/fields/AssociatedDonor/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+    const {selectedOption, loadOptions, mapOptionsForMenu, error} = useDonorAsyncSelect(donationDonorId || null, mode);
+
+    const handleDonorChange = (selectedOption: DonorOption | null) => {
+        setValue('donorId', selectedOption?.value ?? null, {shouldDirty: true});
+    };
+
+    return (
+        <AdminSectionField error={errors.donorId?.message as string}>
+            <label htmlFor="donorId">{__('Donor', 'give')}</label>
+            <p className={styles.fieldDescription}>{__('Link the donation to the selected donor', 'give')}</p>
```
jonwaldstein: This description will be different between screens so we should add a prop for `description`.  

Also, where is the `styles.fieldDescription` coming from?

src/Admin/fields/AssociatedDonor/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * @unreleased add mode prop
+ * @since 4.9.0 Add error prop to all AdminSectionField components
+ * @since 4.8.0 updated to async donor dropdown
+ * @since 4.6.0
+ */
+export default function AssociatedDonor({mode}: {mode: 'test' | 'live'}) {
+    const {watch, setValue} = useFormContext();
+    const {errors} = useFormState();
+    const donationDonorId = watch('donorId');
```
jonwaldstein: As a shared field, we should make this more agnostic without making too many assumptions.  For example, I think we should add a prop for `name` instead of assuming the field name will be `donorId` as that might not always be the case inside a form.  We should also just use the word `donorId` instead of `donationDonorId` as that implies context.

```suggestion
    const donorId = watch(name);
```

src/Admin/fields/AssociatedDonor/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+    const {selectedOption, loadOptions, mapOptionsForMenu, error} = useDonorAsyncSelect(donationDonorId || null, mode);
+
+    const handleDonorChange = (selectedOption: DonorOption | null) => {
+        setValue('donorId', selectedOption?.value ?? null, {shouldDirty: true});
+    };
+
+    return (
+        <AdminSectionField error={errors.donorId?.message as string}>
```
jonwaldstein: If using the name prop this would be like

`errors[name]?.message`

src/Admin/fields/AssociatedDonor/index.tsx
alaca: Resolved: https://github.com/impress-org/givewp/pull/8119/commits/78fbb4bb917f14cc3b7e2280f1acdc6b76181411

src/Admin/fields/AssociatedDonor/index.tsx
alaca: Resolved: https://github.com/impress-org/givewp/pull/8119/commits/78fbb4bb917f14cc3b7e2280f1acdc6b76181411

src/Admin/fields/AssociatedDonor/index.tsx
alaca: Resolved: https://github.com/impress-org/givewp/pull/8119/commits/ef2bce1a9cb856c9258fe1bf2a2f9403765a9435

src/Subscriptions/resources/admin/components/SubscriptionDetailsPage/Tabs/Records/SubscriptionDetails.tsx
alaca: Resolved: https://github.com/impress-org/givewp/pull/8119/commits/ef2bce1a9cb856c9258fe1bf2a2f9403765a9435

src/Admin/fields/AssociatedDonor/index.tsx
jonwaldstein: We'll still need to account for the inner description style here, do you mind if I add something?

alaca: @jonwaldstein, go for it.

jonwaldstein: Thanks! Resolved 5e491649

PR #8120, opened 2025-09-29
Feature: add gateway subscription ID field to the records tab from the subscription details page

glaubersilva:
Resolves [GIVE-2855]

## Description

This PR adds a new field to the records tab from the subscription details page, which allows users to update the gateway subscription ID.

## Affects

The subscription details page.

## Visuals

image

image

image

## Testing Instructions

1. Create a test subscription
2. Go to the records tab from the subscription details page 
3. Edit the new field gateway subscription ID 
4. Save the form, refresh the page, and make sure the data is persisted on the DB

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2855]: https://stellarwp.atlassian.net/browse/GIVE-2855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Admin/fields/LockedTextInput/index.tsx
+++ b/src/Admin/fields/LockedTextInput/index.tsx
@@ -42,7 +42,7 @@ export default function LockedTextInput({name, label, description, placeholder, 
     return (
         <AdminSectionField error={errors[name]?.message as string}>
             <label htmlFor={name}>{label}</label>
-            <p className={styles.description}>{description}</p>
+            <p>{description}</p>
             <div className={styles.inputContainer}>
                 <input
                     id={name}
--- a/src/Admin/fields/LockedTextInput/styles.module.scss
+++ b/src/Admin/fields/LockedTextInput/styles.module.scss
@@ -3,7 +3,7 @@
     align-items: center;
     gap: 20px;
     width: 100%;
-    margin-bottom: 8px;
+    margin-bottom: var(--givewp-spacing-1);
 }
 
 .input {
@@ -26,11 +26,4 @@
         height: 24px;
         transition: stroke 0.2s ease;
     }
-}
-
-.description {
-    color: var(--givewp-grey-600);
-    font-size: 0.875rem;
-    margin: 4px 0 8px 0;
-    line-height: 1.4;
 }


src/Admin/fields/LockedTextInput/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+                {description}
+            </p>
+            <div className={styles.inputContainer}>
+                <input
+                    id={name}
+                    type="text"
+                    className={`givewp-admin-field-input ${styles.input}`}
+                    {...register(name)}
+                    placeholder={placeholder}
+                    disabled={!isEditing}
```
jonwaldstein: This would be something to look up... but I think we should use the `readonly` property here instead of `disabled` as I believe readonly is better for still being able to focus on this input & copy the text

src/Admin/fields/LockedTextInput/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <AdminSectionField error={errors[name]?.message as string}>
+            <label htmlFor={name}>{label}</label>
+            <p className={styles.description}>
+                {description}
+            </p>
+            <div className={styles.inputContainer}>
+                <input
+                    id={name}
+                    type="text"
+                    className={`givewp-admin-field-input ${styles.input}`}
```
jonwaldstein: Don't we input have styles coming from the parent Admin screen component? Also, I don't think we're using classnames like this anywhere else.

src/Admin/fields/LockedTextInput/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+    description,
+    placeholder,
+    warningMessage
+}: LockedTextInputProps) {
+    const {register} = useFormContext();
+    const {errors, isSubmitSuccessful} = useFormState();
+    const [isEditing, setIsEditing] = useState(false);
+
+    const handleEditClick = () => {
+        setIsEditing(!isEditing);
```
jonwaldstein: Let's also set the focus of the input using [react-hook-form](https://react-hook-form.com/docs/useform/setfocus) when clicked so the cursor goes to the input.

jonwaldstein (changes requested):

src/Admin/fields/LockedTextInput/index.tsx
glaubersilva: Resolved in ff9efa306078c28106a2c99529470b11c60e2b26

src/Admin/fields/LockedTextInput/index.tsx
glaubersilva: Ops, seems Cursor added that and I didn't review it. 🫣

Resolved in 7c16dd5fc1def7c3c3bf915631f1928157edd44b

src/Admin/fields/LockedTextInput/index.tsx
glaubersilva: Resolved in d9279d5cc764e3f95790cf3c3e37c960747348c7

PR #8121, opened 2025-09-29
'Netherlands Antilles' is replaced by 'Curaçao' since 2010

Genevieve-K:
See https://en.wikipedia.org/wiki/Cura%C3%A7ao

I don't know if it's necessary to keep  'Netherlands Antilles' for donations with this country.

### DIFF
--- a/includes/country-functions.php
+++ b/includes/country-functions.php
@@ -2333,6 +2333,7 @@ function give_get_spain_states_list() {
 		'CR' => esc_html__( 'Ciudad Real', 'give' ),
 		'CO' => esc_html__( 'C&oacute;rdoba', 'give' ),
 		'CU' => esc_html__( 'Cuenca', 'give' ),
+        'CW' => esc_html__( 'Curaçao', 'give' ),
 		'GI' => esc_html__( 'Girona', 'give' ),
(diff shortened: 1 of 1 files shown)


includes/country-functions.php
```diff
@@ -239,7 +239,7 @@ function give_get_country_list() {
 		'NR' => esc_html__( 'Nauru', 'give' ),
 		'NP' => esc_html__( 'Nepal', 'give' ),
 		'NL' => esc_html__( 'Netherlands', 'give' ),
-		'AN' => esc_html__( 'Netherlands Antilles', 'give' ),
+		'CW' => esc_html__( 'Curaçao', 'give' ),
```
jonwaldstein, line 242: The tricky part about this is the `Netherlands Antilles` country code can be assigned/stored in the give settings for people and removing it could break something for them when updating.  I would suggest keeping the old one and adding this new one so people can choose to update themselves

jonwaldstein (changes requested):

includes/country-functions.php
Genevieve-K: I had asked myself the question and wondered if it was better to keep the old one, which might be in donor addresses.

PR #8125, opened 2025-10-01
Feature: Add Campaign and Form fields to Subscription Details screen

pauloiankoski:
Resolves [GIVE-2858]

## Description

This PR adds Campaign and Form fields to the Subscription Details screen, matching the functionality that was previously available on the Donation Details screen. To achieve this, we first refactored the existing fields from the Donation Details screen by extracting them into a generic reusable component, then further refactored that component to use Async Select dropdowns for better performance and user experience.

The implementation includes:
- Created a generic  component that can be reused across different admin pages
- Implemented  component for better performance with large datasets
- Added  hook for managing async select functionality
- Refactored the Donation Details screen to use the new generic components
- Added Campaign and Form fields to the Subscription Details screen

## Affects

- Subscription Details screen now displays Campaign and Form fields
- Donation Details screen now uses the refactored generic components
- New reusable components available for future admin page implementations
- Improved performance for campaign and form selection with async loading

## Visuals

The Subscription Details screen now includes Campaign and Form fields in the Records tab, providing administrators with better visibility into subscription relationships and allowing for easier management of subscription data.

## Testing Instructions

1. Navigate to the Subscriptions admin page
2. Open any subscription details page
3. Verify that Campaign and Form fields are now visible in the Records tab
4. Test that the fields are editable and save correctly
5. Verify that the Donation Details screen still works as expected with the refactored components
6. Test the async loading functionality with large numbers of campaigns and forms

## Pre-review Checklist

- [x] Includes unit tests

[GIVE-2858]: https://stellarwp.atlassian.net/browse/GIVE-2858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Admin/hooks/useAsyncSelectOption.ts
```diff
@@ -0,0 +1,133 @@
...
+}
+
+export type Option = {
+    value: number;
+    label: string;
+}
+
+export type AsyncSelectOptionsConfig = {
+    recordId: number | null;
+    entity: string;
```
alaca: Is there any reason why we should keep this? Right now is only used in an error message, but we can make that message more generic.

src/API/REST/V3/Routes/Subscriptions/Actions/GetSubscriptionItemSchema.php
```diff
@@ -31,6 +31,11 @@ public function __invoke(): array
                     'default' => 'live',
                     'enum' => ['live', 'test'],
                 ],
+                'campaignId' => [
+                    'type' => 'integer',
+                    'description' => esc_html__('Campaign ID', 'give'),
+                    'nullable' => true,
```
alaca, line 37: Interesting. I didn't know that you could set `nullable` arg. Usually, you would do something like `['integer', 'null']`, but this looks nice 👍

alaca (approved):
Looks good, works as expected 👍 
The only thing I noticed is that when you change the campaign, the change does not reflect on the overview page until you reload the page. I'm fine with this, but I'm almost positive the QA will notice it. 

Love the async select option abstraction 😅

src/Admin/hooks/useAsyncSelectOption.ts
pauloiankoski: Cleaned up on 3292cb077e0cd70ec3dd41f7b0b043e81812c339. Also applied a fix for the reflect changes issue on 90feeb6fa1770832dc65d8e4858e909e01f9731c.

PR #8136, opened 2025-10-10
Feature: Add Donor sorting for "Total Given" column

JoshuaHungDinh:
Resolves [GIVE-2750]

## Description
Refactored the `ListDonors` endpoint to move sorting logic into a dedicated method and added support for sorting by latest donation date and total donation revenue.

### Changes
- **Added donationRevenue sorting**: Sort donors by their total donation amount with proper currency handling

### Technical Details
- Donation revenue calculates the sum of donation amounts with currency decimal precision

## Affects
Donor List Table
Column Sorting
Columns: Donation Revenue (Total Given)

## Visuals
https://github.com/user-attachments/assets/b8332e5e-5da1-4b6e-af68-d262d88b6474
(Updated removed Latest Donation sorting)

## Testing Instructions
- Create several donors and make several donations between them.
- View the Donor List Table
- Verify accurate sorting of each column by clicking on the new sorting buttons added for the "Total Given" column

## Pre-review Checklist
-   [ ] Includes unit tests

[GIVE-2750]: https://stellarwp.atlassian.net/browse/GIVE-2750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ServiceProviders/RestAPI.php
+++ b/src/ServiceProviders/RestAPI.php
@@ -31,6 +31,7 @@ use Give\Donors\Endpoints\ListDonorsStats;
 use Give\Donors\Endpoints\SwitchDonorView;
 use Give\Donors\Endpoints\UpdateStatus;
 use Give\Subscriptions\Endpoints\ListSubscriptions;
+use Give\Subscriptions\Endpoints\ListSubscriptionStats;
 use Give\Subscriptions\Endpoints\SubscriptionActions;
 use Give\Subscriptions\Endpoints\SwitchSubscriptionView;
 
@@ -59,6 +60,7 @@ class RestAPI implements ServiceProvider
         ListDonations::class,
         ListDonationsStats::class,
         ListSubscriptions::class,
+        ListSubscriptionStats::class,
         SwitchDonorView::class,
         SwitchDonationView::class,
         SwitchDonationFormView::class,
--- a/src/Subscriptions/Endpoints/ListSubscriptionStats.php
+++ b/src/Subscriptions/Endpoints/ListSubscriptionStats.php
@@ -1,0 +1,127 @@
+<?php
+
+namespace Give\Subscriptions\Endpoints;
+
+use Give\Framework\Database\DB;
+use Give\Donations\ValueObjects\DonationMetaKeys;
+use WP_REST_Request;
+use WP_REST_Response;
+
+/**
+ * @unreleased
+ */
+class ListSubscriptionStats extends Endpoint
+{
+    /**
+     * @var string
+     */
+    protected $endpoint = 'admin/subscriptions/stats';
+
+    /**
+     * @var WP_REST_Request
+     */
+    protected $request;
+
+    /**
+     * @unreleased
+     */
+    public function registerRoute()
+    {
+        register_rest_route(
+            'give-api/v2',
+            $this->endpoint,
+            [
+                [
+                    'methods' => 'GET',
+                    'callback' => [$this, 'handleRequest'],
+                    'permission_callback' => [$this, 'permissionsCheck'],
+                ],
+                'args' => [
+                    'testMode' => [
+                        'type' => 'boolean',
+                        'required' => false,
+                        'default' => give_is_test_mode(),
+                    ],
+                ],
+            ]
+        );
+    }
+
+    /**
+     * @unreleased
+     */
+    public function handleRequest(WP_REST_Request $request): WP_REST_Response
+    {
+        $this->request = $request;
+
+        $statistics = $this->getSubscriptionStatistics();
+
+        return new WP_REST_Response($statistics);
+    }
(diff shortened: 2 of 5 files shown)


src/Donors/ListTable/Columns/DonationRevenueColumn.php
```diff
@@ -14,6 +14,7 @@
  */
 class DonationRevenueColumn extends ModelColumn
 {
+    protected $sortColumn = 'donationRevenue';
```
glaubersilva: We can use the `totalAmountDonated` here, so we don't need to change anything in the endpoint. Also, let's add an unreleased tag for it.

src/Donors/ListTable/Columns/LatestDonationColumn.php
```diff
@@ -17,6 +17,9 @@
 class LatestDonationColumn extends ModelColumn
 {
 
+    protected $sortColumn = 'latestDonation';
```
glaubersilva: We can use the `id` here, so we don't need to change anything in the endpoint. Also, let's add an unreleased tag for it.

src/Donors/Endpoints/ListDonors.php
```diff
@@ -152,15 +155,11 @@ public function getDonors(): array
...
-        $sortColumns = $this->listTable->getSortColumnById($this->request->get_param('sortColumn') ?: 'id');
-        $sortDirection = $this->request->get_param('sortDirection') ?: 'desc';
 
         $query = give()->donors->prepareQuery();
         $query = $this->getWhereConditions($query);
 
-        foreach ($sortColumns as $sortColumn) {
-            $query->orderBy($sortColumn, $sortDirection);
-        }
+        $this->sortColumns($query);
```
glaubersilva: We don't need to change anything here if we use the proper sort columns that I mentioned in my other comments.

For better context, check the links below to see that the donors query used by this endpoint already has the columns that we want to sort:

https://github.com/impress-org/givewp/blob/5d3211b3b024afabf06a7886de51f243768711b8/src/Donors/Repositories/DonorRepository.php#L406

https://github.com/impress-org/givewp/blob/5d3211b3b024afabf06a7886de51f243768711b8/src/Donors/Repositories/DonorRepository.php#L411


So, all we need to do is pass these same values to the `$sortColumn` properties from the column classes.

src/Donors/ListTable/Columns/LatestDonationColumn.php
pauloiankoski: I think using the donor ID would work for one-time donors, but wouldn’t that assumption break once a donor has multiple donations?

glaubersilva: @pauloiankoski Good point! In that case, we can use the `createdAt` column:

Reference: https://github.com/impress-org/givewp/blob/5d3211b3b024afabf06a7886de51f243768711b8/src/Donors/Repositories/DonorRepository.php#L414

pauloiankoski: Is the donor `date_created` column updated on every donation a donor make?

glaubersilva: @pauloiankoski  Ops! I wrote my last comments thinking of the donation list table. 🫠

So yeah, we'll need to keep using the other approach (or something similar) implemented by @JoshuaHungDinh on this PR for this column.

We can maybe add some logic to change the main query only when necessary, directly in the getDonors method, like I did here: https://github.com/impress-org/givewp/blob/92cd70f6142567e50797b311904ad47097e581f6/src/Subscriptions/Endpoints/ListSubscriptions.php#L166-L175

Or we can just keep the current logic and just tweak it to work with this column only.

src/Donors/ListTable/Columns/LatestDonationColumn.php
```diff
@@ -17,6 +17,9 @@
 class LatestDonationColumn extends ModelColumn
 {
 
+    // protected $sortColumn = 'latestDonation';
+
+
```
glaubersilva: Let's remove this leftover for now to keep it cleaner, and we can re-add it later when necessary.

glaubersilva (approved):
@JoshuaHungDinh Looks good to me. I just left a comment about a minor thing.

src/Donors/ListTable/Columns/LatestDonationColumn.php
JoshuaHungDinh: Thanks! 6c6dc03bb17dcc4a950517f168ea9ca7dd2ad54b

Robincredible (approved):
Passed QA Tests

PR #8138, opened 2025-10-14
Feature: Add trash option to the subscriptions list table

alaca:
## Description

This pull request introduces new bulk actions for managing subscription statuses in the Subscriptions List Table. The  frontend UI has been updated to allow users to trash, restore, and delete subscriptions based on their status.

## Affects

Subscriptions list table
Subscriptions list table endpoint

## Visuals

https://github.com/user-attachments/assets/7fd95d56-4bdf-45c2-8691-7f4b9ea9a753

## Testing Instructions

1. Navigate to **GiveWP > Subscriptions**
2. Check the items show “Edit” and “Trash” as inline actions on hover
3. Confirm those actions work as expected
4. Check the Bulk Actions menu, it may show many options except Restore and Delete
5. Confirm those bulk actions work as expected, specially the Trash option
6. Using the Filter By menu, select “Show trashed” and apply
7. In the filtered list, check the items now show “Restore” and "Delete” as inline actions on hover
8. Confirm those actions work as expected
9. Check the Bulk Actions menu, it may only show Restore and Delete options
10. Confirm those bulk actions work as expected

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Subscriptions/Repositories/SubscriptionRepository.php
+++ b/src/Subscriptions/Repositories/SubscriptionRepository.php
@@ -368,6 +368,11 @@ class SubscriptionRepository
         try {
             $previousStatus = give()->subscription_meta->get_meta($subscription->id, '_wp_trash_meta_status', true);
 
+            // If no previous status was saved, default to 'active'
+            if (empty($previousStatus)) {
+                $previousStatus = SubscriptionStatus::ACTIVE;
+            }
+
             DB::table('give_subscriptions')
                 ->where('id', $subscription->id)
                 ->update(['status' => $previousStatus]);
--- a/tests/Unit/Subscriptions/Repositories/TestSubscriptionRepository.php
+++ b/tests/Unit/Subscriptions/Repositories/TestSubscriptionRepository.php
@@ -287,4 +287,92 @@ class TestSubscriptionRepository extends TestCase
 
         $this->assertSame($campaign2->id, $renewal->campaignId);
     }
+
+    /**
+     * @unreleased
+     *
+     * @throws Exception
+     */
+    public function testTrashShouldMoveSubscriptionToTrash(): void
+    {
+        /** @var Subscription $subscription */
+        $subscription = Subscription::factory()->create([
+            'status' => SubscriptionStatus::ACTIVE(),
+        ]);
+
+        $repository = new SubscriptionRepository();
+
+        $repository->trash($subscription);
+
+        $subscriptionStatus = DB::table('give_subscriptions')
+            ->where('id', $subscription->id)
+            ->value('status');
+
+        $trashMetaStatus = give()->subscription_meta->get_meta($subscription->id, '_wp_trash_meta_status', true);
+        $trashMetaTime = give()->subscription_meta->get_meta($subscription->id, '_wp_trash_meta_time', true);
+
+        $this->assertEquals(SubscriptionStatus::TRASHED, $subscriptionStatus);
+        $this->assertEquals(SubscriptionStatus::ACTIVE, $trashMetaStatus);
+        $this->assertNotEmpty($trashMetaTime);
+    }
+
+    /**
+     * @unreleased
+     *
+     * @return void
+     *
+     * @throws Exception
+     */
+    public function testUnTrashShouldRestorePreviousStatus()
+    {
+        /** @var Subscription $subscription */
+        $subscription = Subscription::factory()->create([
+            'status' => SubscriptionStatus::PENDING(),
+        ]);
+
+        $repository = new SubscriptionRepository();
+
+        // First trash the subscription
+        $repository->trash($subscription);
+
+        // Then untrash it
+        $repository->unTrash($subscription);
+
+        $subscriptionStatus = DB::table('give_subscriptions')
+            ->where('id', $subscription->id)
+            ->value('status');
+
+        $this->assertEquals(SubscriptionStatus::PENDING, $subscriptionStatus);
+    }
+
+    /**
+     * @unreleased
+     *
+     * @throws Exception
+     */
+    public function testUnTrashShouldRestoreActiveStatusWhenNoPreviousStatus(): void
+    {
+        /** @var Subscription $subscription */
+        $subscription = Subscription::factory()->create([
+            'status' => SubscriptionStatus::ACTIVE(),
+        ]);
+
+        // Manually set the status to trashed without using the trash method
+        // This simulates a subscription that was trashed without the proper meta being saved
+        DB::table('give_subscriptions')
+            ->where('id', $subscription->id)
+            ->update(['status' => SubscriptionStatus::TRASHED()->getValue()]);
+
+        $repository = new SubscriptionRepository();
+
+        $repository->unTrash($subscription);
+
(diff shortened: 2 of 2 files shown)


src/Subscriptions/resources/components/SubscriptionsListTable.tsx
```diff
@@ -71,6 +80,56 @@ const bulkActions: Array<BulkActionsConfig> = [
             </>
         ),
     },
+    {
+        label: __('Trash', 'give'),
+        value: 'trash',
+        type: 'warning',
+        isVisible: (data, parameters) => parameters.status !== 'trash',
+        action: async (selected) => {
+            const response = await API.fetchWithArgs('/setStatus', {ids: selected.join(','), status: 'trashed'}, 'POST');
```
jonwaldstein: @alaca trashing is supposed to send a `DELETE` request without the force option.  This should trigger the model `$subscription->trash()` method.  This is a little different than just a status update because it preserves the old status like how `wp_trash_post` and  `wp_untrash_post` works so we can restore the item back to it's original state.

glaubersilva: The subscription repository has the trash method that saves the previous status in a meta key, but there is no restore method. So, we probably will need to implement it as well in the repository to be able to restore the donation with the previous step saved on the meta key.

alaca: Resolved: https://github.com/impress-org/givewp/pull/8138/commits/5ab8d83c6e04bb1ee927a9af99d57e6b351a3e12

src/Subscriptions/Actions/LoadSubscriptionsListTableAssets.php
```diff
@@ -60,6 +63,31 @@ public function __invoke()
...
+            }
+        }
+
+        // Make active status default
+        return array_merge([
+            [
+                'value' => SubscriptionStatus::ACTIVE,
+                'text' => __('Active', 'give'),
+            ],
+        ], $statuses);
```
glaubersilva: I think the default option should display all subscriptions regardless of the status, so I would suggest replacing this logic with something like this:

```
$statuses = [];

foreach (SubscriptionStatus::labels() as $value => $label) {
    $statuses[] = [
        'value' => $value,
        'text' => $label,
    ];
}

return array_merge([
    [
        'value' => 'all',
        'text' => __('All Status', 'give'),
    ],
], $statuses);
```

src/Subscriptions/Endpoints/ListSubscriptions.php
```diff
@@ -111,6 +112,11 @@ public function registerRoute()
                             'columns',
                         ],
                     ],
+                    'status' => [
+                        'enum' => [...array_values(SubscriptionStatus::toArray())],
```
glaubersilva: Let's add support to "all" status this way:

`'enum' => ['all', ...array_values(SubscriptionStatus::toArray())],`

src/Subscriptions/Endpoints/ListSubscriptions.php
```diff
@@ -271,6 +279,8 @@ private function getWhereConditions(QueryBuilder $query): QueryBuilder
             $query->where('payment_mode', $testMode ? SubscriptionMode::TEST : SubscriptionMode::LIVE);
         }
 
+        $query->where('status', $status);
```
glaubersilva: Let's filter by status only if it's different from "all", like this:

```
if ($status === 'all') {
    $query->where('status', SubscriptionStatus::TRASHED, '!=');
} else {
    $query->where('status', $status);
}
```

src/Subscriptions/Actions/LoadSubscriptionsListTableAssets.php
alaca: Good point. Resolved https://github.com/impress-org/givewp/pull/8138/commits/d7828f18ac5261eb2486c6c929b578b2e8bb912c

src/Subscriptions/Actions/LoadSubscriptionsListTableAssets.php
```diff
@@ -60,6 +63,29 @@ public function __invoke()
...
+                'value' => $value,
+                'text' => $label,
+            ];
+        }
+
+        // Make active status default
+        return array_merge([
+            [
+                'value' => 'any',
+                'text' => __('Any', 'give'),
```
glaubersilva: @alaca I think we need to be clearer here and use the same text from campaigns: `All Status`

Right now, the dropdown loads with the `Any` option selected, but "any" what? The user needs to click in the dropdown to understand that this filter is to choose statuses.

image

image

jonwaldstein: I think we might have duplicate work going on here with what @pauloiankoski is working on https://github.com/impress-org/givewp/pull/8144

glaubersilva: @jonwaldstein I remember a conversation in a daily meeting last week about it, where we agreed to keep this dropdown at first and remove it later when the filter-by option gets ready. I think @pauloiankoski  suggested that, but correct me if I'm wrong.

src/Subscriptions/resources/components/SubscriptionsRowActions.tsx
```diff
@@ -18,27 +18,56 @@ export function SubscriptionsRowActions({item, setUpdateErrors, parameters}) {
...
+    const confirmRestoreModal = () => {
+        showConfirmModal(__('Restore', 'give'), confirmRestore, restoreItem, 'normal');
+    };
+
     return (
         <>
             <RowAction
                 href={`edit.php?post_type=give_forms&page=give-subscriptions&id=${item.id}`}
                 displayText={__('Edit', 'give')}
             />
```
glaubersilva: @alaca Let's keep the Edit option only for items not trashed. The default WP list table does not allow editing items in the trash status. 

image

src/Subscriptions/Actions/LoadSubscriptionsListTableAssets.php
pauloiankoski: Yeah, but I think the point here is that the Filter By menu is ready to be merged. I already removed the dropdown menu from the Donors List Table when I applied the Filter By menu there. So would make sense to remove the dropdown in this PR as it's going to be merged after the one implementing the Filter By menu.

src/Subscriptions/resources/components/SubscriptionsRowActions.tsx
pauloiankoski: One important thing to keep in mind is that the status param is now an array, so instead of checking for equality, we’ll need to check if `params?.status?.includes('trashed')`.

pauloiankoski:
@glaubersilva I just removed the _Status_ dropdown filter, which has been replaced by the new Filter By menu. Along with that, I made the following updates:
- Hid other status options from Bulk Actions when viewing trashed items. We were previously able to set a trashed item to active (let me know if that makes sense).
- Updated the _Edit_ inline action so it’s only visible for non-trashed items.

PR #8141, opened 2025-10-15
Feature/: Update subscriptions to reference the correct form id

alaca:
## Description

This pull request introduces a new migration to ensure that subscription records correctly reference the correct form ID after form data has been transferred from version 2 to version 3. This addresses a previous inconsistency where the `give_subscriptions.product_id` field was not updated to reflect changes in the donation meta table. 

## Testing Instructions

The simplest way to test this would be to manually change the product_id to a random value in the subscriptions table and re-run the migration. After re-running the migration, the product_id column should reference the correct form ID.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/tests/Unit/Subscriptions/Migrations/UpdateProductIDTest.php
+++ b/tests/Unit/Subscriptions/Migrations/UpdateProductIDTest.php
@@ -34,6 +34,6 @@ class UpdateProductIDTest extends TestCase
 
         $updatedSubscription = Subscription::find($subscription->id);
 
-        $this->assertEquals($subscription->donationFormId, $updatedSubscription->donationFormId);
+        $this->assertEquals(200, $updatedSubscription->donationFormId);
     }
 }


src/Subscriptions/Migrations/UpdateProductID.php
```diff
@@ -0,0 +1,79 @@
...
+                WHERE subscriptions.product_id != donationMeta2.meta_value
+            SQL;
+
+            DB::query(sprintf(
+                $query,
+                DB::prefix('give_subscriptions'),
+                DB::prefix('give_donationmeta'),
+                DonationMetaKeys::SUBSCRIPTION_ID,
+                DB::prefix('give_donationmeta'),
+                DonationMetaKeys::FORM_ID,
```
jonwaldstein: The tricky part here is that subscriptions have a one-to-many relationship with donations (renewals) so there can be _many_ donations with each `subscription_id`.  I think that will cause unnecessary joins.

Instead, I recommend only joining the initial donation for each subscription which can be achieved by using the `parent_payment_id`

In fact, i'm pretty sure we can grab the `form_id` from the revenue table instead of the meta table as it looks like the revenue table is updated [during the form migration](https://github.com/impress-org/givewp/blob/e11fe2cf3ebc9d16f355f7a5652e312a7d590bc4/src/FormMigration/Actions/TransferDonations.php#L41).

So in conclusion, let's join the revenue table using the `parent_payment_id` and grab the correct `form_id` 😄 

I would also like to see us get in the habit of writing unit tests for our migrations as these are pretty simple to create tests for by creating some test data, instatiating the migration class and calling `$migration->run()`

src/Subscriptions/ServiceProvider.php
```diff
@@ -83,7 +84,8 @@ private function registerMigrations()
             AddPaymentModeToSubscriptionTable::class,
             BackfillMissingCampaignIdForDonations::class,
             AddCampaignIdColumn::class,
-            AddCampaignId::class
+            AddCampaignId::class,
+            UpdateProductID::class
```
jonwaldstein, line 88: I wonder if we should move this migration (and timestamp) to go before the campaignId migrations as the `AddCampaignId` relies on the formId to find the campaign.

alaca: Resolved: https://github.com/impress-org/givewp/pull/8141/commits/595ab546dbcc0d07169f5b97cd852a2a7418b2f6

src/Subscriptions/Migrations/UpdateProductID.php
alaca: Resolved: https://github.com/impress-org/givewp/pull/8141/commits/1f077cb04497fd7f51eaedcc0a8e7dc40860a4a2

tests/Unit/Subscriptions/Migrations/UpdateProductIDTest.php
```diff
@@ -0,0 +1,39 @@
...
+            'donationFormId' => 100,
+        ], [
+            'formId' => 200,
+        ]);
+
+        (new UpdateProductID())->run();
+
+        $updatedSubscription = Subscription::find($subscription->id);
+
+        $this->assertEquals($subscription->donationFormId, $updatedSubscription->donationFormId);
```
jonwaldstein: @alaca shouldn't this be asserting that the updated subscription now equals the initial donation formID?
```suggestion
        $this->assertEquals(200, $updatedSubscription->donationFormId);
```

alaca: Thanks, man. Resolved https://github.com/impress-org/givewp/pull/8141/commits/cf10bc042a3f9d62f6253d05bc5d903ec14371d7

PR #8145, opened 2025-10-15
Fix: prevent override of custom translation files

glaubersilva:
Resolves [GIVE-2947]

## Description

This PR fixes a problem that was preventing V2 forms from using the translation files stored in a custom folder. 

Before, only translation files stored in the `/wp-content/languages/plugins/` and `/wp-content/plugins/give/languages/` folders were working. 

Now, it is also possible to use translated files stored in a custom folder, like the `/wp-content/languages/loco/plugins/` that is usually used to store custom translations created through the Loco Translate Plugin. 

## Affects

Translation files from custom folders.

## Visuals

image

image

## Testing Instructions

1. Create a V2 form with the multi-step template.
2. Set up a new language for the website.
3. Go to `Loco Translate > Plugins > Give - Donation Plugin` and select the target language that you set up before
4. Click in the `Relocate` tab, and move the translation files to the custom folder.
5. Access the V2 form created in the first step and make sure the `Secure Donation` string in the footer is displayed properly translated in the target language.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2947]: https://stellarwp.atlassian.net/browse/GIVE-2947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Helpers/Language.php
+++ b/src/Helpers/Language.php
@@ -2,27 +2,24 @@
 
 namespace Give\Helpers;
 
-use WP_Translation_Controller;
-use WP_Translations;
-
 /**
  * @since 3.0.0
  */
 class Language
 {
     /**
-     * @unreleased Added early return if the textdomain is loaded from a custom folder
+     * @unreleased Added early return if the textdomain is loaded
      * @since 3.0.0
      */
     public static function load()
     {
-        if (self::isTextdomainLoadedFromCustomFolder('give')) {
+        if (is_textdomain_loaded('give')) {
             return;
         }
 
         $giveRelativePath = self::getRelativePath();
 
-        $locale = is_admin() && ! wp_doing_ajax() && function_exists('get_user_locale') ? get_user_locale() : get_locale();
+        $locale = is_admin() && !wp_doing_ajax() && function_exists('get_user_locale') ? get_user_locale() : get_locale();
         $locale = apply_filters('plugin_locale', $locale, 'give'); // Traditional WordPress plugin locale filter.
 
         // Setup paths to current locale file.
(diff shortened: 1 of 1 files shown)


src/Helpers/Language.php
```diff
@@ -72,4 +80,68 @@ public static function switchToLocale(string $locale)
...
+    public static function isTextdomainLoadedFromCustomFolder(string $textdomain): bool
+    {
+        global $l10n;
+        if (!isset($l10n[$textdomain]) || !$l10n[$textdomain] instanceof WP_Translations) {
+            return false;
+        }
+
+        // Try to access the controller through the global instance
+        $controller = WP_Translation_Controller::get_instance();
+        if (!$controller) {
```
alaca: `WP_Translation_Controller::get_instance();` will always return a `WP_Translation_Controller` instance. This check is not required.

src/Helpers/Language.php
```diff
@@ -72,4 +80,68 @@ public static function switchToLocale(string $locale)
...
+        $loadedTranslationsProperty = $reflection->getProperty('loaded_translations');
+        $loadedTranslationsProperty->setAccessible(true);
+        $loadedTranslations = $loadedTranslationsProperty->getValue($controller);
+        
+        $currentLocale = $controller->get_locale();
+        if (!isset($loadedTranslations[$currentLocale][$textdomain])) {
+            return false;
+        }
+
+        $files = $loadedTranslations[$currentLocale][$textdomain];
```
alaca: Since you are already using the reflection class, why don't you just call the `get_files()` method to get the files?

If you ask me, I wouldn't use the reflection class at all; all these methods are protected, which means you can extend the class and use these methods without all this.

glaubersilva: Yeah, I was overcomplicating things with the reflection class. We can keep it much simpler using the native [is_textdomain_loaded](https://developer.wordpress.org/reference/functions/is_textdomain_loaded/) core function, which will check if the textdomain is loaded regardless of the location. 

Initially, I thought that the earlier return in the `load()` method would be useful only for cases where the translation is located in a custom folder. However, WordPress 6.7+  implements [Just-in-Time loading](https://developer.wordpress.org/reference/functions/_load_textdomain_just_in_time/) for translation files, so the erlir return can be useful for cases where the WP loads it automatically even in the default folders.

Resolved in 2106a61490df58f33c3cd8ffb6759fa74f5d2ffb

src/Helpers/Language.php
glaubersilva: Resolved in 2106a61490df58f33c3cd8ffb6759fa74f5d2ffb

glaubersilva:
@alaca This is ready for re-review.

alaca (approved):
I like this solution better, thanks man!

PR #8149, opened 2025-10-17
Feature: Add soft delete (trash/restore) functionality to Donations

pauloiankoski:
Resolves [GIVE-2961]

## Description

This PR implements soft delete functionality for donations in the Donations List Table. Previously, donations could only be permanently deleted, which could result in accidental data loss. This update adds the ability to move donations to trash, restore them from trash, and permanently delete them when needed.

The implementation adds:
- New `trash` and `untrash` actions to the Donation Actions REST API endpoint
- Support for soft delete in the Donation model and repository layer
- Updated UI in the Donations List Table with trash and restore row actions
- A `force` parameter to the delete action to distinguish between soft delete (trash) and permanent deletion

## Affects

- **Donations List Table**: New row actions for trash and restore
- **Donation Actions Endpoint**: New endpoints for trash/untrash operations
- **Donation Repository**: Added methods to support soft delete operations
- **Donation Model**: Updated to track trash status
- Any integrations or add-ons that interact with donation deletion will need to be aware of the soft delete behavior

## Visuals

https://github.com/user-attachments/assets/43015c57-c041-4397-b7e1-35e66f8c771f

## Testing Instructions

1. Navigate to **GiveWP > Donations**
2. Check the items show “Edit” and “Trash” as inline actions on hover
3. Confirm those actions work as expected
4. Check the Bulk Actions menu, it may show many options except Restore and Delete
5. Confirm those bulk actions work as expected, specially the Trash option
6. Using the Filter By menu, select “Show trashed” and apply
7. In the filtered list, check the items now show “Restore” and "Delete” as inline actions on hover
8. Confirm those actions work as expected
9. Check the Bulk Actions menu, it may only show Restore and Delete options
10. Confirm those bulk actions work as expected

## Pre-review Checklist

- [x] Includes unit tests

[GIVE-2961]: https://stellarwp.atlassian.net/browse/GIVE-2961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Donations/Endpoints/DonationActions.php
+++ b/src/Donations/Endpoints/DonationActions.php
@@ -111,7 +111,7 @@ class DonationActions extends Endpoint
 
         switch ($request->get_param('action')) {
             case 'delete':
-                if (! current_user_can('delete_give_payments')) {
+                if (!current_user_can('delete_give_payments')) {
                     return new WP_Error(
                         'rest_forbidden',
                         esc_html__('You don\'t have permission to delete Donation', 'give'),
@@ -122,7 +122,7 @@ class DonationActions extends Endpoint
                 foreach ($ids as $id) {
                     $donation = Donation::find($id);
 
-                    if (! $donation) {
+                    if (!$donation) {
                         $errors[] = $id;
                         continue;
                     }
@@ -141,7 +141,7 @@ class DonationActions extends Endpoint
                 foreach ($ids as $id) {
                     $donation = Donation::find($id);
 
-                    if (! $donation) {
+                    if (!$donation) {
                         $errors[] = $id;
                         continue;
                     }
@@ -160,7 +160,7 @@ class DonationActions extends Endpoint
                 foreach ($ids as $id) {
                     $donation = Donation::find($id);
 
-                    if (! $donation) {
+                    if (!$donation) {
                         $errors[] = $id;
                         continue;
                     }


src/Donations/Endpoints/DonationActions.php
```diff
@@ -113,35 +117,62 @@ public function handleRequest(WP_REST_Request $request)
 
         switch ($request->get_param('action')) {
             case 'delete':
-                if (!current_user_can('delete_give_payments')) {
+                if ( ! current_user_can('delete_give_payments')) {
```
jonwaldstein: Can you check the formatting? It looks like it's adding unnecessary spaces like between the negate operator

src/Donations/Repositories/DonationRepository.php
```diff
@@ -390,6 +390,38 @@ public function trash(Donation $donation): bool
         return true;
     }
 
+    /**
+     * @unreleased
+     *
+     * @throws Exception
+     */
+    public function unTrash(Donation $donation): bool
```
jonwaldstein, line 398: We should add a unit test for this 😄

src/Donations/resources/components/DonationRowActions.tsx
```diff
@@ -4,43 +4,79 @@ import {__, sprintf} from '@wordpress/i18n';
...
 import { store as coreDataStore } from '@wordpress/core-data';
 import { useDispatch } from '@wordpress/data';
+import { useSWRConfig } from 'swr';
+import ListTableApi from '@givewp/components/ListTable/api';
+
+const donationsApi = new ListTableApi(window.GiveDonations);
 
 /**
- * @since 4.6.0 Soft delete donations with Donation v3 API.
+ * @.
```
jonwaldstein: Looks like something weird going on with the docblock here

src/Donations/Endpoints/DonationActions.php
```diff
@@ -18,7 +18,7 @@ class DonationActions extends Endpoint
     /**
      * @inheritDoc
      *
-     * @since 4.10.0 Add force parameter to delete action
```
jonwaldstein, line 21: We should be preserving since tags that were released

src/Donations/Endpoints/DonationActions.php
```diff
@@ -37,15 +37,12 @@ public function registerRoute()
...
+                            'untrash',
                             'setStatus',
                             'resendEmailReceipt',
                         ],
                     ],
-                    'force' => [
-                        'type' => 'boolean',
-                        'default' => false,
-                        'description' => 'Whether to permanently delete (force=true) or move to trash (force=false, default).',
-                    ],
```
jonwaldstein, line 48: Just want to make sure this is intentional? In our REST api we decided to go with a delete request that handles the trash when force is not set.  I'm not as worried about that pattern here with our list tables as they are pretty bespoke but something to consider for consistency.

src/Views/Components/ListTable/FilterBy/hooks/useFilterState.ts
```diff
@@ -40,17 +40,22 @@ export function useFilterState(
...
+                        newFilters[group.apiParam] = filteredValues;
+                        hasChanges = true;
+                    }
                 });
 
-                return newFilters;
+                return hasChanges ? newFilters : prev;
             });
         }
     }, [visibleGroups, groupedOptions]);
```
jonwaldstein: Looking at this file a little bit closer i'm seeing opportunities to improve the efficiency as I think that infinite loop is a good sign of re-rending too many times.  Although you're short-circuiting - it's still being re-rendered when it probably shouldn't.

To start, I suggest removing anything from the deps that should not change like I don't think `groupedOptions` is supposed to change but it's presence in the deps as an array of objects will cause unnecessary re-renders (this happens when comparing objects as they are always false).

It might even be possible to remove some of the useEffects and do things more synchronously.

You can test some of this out by dropping some console logs in the useEffects & enable the re-render watcher in react debug console tools

I'm happy to jump on a call to go through this together or if you want to 😄

src/Donations/Endpoints/DonationActions.php
pauloiankoski: Yes, it is. This follows what's been done in subscriptions and that `force` param is a leftover once using the new approach with trash/untrash.

src/Donations/Endpoints/DonationActions.php
pauloiankoski: That has been copied from the Subscriptions PR. Fixed on 14658cd13f473fdaa013b3d6f676b990e253ae8e but also fixed there.

src/Donations/resources/components/DonationRowActions.tsx
pauloiankoski: Reverted on 6c9a24f5785ea09765c9de6ef27e5efc3828862b

src/Donations/Endpoints/DonationActions.php
pauloiankoski: Reverted on 6c9a24f5785ea09765c9de6ef27e5efc3828862b

src/Views/Components/ListTable/FilterBy/hooks/useFilterState.ts
pauloiankoski: Sure thing. I did a first pass on 3f78d1cf64a2cfb3f4146c47de1cdf86ecf70e44 tho.

src/Donations/Repositories/DonationRepository.php
pauloiankoski: Added on bc50fd600d9d5fc2bb99105b4f7fe0f9ad48b344

src/Donations/Endpoints/DonationActions.php
jonwaldstein: Thanks, there is still an extra space between the statement though - not a big deal but we should figure out how to make our formatting more consistent

src/Donations/Endpoints/DonationActions.php
jonwaldstein: Okay, I guess i'm not understanding why we would change the API for this?  I think we can keep this closer to our v3 API where a DELETE request will trash the item unless force is specified which permanently deletes the item.  Then we could add a `donations/id/restore` endpoint.

src/Views/Components/ListTable/FilterBy/hooks/useFilterState.ts
jonwaldstein: That's looking a lot better, thanks!

src/Donations/Endpoints/DonationActions.php
jonwaldstein: @pauloiankoski Looking closer at this REST API it does not even have a dedicated DELETE endpoint as it's a shared POST and DELETE request that uses an action param to decide how to handle it.  The proper way would be to separate these requests out.  However, I realize we are just modifying the existing old routes to make them work for trashing.

So at this time, what you have is fine 😏 - we just would _not_ do this in our v3 endpoints

jonwaldstein (approved):
@pauloiankoski thanks man!

rickalday (approved):
Passed QA tests.

PR #8164, opened 2025-11-05
Refactor: update campaign REST API endpoints to follow standards

jonwaldstein:
Resolves [GIVE-2910]

## Description

This updates the Campaign endpoints to follow the standards and conventions of our other v3 REST endpoints: 
- Extend the`WP_REST_Controller`
- Updated tests to use the REST test case
- Moved CampaignViewModel into REST domain
- Moved CampaignController into REST domain and removed dependency on the `requestController`
- Updated tests and test file names

Screenshot 2025-11-07 at 1 59 40 PM

## Affects

This touches all the campaign endpoints including campaign collections, campaign item, campaign revenue, campaign statistics, and campaign comments

## Visuals

N/A

## Testing Instructions

ZIP: https://github.com/impress-org/givewp/actions/runs/19240151534

Make sure all CRUD operations are working properly for campaigns including:
- Creating, updating, viewing, merging, and duplicating campaigns

## Pre-review Checklist

-   [x] Includes unit tests

[GIVE-2910]: https://stellarwp.atlassian.net/browse/GIVE-2910?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/API/REST/V3/Routes/Campaigns/CampaignController.php
+++ b/src/API/REST/V3/Routes/Campaigns/CampaignController.php
@@ -501,8 +501,10 @@ class CampaignController extends WP_REST_Controller
 
     /**
      * @unreleased
+     *
+     * @return WP_REST_Response|\WP_Error
      */
-    public function duplicate_item($request): WP_REST_Response
+    public function duplicate_item($request)
     {
         $campaign = Campaign::find((int)$request->get_param('id'));
 


src/API/REST/V3/Routes/Campaigns/CampaignController.php
```diff
@@ -5,308 +5,301 @@
...
-        $this->registerGetCampaign();
-        $this->registerUpdateCampaign();
-        $this->registerGetCampaigns();
-        $this->registerMergeCampaigns();
-        $this->registerCreateCampaign();
-        $this->registerCreateCampaignPage();
-        $this->registerDuplicateCampaign();
+        $this->namespace = CampaignRoute::NAMESPACE;
+        $this->rest_base = CampaignRoute::CAMPAIGNS;
+        $this->requestController = new CampaignRequestController();
```
glaubersilva: I don't think we need this `CampaignRequestController` class. 

In fact, using that is preventing us from using some default methods from `WP_REST_Controller`, like `prepare_item_for_response` and `prepare_response_for_collection`, in the response to the requests.

I would prefer to keep the same pattern used here: 

https://github.com/impress-org/givewp/blob/c7b7fab9461ee1e23bc8b4e0cd4735909d7e7109/src/API/REST/V3/Routes/Subscriptions/SubscriptionController.php#L179-L234

src/API/REST/V3/Routes/Campaigns/CampaignController.php
```diff
@@ -5,308 +5,301 @@
...
-                    },
-                    'permission_callback' => function () {
-                        return current_user_can('manage_options');
-                    },
-                ],
-                'args' => rest_get_endpoint_args_for_schema($this->getSchema(), WP_REST_Server::EDITABLE),
-                'schema' => [$this, 'getSchema'],
-            ]
-        );
+        return $this->requestController->getCampaigns($request);
```
glaubersilva: I think it would be better to implement the getCampaign logic here directly using the $this->prepare_item_for_response($item, $request); method, so we can add some embed links there.

Here is a reference: https://github.com/impress-org/givewp/blob/c7b7fab9461ee1e23bc8b4e0cd4735909d7e7109/src/API/REST/V3/Routes/Subscriptions/SubscriptionController.php#L247-L266

We can do the same (migrate the logic to this controller) for all methods from the requestController: `getCampaign`, `getCampaigns`, `updateCampaign`, `mergeCampaigns`, `createCampaign`, `createCampaignPage`, `duplicateCampaign`

If we need to encapsulate some big logic to query data, I think it would be better to have a query class instead of a new controller class, like this one: https://github.com/impress-org/givewp/blob/c7b7fab9461ee1e23bc8b4e0cd4735909d7e7109/src/Subscriptions/SubscriptionQuery.php#L14

Also, we can encapsulate other operations inside other classes to make this controller shorter, like in this case where the logic to create a new item is outside the `create_item()` method: https://github.com/impress-org/givewp/blob/c7b7fab9461ee1e23bc8b4e0cd4735909d7e7109/src/API/REST/V3/Routes/Subscriptions/SubscriptionController.php#L279-L310

src/API/REST/V3/Routes/Campaigns/CampaignController.php
```diff
@@ -5,308 +5,301 @@
...
+                'type' => 'string',
+                'enum' => ['asc', 'desc'],
+                'default' => 'desc',
+            ],
+            'search' => [
+                'type' => 'string',
+                'default' => '',
+            ],
+        ];
     }
```
glaubersilva, line 578: We can create an action class called `GetCampaignCollectionParams` to make it easy to read and also use the `parent::get_collection_params()` method, like here: https://github.com/impress-org/givewp/blob/c7b7fab9461ee1e23bc8b4e0cd4735909d7e7109/src/API/REST/V3/Routes/Subscriptions/SubscriptionController.php#L524-L538

src/API/REST/V3/Routes/Campaigns/CampaignController.php
```diff
@@ -5,308 +5,301 @@
...
 
     /**
-     * @since 4.13.0 add schema description
-     * @since 4.9.0 Set proper JSON Schema version
-     * @since 4.0.0
+     * @unreleased
      */
-    public function getSchema(): array
+    public function get_item_schema(): array
```
glaubersilva, line 583: We can create an action class called `GetCampaignItemSchema` to make it easy to read and also to use the `add_additional_fields_schema()` method, which allows us to extend the schema in the add-ons, like here:

https://github.com/impress-org/givewp/blob/c7b7fab9461ee1e23bc8b4e0cd4735909d7e7109/src/API/REST/V3/Routes/Subscriptions/SubscriptionController.php#L719-L723

glaubersilva (changes requested):
@jonwaldstein I did some suggestions for the `CampaignController` and some of them would be the same for the other controllers: `CampaignCommentsController`, `CampaignRevenueController`, `CampaignStatisticsController`.

Suggestions: 

- Implement the `prepare_item_for_response()` method with the `self` and `embeddable` links.

- Extract the logic of `get_collection_params()` and `get_item_schema()` methods to action classes to make the controllers easy to read.

- Use query classes and other ones to extract logic when necessary, and take advantage of the support classes that we already have, like these ones: `CURIE`, `Headers`, `Item`, etc. Reference: https://github.com/impress-org/givewp/tree/c7b7fab9461ee1e23bc8b4e0cd4735909d7e7109/src/API/REST/V3/Support

Also, I would say that the `SubscriptionController` is the best sample to use as a reference.

src/API/REST/V3/Routes/Campaigns/CampaignController.php
jonwaldstein: Yo @glaubersilva - I agree the request controller was redundant since we're using the WP controller now.  So, I updated the `CampaignController` to not rely on that anymore and updated all the responses.  

In the future I want to add more relationships for embedding but for now it just has the self link to start.

src/API/REST/V3/Routes/Campaigns/CampaignController.php
jonwaldstein: Yup good call, I updated the `CampaignController` to follow this closer.  In the future we should add a query class but didn't want to overcomplicate this PR lol

src/API/REST/V3/Routes/Campaigns/CampaignController.php
jonwaldstein: This one I didn't feel was necessary since it's only used in one place and is already defining the params

src/API/REST/V3/Routes/Campaigns/CampaignController.php
jonwaldstein: Good call on the additional fields - I added that but didn't feel it was necessary to extract the schema as its only used in one place

jonwaldstein:
@glaubersilva thanks for the suggestions! This is ready for review again 🪇 

BTW I realized there are some endpoints like `campaign-revenue`, `campaign-statistics`, and `campaign-comments` that have no need for preparing items for response and adding links because they are not returning proper resources or subresources - they are essentially just returning aggregate data.  If comments were a proper subresource with its own schema then that would make more sense.

The one exception I found was our `campaign-page` endpoint which I moved into it's own controller because that's an example of an actual subresource.

PR #8166, opened 2025-11-07
Refactor: reset feature flag counter

jonwaldstein:
Resolves [GIVE-2969]

## Description

We originally added a beta features page for Events beta, and a little notification icon to draw attention to the page.  It's currently hard-coded to reset the value to 1 on initial page load - and then reset to 0 when visiting the beta features page.

However, we don't have a plan to move events beta out of beta and this notification has been active for a long time - so it's time to hide that notification until we revisit our strategy.

## Affects

The feature flag / notifications counter

## Visuals

N/A

## Testing Instructions

ZIPL: https://github.com/impress-org/givewp/actions/runs/19240118173

- Install this PR on a fresh WP instance and you should not see a notification.
- Install this on a site with GiveWP that has not dismissed the notification and you should not see a notification.

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-2969]: https://stellarwp.atlassian.net/browse/GIVE-2969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/BetaFeatures/ServiceProvider.php
+++ b/src/BetaFeatures/ServiceProvider.php
@@ -64,7 +64,6 @@ class ServiceProvider implements ServiceProviderInterface
     /**
      * This is intended to add the initial notification count
      *
-     * @unreleased added $count parameter to set the initial notification count
      * @since 3.6.0
      */
     private function setFeatureFlagNotificationCounter(): void


src/BetaFeatures/ServiceProvider.php
```diff
@@ -54,12 +57,19 @@ public function boot(): void
     }
 
     /**
+     * This is intended to add the initial notification count
+     *
+     * @unreleased added $count parameter to set the initial notification count
      * @since 3.6.0
      */
-    private function setFeatureFlagNotificationCounter(): void
+    private function setFeatureFlagNotificationCounter(int $count = 1): void
```
pauloiankoski: @jonwaldstein This method is used to initialize the counter. However, it’s not the method that runs when the badge is shown in the menu. For existing users (that have never visited the Beta page), that value should already be set to `1` at this point, so the badge stays visible. We could add an early return in `FeatureFlagRepository::getNotificationCount` to always return `0`, and stop calling `setFeatureFlagNotificationCounter` in `boot`.

jonwaldstein: @pauloiankoski good thoughts, I actually was thinking this update would _only_ be for new users.  I could be persuaded either way lol

pauloiankoski: Oh okay. No problem then.

pauloiankoski (approved):

src/BetaFeatures/ServiceProvider.php
jonwaldstein: @pauloiankoski I updated this PR with a filter so we can return a count of 0 for now

PR #8170, opened 2025-11-12
Fix: Resolve GiftAid and Company field conflict

pauloiankoski:
Resolves [GIVE-2764]

## Description

This PR fixes a validation conflict between the `GiftAid` and `Company` fields. The issue occurred because `GiftAid` has conditional visibility based on the `Company` field (if `Company` is required or not empty, `GiftAid` must not be shown). When a field has conditional visibility rules and the referenced field (`Company`) is present in the form, an `ExcludeUnless` property is automatically added to that field's validation rules.

In `ConvertFieldAPIRulesToJoi.ts`, fields with no rules receive an "any" type. However, because of the `ExcludeUnless` property, the `GiftAid` field was being processed through the rest of the validation function and was incorrectly validated as a string. When the opted-out value was passed (which was not an empty string), it failed the required validation.

To fix this, we've updated the validation logic to check if a field has only the `excludeUnless` property (and no other validation rules). In such cases, the field now returns `Joi.any()` immediately, preventing it from being incorrectly processed through the string validation logic. This ensures that fields like `GiftAid` (which are object/array types) are not validated as strings when they only have conditional visibility rules.

## Affects

- Frontend validation: Updated `ConvertFieldAPIRulesToJoi.ts` to handle fields with only `excludeUnless` rules properly
- GiftAid add-on compatibility: This change enables proper validation for object-type fields like GiftAid that have conditional visibility

## Visuals
No visual changes - this is a backend validation fix.

## Testing Instructions
   - Create a donation form with both Company and GiftAid fields
   - Fill out the donation form without checking the GiftAid checkbox
   - Submit the form - it should submit successfully without validation errors
   - Reload the donation form
   - Verify that when GiftAid is checked, the field data is properly validated on submission

## Pre-review Checklist

- [ ] Includes unit tests

[GIVE-2764]: https://stellarwp.atlassian.net/browse/GIVE-2764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DonationForms/resources/app/utilities/ConvertFieldAPIRulesToJoi.ts
+++ b/src/DonationForms/resources/app/utilities/ConvertFieldAPIRulesToJoi.ts
@@ -49,7 +49,7 @@ function getJoiRulesForField(field: Field): AnySchema {
 }
 
 /**
- * @unreleased add support for object rules
+ * @unreleased account for custom rules with only excludeUnless property
  * @since 4.13.0 add support for optional false values
  * @since 3.0.0
  */


src/DonationForms/Rules/ObjectRule.php
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * @unreleased
+     */
+    public function __invoke($value, Closure $fail, string $key, array $values)
+    {
+        // Although the field is validated as an object in the frontend, it is passed as an array to the backend.
+        if (!empty($value) && !is_array($value)) {
+            $fail(sprintf(__('%s must be an object', 'give'), '{field}'));
+        }
```
jonwaldstein: I'm not sure this is an actual object rule as it does not validate the value as an object - instead just checking if its not an array

jonwaldstein (approved):
Nice job!

PR #8180, opened 2025-11-20
Fix: Update Campaign blocks to use donor profile avatar

pauloiankoski:
Resolves [GIVE-2971]

## Description

This PR fixes an issue where the Campaign Donations and Campaign Donors blocks were attempting to display donor avatars based on their WordPress User account. However, not all donors have associated WordPress User accounts, which caused avatar display issues.

The solution updates both blocks to:
- Query the `give_donormeta` table to retrieve the `avatarId` meta value directly from the donor profile
- Use `wp_get_attachment_image_url()` to display the avatar when an `avatarId` is found
- Fall back to email-based avatars (via `get_avatar_url()`) when no custom avatar is set
- Maintain the existing anonymous donor handling

This ensures that all donors, regardless of whether they have a WordPress User account, can have their custom avatars displayed correctly in the Campaign blocks.

## Affects

- Campaign Donations block: Donor avatar display in donation listings
- Campaign Donors block: Donor avatar display in donor listings
- Both blocks now query the `give_donormeta` table for avatar information

## Visuals
CleanShot 2025-11-20 at 12 08 29

## Testing Instructions

1. Create a campaign with the Campaign Donations and/or Campaign Donors blocks
2. Make donations from donors with custom avatars set in their donor profiles (stored in donormeta)
3. Make donations from donors without custom avatars set in their donor profiles
4. Make donations from donors without WordPress User accounts
5. Verify that avatars display correctly for all donors:
   - Donors with custom avatars should show their custom avatar
   - Donors without custom avatars should show email-based avatars or default WP picture
   - Anonymous donations should show the default anonymous avatar

## Pre-review Checklist

- [ ] Includes unit tests

[GIVE-2971]: https://stellarwp.atlassian.net/browse/GIVE-2971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Campaigns/Blocks/CampaignDonations/resources/views/render.php
```diff
@@ -51,7 +51,7 @@
 
             <div class="givewp-campaign-donations-block__empty-icon">
                 <?php
-                echo file_get_contents(dirname(__DIR__) . '/icons/empty-state.svg'); ?>
+                echo wp_kses_post(wp_remote_get(dirname(__DIR__) . '/icons/empty-state.svg')); ?>
```
alaca: Hey man, why do we need `wp_remote_get` to get that SVG? We could use an `img` tag with src pointing to SVG file.

src/Campaigns/Blocks/CampaignDonations/resources/views/render.php
```diff
@@ -110,7 +112,7 @@
                         <div class="givewp-campaign-donations-block__donation-ribbon" data-position="<?php
                         echo esc_attr($key + 1); ?>">
                             <?php
-                            echo file_get_contents(dirname(__DIR__) . '/icons/ribbon.svg'); ?>
+                            echo wp_kses_post(wp_remote_get(dirname(__DIR__) . '/icons/ribbon.svg')); ?>
```
alaca: Same here. I'm not a fan of doing this if there is no specific reason. Also, why do we need to sanitize our SVG files?

pauloiankoski: Replaced with an `` tag.

src/Campaigns/Blocks/CampaignDonations/resources/views/render.php
pauloiankoski: Replaced with an `` tag.

alaca (approved):
Looks good, Paulo 👍 
Please resolve merge conflicts before merging.

PR #8194, opened 2026-01-19
Refactor: Conditionally load frontend assets only on pages with Give content

pauloiankoski:
Resolves [GIVE-2427]

## Description

This PR optimizes frontend performance by implementing conditional loading of Give scripts, styles, and REST API entities. Previously, all Give frontend assets were loaded on every page of the site, regardless of whether Give content was present. This caused unnecessary HTTP requests and increased page load times on pages without any Give forms, blocks, or shortcodes.

The solution introduces a new helper class `Give\Helpers\Frontend\Page` that intelligently detects if the current page contains Give content by checking for:
- Give form post types
- Give success/campaign pages  
- Give shortcodes (donation forms, donor dashboard, etc.)
- Give blocks (donation forms, campaigns, donor wall, etc.)
- Give embed contexts (iframes, routes)

This helper is then used to conditionally enqueue assets only when needed, reducing the frontend footprint on non-Give pages while maintaining full functionality on pages that use Give.

## Affects

- **Frontend Performance**: Pages without Give content will no longer load Give scripts/styles
- **Script Loading**: `give.js`, `give-styles.css`, Stripe scripts, and design system styles now load conditionally
- **REST API Entities**: Public entities registration is now conditional
- **Compatibility**: Maintains backward compatibility with all existing Give features

## Visuals

N/A - This is a performance optimization with no visual changes to the user interface.

### Test 1: Verify assets load on Give pages
1. Create a page with a Give donation form (shortcode or block)
2. View the page source
3. Confirm Give scripts and styles are loaded (search for `give.js`, `give-styles.css`)

### Test 2: Verify assets don't load on non-Give pages
1. Create a regular page without any Give content
2. View the page source
3. Confirm Give scripts and styles are NOT loaded

### Test 3: Test all Give content types
Test that assets load correctly for:
- [ ] Give form shortcode `[give_form id="123"]`
- [ ] Give donation form block
- [ ] Give donor dashboard shortcode/block
- [ ] Give donor wall
- [ ] Campaign pages
- [ ] Campaign blocks
- [ ] Give success page
- [ ] Give form single post type
- [ ] Give embed/iframe contexts

### Test 4: Test Stripe gateway
1. Create a donation form with Stripe gateway enabled
2. Verify Stripe scripts load correctly on the form page
3. Complete a test donation to ensure Stripe functionality works

## Pre-review Checklist

- [ ] Includes unit tests
- [ ] Reviewed by the designer (if follows a design) - N/A

[GIVE-2427]: https://stellarwp.atlassian.net/browse/GIVE-2427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/includes/class-give-scripts.php
+++ b/includes/class-give-scripts.php
@@ -63,7 +63,6 @@ class Give_Scripts {
 			add_action( 'admin_enqueue_scripts', [ $this, 'admin_enqueue_styles' ] );
 			add_action( 'enqueue_block_editor_assets', [ $this, 'gutenberg_admin_scripts' ] );
 			add_action( 'admin_head', [ $this, 'global_admin_head' ] );
-
 		} else {
 			add_action( 'wp_enqueue_scripts', [ $this, 'public_enqueue_styles' ] );
 			add_action( 'wp_enqueue_scripts', [ $this, 'public_enqueue_scripts' ] );
@@ -88,9 +87,13 @@ class Give_Scripts {
 	/**
 	 * Registers all plugin styles.
 	 *
+     * @unreleased Prevent loading styles on non-Give pages.
 	 * @since 2.1.0
 	 */
 	public function register_styles() {
+        if (!Give\Helpers\Frontend\Page::hasGiveContent()) {
+            return;
+        }
 
 		// Global WP-admin.
 		wp_register_style( 'give-admin-global-styles', GIVE_PLUGIN_URL . 'build/assets/dist/css/admin-global' . $this->direction . '.css', [], GIVE_VERSION );
@@ -115,9 +118,14 @@ class Give_Scripts {
 	/**
 	 * Registers all plugin scripts.
 	 *
+     * @unreleased Prevent loading scripts on non-Give pages.
 	 * @since 2.1.0
 	 */
 	public function register_scripts() {
+        if (!Give\Helpers\Frontend\Page::hasGiveContent()) {
+            return;
+        }
+
         // WP-Admin.
         EnqueueScript::make(
             'give-admin-scripts',
@@ -454,9 +462,14 @@ class Give_Scripts {
 	/**
 	 * Enqueues public styles.
 	 *
+     * @unreleased Prevent loading styles on non-Give pages.
 	 * @since 2.1.0
 	 */
 	public function public_enqueue_styles() {
+        if (!Give\Helpers\Frontend\Page::hasGiveContent()) {
+            return;
+        }
+
 		wp_enqueue_style( 'give-styles' );
 	}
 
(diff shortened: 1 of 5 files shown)


includes/class-give-scripts.php
```diff
@@ -115,9 +118,14 @@ public function register_styles() {
 	/**
 	 * Registers all plugin scripts.
 	 *
+     * @unreleased Prevent loading scripts on non-Give pages.
 	 * @since 2.1.0
 	 */
 	public function register_scripts() {
+        if (!Give\Helpers\Frontend\Page::hasGiveContent()) {
+            return;
+        }
```
jonwaldstein, line 127: It looks like this method also registers admin scripts which I think would cause issues.  We might consider moving the conditional to the actual enqueue part (not register) or actually refactoring to only register/enqueue admin scripts/styles through the admin_enqueue_scripts action to let WP do its thang and not have to use a conditional.

src/Helpers/Frontend/Page.php
```diff
@@ -0,0 +1,184 @@
...
+    {
+        global $post;
+
+        if (self::isGiveEmbed()) {
+            return true;
+        }
+
+        if (!is_single() && !is_page()) {
+            return false;
+        }
```
jonwaldstein, line 35: I think this is going to cause problems - we don't want to do a general exclusion of generic post type templates as it wouldn't account for archives, search, page builders, pages, FSE etc.

src/Helpers/Frontend/Page.php
```diff
@@ -0,0 +1,184 @@
...
+        // Allow filtering for custom conditions
+        return apply_filters('give_page_has_give_content', false);
+    }
+
+    /**
+     * Check if content has any Give shortcode
+     *
+     * @unreleased
+     */
+    public static function hasGiveShortcode(string $content): bool
```
jonwaldstein, line 77: These hardcoded lists of shortcodes and blocks make me nervous because it looks like we would have to mantain the list - not accounting for add-ons.  If the purpose is just to make sure something from give is present in the post content - perhaps we could do a regex check for the majority of use cases like give & givewp like this:

```php
// Match:
// - Shortcodes: [give_*], [givewp_*], [donation_history*]
// - Blocks: <!-- wp:give/* -->, <!-- wp:givewp/* -->
return (bool) preg_match(
    '/\[(give_|givewp_|donation_history)[^\]]*\]|<!-- wp:(give|givewp)\//',
    $content
);
```

jonwaldstein (changes requested):
@pauloiankoski this is ambitious 😅 

I noticed a few things were broken and left some suggestions - please make sure to test as many different scenarios as you can locally first 🙂 

Image

Image

PR #8223, opened 2026-04-13
Fix: ensure Stripe uses the correct secret key when requesting invoice

rahulacaleffi:
Resolves [GIVE-3132](https://stellarwp.atlassian.net/browse/GIVE-3132)

## Description

When multiple Stripe accounts are connected and assigned to different forms, and a donation comes through a form connected to a secondary account (not the default one), Stripe sends a webhook event to the website (type: `invoice.payment_succeeded`). While processing this event, we request the invoice information from Stripe.

However, during this request preparation, we correctly set the `stripe_account` by retrieving the connected account based on the form ID, but the correct API key (provided by Stripe during the onboarding connection process) was not being used. This caused the logic to fall back to the default key from the default account.

Stripe rejects this request because it targets one `stripe_account` while using another account's API key, which is not allowed when the platform is not using Stripe [Connect](https://stripe.com/connect).

**The fix:**
Passing `api_key` in the options tells the Stripe SDK to authenticate directly with that account’s own key, rather than relying on the globally set default key and the Connect `Stripe-Account` header.

## Affects

Stripe Payment Webhooks - getting the complete invoice from Stripe.

## Visuals

Located in Jira.

## Testing Instructions

Using Stripe Connect, connect to two different Stripe accounts.
Create two V3 donation forms with recurring enabled
Set one donation form to select the default Stripe account (or enable the slider to use the default)
Select the additional non-default Stripe account on the second donation form
Create recurring donations with both donation forms using Stripe as the payment gateway
Both donation forms should work properly and have their status updated after a few seconds

## Pre-review Checklist

-   [ ] Includes unit tests

[GIVE-3132]: https://stellarwp.atlassian.net/browse/GIVE-3132?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sdokus (approved):

bordoni (changes requested):
Changelog?

rahulacaleffi:
> Changelog?

@bordoni , it seems we don't create changelogs at this point, only during the release (from `develop` -> to `release`). At least we don't have that information in docs and the latest PRs don't contain changelog entries either.

kadondibetty:
Great and thanks

PR #8233, opened 2026-05-13
Harbor 1.2: premium plugin exists filter and active add-on detection

pauloiankoski:
## Summary

This change set prepares GiveWP for StellarWP Harbor 1.2 by bumping the `stellarwp/harbor` Composer constraint to `^1.2`, and wiring Harbor’s new `lw_harbor/premium_plugin_exists` boolean filter.

https://github.com/user-attachments/assets/74a7aae5-8f64-463e-a372-d8177d43fa3b

## Changes

- **Composer:** require `stellarwp/harbor` `^1.2` (Harbor 1.2 must be available before `composer install` succeeds).
- **`HasActivePremiumAddons`:** invokable action that returns `true` when another brand already reports premium presence, or when any GiveWP premium add-on is active; otherwise `false`.
- **`HarborServiceProvider`:** registers the action on `lw_harbor/premium_plugin_exists`.
- **Tests:** unit coverage for the other-brands short-circuit, empty premium list, inactive fixture add-on, and active fixture add-on.

### DIFF
--- a/src/VendorOverrides/Harbor/Actions/HasActivePremiumAddons.php
+++ b/src/VendorOverrides/Harbor/Actions/HasActivePremiumAddons.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Give\VendorOverrides\Harbor\Actions;
+
+/**
+ * Return true if another brand already has premium presence or if any GiveWP premium add-on is active.
+ *
+ * Hooked into the `lw_harbor/premium_plugin_exists` filter.
+ *
+ * @unreleased
+ */
+class HasActivePremiumAddons
+{
+    /**
+     * @unreleased
+     */
+    public function __invoke(?bool $otherBrandsExistence): bool
+    {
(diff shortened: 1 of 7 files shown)


src/VendorOverrides/Harbor/Actions/HasActivePremiumAddons.php
```diff
@@ -0,0 +1,35 @@
...
+ * Hooked into the `lw_harbor/premium_plugin_exists` filter.
+ *
+ * @unreleased
+ */
+class HasActivePremiumAddons
+{
+    /**
+     * @unreleased
+     */
+    public function __invoke(?bool $otherBrandsExistence): bool
```
dpanta94: We can declare strictly only bool. We are working on that, so if null ever comes is either our fault or external customization.

pauloiankoski: Resolved b9f3ca001a03e53ee1341ad6d0642270d0c2d002

PR #8237, opened 2026-06-02
Feature: add release-prep commands

jonwaldstein:
## Summary

This adds release-prep commands/scripts instead of having to do this manually. The idea is when you're prepping a release you can simply run composer run `release:prep 4.16.0` with whatever version number and it will handle all/most of the prep needed.

### Details
Adds Composer-invoked release tooling so version bumps, `@since` tag replacement, and changelog generation are scripted and tested instead of manual.

Two namespaces split the developer-facing changelog commands from the release-manager commands:

| Command | Purpose |
|---|---|
| `composer run changelog:add` | Author a changelog entry (auto-names the file from the branch via `--auto-filename`) — run this in each PR |
| `composer run changelog:write -- 4.16.0` | Compile pending entries into `readme.txt` + `changelog.txt` |
| `composer run release:bump-versions 4.16.0` | Update every version declared in `.puprc` → `paths.versions` (`give.php` ×2, `readme.txt`) |
| `composer run release:replace-since-tags 4.16.0` | Replace `@unreleased` / `@since TBD` with `@since 4.16.0` across `.php/.js/.jsx/.ts/.tsx` (descriptions preserved) |
| `composer run release:prep 4.16.0` | Run bump-versions → replace-since-tags → changelog write in order |
| `composer run release:test` | Run the Bats suite |

All commands accept `--dry-run`; `release:prep` / `changelog:write` also forward `--date`.

### Workflow

- **Per PR:** author writes `@unreleased` / `@since TBD` in code, and runs `composer run changelog:add`. The entry YAML is committed with the PR (branch-named, so no conflicts).
- **At release:** `composer run release:prep <version>` bumps versions, resolves the `@since` markers, and folds all pending `changelog/*.yaml` entries into the changelog files.

### Implementation notes

- **Layout:** CLI entrypoints in `bin/`; non-executable helpers in `bin/lib/` (`release-utils.php` for shared arg parsing/semver, `changelog-strategy.js` for the changelog format).
- **Changelog format:** `@stellarwp/changelogger` (dev dependency) loads a custom writing strategy from `bin/lib/changelog-strategy.js`, producing GiveWP's established WordPress format — `= 4.16.0: June 2nd, 2026 =` with `* Fix: …` — rather than the tool's native `= [4.16.0] 2026-06-02 =` / `* Fix - …`.
- **Versioning:** `changelog:write` requires an explicit version. changelogger's native auto-versioning only reads bracketed `= [x.y.z]` headers, so it would mis-detect against the legacy format; the chosen version is always passed explicitly.
- **Bash wrappers** are portable (`set -euo pipefail`, bash 3.2-safe) and fail gracefully if Node/changelogger is missing.
- `changelog/` holds pending entry files and is excluded from the distributed zip (`.distignore`); `/bin` was already excluded.
- Also retires the stale hand-written `= TBD: DATE TBD =` placeholder in `changelog.txt`.

### Testing

18 Bats tests in `tests/bats/` cover all scripts black-box: validation, real file mutation in sandboxes, dry-run no-write, excluded-dir skipping, legacy changelog header/bullet format, and ordinal date formatting. Run with `composer run release:test`.

### DIFF
--- a/bin/release-prep.sh
+++ b/bin/release-prep.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+#
+# Release prep: run the full version-bump pipeline in order.
+#
+#   1. Bump every version declared in .puprc -> paths.versions
+#   2. Replace @unreleased / @since TBD docblock tags with @since <version>
+#   3. Compile pending changelog entries into readme.txt / changelog.txt
+#
+# Usage:
+#   bin/release-prep.sh <version> [--date <date>] [--dry-run]
+#   composer run release:prep 4.16.0
+#
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+VERSION="${1:-}"
+shift || true
+
+if [ -z "$VERSION" ]; then
+  echo "Error: a version number is required." >&2
+  echo "Usage: bin/release-prep.sh <version> [--date <date>] [--dry-run]" >&2
+  exit 1
+fi
+
+# --dry-run is forwarded to every step; any other flags (e.g. --date) apply to
+# the changelog step, which receives the full remaining argument list ("$@").
+DRY_RUN=
+case " $* " in
+  *" --dry-run "*) DRY_RUN=1 ;;
+esac
+
+echo "=== 1/3: Updating version strings ==="
+php bin/update-versions.php "$VERSION" ${DRY_RUN:+--dry-run}
+
+echo ""
+echo "=== 2/3: Replacing @unreleased / @since TBD tags ==="
+php bin/replace-since-tags.php "$VERSION" ${DRY_RUN:+--dry-run}
+
+echo ""
+echo "=== 3/3: Writing changelog ==="
+bin/write-changelog.sh "$VERSION" "$@"
+
+echo ""
+echo "✓ Release prep complete for $VERSION"
--- a/bin/replace-since-tags.php
+++ b/bin/replace-since-tags.php
@@ -1,0 +1,109 @@
+<?php
+/**
+ * Release prep: replace placeholder docblock tags with the release version.
+ *
+ *   "@unreleased"        ->  "@since <version>"   (any trailing description is preserved)
+ *   "@since TBD"         ->  "@since <version>"
+ *
+ * Scans the plugin source (php, js, jsx, ts, tsx) and skips build/vendor dirs.
+ *
+ * Usage:
+ *   php bin/replace-since-tags.php <version> [--dry-run]
+ *   composer run release:replace-since-tags 4.16.0
+ */
+
+declare(strict_types=1);
+
+require __DIR__ . '/release-utils.php';
+
+[$version, $dryRun] = release_parse_args(
+    $argv,
+    'Usage: php bin/replace-since-tags.php <version> [--dry-run]'
+);
+
+$root = release_project_root();
+
+// Directories we never want to touch.
+$excludedDirs = [
+    '.git',
+    'node_modules',
+    'vendor',
+    'bin',
+    'assets/dist',
+    'build',
+];
+
+$extensions = ['php', 'js', 'jsx', 'ts', 'tsx'];
+
+// "@unreleased" (optionally followed by description) and "@since TBD".
+$patterns = [
+    '/@unreleased\b/'        => '@since ' . $version,
+    '/@since\s+TBD\b/'       => '@since ' . $version,
+];
+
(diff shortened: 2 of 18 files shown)


tests/bats/write-changelog.bats
pauloiankoski, line 1: The temporary `zzz-bats-tmp.yaml` file is removed in `teardown`, but if the test suite is killed mid-run (SIGKILL, OOM, lost SSH session) the entry remains and will be compiled into the real changelog on the next changelog:write run. All other tests in the suite use the `make_sandbox` / isolated copy pattern; this one breaks that consistency for a good reason (changelogger must read from the configured changesDir), but the risk is real in CI.

Suggested fix: Either pre-clean `changelog/zzz-bats-tmp.yaml` at the top of `setup()` in addition to relying on `teardown`, or add a suite-level `setup_file` / `teardown_file` guard.

bin/release-prep.sh
```diff
@@ -0,0 +1,45 @@
...
+echo ""
+echo "=== 2/3: Replacing @unreleased / @since TBD tags ==="
+php bin/replace-since-tags.php "$VERSION" ${DRY_RUN:+--dry-run}
+
+echo ""
+echo "=== 3/3: Writing changelog ==="
+bin/write-changelog.sh "$VERSION" "$@"
+
+echo ""
+echo "✓ Release prep complete for $VERSION"
```
pauloiankoski, line 45: After a successful run, the script has modified `give.php`, `readme.txt`, `changelog.txt`, and deleted all the YAML entries, but nothing is staged. There is currently no guidance in the output about the expected next step (`git add`, `git commit`, `git push`), so someone running this for the first time has to guess the intended workflow.

Suggested fix: append a small “Next steps” block to the success output, for example:

```bash id="p6n3vx"
echo ""
echo "Next steps:"
echo "  git add give.php readme.txt changelog.txt changelog/"
echo "  git commit -m \"chore: prepare for release $VERSION\""
echo "  git push origin <your-release-branch>"
```

glaubersilva (approved):
@jonwaldsteinI spent a good amount of time to better understand it and be able to review it. 😅 That said, it looks good to me. Great work, man!

bin/replace-since-tags.php
```diff
@@ -0,0 +1,109 @@
...
+
+    $ext = strtolower($file->getExtension());
+    if (!in_array($ext, $extensions, true)) {
+        continue;
+    }
+
+    $contents = file_get_contents($file->getPathname());
+
+    // Cheap pre-filter: skip the regex work unless a placeholder token is present.
+    if (strpos($contents, '@unreleased') === false && strpos($contents, 'TBD') === false) {
```
pauloiankoski, line 79: Any file containing the substring `TBD` anywhere (code comment, string literal, variable name) skips the early-exit and goes through the full regex work. The actual replacement regex is correct, so no false writes occur, it's a performance minor issue, not a bug. But the intent is clearly to pre-screen for `@since TBD`, so the check should be:

```php
strpos($contents, '@since TBD') === false
```

bin/update-versions.php
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    if ($updated === $contents) {
+        echo "  = $file already at $version ($regex)\n";
+        continue;
+    }
+
+    if (!$dryRun) {
+        file_put_contents($path, $updated);
+    }
```
pauloiankoski, line 100: If the third entry in `.puprc paths.versions` fails (missing file, bad regex) the script exits with code 1, but the first two files have already been written. The working tree is left in a half-updated state with no clear message about which files were actually changed.

Suggested fix: Collect all updates in memory first, then write them all only when `$hadError` is still false. This is a transaction-style change. Alternatively, at minimum, emit a final summary line that lists which files succeeded and which failed, so the developer knows what to roll back.

jonwaldstein:
@glaubersilva thanks! we can test these during a real release soon and make any adjustments if need be 😄

pauloiankoski (commented):
@jonwaldstein Left a few comments, but I’d also like to contribute with some additional release steps I’ve implemented locally that I think would fit well here too.

For example:
- creating the release branch automatically
- triggering ZIP creation
- adding a follow-up release script that can:
  - merge the release branch into develop and master
  - create the GitHub release
  - handle the remaining release automation steps

I think that would make the workflow much more complete and reduce the amount of manual coordination needed during releases.

jonwaldstein:
@pauloiankoski whoops I think you reviewed after merge haha - feel free to open a new PR with your updates 😄

PR #8239, opened 2026-06-03
Fix: wrap admin_url() with esc_url() on donor row action href

thisismyurl:
Hey folks,

The "View Donor" row action link in class-donor-table.php passes admin_url() directly as the %1$s placeholder in sprintf( '<a href="%1$s"...' ) without esc_url(). Adjacent actions in the same array (delete) build their hrefs with # so they don't have this issue, but the view action's admin_url() call needs esc_url() wrapping. One-line change.

(full disclosure, I used AI help with the testing and tweaks - Christopher)

## Summary by CodeRabbit

* **Security & Reliability**
  * Improved URL safety for the “View Donor” link in the admin donor table.
  * Updated release automation permissions to support uploading release assets successfully.

PR #8243, opened 2026-06-12
fix: wrap admin_url() with esc_url() on settings section nav links

thisismyurl:
## Summary

`abstract-admin-settings-page.php` builds navigation links for settings section tabs and places the result of `admin_url()` directly into an `href` attribute without wrapping it in `esc_url()`.

**Line 216 (before):**
```php
$section_list[] = '<li><a href="' . admin_url( 'edit.php?post_type=give_forms&page=' . $this->current_setting_page . '&tab=' . $this->id . '&section=' . sanitize_title( $id ) ) . '" class="...
```

**Line 216 (after):**
```php
$section_list[] = '<li><a href="' . esc_url( admin_url( 'edit.php?post_type=give_forms&page=' . $this->current_setting_page . '&tab=' . $this->id . '&section=' . sanitize_title( $id ) ) ) . '" class="...
```

## Why

Per the [WordPress VIP coding standards](https://docs.wpvip.com/how-tos/php_codesniffer/) and PHPCS `WordPress.Security.EscapeOutput`, any URL placed into an HTML `href` attribute should pass through `esc_url()`. The `wp_kses_post()` call at line 219 already provides a downstream content filter, but `esc_url()` is the correct defence-in-depth layer at the point of URL construction — it normalises the scheme, strips invalid characters, and makes the intent explicit.

I noticed PR #8239 applied the same pattern to `class-donor-table.php`; this follows the same approach for the settings page.

## Testing

1. Navigate to **Donations → Settings** in the WordPress admin.
2. Open any settings tab with multiple sections (e.g. General → Display Options).
3. Confirm the section navigation links render correctly and clicking them works as expected.
4. No change in visible behaviour is expected — only the URL encoding path is affected.

---

> **AI disclosure:** I use [Claude Code](https://claude.ai/code) as part of my development workflow. It helped me identify and verify this issue. I reviewed the code and the fix before opening this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security and reliability of admin section navigation links by safely escaping URLs before display.
  * Preserved existing tab and section navigation behavior.


thisismyurl:
Good points — pushed a follow-up that adds `esc_attr()` to the class attribute value:

```php
esc_attr( $current_section === $id ? 'current' : '' )
```

On `$label`: the full `$section_list` output is already wrapped in `wp_kses_post()` at line 219, which filters `$label` content at the point of rendering. Adding `esc_html()` here would strip intentional markup in section labels (e.g. if any use `<strong>`) and would be redundant with the existing downstream filter. Happy to revisit if the project's convention is to escape at construction rather than at output.

jonwaldstein:
@coderabbitai review

jonwaldstein:
@coderabbitai full review

PR #8246, opened 2026-06-16
Fix: only run Akismet spam check once per donation (SMTNC-1508)

### LINKED ISSUE #8238: GiveWP calls Akismet after every screen in a donation form instead of once.
I work on Akismet, the anti-spam service that GiveWP integrates with if the user has it installed.

We got a notice that donations were being rejected with the message `Something went wrong, please try again or contact support` if Akismet is installed. I did some investigating and found that this is because GiveWP is sending the form submission to be checked by Akismet after every screen in a multi-screen form (e.g., once after selecting the donation amount, again after the user enters a comment in a text field, again after the user enters their name, etc.). Since two or three checks in quick succession from the same IP look very much like the beginning of a flood of fraudulent donations, Akismet begins blocking it at that point.

Ideally, there should only be one Akismet request per donation, sent whenever all of the data that's going to be collected is available. If you have a good reason to check after every screen instead of just once, include `recheck_reason=edit` on the subsequent checks as a signal that the check is not part of a new donation.

To confirm which Akismet requests are being sent, add this to your wp-config.php:

```
define('WP_DEBUG', true);
define('WP_DEBUG_LOG', true);
define('AKISMET_DEBUG', true );
```

Requests and responses from Akismet will then be logged in `wp-content/debug.log`.

Let me know if you have any questions of if I can help in any way.

jonwaldstein:
Resolves: #8238, [SMTNC-1508]
## The problem

Donations were being rejected with *"Something went wrong, please try again or contact support"* when Akismet is installed. Our multi-step v3 forms re-validate on the server after **every** step, which fired the Akismet spam check once per step. Several checks from the same IP in quick succession look like the start of a spam flood, so Akismet began blocking legitimate donations.

## The direction we chose

Per-step server validation exists for **field-level UX feedback** — it has nothing to do with spam checking. So instead of marking the repeated checks with Akismet's `recheck_reason=edit` flag (suggested fallback), we went with the *preferred* approach: **run the Akismet check exactly once, on final submission**, when all the donation data is available.

To keep this clean, the Give layer only expresses a Give concept — whether a validation is the final submission — and the Akismet integration decides what to do with it. `recheck_reason` stays an Akismet-only detail and never leaks into the form DTOs.

## What changed

- The two validation routes now pass an `$isFinalSubmission` flag to the `givewp_donation_form_fields_validated` action (`false` for per-step validation, `true` for final submission).
- The Akismet integration skips non-final validations, so it checks once per donation.
- The flag defaults to `true`, so any other caller of the action still gets checked (fail-safe).

Per-step field validation and its UX feedback are completely unchanged.

## Testing

ZIP: https://github.com/impress-org/givewp/actions/runs/27640983322

Setup Debugging:
```php
define('WP_DEBUG', true);
define('WP_DEBUG_LOG', true);
define('AKISMET_DEBUG', true );
```

- Create a v3 multi-step form
- Install Akismet & enable in Give advanced settings
- Use the [Akismet testing fields](https://akismet.com/support/getting-started/confirm/)
- Make sure akismet is only fired once by checking the debug log

## DEMO

https://www.loom.com/share/1dc8f7123ec247e4b4d1b2688ec1f13a

### DIFF
--- a/src/DonationSpam/ServiceProvider.php
+++ b/src/DonationSpam/ServiceProvider.php
@@ -29,6 +29,10 @@ class ServiceProvider implements ServiceProviderInterface
     }
 
     /**
+     * @since TBD only check the final submission, when all donation data is available. Per-step
+     *               validations exist for field-level UX feedback and are not spam-relevant;
+     *               checking on every step makes Akismet treat the repeated requests from one IP as
+     *               the start of a new (potentially fraudulent) donation flood.
      * @since 3.22.0 updated Akismet validation to use new givewp_donation_form_fields_validated action
      * @since 3.15.0
      *
@@ -39,14 +43,18 @@ class ServiceProvider implements ServiceProviderInterface
     public function boot(): void
     {
         if ($this->isAkismetEnabledAndConfigured()) {
-            add_action('givewp_donation_form_fields_validated', static function (array $data) {
+            add_action('givewp_donation_form_fields_validated', static function (array $data, bool $isFinalSubmission = true) {
+                if (!$isFinalSubmission) {
+                    return;
+                }
+
                 give(ValidateDonation::class)(
                     $data['email'] ?? '',
                     $data['comment'] ?? '',
                     $data['firstName'] ?? '',
                     $data['lastName'] ?? ''
                 );
-            });
+            }, 10, 2);
         }
     }
 
--- a/src/DonationSpam/ServiceProvider.php
+++ b/src/DonationSpam/ServiceProvider.php
@@ -2,8 +2,8 @@
 
 namespace Give\DonationSpam;
 
-use Give\DonationSpam\Akismet\Actions\ValidateDonation;
-use Give\DonationSpam\Exceptions\SpamDonationException;
+use Give\DonationSpam\Akismet\Actions\ValidateDonationOnFinalSubmission;
+use Give\Helpers\Hooks;
 use Give\ServiceProviders\ServiceProvider as ServiceProviderInterface;
 
 /**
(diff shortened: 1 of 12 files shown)


DevinWalker:
You're the man, @jonwaldstein

src/DonationSpam/ServiceProvider.php
```diff
@@ -39,14 +40,18 @@ public function register(): void
...
+                }
+
                 give(ValidateDonation::class)(
                     $data['email'] ?? '',
                     $data['comment'] ?? '',
                     $data['firstName'] ?? '',
                     $data['lastName'] ?? ''
                 );
-            });
+            }, 10, 2);
```
glaubersilva: @jonwaldstein Wouldn't it be better to move this logic into a dedicated action class? 

A few reasons:

- **Stack traces**: anonymous closures show as `{closure}` — a named class shows the exact method, which is much easier to trace when something fails in production.
- **`remove_action`**: impossible to unhook a closure from outside. A class is removable with `remove_action('hook', ActionClass::class)`.
- **Testability**: `ServiceProviderTest` is currently testing business logic through a side effect of hook registration, which means any test for this behavior is implicitly also testing that the hook was registered. Those are two separate concerns. With an action class, the logic lives in `ActionClass::__invoke()` and can be tested directly, without touching the service provider at all.

jonwaldstein: True, if we want to clean up this whole integration than I agree an action class would be better. Something else I realized is the check for akismet enabled is on boot and not inside the validated action itself which would happen on every request.

jonwaldstein: @glaubersilva I updated this to use an action class [0371f79](https://github.com/impress-org/givewp/pull/8246/commits/0371f7912179283115e820cbf2e5e46c81558cb5) and made some other minor updates just to clean things up while we're here 😄

glaubersilva: Great! Much better now.

glaubersilva (approved):
@jonwaldstein Thanks for the changes! Ready for QA. 🚀

PR #8254, opened 2026-06-29
Fix: Sanitize fee recovery amount before saving to DB.

joshAppDev:
### Background

The give-fee-recovery plugin stores the fee amount in the donation meta under '_give_fee_amount'. It uses the `give_sanitize_amount_for_db()` function before saving. This results in the amount having 6 decimal places. (E.g. "4000.000000".)

However, GiveWP automatically re-saves donation meta values if the donation is updated. For example, if the status changes from PROCESSING to COMPLETED. For '_give_fee_amount', it uses only the `formatToDecimal()` method before saving. This has 2 decimal places for the amount. (e.g. "4000.00".)

### The problem

When a currency uses `.` as a thousands separator, it can change how GiveWP parses and formats amounts. This is more noticeable when large amounts are common with the currency, such as with VND or IDR.

Here is an example, from one of our Give sites running 3.22:
image

The amount was initially correct when the donation status was still PROCESSING. But after it changed to COMPLETED, the displayed fee amount became wrong.

### Parsing issue

GiveWP uses several different functions when formatting the fee amount for display, including `give_maybe_santize_amount()`. When this function sees a fee amount with 6 decimal places, it parses correctly. But when the function sees the fee amount with 2 decimal places, it thinks that "4000.00" == "400000". Perhaps there is a way to fix this parsing algorithm, but not in this pull request.

( See give-fee-recovery/includes/admin/class-give-fee-recovery-admin.php :: donation_detail() )

### Sanitizing solution

By adding `give_sanitize_amount_for_db()` to DonationRepository.php, we ensure the fee amount is consistently formatted for the DB across both give core, and give-fee-recovery. As a bonus, it also works around the parsing issue above.

## Testing Instructions

For the IDR currency, we configure zero decimal places in the GiveWP settings.  Additionally, we use the `give_get_zero_based_currencies` and `give_fee_zero_based_currency` filters to set IDR as zero based. 

## Pre-review Checklist

-   [ ] Includes unit tests


kadondibetty:
Great and thanks

src/Donations/Repositories/DonationRepository.php
```diff
@@ -477,7 +477,7 @@ private function getCoreDonationMetaForDatabase(Donation $donation): array
         }
 
         if ($donation->feeAmountRecovered !== null) {
-            $meta[DonationMetaKeys::FEE_AMOUNT_RECOVERED] = $donation->feeAmountRecovered->formatToDecimal();
+            $meta[DonationMetaKeys::FEE_AMOUNT_RECOVERED] = give_sanitize_amount_for_db($donation->feeAmountRecovered->formatToDecimal());
```
jonwaldstein, line 480: I think this would only be a problem on the legacy admin screens but not on the new ones.  `give_sanitize_amount_for_db` is also not something we want to rely on here

jonwaldstein (changes requested):

src/Donations/Repositories/DonationRepository.php
joshAppDev: `give_sanitize_amount_for_db` is being widely used in Give [here](https://github.com/impress-org/givewp/blob/08ca48d3263fcdf0b7a520a118e36f1eeadb2175/src/Donations/Repositories/DonationRepository.php#L442) and applied to all donation amounts. The change in this pull request merely extends that to also apply to the fee recovery amount.

PR #8255, opened 2026-07-06
fix: Remove extra separators rendering multiple times

WayneRocha:
Resolves #[SMTNC-212](https://linear.app/nexcess/issue/SMTNC-212/an-extra-separator-should-not-be-rendered-when-the-donation)

## Description

Some change (that I couldn't track) inserted a regression fixed by https://github.com/impress-org/givewp/pull/8222. The fix was to not render divisors when ::after had some content in it, but later line breaks introduced a space (&nbsp) which IS content.
New fix was to add divisor only as `::after` of `receipt-header-heading`

## Affects

Style from the header of Confirmation page

## Visuals

image

image

https://www.loom.com/share/5df59d73048b49bfa861780bf212f405

## Testing Instructions

*Admin Panel > GiveWP > Campaings > (Create a new or go to "Forms" inside a campaing) > Edit >"Settings" tab > Donation Confirmation > Header*

Edit header, put lines breaks, try to mess with it.

Only one divisor should exist

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/DonationForms/resources/styles/components/_receipt.scss
+++ b/src/DonationForms/resources/styles/components/_receipt.scss
@@ -318,6 +318,10 @@
     &-heading {
         margin-block-end: 1rem;
 
+        > br {
+            display: none;
+        }
+
         &::after {
             position: absolute;
             content: '';
--- a/src/DonationForms/resources/styles/components/_receipt.scss
+++ b/src/DonationForms/resources/styles/components/_receipt.scss
@@ -322,7 +322,7 @@
             display: none;
         }
 
-        &::after {
+        &:not(:empty)::after {
             position: absolute;
             content: '';
             display: block;
--- a/src/DonationForms/resources/styles/components/_receipt.scss
+++ b/src/DonationForms/resources/styles/components/_receipt.scss
@@ -304,7 +304,6 @@
 
     &-heading, &-heading * {
         position: relative;
-        margin-block-end: 1rem;
         font-size: 1.75rem;
         font-weight: 600;
         line-height: 1.2;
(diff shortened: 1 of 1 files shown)


src/DonationForms/resources/styles/components/_receipt.scss
```diff
@@ -314,8 +313,12 @@
         > * {
             color: currentColor;
         }
+    }
+
+    &-heading {
+        margin-block-end: 1rem;
 
-        &:not(:empty)::after {
+        &::after {
```
pauloiankoski: This new `::after` rule has no `:not(:empty)` check. Before this PR, an empty heading (admin clears the "Header" field) showed no separator line. Now it will always show the separator line, even with no text above it. Let's add `:not(:empty)` back, like `&-heading:not(:empty)::after`, so an empty heading still shows nothing.

Image

src/DonationForms/resources/styles/components/_receipt.scss
```diff
@@ -314,8 +313,12 @@
         > * {
             color: currentColor;
         }
+    }
+
+    &-heading {
+        margin-block-end: 1rem;
```
pauloiankoski, line 319: Whenever we have a line break, a `<br>` tag is rendered. However, we are also accounting for spacing after every `<p>` element, which causes the spacing to be duplicated.

Image

To fix that, we could have something like:
```scss
> br {
    display: none;
}
```

Image

WayneRocha: It's a good adjust @pauloiankoski 
Added that in 1dbda6aa841a721e760cd4ad1199392e2a94ebba

WayneRocha: And if somehow the title is empty I will not render the divisor.
14e7f9d676bc1d370274db3772d1a2da03c60c72

rahulacaleffi (approved):
Looks good to me. I saw the adjustments have also been applied.

kadondibetty:
Great and thanks

kadondibetty:
Great and thanks

PR #8259, opened 2026-07-15
Fix: Cannot refund donation when using per form stripe account

WayneRocha:
Resolves [SMTNC-1687](https://linear.app/nexcess/issue/SMTNC-1687/cannot-refund-donation-when-using-per-form-stripe-account)

## Description

**It searches the payment_intent with the default account, and when we don't use the default account, but a different, it does not pass the account_id.
The solutions is to provide the account_id in the refund flow**

- When testing, I needed to expose more info about the error in [DonationController::refund_item](https://github.com/impress-org/givewp/blob/77fd85d9e65584460af5539a2d246459d981f709/src/API/REST/V3/Routes/Donations/DonationController.php#L417), which gave me this info that helped me find the solution:
```json
{
    "message": "Failed to refund donation",
    "error": "Stripe API error: No such payment_intent: 'pi_***************'",
    "code": 0
}
```
- Used `setUpStripeAppInfo` in `StripePaymentElementGateway::refundDonation` like in other flows to setup the necessary info.
- `DonationController::refund_item` now uses `DonationViewModel` like other classes do, for consistent response shape

- I tested the webhook flow too. For example if I refunded directly from Stripe webhook responded and it worked on Give side.
## Affects

**Refunds behavior**
 
## Visuals

Reproducing the error:
https://www.loom.com/share/c9a71091b56640159de05ad41483ff37

## Testing Instructions

1. In GiveWP Payment settings, connect two different Stripe accounts.
1. In a donation form, select the Stripe account that's not the general default.
1. Make a donation.
1. Try to refund a donation

## Pre-review Checklist

-   [x] Includes unit tests
-   [ ] Reviewed by the designer (if follows a design) - **No visual changes**

### DIFF
--- a/src/API/REST/V3/Routes/Donations/DonationController.php
+++ b/src/API/REST/V3/Routes/Donations/DonationController.php
@@ -165,6 +165,19 @@ class DonationController extends WP_REST_Controller
                         'type' => 'integer',
                         'required' => true,
                     ],
+                    'includeSensitiveData' => [
+                        'type' => 'boolean',
+                        'default' => false,
+                    ],
+                    'anonymousDonations' => [
+                        'type' => 'string',
+                        'default' => 'exclude',
+                        'enum' => [
+                            'exclude',
+                            'include',
+                            'redact',
+                        ],
+                    ],
                 ],
             ],
             'schema' => [$this, 'get_public_item_schema'],
@@ -410,9 +423,12 @@ class DonationController extends WP_REST_Controller
                 $handler->handle($donation);
             }
 
+            $includeSensitiveData = $request->get_param('includeSensitiveData');
+            $donationAnonymousMode = new DonationAnonymousMode($request->get_param('anonymousDonations'));
+
             $item = (new DonationViewModel($donation))
-                ->includeSensitiveData(true)
-                ->anonymousMode(new DonationAnonymousMode('include'))
+                ->includeSensitiveData($includeSensitiveData)
+                ->anonymousMode($donationAnonymousMode)
                 ->exports();
 
             $response = $this->prepare_item_for_response($item, $request);
--- a/src/API/REST/V3/Routes/Donations/DonationController.php
+++ b/src/API/REST/V3/Routes/Donations/DonationController.php
@@ -410,7 +410,12 @@ class DonationController extends WP_REST_Controller
                 $handler->handle($donation);
             }
 
-            $response = $this->prepare_item_for_response($donation->toArray(), $request);
+            $item = (new DonationViewModel($donation))
+                ->includeSensitiveData(true)
+                ->anonymousMode(new DonationAnonymousMode('include'))
+                ->exports();
+
+            $response = $this->prepare_item_for_response($item, $request);
 
             return rest_ensure_response($response);
         } catch (\Exception $exception) {
--- a/src/API/REST/V3/Routes/Donations/DonationController.php
+++ b/src/API/REST/V3/Routes/Donations/DonationController.php
@@ -414,7 +414,11 @@ class DonationController extends WP_REST_Controller
 
             return rest_ensure_response($response);
         } catch (\Exception $exception) {
-            return new WP_REST_Response(__('Failed to refund donation', 'give'), 500);
+            return new WP_REST_Response([
+                'message' => __('Failed to refund donation', 'give'),
+                'error' => $exception->getMessage(),
+                'code' => $exception->getCode(),
+            ], 500);
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/API/REST/V3/Routes/Donations/DonationController.php
```diff
@@ -410,11 +410,20 @@ public function refund_item($request)
                 $handler->handle($donation);
             }
 
-            $response = $this->prepare_item_for_response($donation->toArray(), $request);
+            $item = (new DonationViewModel($donation))
+                ->includeSensitiveData(true)
+                ->anonymousMode(new DonationAnonymousMode('include'))
```
jonwaldstein: I don't love that these are hardcoded as they should be using the REST API args - @glaubersilva what do you think?

src/PaymentGateways/Gateways/Stripe/StripePaymentElementGateway/StripePaymentElementGateway.php
```diff
@@ -172,7 +172,9 @@ public function createPayment(Donation $donation, $gatewayData): GatewayCommand
     public function refundDonation(Donation $donation): PaymentRefunded
     {
         try {
-            $refund = $this->refundStripePayment($donation);
+            $this->setUpStripeAppInfo($donation->formId);
+            $stripeConnectAccountId = give_stripe_get_connected_account_id($donation->formId);
+            $refund = $this->refundStripePayment($donation, $stripeConnectAccountId);
```
jonwaldstein: is `refundStripePayment` being used anywhere else? you might want to check give-recurring too.  I'm thinking since we're already passing the `$donation` model as a param we could add this logic right inside the method itself instead of updating its signature with a new param.

src/API/REST/V3/Routes/Donations/DonationController.php
WayneRocha: Makes sense
db01cb4d57f5aeeab7b2d81bb212513df20f5f4c

src/PaymentGateways/Gateways/Stripe/StripePaymentElementGateway/StripePaymentElementGateway.php
WayneRocha: e8fe23621251fa391c4c7a4958ea78ae4f05bcc4

src/API/REST/V3/Routes/Donations/DonationController.php
jonwaldstein: @WayneRocha cool, but in order to use these now you would need to adjust the refund http request to add these args

WayneRocha: @jonwaldstein Really? I made it with default values to not need to add it on the front-end. Tested and it worked, but I'll take another look

jonwaldstein: yeah because the default is `false` and `exclude` so you would want to send `true` and `include` along 😄

jonwaldstein: (unless you don't need data from those args then it doesn't matter)

WayneRocha: Yeah, I don't really need. But I wanted to make this thing in there in case we need something more complex in the front-end in the future. It could even receive a blank response now and would work.

jonwaldstein: Sounds good, you might want to test/QA a situation where a donation is marked as anonymous just to make sure

WayneRocha: [Seems to be](https://www.loom.com/share/b579407548da44aa81b55d3e0470ba9e). Is that what you meant?

jonwaldstein: @WayneRocha not quite, a donation can be marked as "anonymous" by the donor.  This is what that argument is for (to protect against donors wanting to remain anonymous and not display their information publicly).   I was just saying you should test this with a donation that is marked as anonymous and what would happen if that REST argument `anonymousDonations` was either include or exclude.

Screenshot 2026-07-16 at 10 13 27 AM

WayneRocha: Thanks for the clarification @jonwaldstein .
I did some tests here. All of them work, and they have these responses:
```
https://pastebin.com/Jn98Qxef -> REST | includeSensitiveData = true  | anonymousDonations = "include"
https://pastebin.com/FygV7pKq -> REST | includeSensitiveData = true  | anonymousDonations = "exclude"
https://pastebin.com/RWq56JEr -> REST | includeSensitiveData = true  | anonymousDonations = "redact"
https://pastebin.com/qPWDaqUA -> REST | includeSensitiveData = false | anonymousDonations = "redact" | Marked Donation as anonymous
```

WDYT? I didn't really touched `DonationViewModel`, so it should have the behavior it had before

jonwaldstein: @WayneRocha cool, I just wanted to make sure when using the refund button in the UI, with a donation marked as anonymous while using these new args you introduced - nothing weird happens 😄

PR #8277, opened 2026-08-04
Fix: Activation banner doesn't appear when a Give Addon is activated from Harbor

pramodjodhani:
Resolves [SMTNC-1935](https://linear.app/nexcess/issue/SMTNC-1935/givewp-activation-banner-doesnt-appear-when-the-plugin-is-activated)

## Description

Activation banner doesn't appear when a Give Addon is activated from Harbor

## Visuals

image

## Testing Instructions

- Activate the unified license key and install GiveWP
- Activate a new Give addon (Fee recovery, Peer-to-peer, currency switcher) from the Harbor (LW License Manager)
- Notice that the Activation banner for the addon as shown in the screenshot above doesn't appear on the plugins page.
- Switch to this branch, activate the addon again and check the plugins page again. The Activation banner should appear this time.

## Pre-review Checklist

-   [ ] Includes unit tests

---
<a href="https://app.blacksmith.sh/impress-org/codesmith/givewp/pr/8277"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788447653&installation_model_id=426813&pr_number=8277&repository=impress-org%2Fgivewp&return_to=https%3A%2F%2Fgithub.com%2Fimpress-org%2Fgivewp%2Fpull%2F8277&signature=f5f41c123c2eca80878d3c7cb3f8afa26a220cb8e89e78ad80109df08545375a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/changelog/smtnc-1935.yaml
+++ b/changelog/smtnc-1935.yaml
@@ -1,0 +1,5 @@
+significance: minor
+type: fix
+entry: Activation banner for addons doesn't show when addons are activated from
+  the Unified License Manager
+timestamp: 2026-08-04T15:29:01.437Z
--- a/changelog/smtnc-1935.yaml
+++ b/changelog/smtnc-1935.yaml
@@ -1,4 +1,4 @@
-significance: minor
+significance: patch
 type: fix
 entry: Activation banner for addons doesn't show when addons are activated from
   the Unified License Manager
--- a/src/VendorOverrides/Harbor/HarborServiceProvider.php
+++ b/src/VendorOverrides/Harbor/HarborServiceProvider.php
@@ -46,12 +46,20 @@ class HarborServiceProvider implements ServiceProviderContract
         // adds a "licensing" submenu to Give
         lw_harbor_register_submenu('edit.php?post_type=give_forms');
 
-        add_action( 'rest_api_init', function() {
-            $file = GIVE_PLUGIN_DIR . 'includes/admin/plugins.php';
-            if ( file_exists( $file ) ) {
-                require_once $file;
-            }
-        });
+        // Fixes Activation banner doesn't appear when the Give Addon is activated from Harbor.
+        add_action( 'rest_api_init', array($this, 'load_plugin_file'));
+    }
+
+    /**
+     * Load the plugin file.
+     *
+     * @since TBD
+     */
+    public function load_plugin_file() {
+        $file = GIVE_PLUGIN_DIR . 'includes/admin/plugins.php';
+        if ( file_exists( $file ) ) {
+            require_once $file;
+        }
     }
 
 }
--- a/src/VendorOverrides/Harbor/HarborServiceProvider.php
+++ b/src/VendorOverrides/Harbor/HarborServiceProvider.php
@@ -46,20 +46,23 @@ class HarborServiceProvider implements ServiceProviderContract
         // adds a "licensing" submenu to Give
         lw_harbor_register_submenu('edit.php?post_type=give_forms');
 
-        // Fixes Activation banner doesn't appear when the Give Addon is activated from Harbor.
-        add_action( 'rest_api_init', array($this, 'load_plugin_file'));
+        // Fixes Activation banner doesn't appear when a Give Addon is activated from Harbor.
+        add_action('rest_api_init', array($this, 'loadGivePluginFile'));
     }
 
     /**
-     * Load the plugin file.
+     * Load the plugin file so that activation banner appears when
+     * a Give Addon is activated from Harbor.
      *
      * @since TBD
+     *
+     * @return void
      */
-    public function load_plugin_file() {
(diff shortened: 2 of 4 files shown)


src/VendorOverrides/Harbor/HarborServiceProvider.php
```diff
@@ -45,6 +45,24 @@ public function boot()
 
         // adds a "licensing" submenu to Give
         lw_harbor_register_submenu('edit.php?post_type=give_forms');
+
+        // Fixes Activation banner doesn't appear when a Give Addon is activated from Harbor.
+        add_action('rest_api_init', array($this, 'loadGivePluginFile'));
```
jonwaldstein: @pramodjodhani this does not seem like a good practice, this would load a give legacy php file while preparing for any REST request. can you please re-think this approach?

jonwaldstein (changes requested):

jonwaldstein:
Hey @pramodjodhani (here is my AI written comment that should clarify my initial suggestion)

nice find on the root cause here — the `$_REQUEST`-based detection in `give_recently_activated_addons()` was definitely too strict, and the fallback to the `activated_plugin` action's `$plugin` argument is exactly the right fix. I'd like to keep that part as-is. 👍

However, I'd recommend we avoid hooking into `rest_api_init` in `HarborServiceProvider` and discuss an alternative. Here's my thinking:

### The real gap is *where the listener is registered*, not the REST API

Harbor already goes through the standard path — `Plugin_Strategy::do_activate()` calls `activate_plugin()`, which fires `activated_plugin`. The only reason that wasn't enough is that the `add_action( 'activated_plugin', 'give_recently_activated_addons' )` registration lives in `includes/admin/plugins.php`, which is only loaded when `is_admin()` is true (via `LegacyServiceProvider` → `Give_Admin`). So during a REST activation the action fires but nobody is listening.

The `rest_api_init` workaround does close that gap, but it fires on **every** REST request (Gutenberg autosaves, integrations, etc.) and loads the whole legacy admin file up front each time, just in case an activation might happen. It's also Harbor-specific, when WP-CLI, plugin dependencies, and any programmatic `activate_plugin()` call funnel through the same action.

### Suggestion: hook `activated_plugin` directly from always-loaded code

Registering a listener is essentially free (one `$wp_filter` entry — no file loading), and `activated_plugin` only fires in the rare request where an activation actually happens, so the event itself gives us the scoping. Two ways to do it:

**Option A (my preference) — relocate the small listener.** After your change, `give_recently_activated_addons()` no longer has any admin dependencies — it takes the `$plugin` arg and writes an option. We could move it and `give_get_recently_activated_addons()` into an always-loaded file (`includes/misc-functions.php` is loaded unconditionally and already hosts related banner helpers), register the `activated_plugin` hook there, and remove both the registration in `includes/admin/plugins.php` and the `rest_api_init` block from `HarborServiceProvider`. Net cost on normal requests: zero.

**Option B — lazy require inside the callback.** Register from always-loaded code (e.g. `includes/actions.php`):

```php
add_action('activated_plugin', static function ($plugin) {
    if (! function_exists('give_recently_activated_addons')) {
        require_once GIVE_PLUGIN_DIR . 'includes/admin/plugins.php';
    }
    give_recently_activated_addons($plugin);
});
```

(and remove the `add_action` in `includes/admin/plugins.php` so admin-initiated activations don't record twice). This defers the legacy file load to the moment an activation actually happens — though it does register the file's other admin filters at an odd time, so I'd lean toward Option A.

Either way, the banner side needs no changes — `Give_Addon_Activation_Banner` already picks up the `give_recently_activated_addons` option on the next visit to the plugins page, which is the intended UX.

Curious to hear your thoughts!

pramodjodhani:
Thanks @jonwaldstein. It makese sense. I have applied the option A. Can you please give it another look?

changelog/smtnc-1935.yaml
```diff
@@ -0,0 +1,5 @@
+significance: minor
```
jonwaldstein:
```suggestion
significance: patch
```

jonwaldstein (approved):
@pramodjodhani thanks for the updates! lets make sure to show a working demo/loom of this and send to QA 🌈

PR #8288, opened 2026-08-17
[SMTNC-2034] WP admin notice prompting site administrators to update WordPress

WayneRocha:
### 🎫 Ticket

[SMTNC-2034](https://linear.app/nexcess/issue/SMTNC-2034/give-core)

### 🎥 Artifacts

image

### 🗒️ Description

Adds a dismissible WordPress admin notice that prompts site administrators to update WordPress while a core update is available, with the dismissal flag shared across the StellarWP plugins showing the same notice so it only has to be dismissed once.

> Shared flag = `nxs_wp_core_update_notice_dismissed`

### ⚠️ Blockers

None

---
<a href="https://app.blacksmith.sh/impress-org/codesmith/givewp/pr/8288"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789595493&installation_model_id=426813&pr_number=8288&repository=impress-org%2Fgivewp&return_to=https%3A%2F%2Fgithub.com%2Fimpress-org%2Fgivewp%2Fpull%2F8288&signature=9c42b52dd7edab6bc5f6b58c0d9eee76d791c53b8acc7b7a94f3cbff2cac038b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/Notices/WordPressCoreUpdateNotice.php
+++ b/src/Notices/WordPressCoreUpdateNotice.php
@@ -1,0 +1,140 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Give\Notices;
+
+use Give\Vendors\StellarWP\AdminNotices\AdminNotices;
+
+/**
+ * Prompts site administrators to update WordPress while the install is behind the latest release.
+ *
+ * @since TBD
+ */
+class WordPressCoreUpdateNotice
+{
+    /**
+     * Dismissal flag shared with the other StellarWP plugins that display this notice, so a site
+     * running more than one of them only has to dismiss it once.
+     *
+     * @since TBD
+     */
+    const DISMISSED_OPTION = 'nx_wp_core_update_notice_dismissed';
+
+    /**
+     * @since TBD
+     */
+    const NOTICE_ID = 'givewp-wp-core-update';
+
+    /**
+     * @since TBD
+     */
+    const DISMISS_ACTION = 'givewp-dismiss-wp-core-update-notice';
+
+    /**
+     * Registers the notice with the admin notices library.
+     *
+     * @since TBD
+     */
+    public function registerNotice(): void
+    {
+        AdminNotices::show(self::NOTICE_ID, [$this, 'renderNotice'])
+            ->asWarning()
+            ->dismissible()
+            ->withoutAutoParagraph()
+            ->ifUserCan('update_core')
+            ->when([$this, 'shouldDisplay']);
+    }
+
+    /**
+     * Stores the shared dismissal flag when the notice's dismiss control is used.
+     *
+     * @since TBD
+     */
+    public function handleDismissal(): void
+    {
+        if (!isset($_GET[self::DISMISS_ACTION])) {
+            return;
+        }
+
+        check_admin_referer(self::DISMISS_ACTION);
+
+        if (!current_user_can('update_core')) {
+            return;
+        }
+
+        update_option(self::DISMISSED_OPTION, true, false);
+
+        wp_safe_redirect(remove_query_arg([self::DISMISS_ACTION, '_wpnonce']));
+        exit;
+    }
+
+    /**
+     * @since TBD
+     *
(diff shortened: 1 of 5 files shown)


src/Notices/WordPressCoreUpdateNotice.php
```diff
@@ -0,0 +1,140 @@
...
+ */
+class WordPressCoreUpdateNotice
+{
+    /**
+     * Dismissal flag shared with the other StellarWP plugins that display this notice, so a site
+     * running more than one of them only has to dismiss it once.
+     *
+     * @since TBD
+     */
+    const DISMISSED_OPTION = 'nx_wp_core_update_notice_dismissed';
```
defunctl, line 22: You're missing a test case that should fail if this is changed to anything else with a custom message explaining why it should not change.

claudiosanches, line 22:
```suggestion
    const DISMISSED_OPTION = 'nxs_wp_core_update_notice_dismissed';
```

WayneRocha: put `nx` prefix by request of @coutinhocouto.

@coutinhocouto @claudiosanches @rahulacaleffi
What are the real pros vs cons of each prefix based on what we already know?

We need to get into a decision.

defunctl (changes requested):
I don't really understand the purpose of doing this when WordPress has the update notification already, and why plugins would all of the sudden be responsible for that. If this is something mandated, then I think this implementation is flawed:

1.  If multiple plugins implement this, it will still show multiple copies on first load, all would read the shared option as `false` and independently register their own isolated notices. Nothing elects which plugin should show the notice.
2. The shared value is a permanent bool, dismissing it once means it's dismissed for every future core update, it's not version/locale-aware, so once it's gone, it never comes back when their instance is outdated again.
3. This uses `update_option()`, so how is multisite handled? We want this a blog-specific dismissal?
4. The notices library already handles some per-user, per-plugin preferences, doesn't it? There are two competing dismissal mechanisms going on here.

This system would have to be designed to be aware of other instances of this and essentially pick a winner to handle it and decide on dismissal scope explicitly...does dismissing this dismiss for all admins? Just that admin? What about network-wide? etc... 

Don't quote me, but I think https://github.com/stellarwp/harbor does this in some way...also if Harbor is embedded in all of the required plugins already, it might be a better place to implement this?

Again, I don't really understand the purpose of doing this and I think WP should handle core update nags without us, but we'll see once we get more context about the "why" of this.

Edit: I have found out the purpose of this, and it turns out it will very likely be updated in sync in the future, so a proper system does need to be designed so the most recent version from all plugins would be the one shown, e.g. if they have an older version of the notice on Kadence, but this one is a more recent version, it should be shown etc...

src/Notices/WordPressCoreUpdateNotice.php
defunctl: @WayneRocha I think the issue is that your PR description contains this version: 

image

WayneRocha: All your comments are valid @defunctl, but i'm closing this PR since we are moving out from this approach.
More info on [slack](https://lw.slack.com/archives/C0A32QSRQUS/p1786995393967179).

PR #8296, opened 2026-08-20
Fix: scope custom amount minimum and maximum to the custom amount

claudiosanches:
Ticket: SMTNC-1454

The `Min`/`Max` rules built from the custom amount settings were attached to the `amount` field itself. Donation levels and the custom amount input write that same value, so a level below the minimum was rejected.

Admin-configured amounts, the donation levels or the fixed set price, are now exempt from both rules. Legacy v2 behaves this way already: `give_verify_minimum_price()` short-circuits when the submitted amount matches a level, and `give_get_form_minimum_price()` documents the minimum as "only enforced for the custom amount input".

The exemption keys on the submitted amount rather than the `levelId` hidden field, so a crafted POST cannot forge its way under the minimum. It follows that a typed custom amount that happens to equal a level is also accepted, matching v2.

The client needs the same rule or the donor never reaches the server, since the message the donor sees is Joi's. No new serialization was needed: the amount field's JSON already carries `levels` and `fixedAmountValue`.

`Max` reported an over-maximum amount as "must be greater than or equal to" the maximum. Corrected while in the file.

Unchanged by this: with the Currency Switcher on a non-base currency, level values are scaled by the exchange rate client-side while `customAmountMin` is not, so a converted level can still trip the minimum. The same mismatch already affects the minimum itself and `UpdateDonationLevelId`.

## Demo

https://www.loom.com/share/266822d04df44099a09b4645ec2e013e

## Summary by CodeRabbit

* **Bug Fixes**
  * Custom donation amount minimum and maximum validation now correctly allows configured donation levels and fixed prices.
  * Empty minimum values now default to the lowest available donation amount.
  * Empty maximum values allow unlimited custom amounts.
  * Validation messages now accurately describe maximum limits.

* **Documentation**
  * Updated custom amount guidance to clarify how minimums, maximums, and predefined donation levels work.

### DIFF
--- a/src/DonationForms/Actions/ConvertDonationAmountBlockToFieldsApi.php
+++ b/src/DonationForms/Actions/ConvertDonationAmountBlockToFieldsApi.php
@@ -28,6 +28,7 @@ class ConvertDonationAmountBlockToFieldsApi
 {
 
     /**
+     * @since TBD Exempt admin-defined amounts from the custom amount minimum and maximum.
      * @since 4.16.5 Set default value for the levelId hidden field.
      * @since 4.10.0 Replaced generic 'currency' rule with custom CurrencyRule that uses GiveWP's currency list
      * @since 3.0.0
@@ -48,12 +49,16 @@ class ConvertDonationAmountBlockToFieldsApi
             }
 
             if ($block->isCustomAmountEnabled()) {
+                $exemptAmounts = $this->getAdminDefinedAmounts($block);
+
                 if ($block->hasAttribute('customAmountMin')) {
-                    $amountRules[] = new Min($block->getAttribute('customAmountMin'));
+                    $amountRules[] = (new Min($block->getAttribute('customAmountMin')))
+                        ->exemptAmounts(...$exemptAmounts);
                 }
 
                 if ($block->hasAttribute('customAmountMax') && $block->getAttribute('customAmountMax') > 0) {
-                    $amountRules[] = new Max($block->getAttribute('customAmountMax'));
+                    $amountRules[] = (new Max($block->getAttribute('customAmountMax')))
+                        ->exemptAmounts(...$exemptAmounts);
                 }
             }
 
@@ -184,6 +189,31 @@ class ConvertDonationAmountBlockToFieldsApi
             ->label(__('Choose your donation frequency', 'give'))
             ->options(...$options)
             ->rules(new SubscriptionPeriodRule());
+    }
+
+    /**
+     * The amounts the admin configured on the block: the donation levels, or the fixed set price. The custom
+     * amount minimum and maximum never apply to these.
+     *
+     * @since TBD
+     *
+     * @return float[]
+     */
+    private function getAdminDefinedAmounts(DonationAmountBlockModel $block): array
+    {
+        $amounts = $block->getPriceOption() === 'multi'
+            ? array_column($this->prepareLevelsArray($block)['levels'], 'value')
+            : [$block->getSetPrice()];
+
+        // An unset level or set price sanitizes to 0, which must never be exempt from the minimum.
+        return array_values(
+            array_filter(
+                array_map('floatval', $amounts),
+                static function (float $amount): bool {
+                    return $amount > 0;
+                }
+            )
+        );
     }
 
     /**
--- a/src/DonationForms/Actions/ConvertDonationAmountBlockToFieldsApi.php
+++ b/src/DonationForms/Actions/ConvertDonationAmountBlockToFieldsApi.php
@@ -29,7 +29,7 @@ class ConvertDonationAmountBlockToFieldsApi
 
     /**
      * @since TBD Exempt admin-defined amounts from the custom amount minimum and maximum, and fall back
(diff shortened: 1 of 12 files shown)


src/DonationForms/Actions/ConvertDonationAmountBlockToFieldsApi.php
```diff
@@ -39,21 +40,29 @@ public function __invoke(DonationAmountBlockModel $block, string $currency): Don
...
                 if ($block->hasAttribute('customAmountMin')) {
-                    $amountRules[] = new Min($block->getAttribute('customAmountMin'));
+                    $amountRules[] = (new Min($block->getAttribute('customAmountMin')))
+                        ->exemptAmounts(...$exemptAmounts);
                 }
 
                 if ($block->hasAttribute('customAmountMax') && $block->getAttribute('customAmountMax') > 0) {
-                    $amountRules[] = new Max($block->getAttribute('customAmountMax'));
+                    $amountRules[] = (new Max($block->getAttribute('customAmountMax')))
+                        ->exemptAmounts(...$exemptAmounts);
```
jonwaldstein: Hey @claudiosanches something I want to consider here is setting the minimum amount based on the _lowest_ level defined when not using the `customAmountMin`.  Reason being, someone could programmatically submit the form at a lower amount than defined in the levels which is part of donation spam/card testing.  so we still want to prevent those scenarios when the form is configured as such.  let me know if that makes sense / have any questions 😄

src/DonationForms/Actions/ConvertDonationAmountBlockToFieldsApi.php
```diff
@@ -192,6 +193,34 @@ protected function getRecurringAmountPeriodField(DonationAmountBlockModel $block
...
+     *
+     * @param float[] $exemptAmounts
+     *
+     * @return int|null Null when the block defines neither a minimum nor an amount to derive one from.
+     */
+    private function getMinimumAmount(DonationAmountBlockModel $block, array $exemptAmounts): ?int
+    {
+        $customAmountMin = $block->hasAttribute('customAmountMin')
+            ? (int)$block->getAttribute('customAmountMin')
+            : 0;
```
jonwaldstein: Can you actually update `DonationAmountBlockModel` to support `$block->customAmountMin` so we don't have to manually reach into the attributes here?  Honestly, you can probably move most of the custom logic like `getAdminDefinedAmounts` there too.  Outside the scope of this but I had made `DonationAmountBlockType` a while back but did not implement it yet so `DonationAmountBlockModel` is still used.

src/DonationForms/Actions/ConvertDonationAmountBlockToFieldsApi.php
```diff
@@ -54,10 +55,10 @@ public function __invoke(DonationAmountBlockModel $block, string $currency): Don
 
             if ($block->isCustomAmountEnabled()) {
                 $exemptAmounts = $this->getAdminDefinedAmounts($block, $levels);
+                $minimum = $this->getMinimumAmount($block, $exemptAmounts);
```
jonwaldstein: We also want want do this when custom amount is not enabled so the minimum value falls back to the lowest level (regardless of custom amount being enabled)

jonwaldstein (changes requested):

src/DonationForms/Actions/ConvertDonationAmountBlockToFieldsApi.php
claudiosanches: Done. The minimum now applies when custom amount is off too, and `getMinimumAmount()` only honours `customAmountMin` when custom amount is enabled, so a stale value from a form that used to have it on cannot leak through. A fixed set price with custom amount off keeps its `Size` rule, which is stricter than a floor.

Worth flagging while we are here: with the Currency Switcher on a non base currency, `getAmountLevelsWithCurrencySettings` scales the levels by the exchange rate client side while the minimum stays in base currency. On a rate below 1 a converted level lands under the floor. That mismatch already existed for forms that set a minimum, and this widens it to forms that do not. I have left it to the separate currency ticket, let me know if you would rather handle it here.

jonwaldstein:
@coderabbitai full review

src/DonationForms/resources/app/utilities/ConvertFieldAPIRulesToJoi.ts
```diff
@@ -36,10 +36,35 @@ export default function getJoiRulesForForm(form: Form): ObjectSchema {
...
+                return helpers.error('number.min', {limit: min});
+            }
+
+            if (max !== undefined && amount > max) {
+                return helpers.error('number.max', {limit: max});
+            }
+
+            return value;
+        }, 'custom amount range');
+    }
```
jonwaldstein: @claudiosanches should this logic just live inside `convertFieldAPIRulesToJoi`?

claudiosanches:
@jonwaldstein Hello, could you please review this PR again?

jonwaldstein (approved):
@claudiosanches thanks for the updates, looks good to me 👍

claudiosanches:
@jonwaldstein Thanks!

PR #8299, opened 2026-08-24
fix: generate the POT file in the pup zip workflow

jonwaldstein:
## Problem

Zips built through `zip.yml` ship no `languages/give.pot`.

`generate-zip.yml` had an explicit `Generate pot file` step. When #8291 moved zip generation to pup, nothing took its place. `pup i18n` looks like the equivalent, but it downloads *finished* translations from a GlotPress instance rather than producing the source POT — and with no `i18n` section in `.puprc` it exits immediately.

From the one `zip.yml` run to date (32520652251):

```
19:56:22.639  composer -- pup i18n
19:56:22.823  composer -- pup package     ← 0.18s
```

Downloading that run's artifact, `languages/` contains only `.gitkeep`. The `4.16.7.1` zip on WordPress.org — still built by `release.yml`, which kept its own make-pot step — has a 904,160 byte `give.pot`.

## Fix

The shared StellarWP workflow runs `additional_commands` after `pup build` and before `pup package`, which is exactly the slot the POT has to be generated in to be included. `.distignore` has no `languages` entry, so the file ships once it exists.

The exclude list is carried over verbatim from `release.yml`, which is what produces the POT currently shipping.

## Verification before merge

Run this workflow manually from the Actions tab with `production: no`, download the artifact, and confirm `languages/give.pot` is present and comparable in size to the 904 KB one on WordPress.org.

Worth checking the size rather than just the presence: `release.yml` runs `npm run dev`, generates the POT, then runs `npm run build`, whereas pup only does the production build. If the production build turns out to hide strings the dev build exposes, the POT will come back noticeably smaller and this needs the dev build too.

## Related, not included

- `RELEASING.md:46` says publishing a release "pushes the pot file to the translations server". No workflow in this repo does that — 26 add-on repos have a `sync-translations.yml` caller, `givewp` does not, and `release.yml` has never had such a step in its history. Either core needs the caller or the doc needs correcting; that's a separate decision.
- `release.yml` still builds its own zip rather than using pup. Follow-up.

https://claude.ai/code/session_018jugbB3kEcQbDRzWuKfHLr

---
<a href="https://app.blacksmith.sh/impress-org/codesmith/givewp/pr/8299"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790194915&installation_model_id=426813&pr_number=8299&repository=impress-org%2Fgivewp&return_to=https%3A%2F%2Fgithub.com%2Fimpress-org%2Fgivewp%2Fpull%2F8299&signature=ed14738e826de6e1e6fe0ff6b48a1577b0f5917f45984a8cb7789b99a2bd1c64"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/zip.yml
+++ b/.github/workflows/zip.yml
@@ -48,10 +48,8 @@ jobs:
             # pup has no POT generation of its own: `pup i18n` pulls finished translations down from
             # GlotPress, it does not produce the source POT. Generate it here, which the shared workflow
             # runs after `pup build` and before `pup package`, so languages/give.pot ends up in the zip.
+            # WP-CLI is already on the PATH by this point - the shared workflow installs it.
             additional_commands: |
-                curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar
-                chmod +x wp-cli.phar
-                mv wp-cli.phar /usr/local/bin/wp
                 php -d xdebug.mode=off "$(which wp)" i18n make-pot "$GITHUB_WORKSPACE" "$GITHUB_WORKSPACE/languages/give.pot" --exclude="$(cat .distignore | tr "\n" "," | sed 's/,$/ /' | tr " " "\n"),src/**/*.js,*.js.map,blocks/**/*.js"
             slack_channel: ${{ inputs.slack_channel }}
             slack_thread: ${{ inputs.slack_thread }}
--- a/.github/workflows/zip.yml
+++ b/.github/workflows/zip.yml
@@ -45,6 +45,14 @@ jobs:
             production: ${{ inputs.production }}
             # GiveWP supports PHP 7.4 and up, so the vendor directory has to be built on 7.4.
             php_version: '7.4'
+            # pup has no POT generation of its own: `pup i18n` pulls finished translations down from
+            # GlotPress, it does not produce the source POT. Generate it here, which the shared workflow
+            # runs after `pup build` and before `pup package`, so languages/give.pot ends up in the zip.
+            additional_commands: |
+                curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar
+                chmod +x wp-cli.phar
+                mv wp-cli.phar /usr/local/bin/wp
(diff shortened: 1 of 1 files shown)


jonwaldstein:
Heads up on the red CI here: it is not this change.

The run fails at `pup build`, four steps before `additional_commands` is reached:

```
[error] Your github oauth token for github.com contains invalid characters: "***"
[FAIL] Build step failed: composer install --no-dev --prefer-dist --no-progress --optimize-autoloader
```

That is the Strauss/Composer token issue fixed in #8300. This PR touches only `.github/workflows/zip.yml` and #8300 touches only `bin/strauss-installar.sh` and `composer.json`, so there is no conflict between them and no need to stack — but this one cannot produce a green run until #8300 lands.

Suggested order: merge #8300, update this branch from `develop`, then dispatch Generate Zip manually with `production: no` and confirm `languages/give.pot` is in the artifact at roughly the 904 KB the WordPress.org zip ships. Merging on a green CI alone would not prove anything here, since the POT step has not actually executed yet.

jonwaldstein:
Verified on run 32777654437.

The artifact contains `languages/give.pot` at 904,160 bytes, matching the POT in the 4.16.7.1 zip on WordPress.org exactly. Diffing the two, the only difference in the whole file is the generation timestamp:

```diff
-"POT-Creation-Date: 2026-08-24T19:32:33+00:00\n"
+"POT-Creation-Date: 2026-08-24T21:11:47+00:00\n"
```

6,434 msgids in both.

That also answers the open question in the description: the production-only build pup runs produces an identical POT to `release.yml`'s `npm run dev` -> make-pot -> `npm run build` sequence. The dev build is not load-bearing for string extraction, so the eventual `release.yml` migration can drop it.

One caveat on this run being green: #8300 is not merged, and strauss 0.20.1 ran here without hitting the Composer token error. That failure turns out to be intermittent rather than permanent, so this run passing is not evidence the token issue is resolved. It is unrelated to this change either way.

WayneRocha (approved):
https://github.com/impress-org/givewp/actions/runs/32777654437 ✅

Image

PR #8301, opened 2026-08-25
Fix: Prevent fatal error when resuming paused Stripe subscriptions an…

WayneRocha:
Solves [SMTNC-2057](https://linear.app/nexcess/issue/SMTNC-2057/prevent-fatal-error-when-resuming-paused-stripe-subscriptions)

Slack: https://lw.slack.com/archives/C0A32QSRQUS/p1787139786399679

---

## What changed

[include_stripe_files()](https://github.com/impress-org/give-recurring/blob/5a6e1d99daecf9410449b5b7a2bb011e7d09ba4d/give-recurring.php#L305): removed the bailout that skipped loading the Stripe helper files when `give_stripe_is_any_payment_method_active()` returned false.

**Now GiveWP fixes `give_stripe_is_any_payment_method_active()` to return true if stripe is being used.**

That check only looks at the legacy gateway list, so it returns false for the next-gen stripe_payment_element gateway — which is exactly why the function was undefined. The two files it loads are definition-only (4 functions + one lazy class), so loading them unconditionally is safe.

InvoicePaymentSucceeded.php: reverted to the original direct call to give_recurring_stripe_is_subscription_completed().

Now the function is always defined, so the direct call runs the real completion + Stripe-cancel logic instead of silently skipping it.

---
<a href="https://app.blacksmith.sh/impress-org/codesmith/givewp/pr/8301"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790253726&installation_model_id=426813&pr_number=8301&repository=impress-org%2Fgivewp&return_to=https%3A%2F%2Fgithub.com%2Fimpress-org%2Fgivewp%2Fpull%2F8301&signature=bba1955bec79e4bde9ca4202123f257dbe49b41ab49b1f1998a665da22f9cf1c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed a fatal error that could occur when resuming paused Stripe subscriptions.
  - Improved recognition of enabled modern Stripe payment methods, including Stripe Payment Element, alongside legacy Stripe methods.
- **Tests**
  - Added coverage to verify active Stripe payment methods are detected correctly.
- **Documentation**
  - Added a changelog entry documenting the Stripe subscription fix.

### DIFF
--- a/includes/gateways/stripe/includes/admin/admin-helpers.php
+++ b/includes/gateways/stripe/includes/admin/admin-helpers.php
@@ -41,6 +41,7 @@ function give_stripe_supported_payment_methods()
 /**
  * This function is used to check whether a payment method supported by Stripe with Give is active or not.
  *
+ * @since TBD Also check the `gateways_v3` option so modern Stripe gateways are recognized.
  * @since 2.5.5
  *
  * @return bool
@@ -48,8 +49,16 @@ function give_stripe_supported_payment_methods()
 function give_stripe_is_any_payment_method_active()
 {
 	$settings             = give_get_settings();
-    $gateways = $settings['gateways'] ?? [];
     $stripePaymentMethods = give_stripe_supported_payment_methods();
+
+    // The modern Stripe gateways (e.g. Stripe Payment Element) are stored in the
+    // `gateways_v3` option, while the legacy gateways live in the `gateways`
+    // setting. Both must be checked, so a site using only a modern Stripe gateway
+    // is still reported as having an active Stripe payment method.
+    $gateways = array_merge(
+        $settings['gateways'] ?? [],
+        (array) give_get_option('gateways_v3', [])
+    );
 
     $active = false;
     foreach (array_keys($gateways) as $gateway) {
--- a/includes/gateways/stripe/includes/admin/admin-helpers.php
+++ b/includes/gateways/stripe/includes/admin/admin-helpers.php
@@ -50,10 +50,12 @@ function give_stripe_is_any_payment_method_active()
 {
     $stripePaymentMethods = give_stripe_supported_payment_methods();
 
-    // The modern Stripe gateways (e.g. Stripe Payment Element) are stored in the
-    // `gateways_v3` option, while the legacy gateways live in the `gateways`
-    // setting. Both must be checked, so a site using only a modern Stripe gateway
-    // is still reported as having an active Stripe payment method.
+    /*
+     * The modern Stripe gateways (e.g. Stripe Payment Element) are stored in the
+     * `gateways_v3` option, while the legacy gateways live in the `gateways`
+     * setting. Both must be checked, so a site using only a modern Stripe gateway
+     * is still reported as having an active Stripe payment method.
+     */
     $gateways = array_merge(
         (array) give_get_option('gateways', []),
         (array) give_get_option('gateways_v3', [])
--- a/includes/gateways/stripe/includes/admin/admin-helpers.php
+++ b/includes/gateways/stripe/includes/admin/admin-helpers.php
@@ -48,7 +48,6 @@ function give_stripe_supported_payment_methods()
  */
 function give_stripe_is_any_payment_method_active()
 {
-	$settings             = give_get_settings();
     $stripePaymentMethods = give_stripe_supported_payment_methods();
 
     // The modern Stripe gateways (e.g. Stripe Payment Element) are stored in the
@@ -56,7 +55,7 @@ function give_stripe_is_any_payment_method_active()
     // setting. Both must be checked, so a site using only a modern Stripe gateway
     // is still reported as having an active Stripe payment method.
     $gateways = array_merge(
-        $settings['gateways'] ?? [],
+        (array) give_get_option('gateways', []),
         (array) give_get_option('gateways_v3', [])
     );
 
--- a/includes/gateways/stripe/includes/admin/admin-helpers.php
+++ b/includes/gateways/stripe/includes/admin/admin-helpers.php
@@ -48,6 +48,7 @@ function give_stripe_supported_payment_methods()
  */
 function give_stripe_is_any_payment_method_active()
 {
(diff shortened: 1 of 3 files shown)


WayneRocha:
@coderabbitai review

tests/includes/legacy/stripe/tests-give-stripe-admin-helpers.php
```diff
@@ -134,4 +134,38 @@ public function test_give_stripe_is_any_payment_method_active() {
...
+			static function ( $gateways ) {
+				if ( ! in_array( 'stripe_payment_element', $gateways, true ) ) {
+					$gateways[] = 'stripe_payment_element';
+				}
+
+				return $gateways;
+			}
+		);
+
+		$this->assertTrue( give_stripe_is_any_payment_method_active() );
```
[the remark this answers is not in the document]
WayneRocha: Addressed in a782d10a0a5d06bbe16f7b7138e831e319d232d1

includes/gateways/stripe/includes/admin/admin-helpers.php
```diff
@@ -41,16 +41,25 @@ function give_stripe_supported_payment_methods()
...
 	$settings             = give_get_settings();
-    $gateways = $settings['gateways'] ?? [];
     $stripePaymentMethods = give_stripe_supported_payment_methods();
 
+    // The modern Stripe gateways (e.g. Stripe Payment Element) are stored in the
+    // `gateways_v3` option, while the legacy gateways live in the `gateways`
+    // setting. Both must be checked, so a site using only a modern Stripe gateway
+    // is still reported as having an active Stripe payment method.
+    $gateways = array_merge(
+        $settings['gateways'] ?? [],
```
coutinhocouto, line 59: The two halves of the merge are read differently — the legacy one off the already-fetched `$settings` array, the v3 one through `give_get_option()`, which also runs the `give_get_option` filter. Only the v3 half gets the `(array)` guard.

Core's own version of this merge (`includes/gateways/functions.php:81-86`) uses `give_get_option()` with a cast on both sides:

```php
$gateways = array_merge(
    (array) give_get_option('gateways', []),
    (array) give_get_option('gateways_v3', [])
);
```

includes/gateways/stripe/includes/admin/admin-helpers.php
```diff
@@ -41,16 +41,25 @@ function give_stripe_supported_payment_methods()
...
 function give_stripe_is_any_payment_method_active()
 {
 	$settings             = give_get_settings();
-    $gateways = $settings['gateways'] ?? [];
     $stripePaymentMethods = give_stripe_supported_payment_methods();
 
+    // The modern Stripe gateways (e.g. Stripe Payment Element) are stored in the
+    // `gateways_v3` option, while the legacy gateways live in the `gateways`
+    // setting. Both must be checked, so a site using only a modern Stripe gateway
+    // is still reported as having an active Stripe payment method.
```
coutinhocouto, line 57:
```suggestion
    /**
    * The modern Stripe gateways (e.g. Stripe Payment Element) are stored in the
    * `gateways_v3` option, while the legacy gateways live in the `gateways`
    * setting. Both must be checked, so a site using only a modern Stripe gateway
    * is still reported as having an active Stripe payment method.
    */
```

tests/includes/legacy/stripe/tests-give-stripe-admin-helpers.php
```diff
@@ -134,4 +134,44 @@ public function test_give_stripe_is_any_payment_method_active() {
...
```
coutinhocouto, line 175: `LegacyStripeAdapter::addToStripeSupportedPaymentMethodsList()` already registers `stripe_payment_element` on `give_stripe_supported_payment_methods` during `LegacyServiceProvider::register()`. I checked on a booted site and the slug is in the list by the time add-ons load, so this closure never does anything.

The cost is that the test keeps passing if the adapter ever stops registering it. Asserting the dependency instead of stubbing it covers more:

```php
$this->assertContains( 'stripe_payment_element', give_stripe_supported_payment_methods() );
```

coutinhocouto (changes requested):

includes/gateways/stripe/includes/admin/admin-helpers.php
WayneRocha: Addressed in b9869b48910e49af533c7ecbd9caf68b6c80f7a2
Edit: AND changed again in ee111d01327a3a0d03ea6c94f9f47e039f8dfc9c for tests to pass

includes/gateways/stripe/includes/admin/admin-helpers.php
WayneRocha: Addressed in https://github.com/impress-org/givewp/commit/4d75af34bca95feb3b63e187c32a2883c5e8b0f3

tests/includes/legacy/stripe/tests-give-stripe-admin-helpers.php
WayneRocha: Addressed in 51888ffd4282878b68b2fd62c6dbd7bfac0324c0

WayneRocha:
@coderabbitai review

coutinhocouto (approved):

PR #8311, opened 2026-09-01
Feature: external donation form embed script and web component

jonwaldstein:
## Description

Stacked on #8310.

Adds `build/donationFormExternalEmbed.js` — a self-contained bundle (no WordPress packages, iframe-resizer v4 inlined) that registers the `<givewp-donation-form>` custom element for embedding v3 donation forms on non-WordPress sites:

```html
<script src="https://example.org/give/embed/donation-form/script.js" defer></script>
<givewp-donation-form form-id="42"></givewp-donation-form>
```

### Script URL

- New `Route::script($uri, $file)` and `Route::scriptUrl($uri)` on the router serve a built script (any absolute path, so add-ons can use it too) from a URL the plugin controls, so the build path can change without breaking snippets already pasted on other sites. Matching happens on `parse_request` against the path WordPress already resolved, so no rewrite rule is registered and nothing needs flushing. URL shape follows the permalink setting: `/give/embed/donation-form/script.js`, `/index.php/give/embed/donation-form/script.js`, or `/?givewp-route=embed/donation-form/script.js`.
- `ScriptResponse` sends the file with an `ETag` equal to the build hash from the `.asset.php` file (same value `wp_enqueue_script` uses as `ver`), `Cache-Control: public, max-age=3600`, and `304` on `If-None-Match` (tolerates `W/` and the `-gzip` suffix mod_deflate appends).
- The element locates the WordPress site from `document.currentScript.src`, so the snippet no longer needs `wp-url`; it remains as an override. The path has no per-form segment on purpose (one script per page regardless of form count, static response, one cached copy per site); a variant would be a query param folded into the ETag.
- `ScriptResponse::localize()` prints a data object ahead of the file, the way `wp_localize_script()` does for enqueued scripts. `GetExternalEmbedScriptData` supplies the five default labels (`Donate`, `Loading`, `Donation Form`, `Open donation form`, `Close`) translated in the site's locale at request time; the element reads them from `window.givewpDonationFormEmbed`, and the data hash joins the ETag. Verified with the German language pack on wp-env: `Spenden`, `Wird geladen`, `Spendenformular`, `Schließen`.
- Documented in `docs/architecture/external-embeds.md`.

### Element

- iframe rendering with a loading state; reveal happens on the iframe-resizer handshake (`onInit`), since browsers fire `load` even for error pages
- 10s timeout degrades to a plain "Open donation form" link (frame-blocking headers, ad blockers, network failure)
- Offsite-gateway return handling: detects the existing RouteListener args and swaps the iframe to the receipt view
- Strict `postMessage` validation (exact origin, type allowlist, source check, http/https URL check) for the `givewp-navigate` channel
- Login degradation: cross-origin embeds use the inline login form instead of the wp-login redirect flow, whose `redirect_to` rejects external hosts; lost-password opens in a new tab

## Testing
Covered end-to-end by `tests/e2e/external-embeds.spec.ts` in this PR, which loads the script through the new route and relies on the derived home URL (no `wp-url`): cross-origin render, guest donation to receipt, offsite return with cleaned URL params, modal display style, fallback link. The spec serves a fictitious external origin so the iframe crosses origins for real.

Linear: SOFT-4291, SOFT-4292, SOFT-4293, SOFT-4295

## Summary by CodeRabbit

* **New Features**
  * Added support for embedding donation forms on external websites using copy-and-paste snippets.
  * Embedded forms can display inline, in a modal, or in a new browser tab.
  * Donors can log in directly within supported embedded forms.
  * Added localized donor-facing text and payment gateway return handling.
  * Modal forms support keyboard focus management, including Escape-to-close.
  * Forms provide fallback links when they cannot load.
* **Bug Fixes**
  * Improved login and password-reset handling for cross-site embedded forms.


bordoni:
Ideally, we should create a URL that we control for:
`https://example.org/wp-content/plugins/give/build/externalFormEmbed.js` like `https://example.org/give/embed-script.js` 

Which would pass by the server to deliver the JS. Cache heavily, but allow dynamic data if needed.

Potentially allowing: `https://example.org/give/<form-id>/embed-script.js`

src/DonationForms/resources/externalEmbed/index.ts
```diff
@@ -0,0 +1,463 @@
...
+            }
+
+            if (event.key === 'Escape') {
+                hide();
+                return;
+            }
+
+            // Keep Tab inside the dialog. The iframe manages its own inner
+            // focus order; this only stops focus escaping to the host page.
+            if (event.key === 'Tab' && !dialog.contains(document.activeElement)) {
```
glaubersilva: I didn't test it, but AI found a potential issue here, so I'll share the discovery for your own disclosure about how to proceed.

Reproduction: open the modal, Tab to the last focusable element inside the iframe (or Shift+Tab from the close button), then press Tab/Shift+Tab once more. Focus lands on the page behind the overlay for that one keystroke, even though `aria-modal="true"` says it can't.

Why: this handler only checks `!dialog.contains(document.activeElement)` on `keydown`, but the browser moves focus as Tab's default action, which runs after keydown listeners fire. So the check always sees the pre-move state — it can only catch an escape that already happened, one event too late.

Fix: split Escape (stays on `keydown`) from the focus trap (moves to `focusin`, which fires the instant focus lands — no lag). Four edits in this file:

```
// 1. field declaration, alongside keydownHandler
focusinHandler: ((event: FocusEvent) => void) | null = null;

// 2. connectedCallback's re-attach block
if (this.keydownHandler) {
    document.addEventListener('keydown', this.keydownHandler);
}
if (this.focusinHandler) {
    document.addEventListener('focusin', this.focusinHandler);
}

// 3. disconnectedCallback
if (this.keydownHandler) {
    document.removeEventListener('keydown', this.keydownHandler);
}
if (this.focusinHandler) {
    document.removeEventListener('focusin', this.focusinHandler);
}

// 4. openModal, replacing the current keydownHandler block
this.keydownHandler = (event: KeyboardEvent) => {
    if (overlay.style.display === 'none') {
        return;
    }

    if (event.key === 'Escape') {
        hide();
    }
};
document.addEventListener('keydown', this.keydownHandler);

// The iframe manages its own inner focus order; this only pulls focus
// back once it lands outside the dialog (the iframe counts as inside,
// since the host document sees it as one focusable element).
this.focusinHandler = (event: FocusEvent) => {
    if (overlay.style.display === 'none') {
        return;
    }

    if (!dialog.contains(document.activeElement)) {
        dialog.focus();
    }
};
document.addEventListener('focusin', this.focusinHandler);

```

glaubersilva (commented):
@jonwaldstein I found, with the help of AI, one potential issue for your evaluation.

src/DonationForms/resources/externalEmbed/index.ts
jonwaldstein: @glaubersilva thanks, should be resolved now - claude response:

Confirmed and fixed in 50e194a1b, thanks. The keydown check ran before the browser moved focus, and while focus is inside the cross-origin iframe the host document never sees the Tab at all, so it was always one step late.

I went with focus guards rather than a `focusin` listener: a hidden focusable element before the close button and one after the iframe. Tabbing past either end lands on a guard, still inside the dialog, which hands focus to the opposite end. Focus never touches the host page, and it needs no document-level listener. The keydown handler now only handles Escape.

The modal e2e test now presses Shift+Tab from the close button and Tabs 30 times forward from the iframe, asserting focus stays inside the dialog each time. It fails on the old code and passes on the new.

glaubersilva (approved):

jonwaldstein:
> Ideally, we should create a URL that we control for: `https://example.org/wp-content/plugins/give/build/externalFormEmbed.js` like `https://example.org/give/embed-script.js`
> 
> Which would pass by the server to deliver the JS. Cache heavily, but allow dynamic data if needed.
> 
> Potentially allowing: `https://example.org/give/<form-id>/embed-script.js`

@bordoni I don't think a pretty url like `https://example.org/give/embed-script.js` would be reliable in WordPress land as you would need to rely on pretty permalinks being enabled on the customer website.  If the customer changes permalinks for their site it could risk breaking our script.

The only reliable path for something like this is going to be a query string like `https://example.org/?givewp-route=embed-script` (which we use alot in give) but even that would require a WordPress boot per uncached fetch from third-party pages, puts ETag and Cache-Control on us, and means generating JS server-side.  It technically has the benefit of using query vars but with the web component approach we can use whatever dynamic args we want and matches how our shortcodes & blocks work.

The static script `https://example.org/wp-content/plugins/give/build/externalFormEmbed.js` approach here is the safer option and how all our scripts are delivered - similar to how stripe [embeds a buy button](https://docs.stripe.com/payment-links/buy-button). we have the static script that does the embedding and the `<web-component />` to be placed for the rendering.

src/DonationForms/resources/externalEmbed/index.ts
```diff
@@ -0,0 +1,479 @@
...
+
+        /*
+         * Focus guards bracket the dialog's contents. Tabbing past either end
+         * lands on a guard, still inside the dialog, which hands focus to the
+         * opposite end. A key listener cannot do this: the browser moves
+         * focus after keydown, and while focus is inside the cross-origin
+         * iframe the host document sees no key events at all.
+         */
+        const focusables = () => Array.from(dialog.querySelectorAll<HTMLElement>('a[href], button, iframe'));
+        const startGuard = this.createFocusGuard(() => focusables().pop()?.focus());
```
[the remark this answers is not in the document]
jonwaldstein: Valid, fixed in 5ff77d7ce. `focusDialog` now focuses the close button, falling back to the dialog container only if the button is missing. Both open paths go through it, so reopening an existing overlay gets the same behavior. The e2e test now presses Shift+Tab immediately after the modal opens, before any explicit focus, and asserts focus stays inside the dialog. That assertion fails on the previous build and passes on this one.

jonwaldstein:
Call summary @jonwaldstein and @bordoni 
We'll try a dynamic url like `example.org/give/embed/donation-form/script.js` that will use our js file under the hood but gives us a path with more flexibility & control (in case the js path changes in the future).  We just need to account for pretty permalinks not being set in WP settings with proper fallbacks.

jonwaldstein:
@bordoni this was updated to support dynamic urls.

DEMO
https://www.loom.com/share/7af847a4d6244109bb72089b0cb86cef

src/DonationForms/resources/externalEmbed/index.ts
```diff
@@ -0,0 +1,524 @@
...
+
+    disconnectedCallback() {
+        window.removeEventListener('message', this.messageHandler);
+        if (this.keydownHandler) {
+            document.removeEventListener('keydown', this.keydownHandler);
+        }
+    }
+
+    getButtonText(): string {
+        return this.getAttribute('button-text') || 'Donate';
```
glaubersilva: We need to make it translatable with the `_()` function.

src/DonationForms/resources/externalEmbed/index.ts
```diff
@@ -0,0 +1,524 @@
...
+        );
+    }
+
+    renderFallbackLink(loading: HTMLElement) {
+        const link = document.createElement('a');
+        link.href = this.getStandaloneFormUrl();
+        link.target = '_blank';
+        link.rel = 'noopener';
+        link.className = 'givewp-embed__button';
+        link.textContent = this.getAttribute('fallback-text') || 'Open donation form';
```
glaubersilva: We need to make it translatable with the `_()` function.

src/DonationForms/resources/externalEmbed/index.ts
```diff
@@ -0,0 +1,524 @@
...
+        }
+
+        const overlay = document.createElement('div');
+        overlay.className = 'givewp-embed__overlay';
+
+        const dialog = document.createElement('div');
+        dialog.className = 'givewp-embed__dialog';
+        dialog.setAttribute('role', 'dialog');
+        dialog.setAttribute('aria-modal', 'true');
+        dialog.setAttribute('aria-label', this.getAttribute('form-title') || 'Donation Form');
```
glaubersilva: We need to make it translatable with the `_()` function.

src/DonationForms/resources/externalEmbed/index.ts
```diff
@@ -0,0 +1,524 @@
...
+        dialog.tabIndex = -1;
+
+        const close = document.createElement('button');
+        close.type = 'button';
+        close.className = 'givewp-embed__close';
+        close.setAttribute('aria-label', this.getAttribute('close-text') || 'Close');
```
glaubersilva: We need to make it translatable with the `_()` function.

src/DonationForms/resources/externalEmbed/index.ts
```diff
@@ -0,0 +1,524 @@
...
+        url.searchParams.set('receipt-id', params.get('givewp-receipt-id'));
+
+        return url.toString();
+    }
+
+    renderForm(src: string, target: HTMLElement) {
+        const loading = document.createElement('div');
+        loading.className = 'givewp-embed__loading';
+        loading.setAttribute('role', 'status');
+        loading.setAttribute('aria-label', this.getAttribute('loading-text') || 'Loading');
```
glaubersilva: We need to make it translatable with the `_()` function.

src/DonationForms/resources/externalEmbed/index.ts
```diff
@@ -0,0 +1,524 @@
...
+
+        const spinner = document.createElement('span');
+        spinner.className = 'givewp-embed__spinner';
+        loading.appendChild(spinner);
+
+        const iframe = document.createElement('iframe');
+        iframe.src = src;
+        // Matches the title the WordPress embeds use, so tooling and donors see one name.
+        iframe.title = this.getAttribute('form-title') || 'Donation Form';
```
glaubersilva: We need to make it translatable with the `_()` function.

glaubersilva (commented):
@jonwaldstein I left a few comments with some translation inconsistencies that I found in `externalEmbed/index.ts`

Given that, it might be worth also double-checking `docs/architecture/external-embeds.md` while we're at it cause I asked AI to double-check it to find inconsistencies (after I notice the missing translations), and it pointed out two things:

- line 95, _"the route supplies the translated defaults"_ — is this still true? Don't see locale reaching `script.js`.
- line 98, mentions reading a locale param through `getRequestDataByType() `— does `script()` actually use that? Thought that was only `get()`/`post()`.

src/DonationForms/resources/externalEmbed/index.ts
jonwaldstein: Resolved f7c9b453c

src/DonationForms/resources/externalEmbed/index.ts
jonwaldstein: Resolved f7c9b453c

src/DonationForms/resources/externalEmbed/index.ts
jonwaldstein: Resolved f7c9b453c

src/DonationForms/resources/externalEmbed/index.ts
jonwaldstein: Resolved f7c9b453c

src/DonationForms/resources/externalEmbed/index.ts
jonwaldstein: Resolved f7c9b453c

src/DonationForms/resources/externalEmbed/index.ts
jonwaldstein: Resolved f7c9b453c

jonwaldstein:
@glaubersilva thanks, both good catches. I had purposely moved those strings to be from the window but it got dropped during rebasing - but its updated now.

Claude response:

On the docs:
- Line 95 is true with that commit restored. No locale param is needed; the site locale applies at request time.
- Line 98 was forward-looking guidance, not a description of current behavior. `script()` reads no request data today, and the paragraph now says so.

One thing the docs also now spell out: the element's `locale` attribute only reaches the iframe URL, so it changes the form inside the frame, not the launcher/dialog labels. Those follow the site locale unless the text attributes override them.

PR #8312, opened 2026-09-01
Feature: external embed snippet in the form builder embed flow

jonwaldstein:
## Description

Stacked on #8311.

Adds an **External website** tab to the form builder's existing Embed Form modal: display style, button label, and button color controls, the generated snippet shown in a read-only code block, and a one-click copy. Warns when the form contains a required login block (login inside embedded forms is unreliable in some browsers) or has Confirmation Page Redirect enabled.

The snippet carries only per-embed choices: `form-id`, `form-title` (the form's own title, used for the iframe title and modal label), `display-style`, a `button-text` when the admin typed one, and `primary-color` for the launcher button. The color field is seeded with the form's resolved primary color (campaign inheritance included) and labeled as living on the other website, since the launcher renders before any WordPress response and cannot follow later design changes; the form inside the iframe always uses its current design. There is no `secondary-color`, nothing in the embed read it. The script URL comes from `GenerateExternalEmbedScriptUrl` (#8311), the site location is derived from it on the client, and the translated default labels travel with the script, so there is no `wp-url` and no per-language text in the snippet.

```html
<script src="https://example.org/give/embed/donation-form/script.js" defer></script>
<givewp-donation-form form-id="42" form-title="Spring Appeal" primary-color="#0b72d9" display-style="modal"></givewp-donation-form>
```

Localizes configuration only — home URL and embed script URL, no donor data.

Linear: SOFT-4294

## Summary by CodeRabbit

- **New Features**
  - Added support for embedding donation forms on external websites.
  - Added a tabbed interface with a ready-to-copy `<givewp-donation-form>` snippet.
  - Embed snippets include form settings, translated text, display options, campaign colors, and modal controls.
  - Added warnings when donor login or confirmation-page redirects may affect embedded forms.

- **Improvements**
  - Improved clipboard copying with success feedback after the snippet is copied.
  - Made the embed window scrollable with a sticky header for easier navigation.

### DIFF
--- a/src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
@@ -290,12 +290,31 @@ export default function EmbedFormModal({handleClose}: EmbedFormModalProps) {
     }
 
     /**
+     * The snippet carries the selected display style and the translated
+     * donor-facing strings, since the standalone embed script has no access
+     * to WordPress translations.
+     *
      * @since TBD
      */
     const getExternalEmbedSnippet = () => {
+        const attributes = [
+            `form-id="${formId}"`,
+            `wp-url="${homeUrl}"`,
+            `fallback-text="${__('Open donation form', 'give')}"`,
+        ];
+
+        if (isButton) {
+            attributes.push(`display-style="${state.selectedStyle}"`);
+            attributes.push(`button-text="${state.openFormButton || __('Donate', 'give')}"`);
+        }
+
+        if (state.selectedStyle === 'modal') {
+            attributes.push(`close-text="${__('Close', 'give')}"`);
+        }
+
         return [
             `<script src="${externalEmbedScriptUrl}" defer></script>`,
-            `<givewp-donation-form form-id="${formId}" wp-url="${homeUrl}"></givewp-donation-form>`,
+            `<givewp-donation-form ${attributes.join(' ')}></givewp-donation-form>`,
         ].join('\n');
     };
 
--- a/src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
@@ -47,8 +47,29 @@ interface StateProps {
  */
 export default function EmbedFormModal({handleClose}: EmbedFormModalProps) {
 
-    const {formId, homeUrl, externalEmbedScriptUrl, blockData} = getWindowData();
+    const {formId, homeUrl, externalEmbedScriptUrl, blockData, settings, campaignColors} = getWindowData();
     const [isExternalEmbedCopied, setIsExternalEmbedCopied] = useState<boolean>(false);
+
+    /**
+     * The external embed script cannot read form settings, so the snippet
+     * carries the form's colors as attributes. Campaign colors win when the
+     * form inherits them, matching how the form itself resolves colors.
+     *
+     * @since TBD
+     */
+    const getEmbedColors = (): {primary: string; secondary: string} => {
+        try {
+            const parsedSettings = JSON.parse(settings);
+            const inherit = parsedSettings.inheritCampaignColors;
+
+            return {
+                primary: (inherit && campaignColors?.primaryColor) || parsedSettings.primaryColor || '',
+                secondary: (inherit && campaignColors?.secondaryColor) || parsedSettings.secondaryColor || '',
+            };
+        } catch (error) {
+            return {primary: '', secondary: ''};
+        }
+    };
 
     /**
      * Login inside a cross-origin embed is unreliable in some browsers, so
@@ -302,6 +323,14 @@ export default function EmbedFormModal({handleClose}: EmbedFormModalProps) {
             `wp-url="${homeUrl}"`,
             `fallback-text="${__('Open donation form', 'give')}"`,
         ];
+
+        const colors = getEmbedColors();
+        if (colors.primary) {
+            attributes.push(`primary-color="${colors.primary}"`);
+        }
+        if (colors.secondary) {
+            attributes.push(`secondary-color="${colors.secondary}"`);
+        }
 
         if (isButton) {
             attributes.push(`display-style="${state.selectedStyle}"`);
--- a/src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
@@ -4,7 +4,7 @@ import {createPortal} from 'react-dom';
 import {useDispatch, useSelect} from '@wordpress/data';
 import {store} from '@wordpress/core-data';
 import {__, sprintf} from '@wordpress/i18n';
-import {Button, RadioControl, SelectControl, Spinner, TextControl} from '@wordpress/components';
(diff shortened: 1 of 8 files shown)


jonwaldstein:
## Fixes Applied Successfully

Fixed 1 file(s) based on 2 CodeRabbit feedback item(s).

**Files modified:**
- `src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx`

**Commit:** `9b720fd09`

Attribute values in the generated snippet are HTML-encoded, the copy buttons only report success when the clipboard write succeeded, and the fallback textarea is hidden off-screen with valid CSS. The latest changes are on `feature/external-embeds-3-builder-ui`; #8320 was restacked on top.

src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
```diff
@@ -251,6 +322,53 @@ export default function EmbedFormModal({handleClose}: EmbedFormModalProps) {
...
+        }
+
+        if (state.selectedStyle === 'modal') {
+            attributes.push(`close-text="${__('Close', 'give')}"`);
+        }
+
+        return [
+            `<script src="${externalEmbedScriptUrl}" defer></script>`,
+            `<givewp-donation-form ${attributes.join(' ')}></givewp-donation-form>`,
+        ].join('\n');
```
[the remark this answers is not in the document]
jonwaldstein: Fixed in 9b720fd09. Every interpolated attribute in the snippet, including the script URL, goes through an `attribute()` helper that encodes `&`, `"`, `<`, and `>`.

src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
```diff
@@ -212,28 +269,42 @@ export default function EmbedFormModal({handleClose}: EmbedFormModalProps) {
...
+        textArea.focus();
+        textArea.select();
+        try {
+            document.execCommand('copy');
+        } catch (error) {
+            console.error(error);
+        } finally {
+            textArea.remove();
+        }
+    };
```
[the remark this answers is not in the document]
jonwaldstein: Fixed in 9b720fd09. `writeToClipboard` returns a boolean, both copy handlers set the copied state only on `true`, and the fallback textarea is positioned off-screen with `opacity: 0` instead of the invalid `display: hidden`.

src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
```diff
@@ -47,7 +47,64 @@ interface StateProps {
...
-    const {formId} = getWindowData();
+    const {formId, homeUrl, externalEmbedScriptUrl, blockData, settings, campaignColors} = getWindowData();
+    const [isExternalEmbedCopied, setIsExternalEmbedCopied] = useState<boolean>(false);
+
+    const parsedSettings = (() => {
+        try {
+            return JSON.parse(settings);
+        } catch (error) {
+            return {};
+        }
```
bordoni, line 66: We should log this error somewhere

bordoni (changes requested):

src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
jonwaldstein: Resolved [50431e9](https://github.com/impress-org/givewp/pull/8312/commits/50431e95f44ad033e4e5a230d86c359bf9e602b0)

src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
```diff
@@ -47,7 +49,68 @@ interface StateProps {
  */
 export default function EmbedFormModal({handleClose}: EmbedFormModalProps) {
 
-    const {formId} = getWindowData();
+    const {formId, homeUrl, externalEmbedScriptUrl, blockData, settings, campaignColors} = getWindowData();
+    const [isExternalEmbedCopied, setIsExternalEmbedCopied] = useState<boolean>(false);
+
+    const parsedSettings = useMemo((): Partial<FormSettings> => {
```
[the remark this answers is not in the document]
jonwaldstein: Not reachable. `settings` is always `FormSettings::toJson()`, which is `json_encode` of an array merged with fixed keys, so the parsed value is always an object. An undefined or malformed string throws inside the `try` and falls back to `{}`. Skipping.

src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
```diff
@@ -355,6 +448,78 @@ export default function EmbedFormModal({handleClose}: EmbedFormModalProps) {
...
+                                    {__('This form has "Confirmation Page Redirect" enabled in its settings, so donors will leave the site the form is embedded on after donating.', 'give')}
+                                </div>
+                            )}
+
+                            <div className="give-embed-modal-items give-embed-modal-copy">
+                                <div>
+                                    <Button
+                                        icon={isExternalEmbedCopied ? CheckIcon : CopyIcon}
+                                        variant="secondary"
+                                        ref={copyExternalEmbedRef as Ref<HTMLAnchorElement>}
```
glaubersilva: This renders a plain `<button>` here (no `href`), so `Ref<HTMLAnchorElement>` names the wrong element type. Cast to `Ref<HTMLButtonElement>` instead, or just drop the cast — the hook's inferred `Ref<HTMLElement>` already satisfies `Button`'s `any`-typed `ref` prop either way.

src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
```diff
@@ -576,7 +741,7 @@ export default function EmbedFormModal({handleClose}: EmbedFormModalProps) {
                         <Button
                             icon={state.isCopied ? CheckIcon : CopyIcon}
                             variant="secondary"
-                            onClick={handleCopy}
+                            ref={copyShortcodeRef as Ref<HTMLAnchorElement>}
```
glaubersilva: This renders a plain `<button>` here (no `href`), so `Ref<HTMLAnchorElement>` names the wrong element type. Cast to `Ref<HTMLButtonElement>` instead, or just drop the cast — the hook's inferred `Ref<HTMLElement>` already satisfies `Button`'s `any`-typed `ref` prop either way.

glaubersilva (approved):
@jonwaldstein I found two minor things that are not blockers, but you can change them if found it valid. So I'm approving it to move things.

src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
jonwaldstein: Right, the cast was copied from `TemplateTags` without thinking about the element. Dropped both casts and the `Ref` import in 1ffb8b772; the hook's inferred type is enough for `Button`.

src/FormBuilder/resources/js/form-builder/src/components/EmbedForm/index.tsx
jonwaldstein: Right, the cast was copied from `TemplateTags` without thinking about the element. Dropped both casts and the `Ref` import in 1ffb8b772; the hook's inferred type is enough for `Button`.

PR #8313, opened 2026-09-01
Tests: cover external donation form embeds with end-to-end tests

jonwaldstein:
## Description

Stacked on #8312.

Playwright coverage for the external embed path. There is no second server: navigation to a fictitious external origin is fulfilled with fixture HTML that loads the real embed script, so the iframe crosses origins for real.

- renders the form cross-origin (regression gate for the #8310 guards)
- takes a guest donation through to the receipt (Test Gateway)
- swaps to the receipt view when an offsite gateway returns
- degrades to a link when the form cannot load

Chrome's Local Network Access checks are disabled in this spec only — a test-environment artifact of wp-env resolving to 127.0.0.1.

Linear: SOFT-4295

### DIFF
--- a/tests/e2e/external-embeds.spec.ts
+++ b/tests/e2e/external-embeds.spec.ts
@@ -8,6 +8,7 @@ import {
     waitForForm,
 } from './utils/donation-form';
 import {WP_BASE_URL} from './environment';
+import path from 'path';
 
 /**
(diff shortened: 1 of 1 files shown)


jonwaldstein:
Folded into #8311: the spec only exercises the guards and the embed script, so it ships with the embed script PR.

PR #8320, opened 2026-09-04
Enhancement: sign donors in inside externally embedded forms without cookies

jonwaldstein:
## Description

Stacked on #8312.

A form embedded on another website runs in a cross-site iframe, where every current browser drops the WordPress login cookie set by the authentication route (Chrome and Firefox under Lax-by-default, Safari under ITP). Verified in Chromium against wp-env: the login route returned 200 with five `Set-Cookie` headers and the browser stored none. The login appeared to succeed, then a required login block failed server validation with `Login required.` on a field the form had already hidden, so the donor was stuck on step one with no error.

### What changes

- **`AuthenticationRoute`** also returns an `authToken` built with `wp_generate_auth_cookie` under a plugin-specific salt scheme. WordPress signs it, expires it, and backs it with a `WP_Session_Tokens` session, so "log out everywhere" and password changes revoke it. Because the scheme is not `logged_in` or `auth`, the token is not a usable browser cookie if it leaks, and a real login cookie is not a usable token. No new crypto.
- **`AuthenticateFormRequestWithToken`** validates that token with `wp_validate_auth_cookie` and calls `wp_set_current_user`. It runs from the `donate` and `validate` routes only, so the token is never a site-wide credential. It is invoked from the routes rather than on `determine_current_user` because other plugins can resolve the current user before this plugin loads, which was observed locally.
- **Client** keeps the token on `window.givewp.form` (the login template and the form app are separate bundles) and the request builders append it to validate and donate requests. It stays out of form values because the client-side Joi schema rejects unknown keys and blocks submit silently.
- **Form builder** drops the "login is unreliable inside embeds" warning.

Same-origin embeds are unchanged: the cookie is present, so the token path is skipped.

## Testing

- PHPUnit: valid token signs in, logged-in user kept, missing/array token ignored, expired token rejected, tampered token rejected, revoked session rejected, token not valid as any login cookie, login cookie not valid as a token.
- E2E (`external-embeds.spec.ts`): a form with a required login block, embedded cross-origin, signs in with the wp-env admin and completes a donation through to the receipt.
- Manual: required-login embed signs in and donates in Safari.
- Full PHPUnit suite green locally. Same-origin donation form E2E green apart from the offsite Test Gateway case, which is not enabled on my local site and fails before any changed code runs.

Linear: SOFT-4293

## Summary by CodeRabbit

* **New Features**
  * Added secure authentication tokens for embedded donation forms.
  * Embedded forms preserve donor authentication across validation and submission.
  * Successful sign-in stores the authentication token automatically.
  * Added internal and external embed options with customizable snippets and settings.

* **Bug Fixes**
  * Improved external-embed donation flows requiring login.
  * Improved embed-code copying reliability and success feedback.

* **Changes**
  * Removed the external-embed warning about required donor login.

### DIFF
--- a/src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
+++ b/src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace Give\DonationForms\Actions;
+
+/**
+ * Signs the donor in for a donate or validate request from a signed auth
+ * token instead of the login cookie.
+ *
+ * A donation form embedded on another website cannot rely on cookies: the
+ * browser drops WordPress auth cookies set from a cross-site iframe response.
+ * The authentication route therefore also returns the auth cookie value, which
+ * WordPress already signs, expires, and backs with a session token. The form
+ * sends it back with the donation and the route validates it the same way core
+ * validates the cookie.
+ *
+ * This runs from the two form routes rather than on determine_current_user so
+ * it works even when another plugin resolves the current user before this
+ * plugin has loaded, and so the token is never accepted anywhere else.
+ *
+ * @since TBD
+ */
+class AuthenticateFormRequestWithToken
+{
+    const TOKEN_KEY = 'authToken';
+
+    /**
+     * @since TBD
+     */
+    public function __invoke(array $request): void
+    {
+        if (is_user_logged_in()) {
+            return;
+        }
+
+        $token = $request[self::TOKEN_KEY] ?? '';
+
+        if (!is_string($token) || $token === '') {
+            return;
+        }
+
+        $userId = wp_validate_auth_cookie($token, 'logged_in');
+
+        if ($userId) {
+            wp_set_current_user($userId);
+        }
+    }
+}
--- a/src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
+++ b/src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
@@ -46,6 +46,17 @@ class AuthenticateFormRequestWithToken
             return;
         }
 
+        /*
+         * Core extends the expiry by an hour for POST requests, which is meant
+         * for a form that sat open in a browser. This token is a short-lived
+         * credential, so its own expiry is the limit.
+         */
+        $parts = wp_parse_auth_cookie($token, self::SCHEME);
+
+        if (!$parts || (int)$parts['expiration'] < time()) {
+            return;
+        }
+
         $userId = wp_validate_auth_cookie($token, self::SCHEME);
 
         if ($userId) {
--- a/src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
+++ b/src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
@@ -8,10 +8,11 @@ namespace Give\DonationForms\Actions;
  *
  * A donation form embedded on another website cannot rely on cookies: the
  * browser drops WordPress auth cookies set from a cross-site iframe response.
- * The authentication route therefore also returns the auth cookie value, which
- * WordPress already signs, expires, and backs with a session token. The form
- * sends it back with the donation and the route validates it the same way core
- * validates the cookie.
+ * The authentication route therefore also returns a token built with the same
+ * core functions as the auth cookie, which WordPress signs, expires, and backs
+ * with a session token, under a plugin-specific scheme. The form sends it back
+ * with the donation and the route validates it the same way core validates
+ * the cookie.
  *
  * This runs from the two form routes rather than on determine_current_user so
  * it works even when another plugin resolves the current user before this
@@ -22,6 +23,13 @@ namespace Give\DonationForms\Actions;
 class AuthenticateFormRequestWithToken
 {
     const TOKEN_KEY = 'authToken';
+
+    /**
+     * A plugin-specific salt scheme. WordPress derives the salt from the scheme
+     * name, so the token verifies only here and is never a valid login cookie
+     * if it leaks.
+     */
+    const SCHEME = 'givewp_embedded_form';
 
     /**
      * @since TBD
@@ -38,7 +46,7 @@ class AuthenticateFormRequestWithToken
             return;
         }
 
-        $userId = wp_validate_auth_cookie($token, 'logged_in');
+        $userId = wp_validate_auth_cookie($token, self::SCHEME);
 
         if ($userId) {
             wp_set_current_user($userId);
--- a/src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
+++ b/src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
@@ -22,14 +22,22 @@ namespace Give\DonationForms\Actions;
  */
 class AuthenticateFormRequestWithToken
 {
-    const TOKEN_KEY = 'authToken';
+    /**
+     * The request key the form sends the token under. The authentication
+     * route returns it under the same key.
+     *
+     * @since TBD
+     */
+    public const TOKEN_KEY = 'authToken';
 
     /**
      * A plugin-specific salt scheme. WordPress derives the salt from the scheme
      * name, so the token verifies only here and is never a valid login cookie
      * if it leaks.
+     *
+     * @since TBD
      */
-    const SCHEME = 'givewp_embedded_form';
+    public const SCHEME = 'givewp_embedded_form';
 
     /**
      * @since TBD
--- a/src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
+++ b/src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
@@ -59,7 +59,7 @@ class AuthenticateFormRequestWithToken
          * for a form that sat open in a browser. This token is a short-lived
          * credential, so its own expiry is the limit.
          */
-        $parts = wp_parse_auth_cookie($token, self::SCHEME);
+        $parts = wp_parse_auth_cookie($token);
 
         if (!$parts || (int)$parts['expiration'] < time()) {
             return;
--- a/src/DonationForms/Routes/DonateRoute.php
+++ b/src/DonationForms/Routes/DonateRoute.php
@@ -2,8 +2,8 @@
 
 namespace Give\DonationForms\Routes;
 
-
 use Exception;
+use Give\DonationForms\Actions\AuthenticateFormRequestWithToken;
 use Give\DonationForms\Controllers\DonateController;
 use Give\DonationForms\DataTransferObjects\DonateControllerData;
 use Give\DonationForms\DataTransferObjects\DonateFormRouteData;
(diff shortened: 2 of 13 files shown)


jonwaldstein:
## Fixes Applied Successfully

Fixed 2 file(s) based on 1 CodeRabbit feedback item(s).

**Files modified:**
- `src/DonationForms/Actions/AuthenticateFormRequestWithToken.php`
- `tests/Unit/DonationForms/Actions/AuthenticateFormRequestWithTokenTest.php`

**Commit:** `0d399cea5`

The token's embedded expiration is checked before `wp_validate_auth_cookie`, so the one-hour POST grace period no longer applies. The expiry test now runs as a POST with a token expired by five minutes.

src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
```diff
@@ -0,0 +1,55 @@
...
+            return;
+        }
+
+        $token = $request[self::TOKEN_KEY] ?? '';
+
+        if (!is_string($token) || $token === '') {
+            return;
+        }
+
+        $userId = wp_validate_auth_cookie($token, self::SCHEME);
```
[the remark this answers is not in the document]
jonwaldstein: Fixed in 0d399cea5. The token's embedded expiration is now checked with `wp_parse_auth_cookie` before `wp_validate_auth_cookie` runs, so the POST grace period no longer extends it. `testExpiredTokenIsRejected` now runs as a POST with a token expired by five minutes.

src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * This runs from the two form routes rather than on determine_current_user so
+ * it works even when another plugin resolves the current user before this
+ * plugin has loaded, and so the token is never accepted anywhere else.
+ *
+ * @since TBD
+ */
+class AuthenticateFormRequestWithToken
+{
+    const TOKEN_KEY = 'authToken';
```
bordoni: Missng docblock.

src/DonationForms/resources/app/utilities/authToken.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @since TBD
+ */
+export function setAuthToken(token: string): void {
+    window.givewp.form.authToken = token ?? '';
+}
+
+export function getAuthToken(): string {
+    return window.givewp?.form?.authToken ?? '';
+}
```
bordoni, line 24: Missing docblock

src/DonationForms/Routes/DonateRoute.php
```diff
@@ -51,6 +52,8 @@ public function __invoke(array $request)
         // validate signature
         $routeData->validateSignature();
 
+        (new AuthenticateFormRequestWithToken())($request);
```
bordoni, line 55: Should this be store somewhere?

src/DonationForms/Routes/ValidationRoute.php
```diff
@@ -32,6 +33,8 @@ public function __invoke(array $request): bool
         // validate signature
         $routeData->validateSignature();
 
+        (new AuthenticateFormRequestWithToken())($request);
```
bordoni, line 36: Should this be stored here.

src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * A plugin-specific salt scheme. WordPress derives the salt from the scheme
+     * name, so the token verifies only here and is never a valid login cookie
+     * if it leaks.
+     */
+    const SCHEME = 'givewp_embedded_form';
```
bordoni: Missing `@since`

bordoni (changes requested):

src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
jonwaldstein: Resolved [e80153a](https://github.com/impress-org/givewp/pull/8320/commits/e80153afbaac22dff80dea260257fdc65fd618b6)

src/DonationForms/resources/app/utilities/authToken.ts
jonwaldstein: Resolved [e80153a](https://github.com/impress-org/givewp/pull/8320/commits/e80153afbaac22dff80dea260257fdc65fd618b6)

src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
jonwaldstein: Resolved [e80153a](https://github.com/impress-org/givewp/pull/8320/commits/e80153afbaac22dff80dea260257fdc65fd618b6)

src/DonationForms/Routes/DonateRoute.php
jonwaldstein: @bordoni there is nothing to store here

src/DonationForms/Routes/ValidationRoute.php
jonwaldstein: @bordoni there is nothing to store here - are you asking about something else?

src/DonationForms/resources/app/utilities/authToken.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export function setAuthToken(token: string): void {
+    window.givewp.form.authToken = token ?? '';
+}
+
+/**
+ * @since TBD
+ */
+export function getAuthToken(): string {
+    return window.givewp?.form?.authToken ?? '';
```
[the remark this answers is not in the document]
jonwaldstein: Good catch, fixed in 01aa71d09. Both routes now `unset` the token from `$request` right after `AuthenticateFormRequestWithToken` consumes it, so it never reaches the DTO or the failed request log.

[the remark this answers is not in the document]
jonwaldstein: Moved in the follow-up commit: `AuthenticateFormRequestWithToken` now returns the request with the token removed, and both routes reassign `$request` from it. Covered by `testTokenIsStrippedFromTheRequest`.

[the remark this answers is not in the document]
jonwaldstein: Correction: this was never reachable, so the strip is reverted in 102850984. `Log::serializeAndRedactContext` walks the context recursively and replaces any key containing `token` (case-insensitive, see `give_log_redaction_list`) with `[[redacted]]`, so `formData.requestData.authToken` is redacted before it is written. `AuthenticateFormRequestWithToken` is back to `void` and the routes call it as before.

src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
```diff
@@ -0,0 +1,74 @@
...
+        if (!is_string($token) || $token === '') {
+            return;
+        }
+
+        /*
+         * Core extends the expiry by an hour for POST requests, which is meant
+         * for a form that sat open in a browser. This token is a short-lived
+         * credential, so its own expiry is the limit.
+         */
+        $parts = wp_parse_auth_cookie($token, self::SCHEME);
```
glaubersilva: Passing `self::SCHEME` here reads as if it affects parsing, but core's `wp_parse_auth_cookie()` only uses its `$scheme` argument to pick a `$_COOKIE` name when `$cookie` is empty — when a cookie string is passed directly (as here), the argument is inert; the 4-part explode is scheme-agnostic. No functional bug (the real scheme check happens correctly in the following `wp_validate_auth_cookie() `call), just a naming/reading trap for the next person who touches this file.

glaubersilva (changes requested):
While re-checking this PR after the recent rebase, I found that `AuthenticateFormRequestWithToken::__invoke()` no longer strips the `authToken` key from `$request` before `DonateRoute`/`ValidationRoute` build their DTOs from it — and I wanted to flag it with the full history, since the fix already existed on this branch.

**What changed:** `__invoke` is currently `(array $request): void`. Before, it was `(array $request): array`, with `unset($request[self::TOKEN_KEY])` and both routes reassigning `$request = (new AuthenticateFormRequestWithToken())($request)`. That's how it matters: `DonateFormRouteData`/`ValidationRouteData` store the raw request array, and `ValidationRoute::logError()` logs `$formData->toArray()` on validation failures — so today, the token can end up in the application log.

**It's not an active leak right now** — `Log::redact()` recursively redacts any context key matched by `stripos($key, 'token')`, and `authToken` matches, so it's still getting redacted. But that's a generic, independently-changeable keyword list doing the job this feature's own code used to guarantee explicitly.

**Where the fix went:** @jonwaldstein, this was already built — commit `063901e68` ("Enhancement: strip the embedded form auth token inside the authenticate action") did exactly this: made `__invoke` return the stripped array, updated both routes, and added `testTokenIsStrippedFromTheRequest`. It's not reachable from the current tip — `DonateRoute.php` here is byte-identical to that commit's parent (`b6cfae090`). I don't see a commit or comment explaining the removal, so my guess is it fell out during the interactive rebase that restructured the stack's history around the same time, not a deliberate call.

@bordoni — this is also the direct answer to your "Should this be store somewhere?" comments on `DonateRoute.php`/`ValidationRoute.php`: yes, and it already was, briefly.

Since the fix, the routing changes, and the test all already exist verbatim in `063901e68`, this should just be a cherry-pick (or a manual reapply if the rebase makes that awkward) rather than new work. Might also be worth a quick pass over the rest of the rebased commits in case anything else fell out the same way — this one only surfaced because I was tracing a specific line's history for an unrelated reason.

jonwaldstein:
> While re-checking this PR after the recent rebase, I found that AuthenticateFormRequestWithToken::__invoke() no longer strips the authToken key from $request before DonateRoute/ValidationRoute build their DTOs from it — and I wanted to flag it with the full history, since the fix already existed on this branch.

@glaubersilva that was intentional - coderabbit flagged the issue, I updated the PR, then I reverted it and squashed the commit because I realized we already redact the logging information. There is no reason to unset the variable from the request. These AI reviews are tripping over themsleves 😅  

This was the comment for reference https://github.com/impress-org/givewp/pull/8320#discussion_r3982772804

src/DonationForms/Actions/AuthenticateFormRequestWithToken.php
jonwaldstein: Good catch. Dropped the argument in 86ef21c8d so the parse call no longer implies a scheme check that only happens in `wp_validate_auth_cookie()` below.

glaubersilva (approved):
@jonwaldstein This look good, great work! 

I just left some comments about minor details related to translations in a previous PR, as you can see here: https://github.com/impress-org/givewp/pull/8311#pullrequestreview-5182763713

But maybe it would be better to implement these fixes at the top of the stack to make things easier.

PR #8326, opened 2026-09-13
[SMTNC-1971] Fix: create a single copy of each campaign form when duplicating a campaign

WayneRocha:
### 🎫 Ticket

[SMTNC-1971](https://linear.app/nexcess/issue/SMTNC-1971/give-recurring-goal-format-meta-key-should-not-create-new-draft-forms)

### 🎥 Artifacts

[]()

### 🗒️ Description

Stops campaign duplication from creating an extra draft form for every duplicate row in a form's metadata, and makes a grouped model query count its distinct groups instead of the first group's rows.

### ⚠️ Blockers

None.

---
<a href="https://app.blacksmith.sh/impress-org/codesmith/givewp/pr/8326"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791935785&installation_model_id=426813&pr_number=8326&repository=impress-org%2Fgivewp&return_to=https%3A%2F%2Fgithub.com%2Fimpress-org%2Fgivewp%2Fpull%2F8326&signature=9e8fc4aaf8933b3d3293a5d8bfefe3452c7259692d4984327721c5db3df2f3ef"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed campaign duplication so each associated form is copied only once, even when duplicate form metadata keys are present.
  * Prevented duplicate campaign-form records from being created during duplication.
  * Ensured campaign form listings return each associated form once when duplicate metadata entries exist.
  * Improved grouped result counts, including groups containing empty values.

* **Tests**
  * Added coverage for campaign duplication, form listings, and grouped result counting to help prevent regressions.


src/Framework/Models/ModelQueryBuilder.php
```diff
@@ -43,6 +44,18 @@ public function count($column = null): int
...
         $column = ( ! $column || $column === '*') ? '1' : trim($column);
 
+        /*
+         * A grouped query returns one row per group and get_row() reads only the first of them,
+         * so counting the grouped columns is what the caller is actually asking for.
+         */
+        if ($this->groupByColumns) {
+            $groupedColumns = implode(', ', $this->groupByColumns);
+            $this->groupByColumns = [];
+            $this->selects = [new RawSQL("SELECT COUNT(DISTINCT {$groupedColumns}) AS count")];
```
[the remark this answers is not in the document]
WayneRocha: Changes on [10fc63a](https://github.com/impress-org/givewp/pull/8326/commits/10fc63ad7ea73fe17f32ee94dfb4607750095844)

src/Campaigns/Models/Campaign.php
```diff
@@ -88,7 +89,8 @@ public function forms(): ModelQueryBuilder
                     ->leftJoin('give_campaign_forms', 'campaign_forms')
                     ->on('campaign_forms.form_id', 'id');
             })
-            ->where('campaign_forms.campaign_id', $this->id);
+            ->where('campaign_forms.campaign_id', $this->id)
+            ->groupBy('ID');
```
jonwaldstein: Small one: could we qualify this as `forms.ID`? It works unqualified right now only because `give_campaign_forms` has no `id` column, so if a later join brings one in, MySQL will throw an ambiguous column error here.

```suggestion
            ->groupBy('forms.ID');
```

jonwaldstein (commented):
@WayneRocha two notes, neither blocking:

- Is the Recurring-side fix for the duplicate `_give_recurring_goal_format` writes tracked? Core only ever calls `update_meta` for that key, so something on the add-on side is inserting the second row. Until that stops, other `DonationForm::query()` consumers (form list table, `Event::forms()`) will still see duplicates.
- `Aggregate::count()` on the plain `QueryBuilder` has the same grouped-count bug. Since [stellarwp/db](https://github.com/stellarwp/db) was forked from this builder, it probably carries it too. Eventually we want to update Give to use that package instead. Worth checking there when we move over rather than fixing here.

src/Campaigns/Models/Campaign.php
WayneRocha: Applied. The qualification exposed that `Campaign::forms()` runs the V2 form query, which had no table alias, so I aliased the V2 repository's `FROM posts` as `forms` (matching the v3 repo pattern) in `2a3d8ae85`. `GROUP BY forms.ID` is now valid and CI is green across the matrix.

WayneRocha:
Resolving the CodeRabbit reviews on this PR:

**Review 5306182180 — "Use a metadata key joined by `Campaign::forms()` in both tests"** — no change needed; the premise describes the wrong query. `Campaign::forms()` builds the **V2** `DonationFormsRepository::prepareQuery()`, which attaches every form meta key (`formBuilderSettings`, `formBuilderFields`, and all nine V2 keys including `recurringGoalFormat`); the join list appears in the query text (e.g. `give_formmeta_attach_meta_recurringGoalFormat ... meta_key = '_give_recurring_goal_format'`). The settings/fields-only query is the **v3** `DonationFormRepository`, which `Campaign::forms()` does not use. Verified empirically: removing `groupBy('forms.ID')` makes both `testDuplicateMetadataDoesNotCreateAdditionalCampaignForms` and `testCampaignFormsReturnsEachFormOnceWithDuplicateMetadata` **fail** (the duplicate meta rows multiply the join), and with the `groupBy` both pass. The tests do exercise the grouping with the duplicated meta key, so no `SETTINGS` swap is warranted.

**Review 5307132855 — grouped `count($column)` + HAVING aliases** — both fixed in c721fe5af: grouped `count($column)` now sums each group's non-null column counts (`SELECT SUM(nonNullCount) ... FROM (SELECT COUNT(col) AS nonNullCount ... GROUP BY ...) ...`), and `getGroupCountSQL()` keeps the SELECT entries whose aliases HAVING references instead of dropping them. Two tests added in `ModelQueryBuilderTest`. Full matrix CI is green.

PR #8327, opened 2026-09-14
[SMTNC-2333] Fix: donor wall shows donor names, initials, and totals again

WayneRocha:
### 🎫 Ticket

[SMTNC-2333](https://linear.app/nexcess/issue/SMTNC-2333/donor-wall-should-show-the-names-of-the-donors)

### 🎥 Artifacts

Loom from @Robincredible as a reference of what the problem is
https://www.loom.com/share/d41aaabf5b9e4a1c8c106bb5be63a841

Fixed on this branch 👇

Before
image

After
image

### 🗒️ Description

Affected versions : [4.16.7.2](https://github.com/impress-org/givewp/releases/tag/4.16.7.2), [4.16.8](https://github.com/impress-org/givewp/releases/tag/4.16.8), [4.16.8.1](https://github.com/impress-org/givewp/releases/tag/4.16.8.1)

Regression 

The donor wall read every donation meta row through a bare `unserialize()`, which returns `false` for the plain strings nearly all of those rows hold, so names, initials, and totals rendered blank.

https://github.com/impress-org/givewp/blob/3bf0e543103d21195bb2482bab12b469211b7fa1/includes/donors/class-give-donor-wall.php#L418

reading through `Utils::safeUnserialize()` returns non-serialized values untouched while still refusing to return an object.

https://github.com/impress-org/givewp/blob/e6f1e237ac217e9dadd229b6f588fe96cf917a8c/includes/donors/class-give-donor-wall.php#L421-L423

* **Bug Fixes**
  * Restored donor names, initials, and donation totals in the donor wall block and shortcode.
  * Donor names stored as plain text now display correctly.
  * When a donor image is unavailable, the donor’s initials are shown instead.
  * Invalid donor data is prevented from appearing publicly.

* **Tests**
  * Added coverage for donor names, initials, and safe handling of invalid donor data.
  * Added tests to ensure no security regression from [SVUL-82](https://linear.app/nexcess/issue/SVUL-82)

### ⚠️ Blockers

None.

---
<a href="https://app.blacksmith.sh/impress-org/codesmith/givewp/pr/8327"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791951319&installation_model_id=426813&pr_number=8327&repository=impress-org%2Fgivewp&return_to=https%3A%2F%2Fgithub.com%2Fimpress-org%2Fgivewp%2Fpull%2F8327&signature=9ad52a59ea4697e72c349e63e4b7893bc075bc55f01a1abf3dd6a7080e9fc333"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Donor wall now correctly displays donor names, initials, and donation totals.
  - Improved handling of stored donor information to prevent unsafe serialized data from affecting donor wall rendering.

- **Security**
  - Added safeguards to prevent potentially harmful serialized payloads from being instantiated while displaying donor information.

- **Tests**
  - Added coverage for donor names, initials, totals, and protection against unsafe stored data.


tests/Feature/Donors/DonorWallTest.php
```diff
@@ -0,0 +1,136 @@
...
+     * @since TBD
+     */
+    private function denyGravatarLookups(): void
+    {
+        add_filter(
+            'pre_http_request',
+            static function () {
+                return ['response' => ['code' => 404, 'message' => 'Not Found'], 'body' => ''];
+            }
+        );
```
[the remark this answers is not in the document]
WayneRocha: Good catch, applied in e6f1e237a. The callback now takes `($preempt, $parsedArgs, $url)` at priority 10 with 3 args and returns `$preempt` unless the URL contains `gravatar.com`, so any other request the render path makes still goes through normal handling instead of being absorbed by the stub.

includes/donors/class-give-donor-wall.php
```diff
@@ -416,7 +418,9 @@ private function get_donation_data( $atts = [] ) {
 
 			/* @var stdClass $result */
 			foreach ( $results as $result ) {
-				$temp[ $result->{$donation_id_col} ][ $result->meta_key ] = unserialize( $result->meta_value, [ 'allowed_classes' => false ] );
+				$temp[ $result->{$donation_id_col} ][ $result->meta_key ] = is_string( $result->meta_value )
+					? Utils::safeUnserialize( $result->meta_value )
+					: $result->meta_value;
```
jonwaldstein, line 424: @WayneRocha while your here just replace this logic to use our models so we don't have to do any unserializing. This is currently a super legacy way of querying data we want to move away from so might as well do it now 😄

jonwaldstein (changes requested):

includes/donors/class-give-donor-wall.php
jonwaldstein: @WayneRocha 

Thanks for taking a swing at the models here! i'm realizing this is a bigger scope than I thought - so we can circle back one day when its more of a priority 😄 

After looking closer, the template still needs the full legacy meta array (Tributes reads `_give_tributes_*` straight out of it), so the model path ends up building the same array from two queries instead of one. Let's back this out to the original loop and just stop decoding meta at all: the wall only ever renders scalars, so the raw value is what we want.

```php
$temp[ $result->{$donation_id_col} ][ $result->meta_key ] = $result->meta_value;
```

That keeps the single query and ordering, and there's no unserialize() in the class at all. The comment is already in those rows too, so you can read it from `$donation_data[ DonationMetaKeys::COMMENT ]` instead of calling `give_get_payment_meta()` per card. Please keep the tests, they're great.

WayneRocha: Good call haha. way more simple 😀
2b1f048af466715173a46e439fd7e51dc493e3e9

jonwaldstein (approved):

PR #8338, opened 2026-09-18
Fix: donations list slow and showing trashed donations because of an ungrouped OR in the query

jonwaldstein:
## Description

This fixes the donations list table on large sites, where the first page could take minutes to load and trashed donations were showing up alongside live ones. The live/test mode filter and the name search were both missing parentheses around an "or", so the database was ignoring the other filters and scanning everything. On a million-donation test site the list query went from about two minutes to a few milliseconds.

- The remaining load time on very large sites is the total-count query, which gets its own follow-up PR.

## Testing Instructions

1. Go to Donations in the admin (the React list, not the legacy view) with test mode off. Move one live donation to the trash and reload. The trashed donation should no longer appear in the list, and the total count should drop by one.
2. Create a donation in test mode, then view the list with test mode off. It should not appear. Turn test mode on and it should.
3. In the list search box, type a donor's last name that also matches a trashed donation's first name. Only the non-trashed donation should appear.
4. `composer test:serial -- --filter TestListDonations` covers the above.

## Pre-review Checklist

- [x] Acceptance criteria satisfied and tested
- [x] Relevant `@since` tags added to code docblocks
- [x] Changelog entry added
- [x] E2E or unit tests added and passing locally

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
[...]

PR #8339, opened 2026-09-18
Enhancement: fast donations list table on very large sites

jonwaldstein:
## Description

Stacked on #8338.

This makes the donations list table fast on sites with hundreds of thousands of donations, including the last page, search, and the total count. On a million-donation test site every page now loads in well under a second where it used to take minutes. 🚀

- The list now finds the page of donation IDs first and loads the full rows after, the same way the legacy screen always did.
- Searching by name or email now matches from the start of the value (typing "smi" finds "Smith", "jo@" finds "jo@example.org") so the database can use an index instead of scanning every donation.
- A migration replaces the two single-column indexes on the donation meta table with composite ones, the same layout WooCommerce uses for order meta. It took about 30 seconds on a million donations, grows the table by roughly a third rather than doubling it, and skips work that is already done. The same indexes make campaign totals, form goal progress, and donor stats drop from tens of seconds to milliseconds on the same site with no code change.

## Testing Instructions

1. Go to Donations in the admin (React list) and page through to the last page. Each page should load quickly and the total count at the bottom should match the number of non-trashed donations for the current mode.
2. Search for the first few letters of a donor's last name, then of their email address. Matching donations appear; searching a middle fragment of a name (like "mit" for Smith) no longer matches, which is the intended trade-off.
3. Sort by Amount and by Donor and page forward. Order should stay correct across pages.
4. Filter by campaign, by status, and by a date range with test mode on and off. Counts and rows should agree with the legacy Donations screen for the same filters.
5. On a site with an existing database, visit the admin after updating: the "Add composite indexes to the donation meta table" migration should run once from Donations > Tools > Updates with no errors.
6. `composer test:serial -- --filter 'TestListDonations|TestListSubscriptions'` covers the above.

## Pre-review Checklist

- [x] Acceptance criteria satisfied and tested
- [x] Relevant `@since` tags added to code docblocks
- [x] Changelog entry added
- [x] E2E or unit tests added and passing locally

---
<a href="https://app.blacksmith.sh/impress-org/codesmith/givewp/pr/8339?ref=codesmith_pr_footer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792363688&installation_model_id=426813&pr_number=8339&ref=codesmith_pr_footer&repository=impress-org%2Fgivewp&return_to=https%3A%2F%2Fgithub.com%2Fimpress-org%2Fgivewp%2Fpull%2F8339&signature=77b333a70d8fa8cbc4f328a3811fefbeff90325ee0c36f1c6395b92ce9813974"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Performance**
  * Improved donation list loading, pagination, and searches for large sites.
  * Faster filtering by donor details, campaigns, and payment mode.
  * Improved database indexing for donation metadata.

* **Bug Fixes**
  * Corrected donation total counts and sorting across paginated results.
  * Improved separation of live and test donations in donation and subscription lists.
  * Campaign filtering now excludes unrelated campaigns and test donations when test mode is off.
  * Prevented duplicate migration runs during concurrent requests.
  * Migration history now displays the latest runs first.


src/Donations/Migrations/AddIndexesToDonationMetaTable.php
```diff
@@ -0,0 +1,78 @@
...
+        global $wpdb;
+
+        $table = $wpdb->prefix . 'give_donationmeta';
+
+        try {
+            $existing = array_flip(DB::get_col("SHOW INDEX FROM {$table}", 2));
+
+            $clauses = array_diff_key([
+                'donation_meta_key' => 'ADD INDEX donation_meta_key (donation_id, meta_key)',
+                'meta_key_value' => 'ADD INDEX meta_key_value (meta_key, meta_value(191))',
```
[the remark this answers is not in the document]
jonwaldstein: Fixed in 767d3ac. Both indexes now prefix `meta_key` at 191, matching WordPress core and the single-column index this migration replaces.

One correction: the 767-byte limit applies per column prefix, not to the combined key. The combined limit is 3072 bytes. `meta_key(191), meta_value(191)` fits, so the 95/96 split is not needed.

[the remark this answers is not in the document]
jonwaldstein: Tested rather than argued. On MariaDB 12.3 a utf8mb4 InnoDB table created with `ROW_FORMAT=COMPACT` accepted both `(donation_id, meta_key(191))` and `(meta_key(191), meta_value(191))`, 772 and 1528 bytes combined, so the 767-byte limit is per key part there. Where you are right in effect is MyISAM: its limit is 1000 bytes for the whole key and it rejected the value index with `max key length is 1000 bytes`. Tables created on old hosts can be MyISAM, and a failed migration blocks every later one, so 89199e0 switches to the WooCommerce HPOS prefixes, `(donation_id, meta_key(100))` and `(meta_key(100), meta_value(82))`, which pass on both.

src/Framework/Migrations/MigrationsRunner.php
```diff
@@ -86,116 +93,130 @@ public function run()
...
+            require_once ABSPATH . 'wp-admin/includes/class-wp-upgrader.php';
+        }
 
-        foreach ($migrations as $migrationClass) {
-            $migrationId = $migrationClass::id();
+        // Migrations run on every request, so without this a slow one (an ALTER TABLE over millions of
+        // rows) starts again in every request that arrives while it is still running.
+        if ( ! WP_Upgrader::create_lock(self::LOCK_NAME)) {
+            return;
+        }
```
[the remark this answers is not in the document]
jonwaldstein: Kept the one-hour default. The bound is not the lock; it is the migration: `AddIndexesToDonationMetaTable` now treats indexes that appeared while it waited on the table lock as success, so a second runner after expiry blocks on the metadata lock, gets `Duplicate key name`, re-checks `SHOW INDEX` and returns cleanly. The lock exists to stop the pile-up of concurrent requests, which the default covers (16M rows built in about a minute). A longer timeout only delays retry after a fatal mid-run.

tests/Unit/Framework/Migrations/TestMigrationsRunner.php
```diff
@@ -0,0 +1,59 @@
...
+        $register->addMigration(LockedOutMigration::class);
+        $runner = new MigrationsRunner($register, give(MigrationLogFactory::class), give(MigrationLogRepository::class));
+
+        WP_Upgrader::create_lock(MigrationsRunner::LOCK_NAME);
+        $runner->run();
+        $this->assertSame(0, LockedOutMigration::$runs);
+
+        WP_Upgrader::release_lock(MigrationsRunner::LOCK_NAME);
+        $runner->run();
+        $this->assertSame(1, LockedOutMigration::$runs);
```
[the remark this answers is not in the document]
jonwaldstein: Fixed in 89199e0. The test's `setUp()` loads `class-wp-upgrader.php` with the same guard the runner uses.

src/Donations/Endpoints/ListDonations.php
```diff
@@ -192,51 +193,120 @@ public function getDonations(): array
...
+     * @since TBD
+     *
+     * @param string[] $sortColumns
+     *
+     * @return DonationMetaKeys[]
+     */
+    private function getSortDependencies(array $sortColumns): array
+    {
+        $sortSql = implode(' ', $sortColumns);
+        $candidates = [
```
WayneRocha: here you hardcode 5 meta keys, so any third-party sortable column (via the public [givewp_donations_list_table](https://github.com/impress-org/givewp/blob/c07d807fd58e43cb90478cc4175038108f228009/src/Donations/ServiceProvider.php#L45) hook) gets an ORDER BY on a never-joined column → the whole donations list 500s. Derive dependencies from registered columns instead.

changelog/fix-migrations-runner-lock.yaml
```diff
@@ -0,0 +1,4 @@
+significance: patch
+type: fix
+entry: Fixed database migrations starting a second time in requests that arrive while a slow migration is still running, and the migrations list now shows the latest run first
```
WayneRocha:
```suggestion
entry: Database migrations started a second time in requests that arrive while a slow migration is still running. Now it runs one time and the migrations list now shows the latest run first.
```

WOuld be fix: Fixed ...

src/Donations/Migrations/AddIndexesToDonationMetaTable.php
```diff
@@ -0,0 +1,87 @@
...
+
+            // Prefix lengths match WooCommerce HPOS wc_orders_meta. Under utf8mb4 they keep each key part under
+            // the 767-byte InnoDB COMPACT limit and the whole key under the 1000-byte MyISAM limit, so the
+            // ALTER succeeds on tables created by any host GiveWP has ever run on.
+            $clauses = array_diff_key([
+                'donation_meta_key' => 'ADD INDEX donation_meta_key (donation_id, meta_key(100))',
+                'meta_key_value' => 'ADD INDEX meta_key_value (meta_key(100), meta_value(82))',
+            ], $existing);
+
+            $clauses += array_intersect_key([
```
WayneRocha: the new meta-index migration can silently report success when only the DROP clauses failed, leaving redundant indexes forever.

try using a try-catch catching a `DatabaseQueryException` on this full block for security and mark the migrations as failed

WayneRocha (changes requested):
Found small adjustments. Approved after that.

src/Donations/Endpoints/ListDonations.php
jonwaldstein: Fixed in f2fed16. The ID query now attaches whichever donation meta key the sort references, and selects `updatedAt` too. Added a test that registers an add-on column sorting on `company`; it failed with `Unknown column` before the change.

changelog/fix-migrations-runner-lock.yaml
jonwaldstein: Resolved.

src/Donations/Migrations/AddIndexesToDonationMetaTable.php
jonwaldstein: Fixed in f2fed16. The single `ALTER TABLE` is atomic, so the ADD clauses cannot land while the DROP clauses fail. The real gap was a run where the composite indexes already existed and only the DROPs ran: a failed DROP passed the recovery check. The catch now returns success only when both composite indexes exist and both single-column ones are gone. The try/catch already wraps the whole block.

WayneRocha (approved):

PR #8343, opened 2026-09-21
Fix: forms with no campaign fall back to their own goal

jonwaldstein:
## Description

Resolves SOFT-4695.

Forms that aren't linked to a campaign no longer break things. This is the groundwork for standalone forms without campaigns.

- A form set to use its campaign's goal but with no campaign now falls back to its own goal, on the public form page and in the admin forms list.
- Donations to a form with no campaign (or a deleted one) show the form title in the Campaign column of the donations list instead of an error.

## Testing Instructions

1. Create a campaign. There's no way to unlink a form in the admin yet, so open the DB and delete the campaign manually.
2. Open the form on the front end. It loads and shows the form's own goal instead of an error.
3. Go to Donations > Forms. The list loads and the goal column for that form shows its own goal.
4. Make a test donation to that form, then go to Donations. The Campaign column shows the form title with no "Something went wrong" message, and Tools > Logs has no new error entries.
5. `composer test:serial -- --filter 'DonationFormGoalDataTest|TestDonationFormQueryData|CampaignColumnTest'` covers the above.

## Pre-review Checklist

- [x] Acceptance criteria satisfied and tested
- [x] Relevant `@since` tags added to code docblocks
- [x] Changelog entry added
- [x] E2E or unit tests added and passing locally

### DEMO

https://www.loom.com/share/c094059378e0452680c8a18e9da6bcce

## Summary by CodeRabbit

* **Bug Fixes**
  * Unlinked donation forms now use their own goal settings on form pages and in the forms list.
  * Donations without an available campaign now display the associated form title.
  * Donations linked to an existing campaign continue to display the campaign title and link.


WayneRocha (approved):

WayneRocha:
@jonwaldstein Before merging and also @shaynesanderson or @vicskf when QA'ing this make sure no problems having when having P2P active

glaubersilva (commented):
@jonwaldstein I know that it is possible to access the list with all forms in  `https://example.com/wp-admin/edit.php?post_type=give_forms&page=give-forms` address, but not sure if everybody knows, so I think it would be important to mention it in the QA instructions since this PR still doesn't add this menu item and it's useful for tests.

glaubersilva:
@jonwaldstein Another thing: I'm getting the header without a white background, like in your Loom video, so we need to fix it at some point. Mentioned it here just as a reminder.

image

